PR #3, opened 2019-07-30
Simple CMake in AzureDevOps

sergey-shandar:


.gitignore
```diff
@@ -326,5 +326,14 @@ ASALocalRun/
...
 .mfractor/
+
+# CMake
+CMakeCache.txt
+CMakeFiles/
+
+*.vcxproj
+*.vcxproj.filters
+*.sln
+cmake_install.cmake
```
joshfree: missing new line

CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(sdk/keyvault/keyvault)
```
joshfree: missing new line

sdk/keyvault/keyvault/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+cmake_minimum_required (VERSION 3.0)
```
jhendrixMSFT: This minimum version doesn't match what's in the template https://github.com/Azure/azure-sdk-for-c/pull/2/files (3.15)

joshfree: any reason not to just use 3.15.1 since this is a brand new repo?

sergey-shandar: Yes, I started it before. I will update to the latest template as soon as it merged.

sergey-shandar: No, I'm just trying to integrate CMake with AzureDevOps. I will update to the latest template as soon as it merged to master

sergey-shandar: BTW, it looks like, AzureDevOps Ubuntu 16 has a CMake version 3.12. We will need either to use 3.12 or install 3.15 with some kind of `apt-get`.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -0,0 +1,19 @@
...
+      Linux:
+        vm.image: 'ubuntu-16.04'
+      Win:
+        vm.image: 'windows-2019'
+      MacOS:
+        vm.image: 'macOS-10.14'
+  pool:
+    vmImage: $(vm.image)
+  steps:
+  - task: CMake@1
```
sergey-shandar: @bsiegel currently, it tries to build all projects by using `CMakeLists.txt` in the root. I will need to update it to is `$(ServiceDirectory)`.

sergey-shandar:
@jhendrixMSFT, @barcharcraz please, review :-)

CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+cmake_minimum_required (VERSION 3.12)
+project(az)
+add_subdirectory(sdk/storage/storage)
```
jhendrixMSFT: Why is /storage repeated, or is this just temporary to give the build system something to build and will be removed later?

build.ps1
```diff
@@ -0,0 +1,4 @@
+$location = $PSScriptRoot
+$build = Join-Path $location "build"
+CMake -S $location -B $build
+CMake --build $build
```
jhendrixMSFT: Missing new-line.

sdk/storage/storage/build.ps1
```diff
@@ -0,0 +1,4 @@
+$location = $PSScriptRoot
+$build = Join-Path $location "build"
+CMake -S $location -B $build
+CMake --build $build
```
jhendrixMSFT: Missing new-line.

build.ps1
jhendrixMSFT: Also does this mean we're standardizing on PS for cross-plat build scripting?

sergey-shandar: @jhendrixMSFT, @bsiegel  as far as I know, PS core is the only one that is installed by default on all Azure Dev Ops hosts (Windows, Linux, Mac OS X). This specific `build.ps1` is not used by CI and it's for developers. The problem is that by default CMake will output everything in current directory :-)

CMakeLists.txt
sergey-shandar: I'm following a pattern from other SDKs. AFAIK, some services may have several libraries/projects etc.

build.ps1
barcharcraz: IMO it's better to just not add these kinds of build scripts, but that's just me.

sdk/keyvault/keyvault/CMakeLists.txt
barcharcraz: this should be at least 3.12, no reason to revert to a lower version here after setting 3.12 in the parent project

PR #10, opened 2019-08-07
Create CODEOWNERS

sergey-shandar:


bsiegel:
Looks good, but this file should live under `.github/`

Issue #17, opened 2019-09-04
Azure Core: Create error abstraction


jhendrixMSFT:
The core must provide types/abstractions for useful and consistent error reporting.

PR #31, opened 2019-09-19
Azure SDK Core library, data types.

sergey-shandar:
Issue #17


sdk/core/core/README.md
```diff
@@ -0,0 +1,123 @@
...
+typedef struct {
+  double a;
+  double b;
+} az_mytype_bar;
+
+inline az_mytype_bar az_mytype_bar_create(double const a, double const b) {
+  return (az_mytype_bar){ .a = a, .b = b };
+}
+
+struct my_type {
```
barcharcraz: what's this doing here?

sdk/core/core/inc/az_result.h
```diff
@@ -0,0 +1,47 @@
...
+};
+
+typedef int32_t az_result;
+
+#define AZ_MAKE_ERROR(facility, code) ((az_result)(0x80000000 | ((uint32_t)(facility) << 16)) | (uint32_t)(code))
+
+#define AZ_MAKE_RESULT(facility, code) ((az_result)(((uint32_t)(facility) << 16)) | (uint32_t)(code))
+
+inline bool az_failed(az_result result) {
+  return result < 0;
```
barcharcraz: didn't we want to change this to bitwise ops?

sdk/core/core/inc/az_result.h
```diff
@@ -0,0 +1,47 @@
...
+}
+
+inline bool az_succedded(az_result result) {
+  return result >= 0;
+}
+
+#define AZ_RETURN_IF_NOT_OK(exp) \
+  do { \
+    az_result const result = (exp); \
```
barcharcraz: could align \s, clang-format may do it for you

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,82 @@
...
+  // A numer of C characters (bytes) in the string.
+  size_t size;
+} az_const_str;
+
+typedef struct {
+  uint8_t* begin;
+  size_t size;
+} az_str;
+
+AZ_STATIC_ASSERT(CHAR_BIT == 8);
```
barcharcraz: you really don't need this. If your bytes aren't 8bit so many things will break that it's not worth checking for.

sdk/core/core/src/az_core.c
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_assert.h>
```
barcharcraz: what's this header for? is it to test everything in one TU? Is it a PCH file?

sdk/core/core/README.md
sergey-shandar: Yes, we need to move to Design Guidance.

sdk/core/core/inc/az_result.h
sergey-shandar: @vhvb1989 will create a separate PR for Clang format.

sdk/core/core/src/az_core.c
sergey-shandar: Currently, it's just one file that test compilation.

sdk/core/core/inc/az_str.h
sergey-shandar: Some platforms may have `CHAR_BIT` equal to 16. Some really really old platforms had `CHAR_BIT` is equal to 7. Currently, we are saying that this library can't work with such platforms. At least the UTF-8 string library and string literals.

barcharcraz: which platforms have it equal to 16?

sergey-shandar: Some DSPs. I don't think we target them :-) but who knows how people will use the library.

sdk/core/core/inc/az_assert.h
```diff
@@ -0,0 +1,19 @@
...
+#ifndef AZ_ASSERT_H
+#define AZ_ASSERT_H
+
+#include <assert.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define AZ_ASSERT(CONDITION) assert(CONDITION)
```
jhendrixMSFT: Why do we need our own assert macro?

sdk/core/core/inc/az_result.h
```diff
@@ -0,0 +1,47 @@
...
+#endif
+
+enum {
+  AZ_OK            =          0,
+  AZ_ERROR_FLAG    = 0x80000000,
+  AZ_STREAM_RESULT =    0x10000,
+  AZ_JSON_RESULT   =    0x20000,
+};
+
+typedef int32_t az_result;
```
jhendrixMSFT: I presume this is the error type we discussed the other week.  I suggest adding comments that explain the layout of the bits and what they mean (doesn't have to block this PR).

sdk/core/core/inc/az_static_assert.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef AZ_STATIC_ASSERT_H
+#define AZ_STATIC_ASSERT_H
+
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define AZ_STATIC_ASSERT(CONDITION) typedef int _az_static_assert[(CONDITION) ? 1 : -1];
```
jhendrixMSFT: I believe that names beginning with a leading underscore are reserved for the standard library, so while I doubt there is actual worry for collisions we should try to avoid this where possible.

sdk/core/core/README.md
```diff
@@ -0,0 +1,123 @@
+# Azure SDK Core Library
+
+In high-level programming languages, an [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control)
+is commonly used. For example, a generic JSON deserializer may call a user provided function to set an object property value
+(similar to [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern)). The principle requires a call-back
+function or a virtual table. In C, a call-back or a virtual table requires an untyped parameter `void *context`.
+It's very hard to work safely with such functions so Azure SDK Core library for C tries to avoid these technic.
```
joshfree: nit: "these techniques."

sdk/core/core/inc/az_static_assert.h
sergey-shandar: @barcharcraz can we use `_` for private/internal things?

sdk/core/core/README.md
```diff
@@ -0,0 +1,123 @@
+# Azure SDK Core Library
```
joshfree: I would replace this section with information that's useful for an end user to understand the library.  Why do they care or need Azure.Core?  Here is what we can write:

"Azure.Core provides shared primitives, abstractions, and helpers for modern Azure SDK client libraries written in the C programming language.  These libraries follow the Azure SDK Design Guidelines for Embedded C.

Azure.Core allows client libraries to expose common functionality in a consistent fashion, so that once you learn how to use these APIs in one client library, you will know hot to use them in other client libraries.

The main shared concepts of Azure.Core (and so Azure SDK libraries using Azure.Core) include:

* Configuring service clients, e.g configuring retries, logging, tracing.
* Accessing HTTP response details.
* Calling long running operations (LROs).
* Paging and asynchronous streams.
* Reporting errors from service requests in a consistent fashion.
* Abstractions for representing Azure SDK credentials.

Below, you will find sections explaining these shared concepts in more detail."

sdk/core/core/inc/az_assert.h
sergey-shandar: 1. we would like to have some control what does it do.
2. I plan to change the macro to somithing like `AZ_PRECONDITION`.

sdk/core/core/README.md
```diff
@@ -0,0 +1,123 @@
...
+In high-level programming languages, an [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control)
+is commonly used. For example, a generic JSON deserializer may call a user provided function to set an object property value
+(similar to [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern)). The principle requires a call-back
+function or a virtual table. In C, a call-back or a virtual table requires an untyped parameter `void *context`.
+It's very hard to work safely with such functions so Azure SDK Core library for C tries to avoid these technic.
+Instead, the Core library provides safe and useful parts to construct complex solutions.
+
+## 1. Data Types
+
+### 1.1. Const in `struct`
```
joshfree: This should be moved into the Developer Guide markdown for the repo.

sdk/core/core/inc/az_assert.h
jhendrixMSFT: I see.  Before we double-down on our own implementation we should check out the contract system in azure-ulib-c.  If we don't want/need to depend on that repo in full we can copy what they have.  Of course this would also mean they need to make that repo public.

sdk/core/core/inc/az_result.h
```diff
@@ -0,0 +1,55 @@
...
+
+#define AZ_MAKE_ERROR(facility, code) ((az_result)(0x80000000 | ((uint32_t)(facility) << 16)) | (uint32_t)(code))
+
+#define AZ_MAKE_RESULT(facility, code) ((az_result)(((uint32_t)(facility) << 16)) | (uint32_t)(code))
+
+inline bool az_failed(az_result result) {
+  return (result & AZ_ERROR_FLAG) != 0;
+}
+
+inline bool az_succedded(az_result result) {
```
JeffreyRichter: typo: s/b succeded

sdk/core/core/inc/az_result.h
```diff
@@ -0,0 +1,55 @@
...
+  return (result & AZ_ERROR_FLAG) == 0;
+}
+
+#define AZ_RETURN_IF_NOT_OK(exp) \
```
JeffreyRichter: I don't think we want this. The code should have to do this explicitly if it wants to do this. For example, how can I set a breakpoint on the line where we determined a failure occurred? Also, stepping in a debugger would just jump you around without a clear idea as to why.

sdk/core/core/inc/az_static_assert.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef AZ_STATIC_ASSERT_H
+#define AZ_STATIC_ASSERT_H
+
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define AZ_STATIC_ASSERT(CONDITION) typedef int az_static_assert[(CONDITION) ? 1 : -1];
```
JeffreyRichter: We really need some comments in these files. I don't know why or how someone would use this macro.
Also, do we really need this macro? Each thing we add, we MUST test, document, and support FOREVER! So, we want the surface area to be really tiny. The bar to add something like this is really, really high. There must be significant value. When in doubt, leave it out.

sdk/core/core/inc/az_str.h
JeffreyRichter: I also think we should pull this out. We can document that we expect all compilers to treat uint8_t as an unsigned 8-bit value.

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+// A muttable string slice.
+typedef struct {
+  uint8_t *begin;
+  size_t size;
+} az_str;
+
+AZ_STATIC_ASSERT(CHAR_BIT == 8);
+
+// A size of the string literal.
+// Details: to make sure that `S` is a `string literal`, we are appending `""` to `S`.
```
JeffreyRichter: I love the comment!

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+#define AZ_STRING_LITERAL_LEN(S) (sizeof(S "") - 1)
+
+#define AZ_CONST_STR_DECL(NAME, STRING_LITERAL) \
+  static az_const_str const NAME = { .begin = STRING_LITERAL, .size = AZ_STRING_LITERAL_LEN(STRING_LITERAL) };
```
JeffreyRichter: Can the right hand side of the = be replaced with AZ_CONST_STR?
You'd have to reverse the two macros in this file.

Also, are we sure we really even want this macro? Maybe it's better for the developer to explicitly write
     static az_cost_str const NAME = AZ_CONST_STR("hi there")

so that they know exactly what is happening here. I worry that all these macros obfuscates the code. Someone who is unfamiliar with this macro will have no idea what it's doing. It's not always about reducing typing; it's about understanding what is happening so that when bugs happen, the code is understandable and fixable.

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+
+#define AZ_CONST_STR(STRING_LITERAL) \
+  (az_const_str){ .begin = STRING_LITERAL, .size = AZ_STRING_LITERAL_LEN(STRING_LITERAL) }
+
+inline uint8_t const *az_const_str_ptr(az_const_str const buffer, size_t const index) {
```
JeffreyRichter: Again, we must have comments on all these methods/macro indicating what each does.
I'm not sure of the usage for this function. It seems to validate an index with a slice and then return a pointer.
The way slices (in Go) work is that you can slice a slice. In other words, I'd expect a function that takes a source slice, an index (from the begging of the source slice) and a new desired length and returns a new slice structure (or fails/asserts) whose address is oldPtr + newOffset and length is oldSize - newOffset.

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+  AZ_ASSERT(index <= buffer.size);
+  return buffer.begin + index;
+}
+
+inline uint8_t az_const_str_item(az_const_str const buffer, size_t const index) {
```
JeffreyRichter: This returns a byte; not an item. Please rename the function

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+  AZ_ASSERT(index < buffer.size);
```
JeffreyRichter: You could also assert if <0

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+  return buffer.begin[index];
+}
+
+inline uint8_t const *az_const_str_end(az_const_str const buffer) {
```
JeffreyRichter: This returns a pointer to the end of a slice? If so, I don't know why we'd ever need this. I shouldn't ever need to get to the end of a current slice.

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+  return az_const_str_ptr(buffer, buffer.size);
+}
+
+inline size_t az_const_str_index(az_const_str const buffer, uint8_t const *const p) {
```
JeffreyRichter: I don't think we should need this either. The caller can keep track of how far into a slice they are as they do processing. One of the main reason for introducing slices is to NOT use pointers and to avoid pointer arithmetic. The caller should try to avoid having a 'p' and, if they do, then they can't pass it to you to call this function.

sdk/core/core/inc/az_str.h
```diff
@@ -0,0 +1,83 @@
...
+  size_t const result = p - buffer.begin;
+  AZ_ASSERT(result <= buffer.size);
+  return result;
+}
+
+inline az_const_str az_const_substr(az_const_str const buffer, size_t const from, size_t const to) {
```
JeffreyRichter: Yes, this is the function I like. I'd like to pass -1 for to to mean "to the end".
You can also assert <0 for size and <-1 for to.

sdk/core/core/inc/az_str_stream.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
JeffreyRichter: I'm not convinced that this is necessary. Can't it all be done with whatever slice structures/functions we already have?
By the way, we need a function to get the length of a slice (just return its size field).

sdk/core/core/inc/az_str.h
sergey-shandar: The problem is C99. `AZ_CONST_STR` has `(az_const_str)` cast which makes the expression non-constant. For example, this code can't be compiled
```
static as_const_str hello = AZ_CONST_STR("hello");
```
VS error:
```
error C2099: initializer is not a constant
```

JeffreyRichter: But, if I understand this right, the code should not compile because we're trying to put a const value in a non-const variable. But, this would compile, right?

static as_const_str **_const_** hello = AZ_CONST_STR("hello");

sergey-shandar: No, it wouldn't. The problem is not `const` but how C interprets the cast `(az_const_str)`. Here are more examples:
```c
typedef struct {
  int x;
} foo;

// Ok
foo const instance1 = { .x = 5 };

// Compilation error. error C2099: initializer is not a constant
foo const instance2 = (foo){ .x = 5 };

// Compiltion error. error C2099: initializer is not a constant
foor const instance3 = (foo const){ .x = 5 };

void func(foo const f);

// Ok
accept_foo((foo){ .x = 5 });

// Compilation error, syntax error: '{'
accept_foor({ .x = 5 });

foo create_foo();

// Ok
accpept_foo(create_foo());

// Compilation error. error C2099: initializer is not a constant
foo const instance4 = create_foo();
```

sdk/core/core/inc/az_str.h
sergey-shandar: Initially, my implimentation was more like C++ with begin/end iterators but then I changed it to using indeces. You are right, we don't need this function any more.

sdk/core/core/inc/az_str_stream.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/pull/36

PR #52, opened 2019-10-04
HTTPS alpha interface + stopgap implementation

antkmsft:
1. I think I do not understand spans completely yet, but currently there isn't enough time for me to figure it out. We probably shouldn't take both response_buffer and out_response as parameters.

2. There should now be powershell on mac and linux, so this might even work on other platforms, haha.


sdk/core/core/src/az_net.c
```diff
@@ -0,0 +1,260 @@
...
+#else
+#include <unistd.h>
+#endif
+
+#include <_az_cfg_warn.h>
+
+#ifdef _MSC_VER
+// warning C4996: 'strcpy/strcat': This function or variable may be unsafe. Consider using
+// strcpy_s/strcat_s instead.
+#pragma warning(disable : 4996)
```
sergey-shandar: We should move it to `_az_cfg_warn.h`.

sdk/core/core/src/az_net.c
```diff
@@ -0,0 +1,260 @@
...
+  Sleep(milliseconds);
+#else
+  usleep(milliseconds * 1000);
+#endif
+}
+
+static char const * alloc_shell_exec(char const * const cmd) {
+  assert(cmd != NULL);
+
+  FILE * const cmd_output = _popen(cmd, "r");
```
antkmsft: this one has no underscore on posix, make an ifdef

sdk/core/core/src/az_net.c
```diff
@@ -0,0 +1,260 @@
...
+      = 4 * 1024; // trying to determine file size via fseek isn't going to work for the size
+                  // >=~1200 bytes for the reason that we are reading from a pipe. So we allocate
+                  // just large enough buffer and hope it is sufficient.
+
+  char * result = calloc(output_length, sizeof(char));
+  assert(result != NULL);
+
+  fgets(result, (int)output_length, cmd_output);
+
+  _pclose(cmd_output);
```
antkmsft: this one has no underscore in posix, it is only win32 that has

antkmsft:
Will be obsolete soon

PR #71, opened 2019-10-11
Fix implicit fallthrough warning on GCC

antkmsft:
More info: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/
The comment is inspired by C++17's [[fallthrough]] attribute (https://en.cppreference.com/w/cpp/language/attributes/fallthrough).


antkmsft:
https://github.com/Azure/azure-sdk-for-c/pull/75 approach is better

PR #72, opened 2019-10-11
Update and expand the contributing guide

bsiegel:
Based on the changes we made to get CI working, this should get people up and running with our SDK regardless of their OS.


bsiegel:
@sergey-shandar It looks like the steps to actually build are much simpler now (look at the [build definition](https://github.com/Azure/azure-sdk-for-c/blob/master/eng/pipelines/templates/jobs/archetype-sdk-client.yml)). Should we update the build.ps1 files? From what I can tell they're probably out of date now.

sergey-shandar:
@bsiegel yeah, we can remove this file completely. I may keep my own local PowerShell script on my machine, if needed.

PR #74, opened 2019-10-11
Bug fix for test.

sergey-shandar:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -38,4 +38,7 @@ jobs:
   - task: CMake@1
     inputs:
       cmakeArgs: --build .
-    displayName: cmake build
+      displayName: cmake build
```
bsiegel: This line diff needs to be reverted. Thanks yaml...

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -39,3 +39,7 @@ jobs:
     inputs:
       cmakeArgs: --build .
     displayName: cmake build
+  - task: CmdLine@2
```
bsiegel: Only suggestion here is for format this task as:

```yaml
- script: |
    ctest -C Debug
  workingDirectory: build
  displayName: ctest
```

bsiegel: Or feel free to give it a more useful display name

PR #79, opened 2019-10-14
Extra az_span functions mainly for interfacing with other libraries - to c_str(), etc.

antkmsft:


sdk/core/core/inc/az_span.h
```diff
@@ -14,6 +15,13 @@
 
 #include <_az_cfg_prefix.h>
 
+#define AZ_CONTRACT_ARG_VALID_SPAN(arg) \
+  do { \
+    if ((arg).begin == NULL && (arg).size > 0) { \
```
sergey-shandar: Use inline function inside this macro. We shouldn't call `arg` twice, it may have side effects.

sdk/core/core/inc/az_span.h
```diff
@@ -98,15 +108,68 @@ AZ_INLINE az_const_span az_to_const_span(az_span const span) {
  * span.
  */
 AZ_INLINE bool az_const_span_eq(az_const_span const a, az_const_span const b) {
-  return a.size == b.size && memcmp(a.begin, b.begin, a.size) == 0;
+  return a.size == b.size && memcmp(a.begin, b.begin, a.size * sizeof(*b.begin)) == 0;
```
sergey-shandar: Should we just use `AZ_STATIC_ASSERT(sizeof(uint8_t) == 1)` ?

sdk/core/core/inc/az_span.h
```diff
@@ -101,12 +113,69 @@ AZ_INLINE bool az_const_span_eq(az_const_span const a, az_const_span const b) {
...
+    memmove((void *)buffer.begin, (void const *)src.begin, src.size * sizeof(*src.begin));
+  }
+
+  out_result->begin = buffer.begin;
+  out_result->size = src.size;
+
+  return AZ_OK;
+}
+
+AZ_INLINE uint8_t * az_span_end(az_span const span) { return &span.begin[span.size]; }
```
sergey-shandar: `span.begin + span.size`. I would avoid `span.begin[span.size]` even if it's temporary.

antkmsft:
@vhvb1989, this is what we were talking about yesterday. You need az_span_to_c_str() when giving strings to libcurl.

PR #80, opened 2019-10-15
Visitors :-)

sergey-shandar:
It looks like visitors in C is much easier to construct than iterators


sdk/core/core/inc/az_callback.h
```diff
@@ -0,0 +1,34 @@
...
+typedef ptrdiff_t az_callback_data;
+
+#define AZ_CAT(A, B) A##B
+
+#define AZ_CALLBACK_ARG(NAME) AZ_CAT(NAME, _arg)
+
+#define AZ_CALLBACK_DECL(NAME, ARG) \
+  typedef ARG AZ_CALLBACK_ARG(NAME); \
+  typedef struct { \
+    az_result (*func)(az_callback_data const, ARG const); \
```
antkmsft: Should you use AZ_CALLBACK_ARG(NAME) instead of ARG?

sdk/core/core/test/curl_easy_perform_poc.c
```diff
@@ -4,6 +4,8 @@
 #include <stdio.h>
 #include <curl/curl.h>
 
+#pragma warning(disable: 4996)
```
antkmsft: please rebase against latest master, but in either case, you should've included _az_cfg.h instead of disabling.

antkmsft:
The implementation (not usage) is hard to understand and read, but I assume once it is debugged, there's not much need to look into it/debug it. I'd love to see some minimum viable product code that makes a couple of real REST api calls and reads response to see/prove how exactly readable/useful this is.

PR #108, opened 2019-10-23
Renaming az_write_span_iter to az_span_builder

sergey-shandar:
- az_write_span_iter => az_span_builder.
- fix security bug.


sdk/core/core/inc/az_span_builder.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef AZ_SPAN_BUILDER_H
+#define AZ_SPAN_BUILDER_H
+
+#include <az_span.h>
+
+#include <_az_cfg_prefix.h>
+
+typedef struct {
+  az_span span;
+  size_t i;
```
gilbertw: Can you add some comments.  The size on the builder is unnecessary.  I expect the size of the span in the builder to be the same as the size of the span it contains.  Why track the size twice?

JeffreyRichter:
I agree that comments would be great.
I *think* that Sergey is trying to make a span that you can append to.
In Go, a slice contains a reference to the start, a length, and a capacity. The capacity is the max size and the length is the currently-consumed size. You can append to a slice up to its capacity and then a re-allocation must occur (we wouldn’t do this in our non-allocating C SDK).

I think better variable/field naming could lead to less comments. But comments describing the purpose of the structure would be very helpful.


sdk/core/core/inc/az_span_builder.h
sergey-shandar: @gilbertw I've added some comments. The builder has a buffer (previously known as a `.span`) and a position (previously known as `.i`). The buffer (the type is a `az_span`) has a size which is capacity. `.position` is a current write position and should never be bigger than `.buffer.size`. We may rename it to `.size` I don't care much.

PR #111, opened 2019-10-24
change pointer alignment and braces

danewalton:
Seems like we are aligned on the pointer alignment. I also have added the break on brace option as was mentioned in the issued bug. The IoT client team prefers it and seems some others in that discussion do as well so if anyone else has any objections to that addition please feel free to voice concerns.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -5,7 +5,7 @@ TabWidth: 2
 IndentWidth: 2
 IncludeBlocks: Preserve
 NamespaceIndentation: None
-PointerAlignment: Middle
+PointerAlignment: Left
 AlignAfterOpenBracket: AlwaysBreak
 AlignEscapedNewlines: DontAlign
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -15,6 +15,7 @@ BinPackParameters: false
 IndentCaseLabels: true
 BreakBeforeTernaryOperators: true
 BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Allman
```
sergey-shandar: I applied this 'braces' to `az_json_read.c`. It was 456 lines, now it's 533. IMHO, `Allman` style is not very good for a lot of small functions and code blocks but it's good for huge unreadable C-style 
functions :-)

sergey-shandar:
Please, don't merge this now (next few weeks) because it can introduce a lot of merge conflicts with changes that we are working on. It may introduce more disruption than value. In a long-term, I don't care much.

PR #126, opened 2019-10-30
An HTTP response parser using JSON parser API and some parts of JSON parser implementation.

sergey-shandar:
This PR also includes 
- renaming
  - az_const_span => az_span
  - az_callback => az_action
  - az_span => az_mut_span
- CURL_SLIST appender using emitters.


sdk/core/core/inc/az_action.h
```diff
@@ -0,0 +1,47 @@
...
+
+#define AZ_CAT(A, B) A##B
+
+#define AZ_ACTION_ARG(NAME) AZ_CAT(NAME, _arg)
+
+/**
+ * Defines an action @NAME type which accepts an argument of type @ARG.
+ */
+#define AZ_ACTION_TYPE(NAME, ARG) \
+  typedef ARG AZ_ACTION_ARG(NAME); \
```
barcharcraz: I really don't like putting this stuff in a macro, especially one that defines structs.

sdk/core/core/inc/az_action.h
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+/**
+ * Defines a function @NAME##_action which creates an action of type @ACTION
+ * using the given @NAME function.
+ *
+ * The function is a safe way to create an action with the same type @SELF for 
+ * the first argument of the function @NAME and @ACTION.self.
+ */
+#define AZ_ACTION_FUNC(NAME, SELF, ACTION) \
```
barcharcraz: ditto about the macros.

It's really, really, really hard to figure out what code is actually doing when it starts to use and combine macros like this.

sdk/core/core/inc/az_contract.h
```diff
@@ -10,13 +10,18 @@
 
 #include <_az_cfg_prefix.h>
 
-#define AZ_CONTRACT_ARG_NOT_NULL(arg) \
+#define AZ_CONTRACT(condition, error) \
   do { \
-    if ((arg) == NULL) { \
+    if (!(condition)) { \
```
barcharcraz: this should do the thing we discussed where it calls a "panic" function, and returns its result (assuming it doesn't crash)

sdk/core/core/inc/az_action.h
sergey-shandar: Me neither, but I don't know a better way to do it without buggy unsafe `void *` casting.

sdk/core/core/inc/az_contract.h
```diff
@@ -10,13 +10,18 @@
...
       return AZ_ERROR_ARG; \
     } \
   } while (0)
 
+#define AZ_CONTRACT_ARG_NOT_NULL(arg) AZ_CONTRACT((arg) != NULL, AZ_ERROR_ARG)
```
barcharcraz: I prefer just writing this out in code, instead of wrapping it in another macro, but this one is defensible.

sdk/core/core/inc/az_contract.h
sergey-shandar: Yes, we have a placeholder for this function now. So it's an improvement ;-)

sdk/core/core/inc/az_contract.h
```diff
@@ -10,13 +10,18 @@
...
+
+#define AZ_CONTRACT_ARG_VALID_SPAN(span) AZ_CONTRACT(az_span_is_valid(span), AZ_ERROR_ARG)
```
barcharcraz: ditto

sdk/core/core/inc/az_curl_slist.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <curl/curl.h>
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * Appends an HTTP header to the given CURL list.
+ */
+AZ_NODISCARD az_result
+az_curl_slist_append_header(struct curl_slist ** const pp_list, az_pair const header);
```
barcharcraz: I somewhat perfer just calling curl_slist_append from the curl PAL, but this is OK I suppose (I haven't gotten to its implementation yet).

sdk/core/core/inc/az_http_header.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef AZ_HTTP_HEADER_H
+#define AZ_HTTP_HEADER_H
+
+#include <az_span_emitter.h>
+#include <az_pair.h>
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * Emits an HTTP header as a sequence of spans in a format "%{key}: %{value}"
```
barcharcraz: is that replacement format literal, or does the %{} stand in for something.

sdk/core/core/inc/az_action.h
sergey-shandar: I put some notes about principals how to work with actions here https://github.com/Azure/azure-sdk-for-c/blob/sergey/refactoring/emitter-design.md It's work in progress. Anyway, we already had this macro in `az_callback.h`

sdk/core/core/inc/az_http_header.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+AZ_NODISCARD az_result
+az_http_header_emit_span_seq(az_pair const * const self, az_span_action const action);
+
+/**
+ * @az_http_header_emit_span_sequence as @az_span_emitter.
+ */
+AZ_ACTION_FUNC(az_http_header_emit_span_seq, az_pair const, az_span_emitter)
```
barcharcraz: again, I really prefer just writing out the typedef here. That way when I see az_http_header_emit_span_seq_action I can simply "go to definition" and see the function pointer definition.

sdk/core/core/inc/az_http_header.h
sergey-shandar: No, I just need some kind of syntax to explain what the resulted string.

sdk/core/core/inc/az_http_query.h
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
barcharcraz: General note:

These headers are really small. This is fine, but can create extra work down the line. Will anyone actually want to use just some of these headers?

sdk/core/core/inc/az_http_response_action.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <_az_cfg_prefix.h>
+
+typedef struct {
+  az_span_action header_name_part_action;
+  az_span_action header_value_part_action;
+  az_span_action body_part_action;
+} az_http_response_action;
+
+AZ_ACTION_TYPE(az_http_response_emitter, az_http_response_action)
```
barcharcraz: see I've already forgotten what this macro does. I would _really_ prefer just writing out the type.

sdk/core/core/inc/az_http_response_parser.h
```diff
@@ -0,0 +1,75 @@
...
+typedef struct {
+  az_span_reader reader;
+  az_http_response_kind kind;
+} az_http_response_parser;
+
+/**
+ * Initializes an HTTP response parser.
+ */
+AZ_NODISCARD az_result
+az_http_response_parser_init(az_span const buffer, az_http_response_parser * const out);
```
barcharcraz: az_http_response_parser should be the first parameter

sdk/core/core/inc/az_json_read.h
```diff
@@ -61,7 +59,7 @@ typedef struct {
   az_json_stack stack;
 } az_json_state;
 
-AZ_NODISCARD az_json_state az_json_state_create(az_const_span const buffer);
+AZ_NODISCARD az_json_state az_json_state_create(az_span const buffer);
```
barcharcraz: these may want to be _init methods returning an az_result. Although I suppose if we ever need that the _init name is there.

sdk/core/core/inc/az_mut_span.h
```diff
@@ -0,0 +1,169 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * A mutable span of bytes.
+ */
+AZ_SPAN_TYPE(az_mut_span, uint8_t)
```
barcharcraz: I don't like this. Just write the type out.

sdk/core/core/inc/az_mut_span.h
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/**
+ * Cast the given mutable span to an immutable span.
+ */
+AZ_INLINE az_span az_mut_span_to_span(az_mut_span const span) {
+  return (az_span){ .begin = span.begin, .size = span.size };
+}
+
+AZ_INLINE bool az_mut_span_is_overlap(az_mut_span const a, az_mut_span const b) {
```
barcharcraz: is this used for anything but picking memmove/memcpy, if it is juse eliminate and always use memmove

sdk/core/core/inc/az_mut_span.h
```diff
@@ -0,0 +1,169 @@
...
+  if (az_span_is_overlap(az_mut_span_to_span(buffer), src)) {
+    return AZ_ERROR_ARG;
+  }
+
+  if (buffer.size < src.size) {
+    return AZ_ERROR_BUFFER_OVERFLOW;
+  }
+
+  if (!az_span_is_empty(src)) {
+    memcpy((void *)buffer.begin, (void const *)src.begin, src.size);
```
barcharcraz: just do memmove

sdk/core/core/inc/az_mut_span.h
```diff
@@ -0,0 +1,169 @@
...
+
+      ++a_ptr;
+      ++b_ptr;
+    } while (a_ptr != a_end);
+  }
+
+  return AZ_OK;
+}
+
+AZ_INLINE az_result az_mut_span_to_str(
```
barcharcraz: most of these should probably be regular functions. or at least not forced inline.

Some are large enough that the optimizer may actually be justified in deciding not to inline them (esp in -Os mode)

barcharcraz: that said, my personal style veers strongly toward non-inline functions (I write a ton of non-inline c++ templates too). So take that advice with a grain of salt.

sdk/core/core/inc/az_json_read.h
sergey-shandar: Should we introduce some relaxed rules for internal API to allow `_create` method? It's a little bitannoying to write simple initializers with `az_result` conventions.

sdk/core/core/inc/az_mut_span.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/128

sdk/core/core/inc/az_mut_span.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/128

sdk/core/core/inc/az_pair.h
```diff
@@ -4,39 +4,34 @@
...
+  az_span key;
+  az_span value;
 } az_pair;
 
 /// A span of pairs.
-typedef struct {
-  az_pair const * begin;
-  size_t size;
-} az_pair_span;
+AZ_SPAN_TYPE(az_pair_span, az_pair const)
```
barcharcraz: do these new types have the same "methods" as the regular span and mut span? if not I strongly suggest, again writing out the type here. If so you're using macros to implement c++ templates and that doesn't tend to end that well.

sdk/core/core/inc/az_pair.h
```diff
@@ -4,39 +4,34 @@
...
 
 /// @az_pair_visitor is a callback with one argument @az_pair.
-AZ_CALLBACK_TYPE(az_pair_visitor, az_pair)
+AZ_ACTION_TYPE(az_pair_action, az_pair)
 
-/// @az_pair_seq is a @az_pair sequence visitor.
-AZ_CALLBACK_TYPE(az_pair_seq, az_pair_visitor)
+/// @az_pair_emitter is an emitter of @az_pair sequences.
+AZ_ACTION_TYPE(az_pair_emitter, az_pair_action)
```
barcharcraz: I'm not sure doxygen will be able to figure this out, it may not matter though.

sdk/core/core/inc/az_span.h
```diff
@@ -15,51 +15,30 @@
...
-  /**
-   * A number of bytes in the span.
-   */
-  size_t size;
-} az_const_span;
+#define AZ_SPAN_TYPE(NAME, VALUE) \
+  typedef struct { \
+    VALUE * begin; \
+    size_t size; \
+  } NAME;
```
barcharcraz: feelings of ODR intensify.

sdk/core/core/inc/az_span.h
```diff
@@ -15,51 +15,30 @@
...
-AZ_INLINE bool az_const_span_is_valid(az_const_span const span) {
-  return span.begin == NULL ? span.size == 0
-                            : ((span.size == 0 || span.size > 0)
-                               && (span.size == 0 || span.begin <= (span.begin + span.size - 1)));
+  return span.size == 0 || (span.begin != NULL && span.begin <= span.begin + span.size - 1);
 }
 
 /**
  * Returns a byte in `index` position.
  * Returns `AZ_ERROR_EOF` if the `index` is out of the span range.
```
barcharcraz: formatting here, although I think the returns annotation got mangled someplace, so the design guide needs updating there

sdk/core/core/inc/az_span.h
```diff
@@ -71,209 +50,69 @@ AZ_INLINE az_result_byte az_const_span_get(az_const_span const span, size_t cons
...
-AZ_NODISCARD az_result az_span_replace(
-    az_span const buffer,
-    az_const_span const src,
-    uint8_t (*const func)(uint8_t const),
-    az_span * const out_result);
-
 #define AZ_ARRAY_SIZE(ARRAY) (sizeof(ARRAY) / sizeof(*ARRAY))
 
-#define AZ_SPAN(ARRAY) \
+#define AZ_SPAN_FROM_ARRAY(ARRAY) \
```
barcharcraz: this is asking for bugs because people will use it on arrays that have decayed to pointers.

barcharcraz: you may be able to do this with an inline function and the whole [static] c99 thing, but I'm not sure.

sdk/core/core/inc/az_span.h
```diff
@@ -71,209 +50,69 @@ AZ_INLINE az_result_byte az_const_span_get(az_const_span const span, size_t cons
...
- * } az_span_visitor;
+ *   az_result (* func)(void *, az_const_span);
+ *   void * self;
+ * } az_span_action;
  * ```
  *
  * Example of usage
  *
  * ```c
- * az_span_visitor const visitor = ...;
```
barcharcraz: there's a new examples section of the docs guidelines btw.

Since this isn't a full example (it doesn't compile) that may not be appropriate here.

sdk/core/core/inc/az_mut_span.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/129

sdk/core/core/inc/az_span_emitter.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+AZ_ACTION_FUNC(az_span_span_emit, az_span_span const, az_span_emitter)
+
+/**
+ * Calculates a size of a contigous buffer to store all spans from @self.
+ */
+AZ_NODISCARD az_result az_span_emitter_size(az_span_emitter const self, size_t * const out_size);
+
+/**
+ * The function creates a temporary zero-terminated string from @emmiter in dynamic memory.
```
barcharcraz: I find this whole "emitter" thing rather confusing. I fee like I need to trek through 2-3 layers of macros to see what's going on. It may be worth it, but it's confusing.

sdk/core/core/inc/az_span_reader.h
```diff
@@ -34,6 +36,25 @@ AZ_INLINE void az_span_reader_next(az_span_reader * const p_reader) {
   p_reader->i += 1;
 }
 
+// Parsing utilities
+
+AZ_NODISCARD AZ_INLINE az_result az_error_unexpected_char(az_result_byte const c) {
+  return az_failed(c) ? c : AZ_ERROR_UNEXPECTED_CHAR;
```
barcharcraz: I still really don't like the `az_result_foo` thing

sdk/core/core/inc/az_span.h
sergey-shandar: yes. Do you have a better idea how we can know the size of the array?

sdk/core/core/inc/az_str.h
```diff
@@ -28,7 +29,18 @@
...
 // `void some_function(az_const_span const span);`
-#define AZ_STR(STRING_LITERAL) (az_const_span) AZ_CONST_STR(STRING_LITERAL)
+#define AZ_STR(STRING_LITERAL) (az_span) AZ_CONST_STR(STRING_LITERAL)
+
+/**
+ * A callback that accepts zero-terminated string `char const *`.
+ */
+AZ_ACTION_TYPE(az_str_action, char const *)
+
+#define AZ_CR '\r'
```
barcharcraz: are you sure these should not all be "" literals?

Also: I encourage just writing them out again.

sdk/core/core/inc/az_url.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * https://en.wikipedia.org/wiki/URL
+ */
+typedef struct {
+  az_span host;
+  az_span path;
+  az_pair_emitter query;
```
barcharcraz: still not a fan of having these lazy callbacks.

sdk/core/core/src/az_base64.c
```diff
@@ -107,22 +107,21 @@ AZ_INLINE uint8_t base64_as_uint6(uint8_t const base64) {
...
-    az_const_span const input,
-    az_const_span * const out_result) {
+    az_mut_span const buffer,
+    az_span const input,
+    az_span * const out_result) {
   AZ_CONTRACT_ARG_NOT_NULL(out_result);
-  if (!az_span_is_valid(buffer) || !az_const_span_is_valid(input)) {
-    return AZ_ERROR_ARG;
-  }
+  AZ_CONTRACT_ARG_VALID_MUT_SPAN(buffer);
```
barcharcraz: would again prefer `AZ_CONTRACT(az_mut_span_is_valid(buffer))`

sdk/core/core/src/az_curl_slist.c
```diff
@@ -0,0 +1,44 @@
...
+ * Appends a zero-terminated string to the given CURL list.
+ */
+AZ_NODISCARD az_result
+az_curl_slist_append(struct curl_slist ** const self, char const * const str) {
+  AZ_CONTRACT_ARG_NOT_NULL(self);
+  AZ_CONTRACT_ARG_NOT_NULL(str);
+
+  struct curl_slist * const p_list = curl_slist_append(*self, str);
+  if (p_list == NULL) {
+    return AZ_HTTP_ERROR_PAL;
```
barcharcraz: this eats errors. But I'm not sure that's avoidable.

sdk/core/core/inc/az_curl_slist.h
sergey-shandar: It has a different signature so we have to wrap it for our callbacks.

sdk/core/core/src/az_curl_slist.c
```diff
@@ -0,0 +1,44 @@
...
+AZ_NODISCARD az_result
+az_curl_slist_append_header(struct curl_slist ** const self, az_pair const header) {
+  AZ_CONTRACT_ARG_NOT_NULL(self);
+
+  az_span_emitter const header_span_emitter = az_http_header_emit_span_seq_action(&header);
+  az_str_action const curl_slist_append_action = az_curl_slist_append_action(self);
+
+  // the function creates a temporary dynamic zero-terminated string from `header_span_emitter`
+  // and passes it to `curl_slist_append_action`.
+  return az_span_emitter_to_tmp_str(header_span_emitter, curl_slist_append_action);
```
barcharcraz: I'm really concerned about memory leaks with this emitter stuff.

You should run your tests under leaksan/valgrind/crt debug heap

sdk/core/core/src/az_curl_slist.c
sergey-shandar: That's from CURL docs. They may have additional error info somewhere but I suspect it's always OUT_OF_MEMORY.

sdk/core/core/src/az_http_request_builder.c
```diff
@@ -35,65 +35,69 @@ AZ_INLINE uint16_t get_headers_max(az_span const buffer, az_pair * const headers
...
   }
 
   if (buffer.size < max_url_size) {
     return AZ_ERROR_BUFFER_OVERFLOW;
   }
 
-  AZ_RETURN_IF_FAILED(az_span_set(buffer, '\0')); // zero the buffer; we don't have to do this
+  AZ_RETURN_IF_FAILED(az_mut_span_set(buffer, '\0')); // zero the buffer; we don't have to do this
 
-  az_span uri_buf = { 0, 0 };
```
barcharcraz: you only need one zero here.

sdk/core/core/inc/az_str.h
sergey-shandar: RFC says `CR` so I use `CR`, not `\r` :-)

sdk/core/core/src/az_http_response_parser.c
```diff
@@ -0,0 +1,242 @@
...
+  AZ_RETURN_IF_FAILED(az_span_reader_expect_char(p_reader, ' '));
+
+  // status-code = 3DIGIT
+  {
+    uint16_t status_code = 0;
+    for (int i = 3; i > 0;) {
+      --i;
+      uint8_t digit = 0;
+      AZ_RETURN_IF_FAILED(az_span_reader_expect_digit(p_reader, &digit));
+      status_code = status_code * 10 + digit;
```
barcharcraz: why not use `strtoll` or `strtol` (these are fine, it's `atoi` and friends that are bad)

sdk/core/core/src/az_http_response_parser.c
```diff
@@ -0,0 +1,242 @@
...
+  AZ_RETURN_IF_FAILED(az_http_response_parser_set_kind(self));
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD bool az_is_http_whitespace(az_result_byte const c) {
+  switch (c) {
+    case ' ':
+    case '\t':
+      return true;
```
barcharcraz: this may need a default: case or annotation to avoid falling afowl of static analysis.

sdk/core/core/src/az_http_response_parser.c
```diff
@@ -0,0 +1,242 @@
...
+  {
+    size_t const field_name_begin = p_reader->i;
+
+    // https://tools.ietf.org/html/rfc7230#section-3.2.6
+    // token = 1*tchar
+    // tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" /
+    //         "_" / "`" / "|" / "~" / DIGIT / ALPHA;
+    // any VCHAR,
+    //    except delimiters
+    while (true) {
```
barcharcraz: doing this parsing ourselves makes me very uneasy, unless we can confirm we're always talking to azure. (Note: just always using https does not do this, unless we validate specifically against azure's cert, AND we need to make sure there's no way for an attacker to control the http responses azure returns (for example by directing us to an attacker controlled azure endpoint).

sdk/core/core/src/az_uri.c
```diff
@@ -7,25 +7,22 @@
...
   switch (c) {
-    default:
-      return !(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'));
     case '-':
     case '_':
     case '.':
     case '~':
       return false;
+    default:
+      return !(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'));
```
barcharcraz: you might be able to get in trouble here if the source/execution encoding is something like EBCDIC, and MAYBE ShiftJIS.

sdk/core/core/test/test_http_response_parser.h
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
barcharcraz: can we also put through headers from an HTTP fuzzer?

sdk/core/core/test/test_http_response_parser.h
```diff
@@ -0,0 +1,97 @@
...
+// SPDX-License-Identifier: MIT
+
+#include <az_http_response_parser.h>
+#include <az_span.h>
+
+#include "./az_test.h"
```
barcharcraz: did you mean to do "./" here? it's not necessary.

sdk/core/core/src/az_uri.c
antkmsft: We put all these assumptions to the https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/src/az_assumptions.c file.
We also rely on that assumption in az_base64.c.
Not that it can't be done, but should you try to compile on EBCDIC, there will be many other problems: we send HTTP requests, and the other side expects ASCII/UTF8.

sergey-shandar:
https://github.com/Azure/azure-sdk-for-c/issues/134

sdk/core/core/inc/az_json_read.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/133

sdk/core/core/inc/az_span.h
sergey-shandar: Fixed

sdk/core/core/inc/az_span.h
sergey-shandar: @barcharcraz It's easy in C++ but I have no idea how to do it in C99 safely.

sergey-shandar: Reopen it if you have a solution

sdk/core/core/inc/az_span_emitter.h
sergey-shandar: This PR is a refactoring of some earlier ideas of `push` iterator. Here's some thoughts https://github.com/Azure/azure-sdk-for-c/blob/sergey/refactoring/emitter-design.md

sdk/core/core/inc/az_span_reader.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/130

sdk/core/core/inc/az_contract.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/135

sdk/core/core/inc/az_contract.h
sergey-shandar: https://github.com/Azure/azure-sdk-for-c/issues/135

sdk/core/core/inc/az_http_header.h
sergey-shandar: For this specific one, I don't have a safe solution without a macro. I may consider to remove `AZ_ACTION_TYPE` but it's almost impossible to do safe cast wtihout `AZ_ACTION_FUNC`.

sdk/core/core/inc/az_span.h
sergey-shandar: Will fix it as soon as we have doxygen running.

sdk/core/core/src/az_curl_slist.c
sergey-shandar: It's using `malloc` and `free` only in one funcion and there is no branches between them. It's safe then wrap them since C has no RAII.

sdk/core/core/src/az_http_response_parser.c
sergey-shandar: span is not zero-terminated string, we can't use `str...` functions here.

sdk/core/core/test/test_http_response_parser.h
sergey-shandar: Until we solve test story, I'm ok with this. It's easier than altering CMake

sdk/core/core/test/test_http_response_parser.h
sergey-shandar: Sure, could you give a good example how we can get them? In the future, we may also have a non-deterministic stress-test from some HTTP fuzzer app.

sdk/core/core/src/az_http_response_parser.c
sergey-shandar: @JeffreyRichter any comment on this?

sdk/core/core/inc/az_str.h
```diff
@@ -27,10 +28,21 @@
...
+
+#define AZ_LF '\n'
+
+#define AZ_CRLF "\r\n"
```
antkmsft: Can you see if you can define and use AZ_STR_CRLF the same way we define AZ_STR_ZERO?

sergey-shandar: I need this as a string as well so I can concatenate strings at compile time.

sdk/core/core/src/az_http_response_parser.c
JeffreyRichter: @barcharcraz , are you proposing an alternate solution?
I'm not sure what you think the problem is. Are you concerned that the buffer doesn't have well-formed http response payload in it? If so, how would you expect we be resilient to that?

PR #131, opened 2019-10-30
HTTP response parser with direct calls instead of discriminated unions

sergey-shandar:


sdk/core/core/inc/az_http_response_parser.h
```diff
@@ -61,14 +49,28 @@ typedef struct {
...
+ * An HTTP status line.
+ */
+AZ_NODISCARD az_result az_http_response_parser_get_status_line(
+    az_http_response_parser * const self,
+    az_http_response_status_line * const out);
+
+/**
+ * An HTTP header.
+ */
+AZ_NODISCARD az_result az_http_response_parser_get_header(
```
JeffreyRichter: This probably needs to be named az_http_response_parser_get_next_header

sdk/core/core/inc/az_http_response_parser.h
```diff
@@ -61,14 +49,28 @@ typedef struct {
...
+    az_http_response_parser * const self,
+    az_http_response_header * const out);
 
 /**
- * Returns the next HTTP response value.
+ * An HTTP body.
  */
-AZ_NODISCARD az_result az_http_response_parser_read(
+AZ_NODISCARD az_result az_http_response_parser_get_body(
```
JeffreyRichter: This seems like it would be inefficient since it would have to parse through all the headers to get to the body.
My design had you passing in the lastheader in order to get the body.
Your design allows a call to this method before any calls to az_http_response_parser_get_next_header.

sdk/core/core/src/az_http_response_parser.c
```diff
@@ -189,54 +203,18 @@ AZ_NODISCARD az_result az_http_response_parser_read_header(
...
 
-AZ_NODISCARD az_result az_http_response_parser_read_body(
+AZ_NODISCARD az_result az_http_response_parser_get_body(
     az_http_response_parser * const self,
     az_http_response_body * const out) {
   AZ_CONTRACT_ARG_NOT_NULL(self);
   AZ_CONTRACT_ARG_NOT_NULL(out);
 
+  if (self->kind != AZ_HTTP_RESPONSE_BODY) {
+    return AZ_HTTP_ERROR_INVALID_STATE;
```
JeffreyRichter: I see that you have this here but your design means a run-time failure while my design is more of a compile-time error since you had to pass the &lastHeader in as an argument to this method.

sdk/core/core/inc/az_http_response_parser.h
sergey-shandar: It the same. The function works in `O(1)` because the first parameter is a parser which maintains the state.

sdk/core/core/inc/az_http_response_parser.h
```diff
@@ -61,14 +49,28 @@ typedef struct {
...
+AZ_NODISCARD az_result az_http_response_parser_get_status_line(
+    az_http_response_parser * const self,
+    az_http_response_status_line * const out);
+
+/**
+ * An HTTP header.
+ */
+AZ_NODISCARD az_result az_http_response_parser_get_header(
+    az_http_response_parser * const self,
+    az_http_response_header * const out);
```
JeffreyRichter: Shouldn't this just be an az_pair?
Do we need a special az_http_response_header data type?

sdk/core/core/src/az_http_response_parser.c
sergey-shandar: What will happen if someone pass not the last header to your function?

sdk/core/core/inc/az_http_response_parser.h
```diff
@@ -61,14 +49,28 @@ typedef struct {
  * Initializes an HTTP response parser.
  */
 AZ_NODISCARD az_result
-az_http_response_parser_init(az_span const buffer, az_http_response_parser * const out);
+az_http_response_parser_init(az_http_response_parser * const out, az_span const buffer);
```
JeffreyRichter: Obviously, your design requires the user allocate an az_http_response_parser structure and the user must call az_http_response_parser_init on it before any of the other functions work.
My design does require this additional struct, the allocation of it, and the extra function definition & call.

So, for these additional concepts, code, documentation, testing, support costs, what and more customer code as well, what is the benefit that the customer is getting? You really have to answer this question: whenever we do anything, what is the benefit compared to the costs? Right now, I'm not seeing any benefit to what you have done.

sergey-shandar: What about code reusage?

JeffreyRichter: What is the code reusage you're think of here?

sergey-shandar: `az_http_response_parser` uses `az_span_reader`.

JeffreyRichter: I coded up the customer experience with both designs (removing error handling). We can show to the whole team and discuss:

#include <az_http_response_parser.h>
#include <az_span.h>

static void test_http_response_parser() {
    az_span const response = AZ_STR( //
        "HTTP/2.0 205 \r\n"
        "header1: some value\r\n"
        "Header2:something\t   \r\n"
        "\r\n");

///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////

    // Jeff's design:
    az_http_response_get_status_line(response, &status_line);

    az_pair header = { 0 };             // Jeff requires initialization
    while (az_http_response_get_next_header(&header) == AZ_OK) {
        // Process header...
    }

    az_span body = { 0 };
    az_http_response_get_body(&header, &body);  // You can't readget body unless you have a header
    // Process body...

///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////

    // Sergey's design:
    az_http_response_parser parser;                                 // Sergey 
    az_http_response_parser_init(&parser, response);                // Sergey 

    az_http_response_status_line status_line = { 0 };               // Both
    az_http_response_parser_get_status_line(&parser, &status_line); // Both

    az_http_response_header header = { 0 };                         // Sergey does NOT require initialization
    while (az_http_response_parser_get_next_header(&parser, &header) == AZ_OK) { // Sergey has extra parameter
        // Process header...
    }

    az_http_response_body body = { 0 };                             // Both
    az_http_response_parser_get_body(&parser, &body);               // 1st parameter is different
    // Process body...
}

sdk/core/core/src/az_http_response_parser.c
JeffreyRichter: Yes, this is a potential problem but I think it is less likely a customer will make the error because the API requires a header in order to call it. Your design does not have this requirement and so it is easy for a customer to think they can just get the body without enumerating through the headers.

sdk/core/core/inc/az_http_response_parser.h
sergey-shandar: @JeffreyRichter  there is one more difference, in Sergey's design a `header` can be a local variable inside the while loop. A user doesn't need to keep it for `az_http_response_parser_get_body` and doesn't need to keep for the next iteration.

JeffreyRichter: Code reuse is nice but very low priority. It would not be the reason for API design unless the additional code was VERY substantial.

JeffreyRichter: Yes, good point about the header being scoped in your design - I do prefer good scoping.
But, your additional parser variable is violating the scoping similar to my header variable - so I think this is a wash.

sergey-shandar: My impression was a little bit different, reusage and consistency are quite important. Including safety, which is very important in C language where it's very easy to destroy civilization with incorrect index.

sergey-shandar: @JeffreyRichter there is another very minor issue in Jeff's design but it's worth to mention to be fare. It scans the status lines twice. It doesn't change `O` unless a user does a mistake and ask the first header multiple times by providing empty header every time. It can happen with Sergey's design as well but it's more explicit, because user has to create a parser first and ask a status line.

BTW, I like the idea of reusing a header as a state, it's clever.

sergey-shandar: @JeffreyRichter I put an example into test, it's a little bit close to the real world (handling errors, checking headers, protocol and status code) https://github.com/Azure/azure-sdk-for-c/blob/sergey/http-response-parser-direct-calls/sdk/core/core/test/test_http_response_parser_example.h#L11

It's based on Sergey's design but I'm going to prepare also Jeff's design as well.

sergey-shandar: @JeffreyRichter While I was trying to implement Jeff's design, I've found one issue in this function
```c
az_http_response_get_next_header(&header)
```

Actually, it should accept `response` as well. Otherwise we may read memory which we suppose not to read because just a header doesn't have a sign of the end of response.

sergey-shandar: Please, correct me if I'm wrong. Here's the updated Jeff's design:

```c
/**
 * Get an HTTP status line.
 */
AZ_NODISCARD az_result
az_http_response_get_status_line(az_span const self, az_http_response_status_line * const out);

/**
 * Get next header HTTP header.
 *
 * @p_header has to be previous header or an empty one for the first line.
 */
AZ_NODISCARD az_result
az_http_response_get_status_get_next_header(az_span const self, az_pair * const p_header);

/**
 * Get an HTTP body.
 *
 * @p_header has to be the last header!
 */
AZ_NODISCARD az_result az_http_response_get_status_get_body(
    az_span const self,
    az_pair * const p_header,
    az_span * const body);
```

So `get_body` should accept 3 parameters. :-(

JeffreyRichter: Yes, you're aright about get_next_header. It needs the az_response_buffer. Shouldn't all these methods take an az_response_buffer instead of just any az_span?

I'd change the p_header parameter name to p_lastHeader for az_http_response_get_status_get_body since it MUST be the last header.

JeffreyRichter: Actually, az_http_response_get_status_get_body does only need 2 parameters but I think accepting 3 is OK. It does "feel like" methods that operate on the http response should take an http_response for their 1st parameter.

sergey-shandar: @jeffreychenmsft `az_response_buffer` is equal to `az_span`, AFAIK.

sergey-shandar: @JeffreyRichter I've changed `self` to `http_response`. I followed by `self` convention, when the first parameter if a subject of method signature. So `az_http_response_get_...(az_span const self, …)` means that `self` is `http_response` and operation is `get_...`

JeffreyRichter: All structures are, of course, byte arrays. And so an az_span _can_ be used for everything. However, these functions ONLY work on a buffer (az_span) that is in a certain format (an HTTP response) and so we want an abstraction (data type) that indicates this: az_http_response and then these methods REQUIRE this data type. This gives us some degree of compile-time type-safety and some API self-documentation.

JeffreyRichter: Oh, if this is a convention/guideline that we've previously agreed to, then using 'self' is fine. I was not aware or forgot that we had this as a guideline.

sergey-shandar: @JeffreyRichter no, we didn't discuss this convention yet. Don't worry. We may discuss it on a design meeting.

sergey-shandar:
@gilbertw, @antkmsft, @vhvb1989, @JeffreyRichter please review the PR. It has both implementations 
- reading an HTTP Response using a parser. Example https://github.com/Azure/azure-sdk-for-c/blob/sergey/http-response-parser-direct-calls/sdk/core/core/test/test_http_response_parser_example.h#L11
- reading an HTTP Response using `get` function and the last `header` as a parser state. Example https://github.com/Azure/azure-sdk-for-c/blob/sergey/http-response-parser-direct-calls/sdk/core/core/test/test_http_response_parser_example.h#L64

sdk/core/core/inc/az_http_response_parser.h
sergey-shandar: @JeffreyRichter one more thing I learnt when implementing Jeff's design (getters). In the future it would be harder to change it. For example, if we have to change from a buffer to any kind of stream (iterator, emitter, IOT streams whatever) then we need to redesign complete API and revert back to the same parser API where a parser is a black box with clear interface.

JeffreyRichter: I don't understand what it's your head here so I'm not sure I agree with it. Regardless, I'm not sure it matters though; any change like you're proposing is a breaking change to the customer so we either feel it's worth the change and break them or it's not.

JeffreyRichter:
I feel that the 2 are very similar now so I don’t care all that much.
You should delete line 65 from https://github.com/Azure/azure-sdk-for-c/blob/sergey/http-response-parser-direct-calls/sdk/core/core/test/test_http_response_parser_example.h#L64 which makes mine a line shorter.
I wish you didn’t have all the error checking (AZ_RETURN_IF_FAILED) in both examples as it adds no value for this discussion and takes focus away from the critical parts.
I think the main differences are:

·                     Sergey con: You introduce the concept of a parser struct.

·                     Sergey con: The parser requires a function call to initialize it and they could forget to call this.

·                     Jeff con: My header must be initialized to 0 up front; a customer could forget to do this.

·                     Jeff pro: getting the body requires a header and it is more likely that a customer will get the headers first before calling this method.

FYI: Getting the body will probably require changes because the current strategy will NOT work for downloading blobs/files. But, it does work for everything else. Right now, I’m focusing on Key Vault and so we’re fine with this for now.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://na01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fazure.microsoft.com%2Ftraining%2Fdistributed-cloud-apps%2F&data=02%7C01%7Cjeffreyr%40microsoft.com%7C948cf3366acf45f6ef8d08d62e3b3b00%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636747229263047172&sdata=sg%2FtzHWpscWQDr7foOSsaNrKDP5vNSHE3b72HJG8swU%3D&reserved=0>)

PR #175, opened 2019-11-08
Change URI encoding functions to accept span builder

antkmsft:


sdk/core/core/inc/az_mut_span.h
```diff
@@ -67,6 +67,28 @@ AZ_INLINE az_result az_mut_span_copy(
...
+    memmove((void *)buffer.begin, (void const *)src.begin, src.size);
+  }
+
+  out_result->begin = buffer.begin;
+  out_result->size = src.size;
+
+  return AZ_OK;
+}
+
+
```
antkmsft:
```suggestion
```

sdk/core/core/inc/az_result.h
```diff
@@ -34,6 +34,9 @@ enum {
     } \
   } while (0)
 
+#define AZ_IGNORE_RESULT(exp) \
+  if (exp) {};
```
sergey-shandar: It should be `do { … } while(0)`

sdk/core/core/src/az_curl_adapter.c
```diff
@@ -229,7 +229,7 @@ AZ_NODISCARD az_result setup_url(CURL * const p_curl, az_http_request_builder co
   }
 
   // free used buffer before anything else
-  az_mut_span_set(writable_buffer, 0);
+  AZ_IGNORE_RESULT(az_mut_span_memset(writable_buffer, 0));
```
sergey-shandar: IMHO, we should either check the result or change the signature to `void az_mut_span_memset`

sdk/core/core/inc/az_result.h
antkmsft: I don't think it will silence the GCC warning, but I can give it a try.
The reason I am introducing this macro is that on our current GCC version, and this is likely a bug, when you have AZ_NODISCARD. and  silence it with `(void)fn();`, it gives a warning.

sdk/core/core/src/az_curl_adapter.c
antkmsft: I don't think so. There could be other contexts when it could be important to know that the function call is succeeded, and while some part of me want elegant code, some other part of me feels that making it void could be similar to adjusting the results to the desired outcome.

sdk/core/core/inc/az_result.h
antkmsft:
```suggestion
  do { exp; } while (0);
```

sdk/core/core/inc/az_result.h
```diff
@@ -34,6 +34,9 @@ enum {
     } \
   } while (0)
 
+#define AZ_IGNORE_RESULT(exp) \
+  do { exp; } while (0);
```
antkmsft:
```suggestion
  if (exp) {}
```

sdk/core/core/src/az_curl_adapter.c
sergey-shandar: Ok for now, I think we need to discuss pure functions, or non-throwable functions and default initialization to make our code more deterministic.

PR #180, opened 2019-11-12
Size bug fix for auth

antkmsft:


sdk/core/core/src/az_auth.c
```diff
@@ -125,7 +125,7 @@ AZ_NODISCARD az_result az_auth_get_token(
     AZ_RETURN_IF_FAILED(az_span_builder_append(&builder, auth_url2));
 
     auth_url = az_span_builder_result(&builder);
-    builder = az_span_builder_create(az_mut_span_take(response_buf, auth_url.size));
+    builder = az_span_builder_create(az_mut_span_drop(response_buf, auth_url.size));
```
gilbertw: Can you add test cases for this

PR #201, opened 2019-11-19
Polymorphic JSON builder.

sergey-shandar:


sdk/core/core/src/az_json_builder.c
```diff
@@ -0,0 +1,116 @@
...
+// SPDX-License-Identifier: MIT
+
+#include <az_json_builder.h>
+
+#include <az_str.h>
+
+#include <_az_cfg.h>
+
+AZ_NODISCARD az_result
+az_json_builder_init(az_json_builder * const out, az_span_action const write) {
```
JeffreyRichter: I don't understand that use/value of az_span_action. What is the purpose of this?

JeffreyRichter: Also, I think *_init takes a mutable span as an argument and associates it with the az_json_builder.

sdk/core/core/src/az_json_builder.c
```diff
@@ -0,0 +1,116 @@
...
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD static az_result
+az_json_builder_write_str(az_json_builder * const self, az_span const value) {
+  AZ_CONTRACT_ARG_NOT_NULL(self);
+
+  az_span_action const write = self->write;
+  AZ_RETURN_IF_FAILED(az_span_action_do(write, AZ_STR("\"")));
```
JeffreyRichter: Maybe az_span_action should take varargs and then we could call it like this (1 call instead of 3):
az_span_action(write, AZ_STR("\""), value, AZ_STR("\""));

sdk/core/core/src/az_json_builder.c
```diff
@@ -0,0 +1,116 @@
...
+      self->need_comma = true;
+      return az_json_builder_write_str(self, value.data.string);
+    }
+    case AZ_JSON_VALUE_NUMBER: {
+      self->need_comma = true;
+      // TODO:
+      return AZ_ERROR_NOT_IMPLEMENTED;
+    }
+    case AZ_JSON_VALUE_OBJECT: {
+      self->need_comma = false;
```
JeffreyRichter: If I understand the 'need_comma' thing right, I think you always want a comma unless the last character in the span is a { or a [, right? If so, kill the 'need_comma' field and just look at the last character in the span; this will also kill all the code that sets the field.

sdk/core/core/src/az_json_builder.c
```diff
@@ -0,0 +1,116 @@
...
+      self->need_comma = true;
+      return az_span_action_do(write, AZ_STR("null"));
+    }
+    case AZ_JSON_VALUE_BOOLEAN: {
+      self->need_comma = true;
+      return az_span_action_do(write, value.data.boolean ? AZ_STR("true") : AZ_STR("false"));
+    }
+    case AZ_JSON_VALUE_STRING: {
+      self->need_comma = true;
+      return az_json_builder_write_str(self, value.data.string);
```
JeffreyRichter: I think you want this to be your "master" function. That is, this is the function that writes to the json_builder. And, then you have convenience methods that call this method. So, put in this case the code to write the string to the json_builder and have the convenient az_json_builder_write_str function call this function.

sdk/core/core/test/test_json_builder.c
```diff
@@ -0,0 +1,56 @@
...
+  {
+    AZ_RETURN_IF_FAILED(
+        az_json_builder_write_object_member(&builder, AZ_STR("foo"), az_json_value_create_array()));
+    AZ_RETURN_IF_FAILED(
+        az_json_builder_write_array_item(&builder, az_json_value_create_string(AZ_STR("bar"))));
+    AZ_RETURN_IF_FAILED(
+        az_json_builder_write_array_item(&builder, az_json_value_create_null()));
+    AZ_RETURN_IF_FAILED(az_json_builder_write_array_close(&builder));
+  }
+
```
JeffreyRichter: Here's an idea but I'm not sure it will completely work:
az_json_builder_write_object_member(&builder, 
   AZ_STR("name"), az_json_value_create_boolean(true)));

// *_array is a varargs function; it automatically adds the [ and the ] at the beginning and end; I just need to list the array items
az_json_builder_write_array(&builder, AZ_STR("foo"), 
   az_json_value_create_string(AZ_STR("bar"))
   az_json_value_create_null())

// We could do the same thing for an object 

As I write this, I see that it only works for scalars and not for items that may be complex objects but maybe there is some idea here that could work for common cases. Maybe if we did something like this:

az_json_builder_write(&builder, AZ_STR("foo"), 
   az_json_array_start(),
   az_json_value_create_string(AZ_STR("bar"))
   az_json_value_create_null(),
   az_json_array_end())

sdk/core/core/src/az_json_builder.c
sergey-shandar: In some modern languages, like C#, JavaScript, varargs is safe. But in C, it's a very error-prone. I would avoid as much as I can. In the same time, to make polymorphic functions faster, we can accept a span of spans instead of one span.

sdk/core/core/test/test_json_builder.c
sergey-shandar: I'm ok to add some utility functions that will work as DOM. In the same time, I would like to keep current functions for streaming as well.

sdk/core/core/src/az_json_builder.c
sergey-shandar: Currently, I'm using polymorphic writer which can't return back the last character. I may use it if I use `az_span_builder` instead of `write` but even in this case we would like to keep `az_span_builder` as a black box with limited set of functions.

Issue #216, opened 2019-11-21
[FEATURE REQ] Clear distinguish between public and private API.


sergey-shandar:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [ ] Description Added
- [ ] Expected solution specified

barcharcraz:
Can you fill in additional information about this request please? It's really hard to take action on an issue with just the default template.

sergey-shandar:
@barcharcraz I think, we discussed this one in our meeting. We should use `_az_` for internal things (functions, constants, etc.).

gilbertw:
Internal is distinguished with _az_

JeffreyRichter:
We agreed to _az_* (leading underscore)

PR #221, opened 2019-11-22
AZ_JSON_VALUE_SPAN

sergey-shandar:
AZ_JSON_VALUE_SPAN allows to write any string as a JSON string.


sergey-shandar:
Issue #198

sdk/core/core/inc/az_json_string.h
```diff
@@ -15,6 +15,24 @@
...
+/**
+ * Encodes the given character into a JSON escape sequence. The function returns an empty span if
+ * the given character doesn't require to be escaped.
+ */
+AZ_NODISCARD az_span az_json_esc_encode(az_result_byte const c);
+
+enum {
+  AZ_HEX_LOWER_OFFSET = 'a' - 10,
+  AZ_HEX_UPPER_OFFSET = 'A' - 10,
+};
```
antkmsft: ???
Where is the lower offset used? what does 'a' - 10 is 'W', 'A' - 10 is '7', what??? :)
Of course I understand how it ends up working, but it is confusing and it is not the most straightforward way to implement.
Enum is public, but it is not useful.
I would suggest to implement in in a way I implemented it in base64.c:
```
AZ_NODISCARD AZ_INLINE uint8_t int4_to_hex(uint8_t const int4) {
  assert(int4 <= 0x0F);
  return (int4 < 10) ? ('0' + int4) : ('A' + (int4 - 10));
}
```

sdk/core/core/inc/az_json_value.h
```diff
@@ -18,6 +18,9 @@ typedef enum {
   AZ_JSON_VALUE_STRING = 4,
   AZ_JSON_VALUE_OBJECT = 5,
   AZ_JSON_VALUE_ARRAY = 6,
+  // A special case for non-JSON strings. This field is used to serialize `az_span` into JSON
+  // string. Currently, our JSON parser doesn't return this kind of values.
```
antkmsft: Should we call it 'non-encoded/non-escaped string'?

sdk/core/core/inc/az_json_string.h
sergey-shandar: @antkmsft, @gilbertw, @vhvb1989 I have conflicting proposals here. There are several problems:
1. If inside inline function I would like to use something that is private, how can I do this? I think, the answer should be provided be in this issue #216 
2. premature optimization, when we use constants assuming that they are better then a constant expression, like `('A' - 10)`.

sdk/core/core/inc/az_json_string.h
```diff
@@ -15,6 +15,24 @@
...
+
+/**
+ * @digit should be in the range [0..15].
+ */
+AZ_NODISCARD AZ_INLINE uint8_t az_digit_to_upper_hex(uint8_t const digit) {
```
antkmsft: If this function is public, the name is a bit confusing, I think, `int4 to upper hex` describes it better.
And also if this is public, we probably should make 'to lower hex' version as well.

sdk/core/core/inc/az_json_string.h
antkmsft: My opinion: based on az_assumptions.c, 'A', is a number, it is just another way to write 65*. I think there should be no constant in this case, the same reason you con't create constants like ONE = 1, TWO = 2.
We would make 1 a constant i one case - if it is a point of configuration, something that we may change.
In the context of converting to hex, it is '0'..'9' 'A' ..'F'/'a'..'f' by definition, it is never going to change on planet earth. Maybe if Martians have their alphabet arranged in different order - then yes, but not until then.
--
* - As to why we only support compiling on ASCII-order for latin alphabet: we assume/only work correctly on ASCII/UTF8, because all the JSON responses we are going to get from azure are going to be ASCII/UTF8; and if we want to compile on EBCDIC, for example - there will be far more places you would need to make changes and write additional code, other than to remap 'A' to something else. Plus, EBCDIC is long dead, and UTF is here to stay, so probably will hever gonna happen

sdk/core/core/inc/az_json_string.h
sergey-shandar: I've renamed it to `az_number_to_upper_hex`.

sdk/core/core/src/az_json_builder.c
```diff
@@ -28,6 +31,80 @@ AZ_NODISCARD static az_result az_json_builder_write_str(
   return AZ_OK;
 }
 
+typedef struct {
+  az_span_action write;
+  az_span value;
+  size_t begin;
+  size_t i;
```
antkmsft: I personally don't like the name `i` in this context. `pos`, `position`, `idx`, `index` is better, IMO.

antkmsft: And also given that `begin` is not a pointer, something makes me want to see it named as `begin_something`, such as `begin_pos`, or at least `start`

sdk/core/core/src/az_json_builder.c
```diff
@@ -28,6 +31,80 @@ AZ_NODISCARD static az_result az_json_builder_write_str(
...
+    // no need to escape, process to the next character.
+    state.i += 1;
+  }
+  AZ_RETURN_IF_FAILED(az_json_value_span_state_flush(&state));
+  AZ_RETURN_IF_FAILED(az_span_action_do(state.write, AZ_STR("\"")));
+  return AZ_OK;
+}
+
 // 2^53 - 1
 #define AZ_JSON_UINT_MAX 0x1FFFFFFFFFFFFFull
```
antkmsft: unrelated, but I think that using lowercase for long constant will eventually make a problem when we have a constant of `10l`

antkmsft:
@sergey-shandar , can you follow up on the PR feedback?

sdk/core/core/inc/az_json_value.h
sergey-shandar: In this case, I would prefer to rename `AZ_JSON_VALUE_STRING` as well.

sdk/core/core/src/az_json_builder.c
sergey-shandar: I will keep it as is for now because it's an internal structure.

sdk/core/core/src/az_json_builder.c
sergey-shandar: The only problem is uppercase hex letters `0xFFFFULL`.

PR #244, opened 2019-12-03
AZ_SPAN_NULL => az_span_create_empty()

sergey-shandar:
- rename `AZ_SPAN_NULL` to `az_span_create_empty()`. Please, avoid static variables, even if they are declared as `const`.
- internal constants `AZ_HEX_...` => `_az_HEX_...`.
- Filename rename `..._POC` to `..._poc1`. We shouldn't use upper case letters in C file names.


sdk/core/core/inc/az_hex.h
```diff
@@ -9,15 +9,15 @@
...
-  AZ_HEX_LOWER_OFFSET = 'a' - 10,
-  AZ_HEX_UPPER_OFFSET = 'A' - 10,
+  _az_HEX_LOWER_OFFSET = 'a' - 10,
+  _az_HEX_UPPER_OFFSET = 'A' - 10,
 };
 
 /**
  * Converts a number [0..15] into uppercase hexadecimal digit character (base16).
  */
 AZ_NODISCARD AZ_INLINE uint8_t az_number_to_upper_hex(uint8_t const number) {
```
barcharcraz: why not just use snprintf?

sdk/core/core/inc/az_span.h
```diff
@@ -25,7 +25,7 @@ typedef struct {
 
 typedef int32_t az_result_byte;
 
-extern az_span const AZ_SPAN_NULL;
+AZ_NODISCARD AZ_INLINE az_span az_span_create_empty() { return (az_span){ 0 }; }
```
barcharcraz: probably not needed, just have people write the zero span.

sdk/core/core/src/az_json_string.c
```diff
@@ -13,10 +13,10 @@ AZ_NODISCARD AZ_INLINE az_result_byte az_hex_to_digit(az_result_byte const c) {
     return c - '0';
   }
   if ('a' <= c && c <= 'f') {
-    return c - AZ_HEX_LOWER_OFFSET;
```
barcharcraz: again, just use a builtin function here

sdk/core/core/inc/az_hex.h
sergey-shandar: No way :-)

sdk/core/core/src/az_json_string.c
sergey-shandar: I've tried but it fails in VS if `c` is less than -1.

sdk/core/core/inc/az_span.h
JeffreyRichter: I really like AZ_SPAN_NULL; What if this were a define like NULL (since you don't like global const)?

sergey-shandar: @JeffreyRichter as far as I understood, there are two questions:
1. null vs empty. We are already have an empty span and currently it's a span that has `size == 0`. For example, we have a function `az_span_is_empty()`. Usually, NULL applies for pointers. A span is not a pointer (like C# string) but it's a pair of begin/end values. For example, a C++ range or a Rust slice don't have NULL range. We can introduce a special value but I'm afraid it can be similar to a Tony Hoare's billion-dollar mistake.
2. Use #define instead of inline function. As far as I remember, we should use inline functions and enum values where we can instead of macros. @barcharcraz may comment on it.

sergey-shandar: @barcharcraz Usually, we use `{ 0 }` to initialize it to a deterministic value, for example, before calling `_init` method. We also are trying to design our types that `{ 0 }` means a default valid value. But I'm ok with having empty/null constants/inline functions if people would like to use it to clearly specify a meaning of the value, even if the value is `{ 0 }`.

sergey-shandar: BTW, new C# Span is a value type and can't be NULL https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=netcore-3.0

antkmsft: In case we wanted a constant, I'd prefer `(az_span){ 0 }` to `AZ_SPAN_NULL`

JeffreyRichter:
What I think I want is a way to see in code something like this:
if (some_span == AZ_SPAN_NULL) { … }

I just think is very readable and it logically says if no SPAN is passed.
If you don’t like the use of NULL here, then I can accept another name, like AZ_SPAN_NONE.
I’d prefer NOT to have a function that creates an empty span.
I also don’t think we need functions like
bool is_empty(az_span span)
Or
Bool is_none(az_span span)

But I do think having a function that returns the size is useful:
int az_span_get_size(az_span span)

and the result of this can be compared to 0 for an empty span (which an AZ_SPAN_NONE [or AZ_SPAN_NULL] span is).


sergey-shandar:
@JeffreyRichter C doesn't allow to define `==` operator for structures :-(

JeffreyRichter:
Good point. That is a shame.

PR #257, opened 2019-12-10
Detect resource URL when getting token

antkmsft:
Detect resource URL when getting token. Warning: using the same credential object for different resources may defeat the whole purpose of token caching, and may result in a number of get-token requests. Other than that, it is safe. Don't do this. Don't use the same token credential for different clients. This change essentially makes it easy for you to do that :-/ With older design, you'd have to be super intentional/creative to defeat the token caching.


sdk/core/core/src/az_client_secret_credential.c
```diff
@@ -219,21 +224,80 @@ _az_client_secret_credential_ensure_token_credential(az_client_secret_credential
...
+    AZ_RETURN_IF_FAILED(_az_client_secret_credential_ensure_token_credential(self));
+  } else {
+    resource_builder = az_span_builder_create(
+        (az_mut_span)AZ_SPAN_FROM_ARRAY(self->_token_credential._resource_buf));
+
+    AZ_RETURN_IF_FAILED(az_span_builder_append(&resource_builder, request_resource));
+    self->_token_credential._resource_size = az_span_builder_result(&resource_builder).size;
+
+    AZ_RETURN_IF_FAILED(_az_client_secret_credential_renew_token_credential(self));
+  }
```
schaabs: Without any mutex around this it introduces race conditions if a credential is used by different clients with different resource strings. If different clients are attempting to retrieve tokens for different resources it's possible that one would get a token for the wrong resource. So it would require any callers to call this inside a critical section. However, this becomes a moot point if we refactor as suggested in the comments on https://github.com/Azure/azure-sdk-for-c/pull/250.

antkmsft: I'll complete this PR so that it goes to my branch with auth refactoring, and then change that branch according to your suggestions.

PR #258, opened 2019-12-10
Fix extra Clang warnings

antkmsft:
https://github.com/Azure/azure-sdk-for-c/issues/256

### DIFF
--- a/sdk/core/core/src/az_base64.c
+++ b/sdk/core/core/src/az_base64.c
@@ -150,7 +150,7 @@ AZ_NODISCARD az_result az_base64_encode(
           octet2 = input.begin[oct + TRIBYTE_OCTET2_INDEX];
 
           buffer.begin[sxt + TRIBYTE_SEXTET3_INDEX]
-              = uint6_as_base64(base64url, octet2 & TRIBYTE_OCTET2_SEXTET3_MASK);
+              = uint6_as_base64(base64url, (uint8_t)(octet2 & TRIBYTE_OCTET2_SEXTET3_MASK));
         } else {
           buffer.begin[sxt + TRIBYTE_SEXTET3_INDEX] = BASE64_PADDING_CHAR;
         }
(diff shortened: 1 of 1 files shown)


sdk/core/core/src/az_base64.c
```diff
@@ -160,35 +160,36 @@ AZ_NODISCARD az_result az_base64_encode(
       buffer.begin[sxt + TRIBYTE_SEXTET2_INDEX] = (oct + TRIBYTE_OCTET1_INDEX < input.size)
           ? uint6_as_base64(
               base64url,
-              ((octet1 & TRIBYTE_OCTET1_SEXTET2_MASK) << TRIBYTE_OCTET1_SEXTET2_LSHIFT)
-                  | ((octet2 & TRIBYTE_OCTET2_SEXTET2_MASK) >> TRIBYTE_OCTET2_SEXTET2_RSHIFT))
+              (uint8_t)(
```
gilbertw: It feels strange to cast to a unit8_t then call a method which is named uint6.  Also why call it uint6_.... when the parameter passed in is a uint8_t?

antkmsft: There isn't `uint6_t`, so we use uint8_t type for parameters. We discussed it with Sergey back when I was writing it, he suggested the name and I agree with the reasoning. Even though there isn't a 6 bit integer type, the function name suggests it only operates on the values from 0 to 0x3F (because base64 encoding essentially splits uint8 values into uint6 values).

antkmsft: Another name would be `sextet` (6 bits) as opposed to `octet` (8 bits) but we thought `uint6` is bit more understandable in this context... No? This is actually interesting feedback, let me know if you'd prefer the name to be `sextet_as_base64()`. We then can collect more opinions across the room, what name is more understandable... Because at a time, there were just two of us and we both agreed on `int6`, so I didn't bother to ask for another opinions.

gilbertw: Its a NIT.  I can live with the current naming since there isn't anything better.

antkmsft:
https://github.com/Azure/azure-sdk-for-c/pull/266

PR #266, opened 2019-12-12
Clang warnings

antkmsft:
@danieljurek, I can no longer see my commit in your branch - I think it was lost after a force push. This PR is the same as previous one, only that it is now based on the most recently rebased commit history of your branch.


antkmsft:
https://github.com/Azure/azure-sdk-for-c/pull/252

antkmsft:
https://github.com/Azure/azure-sdk-for-c/issues/256

PR #298, opened 2020-01-09
Fix local MSVC warnings

antkmsft:


sdk/core/core/inc/_az_cfg_test_include_prefix.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
sergey-shandar: IMHO, we should use the same `_az_cfg_prefix.h` everywhere.

PR #299, opened 2020-01-09
Fix core and test warning for MSVC

sergey-shandar:


sdk/core/core/test/main.c
```diff
@@ -660,7 +660,7 @@ int main() {
     TEST_ASSERT(az_span_is_equal(az_span_builder_result(&builder), uri_decoded));
   }
   {
-    int16_t const url_max = 100;
+    // int16_t const url_max = 100;
```
gilbertw: remove rather than comment

sdk/core/core/CMakeLists.txt
```diff
@@ -73,7 +73,10 @@ add_executable (
     test/test_json_builder.c
     test/test_json_get_by_pointer.c
     test/test_json_pointer.c
-    test/test_json_string.c  
+    test/test_json_string.c
+    test/test_pair_span.c
```
barcharcraz: this seems like an unrelated change.

PR #320, opened 2020-01-29
Retry policy

antkmsft:
The logic is mostly taken from https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/azcore/policy_retry.go;
I also did look at https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs as another point of reference.


antkmsft:
@JeffreyRichter , my main inspiration was the Go retry logic (https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/azcore/policy_retry.go)

sdk/core/core/inc/az_http_policy_retry_options.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+typedef struct {
+  int16_t max_tries;
+  int32_t retry_delay_msec;
+  int32_t max_retry_delay_msec;
+} az_http_policy_retry_options;
+
+AZ_INLINE az_http_policy_retry_options az_http_policy_retry_options_create() {
+  return (az_http_policy_retry_options){
+    .max_tries = 4,
```
JeffreyRichter: You should have a define for '4' too up top

sdk/core/core/inc/az_pal.h
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_PAL_H
+#define _az_PAL_H
+
+#include <stdint.h>
+
+#include <_az_cfg_prefix.h>
+
+void az_pal_wait(int32_t milliseconds);
```
JeffreyRichter: Let's rename this to az_pal_sleep to be more similar with Windows & pthreads.

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
     void * const data,
     az_http_request_builder * const hrb,
     az_http_response * const response) {
-  (void)data;
-
   // reset response to be written from the start
-  AZ_RETURN_IF_FAILED(az_http_response_reset(response));
+  if (data == NULL) {
```
JeffreyRichter: data should never be NULL so you can remove this test

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+    return az_http_pipeline_nextpolicy(p_policies, hrb, response);
+  }
 
-  return az_http_pipeline_nextpolicy(p_policies, hrb, response);
+  az_http_policy_retry_options const * const retry_policy
+      = (az_http_policy_retry_options const *)data;
+
+  int16_t const max_tries = retry_policy->max_tries;
+
+  bool const custom_retry_delay_limits = retry_policy->retry_delay_msec > 0
```
JeffreyRichter: It would be better to validate this when the options are passed to initialize a client and then fail client creation instead of silently ignoring the options while requests are under way and using values the user did NOT specify.

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+
+  bool const should_log = az_log_should_write(AZ_LOG_HTTP_RETRY);
+  az_result result = AZ_OK;
+  int16_t attempt = 1;
+  while (true) {
+    AZ_RETURN_IF_FAILED(az_http_response_reset(response));
+    result = az_http_pipeline_nextpolicy(p_policies, hrb, response);
+
+    // Even HTTP 429, or 502 are expected to be AZ_OK, so the failed result is not retriable.
+    if (attempt > max_tries || az_failed(result) || response == NULL
```
JeffreyRichter: I'm very sure that response can never be NULL.

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+        case 408:
+        case 429:
+        case 500:
+        case 502:
+        case 503:
+        case 504: {
+          // Try to get the value of retry-after header, if there's one.
+          az_pair header = { 0 };
+          while (az_http_response_parser_read_header(&response_parser, &header) == AZ_OK) {
+            if (az_span_is_equal(header.key, AZ_STR("retry-after-ms"))
```
JeffreyRichter: These comparisons need to be case-insensitive. I suggest putting the code to find the headers in a separate private function to make the retry logic cleaner. This code tends to have bugs and be hard to follow. Breaking it into discreet chunks helps with code maintainability a lot.

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+              if (az_succeeded(az_span_to_uint64(header.value, &msec))) {
+                retry_after_msec
+                    = msec < INT32_MAX ? (int32_t)msec : INT32_MAX; // int32_t max == ~24 days
+                break;
+              }
+            } else if (az_span_is_equal(header.key, AZ_STR("Retry-After"))) {
+              // The vaule is either seconds or date.
+              uint64_t seconds;
+              if (az_succeeded(az_span_to_uint64(header.value, &seconds))) {
+                retry_after_msec = seconds < INT32_MAX / 1000 ? (int32_t)seconds * 1000 : INT32_MAX;
```
JeffreyRichter: 1000 is a magic number; move it to an enum value at the top of this file.

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+      }
+    }
+
+    ++attempt;
+
+    if (retry_after_msec < 0) { // there wasn't any kind of "retry-after" response header
+      int32_t const exponential_retry_after = (int32_t)(
+          (double)retry_delay_msec
+          * (double)(attempt <= 30 ? (1 << attempt) : INT32_MAX) // scale exponentially
+          * (((1.0 / RAND_MAX) * rand()) / 2 + 0.8)); // add jitter factor from 0.8 to 1.3
```
JeffreyRichter: rand may require the pal layer

sdk/core/core/src/az_http_policy.c
```diff
@@ -61,12 +65,110 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
...
+
+      retry_after_msec = exponential_retry_after > max_retry_delay_msec ? max_retry_delay_msec
```
JeffreyRichter: I also suggest moving all the code that calculates the delay into a separate private function.

sdk/core/core/src/az_log.c
```diff
@@ -176,6 +176,28 @@ static az_result _az_log_http_response_msg(
...
+  AZ_RETURN_IF_FAILED(az_span_builder_append(log_msg_bldr, AZ_STR("HTTP Retry attempt")));
+
+  if (attempt >= 0) {
+    AZ_RETURN_IF_FAILED(az_span_builder_append(log_msg_bldr, AZ_STR(" #")));
+    AZ_RETURN_IF_FAILED(az_span_builder_append_uint64(log_msg_bldr, (uint64_t)attempt));
+  }
+
+  AZ_RETURN_IF_FAILED(az_span_builder_append(log_msg_bldr, AZ_STR(" will be made")));
+
+  if (delay_msec >= 0) {
```
JeffreyRichter: Take this if out. If we're logging retry logic then let's see the 0msec log entry too.

sdk/core/core/src/az_log.c
```diff
@@ -176,6 +176,28 @@ static az_result _az_log_http_response_msg(
...
+    AZ_RETURN_IF_FAILED(az_span_builder_append(log_msg_bldr, AZ_STR(" in ")));
+    AZ_RETURN_IF_FAILED(az_span_builder_append_uint64(log_msg_bldr, (uint64_t)delay_msec));
```
JeffreyRichter: We have a brand new append_double method now in the new az_span.h file. Use this instead of casting the double to a uint64_t.

sdk/core/core/src/az_log.c
```diff
@@ -176,6 +176,28 @@ static az_result _az_log_http_response_msg(
...
+    AZ_RETURN_IF_FAILED(az_span_builder_append(log_msg_bldr, AZ_STR("ms")));
+  }
+
+  return az_span_builder_append_byte(log_msg_bldr, '.');
```
JeffreyRichter: Remove this and add the '.' to the previous 'ms' string.

sdk/core/core/src/az_log_private.h
```diff
@@ -19,6 +19,8 @@ void _az_log_http_response(
     uint64_t const duration_msec,
     az_http_request_builder const * const hrb);
 
+void _az_log_http_retry(int16_t const attempt, int32_t const delay_msec);
```
JeffreyRichter: I don't think we need this in any header file if it's just called from the retry policy. Maybe we want to move the retry policy and all its supporting functions to its own .c file. Then, only the retry policy function and its options struct need to be in a header file.

sdk/core/core/src/az_pal.c
```diff
@@ -0,0 +1,23 @@
...
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+#include <_az_cfg.h>
+
+#ifdef _WIN32
+void az_pal_wait(int32_t milliseconds) {
```
JeffreyRichter: This can't stay like this. Each pal implementation MUST be inits own .c file. We can wait for Charlie  to get back to us.

sdk/core/core/src/az_http_policy.c
antkmsft: It is a number of milliseconds in a second.

sdk/keyvault/keyvault/src/az_keyvault_client.c
```diff
@@ -33,8 +33,9 @@ AZ_NODISCARD AZ_INLINE az_span az_keyvault_client_constant_for_application_json(
 az_keyvault_keys_client_options const AZ_KEYVAULT_CLIENT_DEFAULT_OPTIONS
     = { .service_version = AZ_CONST_STR("7.0"),
```
JeffreyRichter: Rick was supposed to remove .service_version. Can you follow up with him?

sdk/core/core/src/az_pal.c
antkmsft: I know, this is temporary, until we have a solution from Charlie.

JeffreyRichter:
So make a private constant int variable (ms_per_sec) for this then.


sdk/core/core/src/az_log.c
antkmsft: Will do; but this is not double.

sdk/core/core/src/az_log.c
antkmsft: The idea was to print "HTTP Retry attempt( #3) will be made( in 3000ms).", so that we raher won't print non-positive values, because the behavior is unknown in this case, so we don't tell what we don't know. Maybe it is better to simplify.

sdk/core/core/src/az_log_private.h
antkmsft: I think it is more convenient to keep logging-related private functions in a separate file so that they could reuse/share code/formatting parts if necessary.

sdk/keyvault/keyvault/src/az_keyvault_client.c
antkmsft: Will do.

sdk/core/core/src/az_log.c
antkmsft: I'll also see if I can improve this code so that the cast is not necessary. Currently, it is technically not required here, but it would be required if that was a negative value.

sdk/core/core/src/az_log_private.h
JeffreyRichter: No, this is not the right thing to do. It is better to keep all retry-related logic together.
Logging is a cross-cutting concern; it may appear anywhere and everywhere in the code base. We should NOT have to edit a logging .c file to add a log entry to a new classification somewhere else in the code base.

sdk/keyvault/keyvault/src/az_keyvault_client.c
antkmsft: @gilbertw , can I remove the service version?

gilbertw: I fixed the service version for storage.  I'll submit a fix for this in KeyVault

gilbertw: @antkmsft , https://github.com/Azure/azure-sdk-for-c/pull/322

gilbertw: I've submitted a change to master which addresses this.

sdk/core/core/internal/az_http_policy_retry_options_internal.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef _az_HTTP_POLICY_RETRY_OPTIONS_INTERNAL_H
+#define _az_HTTP_POLICY_RETRY_OPTIONS_INTERNAL_H
+
+#include <az_http_policy_retry_options.h>
+
+#include <stdbool.h>
+
+#include <_az_cfg_prefix.h>
+
+AZ_NODISCARD bool az_http_policy_retry_options_validate(
```
gilbertw: _ private function
I don't expect customers to call this function

sdk/core/core/src/az_http_policy.c
```diff
@@ -37,6 +32,14 @@ AZ_NODISCARD AZ_INLINE az_result az_http_pipeline_nextpolicy(
   return p_policies[0].pfnc_process(&(p_policies[1]), p_policies[0].data, hrb, response);
 }
 
+// Non-inlinable version to create function pointers to
+static AZ_NODISCARD az_result _az_http_pipeline_nextpolicy(
+    az_http_policy * const p_policies,
+    az_http_request_builder * const hrb,
+    az_http_response * const response) {
+  return az_http_pipeline_nextpolicy(p_policies, hrb, response);
```
gilbertw: az_http_pipeline_nextpolicy is an internal function.  It should be named with a leading _

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,213 @@
...
+
+az_http_policy_retry_options az_http_policy_retry_options_create() {
+  return (az_http_policy_retry_options){
+    .max_tries = _az_HTTP_POLICY_RETRY_OPTIONS_DEFAULT_MAX_TRIES,
+    .retry_delay_msec = _az_HTTP_POLICY_RETRY_OPTIONS_DEFAULT_RETRY_DELAY_MSEC,
+    .max_retry_delay_msec = _az_HTTP_POLICY_RETRY_OPTIONS_DEFAULT_MAX_RETRY_DELAY_MSEC,
+  };
+}
+
+AZ_NODISCARD bool az_http_policy_retry_options_validate(
```
gilbertw: _az_http_policy_retry_options_validate

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,213 @@
...
+  bool const should_log = az_log_should_write(AZ_LOG_HTTP_RETRY);
+  az_result result = AZ_OK;
+  int16_t attempt = 1;
+  while (true) {
+    AZ_RETURN_IF_FAILED(az_http_response_reset(response));
+    AZ_RETURN_IF_FAILED(az_http_request_builder_remove_retry_headers(hrb));
+    result = next_http_policy(p_policies, hrb, response);
+
+    // Even HTTP 429, or 502 are expected to be AZ_OK, so the failed result is not retriable.
+    if (attempt > max_tries || az_failed(result) || response->builder.length == 0) {
```
gilbertw: This function has a bug.  It calls next_http_policy max_tries+1 before failing.  It should return if attempt >= max_tries

sdk/identity/identity/src/az_identity_client_secret_credential.c
```diff
@@ -85,10 +87,16 @@ _az_identity_client_secret_credential_ms_oauth2_send_get_token_request(
   AZ_RETURN_IF_FAILED(az_http_request_builder_init(
       &hrb, hrb_buf, (uint16_t)auth_url.size, AZ_HTTP_METHOD_VERB_POST, auth_url, auth_body));
 
+  static az_http_policy_retry_options default_retry_policy = {
```
gilbertw: auth_retry_policy_options is a better name for this structure.  Its not default because its specific to the auth pipeline

joshfree:
@antkmsft what works remain to get this PR merged for the March milestone (Fri Mar 6)?

antkmsft:
> @antkmsft what works remain to get this PR merged for the March milestone (Fri Mar 6)?

@joshfree , I am working on this now, I need to adapt this logic to the refactored core that we have now. I should complete it by EOW; we'll cetainly meet the Mar 6 deadline.

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
```diff
@@ -26,8 +27,9 @@ enum {
 
 az_storage_blobs_blob_client_options const AZ_STORAGE_BLOBS_BLOB_CLIENT_DEFAULT_OPTIONS
     = { .retry = {
-            .max_retry = 5,
-            .delay_in_ms = 1000,
+            .max_tries = 5,
```
gilbertw: max_retries

gilbertw: FWIW there doesn't seem to be a standard terminology for the retry policies.  Unfortunate that each language has its own terminology for what is the same thing @JeffreyRichter

.NET
```
private readonly TimeSpan _delay;
private readonly TimeSpan _maxDelay;
private readonly int _maxRetries;
```
JS
```
retryCount?: number,
retryInterval?: number,
maxRetryInterval?: number
```
Python
```
'total': options.pop("retry_total", self.total_retries),
'connect': options.pop("retry_connect", self.connect_retries),
'read': options.pop("retry_read", self.read_retries),
'status': options.pop("retry_status", self.status_retries),
'backoff': options.pop("retry_backoff_factor", self.backoff_factor),
'max_backoff': options.pop("retry_backoff_max", self.BACKOFF_MAX),
'methods': options.pop("retry_on_methods", self._method_whitelist),
'history': []
```
```
    /**
     * Max number of retry attempts to be make.
     *
     * @return The max number of retry attempts.
     */
    int getMaxRetries();

    /**
     * Computes the delay between each retry.
     *
     * @param retryAttempts The number of retry attempts completed so far.
     * @return The delay duration before the next retry.
     */
    Duration calculateRetryDelay(int retryAttempts);
```

JeffreyRichter: Yes, the inconsistency is unfortunate. I think we should use:
max_retries
retry_delay_msec
max_retry_delay_msec

sdk/core/core/internal/az_clock_internal.h
```diff
@@ -13,7 +15,7 @@ AZ_INLINE uint64_t _az_clock_msec() {
   // Convert clock_t to milliseconds
   // Floating point arithmetic is used to cover CLOCKS_PER_SEC all values
   // including 12000, 2000, 1500, 300, 500, 100, 2
-  return (uint64_t)(clock() / (CLOCKS_PER_SEC / 1000.0));
+  return (uint64_t)(clock() / (CLOCKS_PER_SEC / (double)_az_TIME_MILLISECONDS_PER_SECOND));
```
CIPop: I've seen one libc for ARM where clock() wasn't properly connected to the sys-tick timer (always 0xFF..FF) - (Might have been libc from Arm GCC and FreeRTOS.)
Documenting the C99 requirement for time.h might be sufficient instead of extracting it to a PAL.

CIPop: On second thought - clock() may not have sufficient resolution for what you're using it.
Also, forcing uint64_t on an 8bit MCU is not a good idea.

sdk/core/core/src/az_http_policy_logging.c
```diff
@@ -0,0 +1,182 @@
...
+#include <az_str.h>
+
+#include <stdbool.h>
+#include <stddef.h>
+
+#include <_az_cfg.h>
+
+enum {
+  _az_LOG_VALUE_MAX_LENGTH
+  = 50, // When we print values, such as header vaules, if they are longer than
```
CIPop: Placing the comment above may look better and render much better search results in most IDEs which display a single line (no context) when displaying search results.

sdk/core/core/src/az_pal.c
```diff
@@ -0,0 +1,24 @@
...
+
+#include <_az_cfg.h>
+
+#ifdef _WIN32
+void az_pal_sleep(int32_t milliseconds) {
+  if (milliseconds > 0) {
+    Sleep(milliseconds);
+  }
+}
+#else
```
CIPop: This only works on Linux. Should the condition be an `#elif` and a `#error` with a clear message that the PAL needs an implementation on other systems?

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,213 @@
...
+
+    if (!should_retry) {
+      return result;
+    }
+
+    ++attempt;
+
+    if (retry_after_msec < 0) { // there wasn't any kind of "retry-after" response header
+      retry_after_msec
+          = _az_http_policy_retry_calc_delay(attempt, retry_delay_msec, max_retry_delay_msec);
```
CIPop: I can only find `_az_http_policy_retry_calc_delay` here: any reason it does the calculation as a `double` since the result is converted to an `int32_t` anyway? 
That immediately makes it extremely heavy (size and execution time) for anything that doesn't have an FPU.

CIPop: @JeffreyRichter This retry policy calculation would work well for MQTT. Should it be promoted to public API?

sdk/core/core/internal/az_clock_internal.h
antkmsft: These sources are based on the old implementation, we have the PAL layer now, I am going to use that. The other question is that we may still redesign the PAL layer to use an equivalent of clock_t convertible to milliseconds, instead of pure milliseconds as we do now; but that would be a separate commit/PR.

sdk/core/core/src/az_pal.c
antkmsft: We now have POSIX, Win32, and "empty" PAL. Please wait for the updated PR where I am going to use it.

sdk/core/core/src/az_http_policy_retry.c
antkmsft: I'll rewrite this piece to not use FPU arithmetic, it looks avoidable.

antkmsft: I like that idea but I would cover that refactoring as a separate PR - let's see how much common code is there, I am sure there's a significant part. But currently, this code relies on HTTP headers, and I don't have any MQTT sample in front of me to generalize it into reusable piece. Not that we can't do it all in one PR, but I don't think we lose anything if we do it as a separate one, which is a bit easier to handle.

JeffreyRichter:
We can promote it to internal; not public.


sdk/core/core/src/az_http_policy_retry.c
antkmsft: This is how Go does it: even if Max is 0, we would execute at least once.

antkmsft: https://github.com/Azure/azure-sdk-for-go/blob/cbb75bf1b53127c8f25a63b1d9e929d0d638246b/sdk/azcore/policy_retry.go

JeffreyRichter:
Anton,

Chris just means to move _az_http_policy_retry_calc_delay to internal, not the http policy.

And, rename the function to _az_retry_calc_delay


JeffreyRichter:
Go is going to change how it does it to match what we do in C. For example, it will be max retries and not max tries. We should always do at least one http request and then at most max_retries additional. If the customer sets max retries to 0 then we do 1 http request only and 0 retries.

Go will match this behavior.


antkmsft:
I made a new PR that is based on the new sources. I think I've covered all of your feedback there: https://github.com/Azure/azure-sdk-for-c/pull/413

@CIPop thank you for your feedback, unfortunately I can't simply invite you to the new PR, so I am going to tag you in this comment. Please use the link above to join the new code review!

PR #322, opened 2020-01-29
API version is fixed to the release of the SDK

gilbertw:


sdk/keyvault/keyvault/src/az_keyvault_client.c
```diff
@@ -137,13 +135,13 @@ AZ_NODISCARD az_result az_keyvault_keys_key_create(
...
   AZ_RETURN_IF_FAILED(az_http_request_builder_set_query_parameter(
-      &hrb, AZ_STR("api-version"), client->retry_options.service_version));
+      &hrb, AZ_STR("api-version"), AZ_KEYVAULT_API_VERSION));
 
   AZ_RETURN_IF_FAILED(az_http_request_builder_append_path(&hrb, key_name));
 
   // add extra header just for testing append_path after another query
-  AZ_RETURN_IF_FAILED(az_http_request_builder_set_query_parameter(
-      &hrb, AZ_STR("ignore"), client->retry_options.service_version));
+  AZ_RETURN_IF_FAILED(
```
JeffreyRichter: Do you want to get rid of this?

sdk/keyvault/keyvault/src/az_keyvault_client.c
```diff
@@ -137,13 +135,13 @@ AZ_NODISCARD az_result az_keyvault_keys_key_create(
 
   // add version to request
   AZ_RETURN_IF_FAILED(az_http_request_builder_set_query_parameter(
-      &hrb, AZ_STR("api-version"), client->retry_options.service_version));
+      &hrb, AZ_STR("api-version"), AZ_KEYVAULT_API_VERSION));
```
JeffreyRichter: Another option here is to create an ApiVersion policy and configure it with the AZ_KEYVAULT_API_VERSION option and then ALL operations will just get this header added automatically so there is no need to repeat this code for every operation. You could use the same policy for storage too. In this case, I'd make the query parameter name be an option along with the version string.

gilbertw: That's a good idea.  I'll make that change in a subsequent PR.

sdk/keyvault/keyvault/src/az_keyvault_client.c
gilbertw: Yes, this should only be in the tests

PR #338, opened 2020-02-07
Fix HTTP request sending in auth code

antkmsft:
@JeffreyRichter, _az_credential now needs the http_client_fn to send requests, this PR adds that.

@vhvb1989, there was a bug in our CURL code - when we used AZ_RETURN_IF_CURL_FAILED against az_result (i.e. instead of AZ_RETURN_IF_FAILED). I've made naming changes as well, and also marked all the local non-AZ_INLINE functions as static, so these internal function are never visible to linker outside of that C file (but that is not critical for the bugfix).

[ATTENTION: Make sure you see the changes to the az_curl.c (at the very end of this PR) - Github hides the change claiming it is "large" to show by default, so you need to click to expand]

For some reason, I am getting HTTP 400 (Bad Request), when I include Content-Type header for token request. But everything works fine if I don't add any header. But it works in Postman, so I think it is highly likely that there's a bug is in the code that handles HTTP headers - either CURL http client implementation, or in http_request.

With Content-Type commented out, this actually fixes auth token retrieval, although I would like to be able to include it for auth requests. Plus, if I am right, you won't be able to send any HTTP request anyways, if there's all-affecting http header bug in our code. If so, let's fix it tomorrow. 

Also, do we use pointer to pointer to http_client client intentionally? We use it in all client options, plus http policy expects not the pointer to http client, but a pointer to pointer. Therefore, I also wrote credential_policy to stick tho that format. Is this intentional to allow switching http clients in run time? (i.e. you update keyvault client's http client after it was initialized, and it starts sending all further http requests via another client)

### DIFF
--- a/sdk/core/core/src/az_aad.c
+++ b/sdk/core/core/src/az_aad.c
@@ -81,11 +81,17 @@ AZ_NODISCARD az_result _az_aad_build_body(
   return AZ_OK;
 }
 
-AZ_NODISCARD az_result _az_aad_request_token(az_http_request * ref_request, _az_token * out_token) {
-  AZ_RETURN_IF_FAILED(az_http_request_append_header(
+AZ_NODISCARD az_result _az_aad_request_token(
+    az_http_client_fn * http_client,
+    az_http_request * ref_request,
+    _az_token * out_token) {
+  // FIXME: If you uncomment the line below, we'll start getting HTTP 400 Bad Request instead of 200
+  // OK. I suspect, it is because there's a bug in the code that adds headers. Could be something
+  // else, of course.
+  /*AZ_RETURN_IF_FAILED(az_http_request_append_header(
       ref_request,
       AZ_SPAN_FROM_STR("Content-Type"),
-      AZ_SPAN_FROM_STR("application/x-www-url-form-urlencoded")));
+      AZ_SPAN_FROM_STR("application/x-www-url-form-urlencoded")));*/
 
   uint8_t response_buf[_az_AAD_RESPONSE_BUF_SIZE] = { 0 };
   az_http_response response = { 0 };
@@ -96,7 +102,7 @@ AZ_NODISCARD az_result _az_aad_request_token(az_http_request * ref_request, _az_
       .p_policies = {
         { .process = az_http_pipeline_policy_retry, .p_options = NULL },
         { .process = az_http_pipeline_policy_logging, .p_options = NULL },
-        { .process = az_http_pipeline_policy_transport, .p_options = NULL },
+        { .process = az_http_pipeline_policy_transport, .p_options = http_client },
       },
     };
   AZ_RETURN_IF_FAILED(az_http_pipeline_process(&pipeline, ref_request, &response));
--- a/sdk/core/core/src/az_aad.c
+++ b/sdk/core/core/src/az_aad.c
@@ -7,6 +7,8 @@
 #include <az_pal_clock_internal.h>
 #include <az_time_internal.h>
 
+#include <stddef.h>
+
 #include <_az_cfg.h>
 
 AZ_NODISCARD bool _az_token_expired(_az_token const * token) {
@@ -82,7 +84,7 @@ AZ_NODISCARD az_result _az_aad_build_body(
 }
 
 AZ_NODISCARD az_result _az_aad_request_token(
-    az_http_client_fn * http_client,
+    az_http_transport_options * http_transport_options,
     az_http_request * ref_request,
     _az_token * out_token) {
   // FIXME: If you uncomment the line below, we'll start getting HTTP 400 Bad Request instead of 200
@@ -102,7 +104,7 @@ AZ_NODISCARD az_result _az_aad_request_token(
       .p_policies = {
         { .process = az_http_pipeline_policy_retry, .p_options = NULL },
         { .process = az_http_pipeline_policy_logging, .p_options = NULL },
-        { .process = az_http_pipeline_policy_transport, .p_options = http_client },
+        { .process = az_http_pipeline_policy_transport, .p_options = http_transport_options },
       },
     };
   AZ_RETURN_IF_FAILED(az_http_pipeline_process(&pipeline, ref_request, &response));
--- a/sdk/core/core/src/az_aad_private.h
+++ b/sdk/core/core/src/az_aad_private.h
@@ -30,7 +30,7 @@ AZ_NODISCARD az_result _az_aad_build_body(
     az_span * out_body);
 
 AZ_NODISCARD az_result _az_aad_request_token(
-    az_http_client_fn * http_client,
+    az_http_transport_options * http_transport_options,
     az_http_request * request,
     _az_token * out_token);
 
--- a/sdk/core/core/src/az_aad_private.h
+++ b/sdk/core/core/src/az_aad_private.h
@@ -29,7 +29,10 @@ AZ_NODISCARD az_result _az_aad_build_body(
     az_span client_secret,
     az_span * out_body);
 
-AZ_NODISCARD az_result _az_aad_request_token(az_http_request * request, _az_token * out_token);
+AZ_NODISCARD az_result _az_aad_request_token(
+    az_http_client_fn * http_client,
+    az_http_request * request,
+    _az_token * out_token);
 
 AZ_NODISCARD bool _az_token_expired(_az_token const * token);
 AZ_NODISCARD _az_token _az_token_get(_az_token const * self);
--- a/sdk/core/core/src/az_credentials.c
+++ b/sdk/core/core/src/az_credentials.c
@@ -5,6 +5,8 @@
 #include <az_credentials.h>
 #include <az_http.h>
 
+#include <stddef.h>
+
 #include <_az_cfg.h>
 
 static AZ_NODISCARD az_result
@@ -28,7 +30,7 @@ _az_client_secret_credential_request_token(az_client_secret_credential * credent
       &request, az_http_method_post(), url, AZ_SPAN_FROM_BUFFER(header_buf), body));
 
   return _az_aad_request_token(
-      credential->_internal.credential._internal.http_client,
+      &credential->_internal.credential._internal.http_transport_options,
       &request,
       &credential->_internal.token);
 }
@@ -67,7 +69,7 @@ AZ_NODISCARD az_result az_client_secret_credential_init(
     ._internal = {
       .credential = {
         ._internal = {
-          .http_client = NULL,
+          .http_transport_options = NULL,
           .apply_credential = (_az_credential_apply_fn)_az_client_secret_credential_apply,
           .set_scopes = (_az_credential_set_scopes_fn)_az_client_secret_credential_set_scopes,
           },
--- a/sdk/core/core/src/az_credentials.c
+++ b/sdk/core/core/src/az_credentials.c
@@ -27,7 +27,10 @@ _az_client_secret_credential_request_token(az_client_secret_credential * credent
   AZ_RETURN_IF_FAILED(az_http_request_init(
       &request, az_http_method_post(), url, AZ_SPAN_FROM_BUFFER(header_buf), body));
 
-  return _az_aad_request_token(&request, &credential->_internal.token);
+  return _az_aad_request_token(
+      credential->_internal.credential._internal.http_client,
+      &request,
+      &credential->_internal.token);
 }
 
 // This gets called from the http credential policy
@@ -62,8 +65,9 @@ AZ_NODISCARD az_result az_client_secret_credential_init(
     az_span client_secret) {
   *self = (az_client_secret_credential){
     ._internal = {
-      .vtbl = {
+      .credential = {
         ._internal = {
+          .http_client = NULL,
           .apply_credential = (_az_credential_apply_fn)_az_client_secret_credential_apply,
           .set_scopes = (_az_credential_set_scopes_fn)_az_client_secret_credential_set_scopes,
           },
--- a/sdk/keyvault/keyvault/src/az_keyvault_client.c
+++ b/sdk/keyvault/keyvault/src/az_keyvault_client.c
@@ -7,6 +7,8 @@
 #include <az_keyvault.h>
 #include <az_span.h>
 
+#include <stddef.h>
+
 #include <_az_cfg.h>
 
 /**
@@ -38,11 +40,11 @@ AZ_NODISCARD AZ_INLINE az_span az_keyvault_client_constant_for_application_json(
 }
 
 AZ_NODISCARD az_keyvault_keys_client_options
-az_keyvault_keys_client_options_default(az_http_client_fn http_client) {
+az_keyvault_keys_client_options_default(az_http_transport_options const * http_transport_options) {
 
   az_keyvault_keys_client_options options = (az_keyvault_keys_client_options){
-    ._internal
-    = { .http_client = http_client, .api_version = az_http_policy_apiversion_options_default() },
+    ._internal = { .http_transport_options = *http_transport_options,
+                   .api_version = az_http_policy_apiversion_options_default() },
     .retry = az_http_policy_retry_options_default(),
   };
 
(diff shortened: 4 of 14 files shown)


antkmsft:
@JeffreyRichter , what is better ? : - to initialize credential with the client's http_client during the client's init() [this is what I do now], or to request another http_client parameter in the credential's init?

sdk/core/core/src/az_aad.c
```diff
@@ -96,7 +102,7 @@ AZ_NODISCARD az_result _az_aad_request_token(az_http_request * ref_request, _az_
       .p_policies = {
         { .process = az_http_pipeline_policy_retry, .p_options = NULL },
         { .process = az_http_pipeline_policy_logging, .p_options = NULL },
-        { .process = az_http_pipeline_policy_transport, .p_options = NULL },
+        { .process = az_http_pipeline_policy_transport, .p_options = http_client },
```
JeffreyRichter: Let's do this: define a transport policy options struct similar to what we have for some other policies. Put the http_client as the only field in this options structure. The, this p_options is a pointer to the that options structure since, in the future, the structure might get more fields in it. 
(this is why I had ** here; I was thinking about the future of having a struct and keeping it consistent with ALL the other policies).

sdk/core/core/src/az_aad.c
```diff
@@ -81,11 +81,17 @@ AZ_NODISCARD az_result _az_aad_build_body(
   return AZ_OK;
 }
 
-AZ_NODISCARD az_result _az_aad_request_token(az_http_request * ref_request, _az_token * out_token) {
-  AZ_RETURN_IF_FAILED(az_http_request_append_header(
+AZ_NODISCARD az_result _az_aad_request_token(
+    az_http_client_fn * http_client,
```
JeffreyRichter: After you do what I say below (see next comment), then this first parameter should be a pointer to a transport_policy_options structure.

sdk/core/core/src/az_aad_private.h
```diff
@@ -29,7 +29,10 @@ AZ_NODISCARD az_result _az_aad_build_body(
     az_span client_secret,
     az_span * out_body);
 
-AZ_NODISCARD az_result _az_aad_request_token(az_http_request * request, _az_token * out_token);
+AZ_NODISCARD az_result _az_aad_request_token(
+    az_http_client_fn * http_client,
```
JeffreyRichter: This will have to match

sdk/core/core/src/az_credentials.c
```diff
@@ -62,8 +65,9 @@ AZ_NODISCARD az_result az_client_secret_credential_init(
     az_span client_secret) {
   *self = (az_client_secret_credential){
     ._internal = {
-      .vtbl = {
+      .credential = {
         ._internal = {
+          .http_client = NULL,
```
JeffreyRichter: this will be a transport_policy_options struct (not a pointer to it). http_client will be its only field (for now).

sdk/keyvault/keyvault/src/az_keyvault_client.c
```diff
@@ -61,7 +61,8 @@ AZ_NODISCARD az_result az_keyvault_keys_client_init(
   AZ_CONTRACT_ARG_NOT_NULL(self);
   AZ_CONTRACT_ARG_NOT_NULL(options);
 
-  _az_credential_vtbl * const cred = (_az_credential_vtbl *)credential;
+  _az_credential * const cred = (_az_credential *)credential;
+  cred->_internal.http_client = &self->_internal.options._internal.http_client;
```
JeffreyRichter: Of course, the key_vault_key_client_options structure also needs to change the http_client to a transport_policy_options struct as well.

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
```diff
@@ -41,7 +41,8 @@ AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
   AZ_CONTRACT_ARG_NOT_NULL(client);
   AZ_CONTRACT_ARG_NOT_NULL(options);
 
-  _az_credential_vtbl * const cred = (_az_credential_vtbl *)credential;
+  _az_credential * const cred = (_az_credential *)credential;
+  cred->_internal.http_client = &client->_internal.options._internal.http_client;
```
JeffreyRichter: And here too.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -65,10 +65,10 @@ AZ_NODISCARD AZ_INLINE az_result az_curl_done(CURL ** pp) {
   return AZ_OK;
 }
 
-uint8_t zerobyte[] = { 0 };
-az_span const AZ_SPAN_ZEROBYTE = AZ_SPAN_LITERAL_FROM_INITIALIZED_BUFFER(zerobyte);
+static uint8_t zerobyte[] = { 0 };
```
JeffreyRichter: Please put a 1 in the square brackets. The syntax is confusing without it.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -79,8 +79,8 @@ az_span const AZ_HTTP_REQUEST_BUILDER_HEADER_SEPARATOR = AZ_SPAN_LITERAL_FROM_ST
  * @param separator symbol to be used between key and value
  * @return az_result
  */
-AZ_NODISCARD az_result
-az_write_to_buffer(az_span writable_buffer, az_pair header, az_span separator) {
+static AZ_NODISCARD az_result
+_az_span_write_to_buffer(az_span writable_buffer, az_pair header, az_span separator) {
```
JeffreyRichter: Can we rename this to _az_span_append_header_to_buffer because it calls append internally and this indicates what the function actually appends to the span

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -89,7 +89,8 @@ az_write_to_buffer(az_span writable_buffer, az_pair header, az_span separator) {
   return AZ_OK;
```
JeffreyRichter: Is the line above the bug? You probably don't append a zero-byte; don't you append a CRLF

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -121,20 +124,20 @@ az_add_header_to_curl_list(az_pair header, struct curl_slist ** p_list, az_span
     size_t const buffer_size = (size_t)az_span_length(header.key)
         + (size_t)az_span_length(separator) + (size_t)az_span_length(header.value) + 1;
```
JeffreyRichter: Why +1 at the end? Add comment or use sizeof(???). Also, if this is CRLF, then it is +2; not +1.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -145,17 +148,18 @@ az_add_header_to_curl_list(az_pair header, struct curl_slist ** p_list, az_span
...
-AZ_NODISCARD az_result
-az_build_headers(az_http_request * p_request, struct curl_slist ** p_headers) {
+static AZ_NODISCARD az_result _az_http_client_curl_build_headers(
+    az_http_request * p_request,
+    struct curl_slist ** p_headers) {
   AZ_CONTRACT_ARG_NOT_NULL(p_request);
 
   az_pair header;
   for (int32_t offset = 0;
        offset < (int32_t)(az_span_length(p_request->_internal.headers) / sizeof(az_pair));
```
JeffreyRichter: Don't we have functions for this now? The CURL code should NOT be accessing internal fields directly.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -169,7 +173,8 @@ az_build_headers(az_http_request * p_request, struct curl_slist ** p_headers) {
  * @param url_from_request a url that is not zero terminated
  * @return az_result
  */
-AZ_NODISCARD az_result az_write_url(az_span writable_buffer, az_span url_from_request) {
+static AZ_NODISCARD az_result
+_az_http_client_curl_write_url(az_span writable_buffer, az_span url_from_request) {
```
JeffreyRichter: rename 'write' to 'append' to be consistent

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -169,7 +173,8 @@ az_build_headers(az_http_request * p_request, struct curl_slist ** p_headers) {
...
 
   AZ_RETURN_IF_FAILED(az_span_append(writable_buffer, url_from_request, &writable_buffer));
   AZ_RETURN_IF_FAILED(az_span_append(writable_buffer, AZ_SPAN_ZEROBYTE, &writable_buffer));
```
JeffreyRichter: Is zero-byte right? I think it is CRLF again

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -214,49 +223,54 @@ AZ_NODISCARD az_result az_curl_send_get_request(CURL * p_curl) {
...
-AZ_NODISCARD az_result az_curl_send_post_request(CURL * p_curl, az_http_request const * p_request) {
+static AZ_NODISCARD az_result
+_az_http_client_curl_send_post_request(CURL * p_curl, az_http_request const * p_request) {
   AZ_CONTRACT_ARG_NOT_NULL(p_curl);
   AZ_CONTRACT_ARG_NOT_NULL(p_request);
 
   // Method
   az_span body = { 0 };
   int32_t required_length
       = az_span_length(p_request->_internal.body) + az_span_length(AZ_SPAN_ZEROBYTE);
```
JeffreyRichter: CURL should NOT be accessing internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -277,46 +291,47 @@ int32_t _az_curl_upload_read_callback(void * ptr, size_t size, size_t nmemb, voi
...
-az_curl_send_upload_request(CURL * p_curl, az_http_request const * p_request) {
+static AZ_NODISCARD az_result
+_az_http_client_curl_send_upload_request(CURL * p_curl, az_http_request const * p_request) {
   AZ_CONTRACT_ARG_NOT_NULL(p_curl);
   AZ_CONTRACT_ARG_NOT_NULL(p_request);
 
   // Method
   az_span body = { 0 };
   int32_t required_size
       = az_span_length(p_request->_internal.body) + az_span_length(AZ_SPAN_ZEROBYTE);
```
JeffreyRichter: CURL should not access internal fields directly; we may need to add a function to make this work.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -277,46 +291,47 @@ int32_t _az_curl_upload_read_callback(void * ptr, size_t size, size_t nmemb, voi
...
-  AZ_RETURN_IF_FAILED(az_span_malloc(required_size, &body));
+  AZ_RETURN_IF_FAILED(_az_span_malloc(required_size, &body));
 
   char * b = (char *)az_span_ptr(body);
   az_result res_code = az_span_to_str(b, required_size, p_request->_internal.body);
```
JeffreyRichter: Don't access internal field directly.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -277,46 +291,47 @@ int32_t _az_curl_upload_read_callback(void * ptr, size_t size, size_t nmemb, voi
...
+_az_http_client_curl_send_upload_request(CURL * p_curl, az_http_request const * p_request) {
   AZ_CONTRACT_ARG_NOT_NULL(p_curl);
   AZ_CONTRACT_ARG_NOT_NULL(p_request);
 
   // Method
   az_span body = { 0 };
   int32_t required_size
       = az_span_length(p_request->_internal.body) + az_span_length(AZ_SPAN_ZEROBYTE);
-  AZ_RETURN_IF_FAILED(az_span_malloc(required_size, &body));
+  AZ_RETURN_IF_FAILED(_az_span_malloc(required_size, &body));
```
JeffreyRichter: Do we have to allocate and copy the body? Can we just pass the address of the request buffer directly to curl?

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -329,7 +344,8 @@ az_curl_send_upload_request(CURL * p_curl, az_http_request const * p_request) {
  * @param p_request an http request builder
  * @return az_result
  */
-AZ_NODISCARD az_result setup_headers(CURL * p_curl, az_http_request * p_request) {
+static AZ_NODISCARD az_result
+_az_http_client_curl_setup_headers(CURL * p_curl, az_http_request * p_request) {
   AZ_CONTRACT_ARG_NOT_NULL(p_curl);
   AZ_CONTRACT_ARG_NOT_NULL(p_request);
```
JeffreyRichter: Below: don't access internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -365,20 +382,20 @@ AZ_NODISCARD az_result setup_url(CURL * p_curl, az_http_request const * p_reques
     size_t const extra_space_for_zero = az_span_length(AZ_SPAN_ZEROBYTE);
     size_t const url_final_size = az_span_length(p_request->_internal.url) + extra_space_for_zero;
```
JeffreyRichter: Don't access internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -365,20 +382,20 @@ AZ_NODISCARD az_result setup_url(CURL * p_curl, az_http_request const * p_reques
...
     // allocate buffer to add \0
-    AZ_RETURN_IF_FAILED(az_span_malloc(url_final_size, &writable_buffer));
+    AZ_RETURN_IF_FAILED(_az_span_malloc(url_final_size, &writable_buffer));
```
JeffreyRichter: Do we have to allocate for CURL; can't we just use our http_response buffer instead?

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -365,20 +382,20 @@ AZ_NODISCARD az_result setup_url(CURL * p_curl, az_http_request const * p_reques
...
   }
 
   // write url in buffer (will add \0 at the end)
-  az_result result = az_write_url(writable_buffer, p_request->_internal.url);
+  az_result result = _az_http_client_curl_write_url(writable_buffer, p_request->_internal.url);
```
JeffreyRichter: Don't access internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -365,20 +382,20 @@ AZ_NODISCARD az_result setup_url(CURL * p_curl, az_http_request const * p_reques
...
   if (az_succeeded(result)) {
     char * buffer = (char *)az_span_ptr(writable_buffer);
-    result = az_curl_code_to_result(curl_easy_setopt(p_curl, CURLOPT_URL, buffer));
+    result = _az_http_client_curl_code_to_result(curl_easy_setopt(p_curl, CURLOPT_URL, buffer));
   }
 
   // free used buffer before anything else
   memset(az_span_ptr(writable_buffer), 0, az_span_capacity(writable_buffer));
-  az_span_free(&writable_buffer);
+  _az_span_free(&writable_buffer);
```
JeffreyRichter: If we can avoid malloc, we can avoid free.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -390,13 +407,18 @@ AZ_NODISCARD az_result setup_url(CURL * p_curl, az_http_request const * p_reques
  * @param response_builder an http request builder holding all http request data
  * @return az_result
  */
-AZ_NODISCARD az_result setup_response_redirect(CURL * p_curl, az_span * response_builder) {
+static AZ_NODISCARD az_result
+_az_http_client_curl_setup_response_redirect(CURL * p_curl, az_span * response_builder) {
```
JeffreyRichter: Was this code ever tested?

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -412,26 +434,27 @@ AZ_NODISCARD az_result setup_response_redirect(CURL * p_curl, az_span * response
...
 
-  AZ_RETURN_IF_CURL_FAILED(setup_headers(p_curl, p_request));
+  AZ_RETURN_IF_FAILED(_az_http_client_curl_setup_headers(p_curl, p_request));
 
-  AZ_RETURN_IF_CURL_FAILED(setup_url(p_curl, p_request));
+  AZ_RETURN_IF_FAILED(_az_http_client_curl_setup_url(p_curl, p_request));
 
-  AZ_RETURN_IF_CURL_FAILED(setup_response_redirect(p_curl, &response->_internal.http_response));
+  AZ_RETURN_IF_FAILED(
+      _az_http_client_curl_setup_response_redirect(p_curl, &response->_internal.http_response));
```
JeffreyRichter: Don't access internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -412,26 +434,27 @@ AZ_NODISCARD az_result setup_response_redirect(CURL * p_curl, az_span * response
...
 
   if (az_span_is_equal(p_request->_internal.method, az_http_method_get())) {
-    result = az_curl_send_get_request(p_curl);
+    result = _az_http_client_curl_send_get_request(p_curl);
```
JeffreyRichter: This doesn't seem right to me: A get request is very likely to have headers with it - how would they get passed?

Also, I think you don't actually need separate functions for each HTTP method verb. I think you can do this all with the same function and just set the passed-in request method verb.

And finally, Don't access internal fields directly

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -412,26 +434,27 @@ AZ_NODISCARD az_result setup_response_redirect(CURL * p_curl, az_span * response
...
+    result = _az_http_client_curl_send_post_request(p_curl, p_request);
   } else if (az_span_is_equal(p_request->_internal.method, az_http_method_delete())) {
-    result = az_curl_send_delete_request(p_curl, p_request);
+    result = _az_http_client_curl_send_delete_request(p_curl, p_request);
   } else if (az_span_is_equal(p_request->_internal.method, az_http_method_put())) {
-    result = az_curl_send_upload_request(p_curl, p_request);
+    result = _az_http_client_curl_send_upload_request(p_curl, p_request);
   } else {
     return AZ_ERROR_HTTP_INVALID_METHOD_VERB;
   }
```
JeffreyRichter: below: Don't access internal fields directly and I'm questioning AZ_SPAN_ZEROBYTE

sdk/transport_policies/curl/src/az_curl.c
antkmsft: Ah! And here's possibly why it now fails when I add Content-Type header for the auth!

sdk/transport_policies/curl/src/az_curl.c
antkmsft: At some point, we should ultimately add '\0', because curl expects 'char *' strings to be null-terminated. But as far as header separator, I think you're right, there's a bug and this is likely why I currently can't set Content-Type header for the token request.

JeffreyRichter: do curl's doc say this explicitly? Maybe curl wants some length which doesn't require 0-termination. We should know; not guess.

PR #341, opened 2020-02-10
change pointer alignment and braces

danewalton:
Reissuing this PR based on our meeting on Friday. Thanks for helping us come to a middle ground here.


.clang-format
```diff
@@ -5,7 +5,7 @@ TabWidth: 2
 IndentWidth: 2
 IncludeBlocks: Preserve
 NamespaceIndentation: None
-PointerAlignment: Middle
+PointerAlignment: Left
```
CIPop: I don't find this acceptable: all OS kernels we're going to work with have this to the right: Linux Kernel, ThreadX, FreeRTOS, uC/OS, etc.
Creating a different style would cause confusion between the IoT SDK core and the ports.

As discussed offline, Middle is a good compromise (although deep inside, I know right is right 😜 )

antkmsft: @danewalton, so Middle or Left? :)

danewalton: Cris and I talked offline and believe we can go left (maintains consistent style with the iot sdk). I recognize that it won't make everyone happy but I think ultimately middle leaves no one happy. It looks a lot like multiplication and very very few people use it. For the sake of choosing one and moving on to more important things, the majority of the IoT team is good with left.

CIPop: If we can't do right let's at least remove the extra space...

antkmsft: We're about to wrap up most of the planned work soon. I'll apply this PR and re-format all the code.

CIPop: Thanks @antkmsft - really appreciate the flexibility, allowing our preference as well as handling the extra work this generated!

antkmsft:
https://github.com/Azure/azure-sdk-for-c/pull/377

PR #344, opened 2020-02-11
Add security policy

heaths:


joshfree:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

@danieljurek can you PTAL?

PR #349, opened 2020-02-12
Minor cleanups

antkmsft:


sdk/core/core/src/az_span.c
```diff
@@ -197,7 +197,7 @@ AZ_NODISCARD az_result az_span_to_str(char * s, int32_t max_size, az_span span)
 
   memmove((void *)s, (void const *)az_span_ptr(span), span_length);
 
-  s[span_length] = 0;
+  s[span_length] = '\0';
```
JeffreyRichter: I GREATLY prefer just 0 without the apostrophes and the slash. Why do you prefer this?

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -377,8 +377,8 @@ _az_http_client_curl_setup_url(CURL * p_curl, az_http_request const * p_request)
   az_span writable_buffer;
   {
     // set URL as 0-terminated str
-    size_t const url_final_size
-        = az_span_length(p_request->_internal.url) + az_span_length(AZ_SPAN_FROM_STR("\0"));
+    size_t const url_final_size = (size_t)az_span_length(p_request->_internal.url)
```
JeffreyRichter: I would change this size_t to a int32_t and then remove all the casting.
I'd also just add 1 instead of creating a span of "\0" and then getting its length; just add a comment about what your doing and shrink the code by several instructions.

sdk/core/core/src/az_span.c
antkmsft: I have no strong preference, so I'll change it to 0.
I prefer it, because it reads as "null character", meaning of which is known. 0 reads as just "zero" to me, indistinguishable from ints. I know that '\0' is 0, but when dealing with strings, and characters, I prefer to deal with characters, the almost the similar way I write 'A' and not 65 (the example is clearly an exaggeration, but you asked me to explain, so I am trying my best :)).

sdk/transport_policies/curl/src/az_curl.c
antkmsft: Done.

sdk/core/core/src/az_span.c
antkmsft: Done.

JeffreyRichter: Maybe it's my C training from decades ago but I was always taught that a string is a zero-terminated array of chars. To me, this means a zero byte terminating the array; not a 0 character. In fact, 0 can't be a character and so it confuses me to see it being shown in the code as a zero-character. I was always taught that a 0 BYTE terminates the string.

That being said, I just did some web searches to see what other people do and I found some using 0 and some using '\0'. So, I feel a little less strongly about it now. But, I still prefer the 0 as its simpler: no question about "is the slash going the right way?" or is that the letter 'O' and not a 0? 

Thanks for making the change.

PR #360, opened 2020-02-18
Add the transport policy as a copy.

gilbertw:
Transport policies should be stored by value so that users can reuse/release the structure.

Various bug fixes along the way:
- Correct some verbs
- Fixes to upload callback after the refactor


sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -291,9 +292,10 @@ static int32_t _az_http_client_curl_upload_read_callback(
 
   int32_t data_length = az_span_length(*mut_span);
   int32_t to_copy = (data_length < curl_size) ? data_length : curl_size;
-  AZ_RETURN_IF_FAILED(
-      az_span_append(*mut_span, az_span_init((uint8_t *)ptr, to_copy, to_copy), mut_span));
-
+  memcpy(ptr, mut_span->_internal.ptr, to_copy);
```
JeffreyRichter: NO code but the span code itself should be directly accessing _internal members of a span structure. Can this be modified to avoid doing this?

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -314,6 +316,7 @@ _az_http_client_curl_send_upload_request(CURL * p_curl, _az_http_request const *
 
   char * b = (char *)az_span_ptr(body);
   az_result res_code = az_span_to_str(b, required_size, p_request->_internal.body);
+  body._internal.length = required_size;
```
JeffreyRichter: Similarly, this should not access span's _internal fields.

sdk/transport_policies/curl/src/az_curl.c
gilbertw: I agree.  The old code didn't transfer over 1-1 due to the mutating buffer being initialized when we map to it.  The old code took advantage of behaviors in the mutating span which are no longer present.  I'll play with it and see if I can make it better.  At a minimum I can update this code to use the access helpers

sdk/core/core/src/az_http_policy.c
```diff
@@ -174,3 +174,12 @@ AZ_NODISCARD az_result az_http_pipeline_policy_transport(
 
   return send_request(p_request, p_response);
 }
+
+AZ_NODISCARD az_http_transport_options
+az_http_transport_options_default(az_http_client_send_request_fn send_request) {
```
vhvb1989: it is just a kind of weird that it gives the `default` options but it only sets the value it gets as parameter XD.
Looks more like an `init` for its implementation.
But I got it.  It is because of all other options we have with this `default` approach...

sdk/transport_policies/curl/src/az_curl.c
vhvb1989:
```suggestion
  memcpy(ptr, az_span_ptr(mut_span), to_copy);
```

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -291,9 +292,10 @@ static int32_t _az_http_client_curl_upload_read_callback(
 
   int32_t data_length = az_span_length(*mut_span);
   int32_t to_copy = (data_length < curl_size) ? data_length : curl_size;
-  AZ_RETURN_IF_FAILED(
-      az_span_append(*mut_span, az_span_init((uint8_t *)ptr, to_copy, to_copy), mut_span));
-
+  memcpy(ptr, mut_span->_internal.ptr, to_copy);
+  mut_span->_internal.capacity -= to_copy;
+  mut_span->_internal.ptr += to_copy;
```
vhvb1989:
```suggestion
  // Move span to the remaining (after `to_copy`)
  AZ_RETURN_IF_FAILED(az_span_slice(*mut_span, to_copy, -1, mut_span);
```

vhvb1989: probably is not the mut_span that needs to be updated but the pointer to user_data??

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -291,9 +292,10 @@ static int32_t _az_http_client_curl_upload_read_callback(
 
   int32_t data_length = az_span_length(*mut_span);
   int32_t to_copy = (data_length < curl_size) ? data_length : curl_size;
-  AZ_RETURN_IF_FAILED(
```
vhvb1989: I think we need a super great and amazing comment for this method to mention what it does.
It is being called as many times as it requires to upload something right?
Every time it gets called, we would upload a chunk of data?  that chuck needs to be copied to ptr with max = data_lenght?
And user_data is pointing to the start of what we need to upload?
If so, every time this is called, we need to move the userdata pointer += curl_size ?  so the next time we add the next chunck?  or is it something else?

sdk/transport_policies/curl/src/az_curl.c
gilbertw: They are one and the same.  The userdata void* pointer is cast to mut_span.

sdk/transport_policies/curl/src/az_curl.c
gilbertw: More less that is what its doing.  I've added a comment to that sense with references back to the curl documentation.

sdk/transport_policies/curl/src/az_curl.c
```diff
@@ -314,6 +316,7 @@ _az_http_client_curl_send_upload_request(CURL * p_curl, _az_http_request const *
 
   char * b = (char *)az_span_ptr(body);
```
vhvb1989: it looks like this method `upload` was trying to imitate what method `post` does.
And `post` is imitating what `set_url` does.
But this is different completely.  For `set_url` we need to allocate a new buffer in memory because we need it to be 0-terminated.
I think for posting a body, or uploading a body, we shouldn't need to add a 0 at the end of it (because it changes the actual body of it).  
It should be enough to use the body span directly.
Like 

```c
static AZ_NODISCARD az_result
_az_http_client_curl_send_upload_request(CURL * p_curl, _az_http_request const * p_request) {
  AZ_CONTRACT_ARG_NOT_NULL(p_curl);
  AZ_CONTRACT_ARG_NOT_NULL(p_request);

  // Method
  az_span body = p_request->_internal.body;

  AZ_RETURN_IF_CURL_FAILED(curl_easy_setopt(p_curl, CURLOPT_UPLOAD, 1L));

  AZ_RETURN_IF_CURL_FAILED(
      curl_easy_setopt(p_curl, CURLOPT_READFUNCTION, _az_http_client_curl_upload_read_callback));

  AZ_RETURN_IF_CURL_FAILED(curl_easy_setopt(p_curl, CURLOPT_READDATA, body));

  AZ_RETURN_IF_CURL_FAILED(curl_easy_setopt(p_curl, CURLOPT_INFILESIZE, az_span_length(body)));

  // Do the curl work
  return _az_http_client_curl_code_to_result(curl_easy_perform(p_curl));
}
```

vhvb1989: Actually `post` method does need to allocate because it uses `CURLOPT_POSTFIELDS`  and it is expected to be 0-terminated. : (https://curl.haxx.se/libcurl/c/CURLOPT_POSTFIELDS.html).

JeffreyRichter:
If Charlie comes up with a good pal build pattern, then this parameter will go away


JeffreyRichter:
Yes, this is OK; it was actually a bug that we used the same one before – I can explain why in person.
However, all of this is temporary (I hope). Once we get the PLA stuff sorted out, I think we’ll hardcode that transport policy in Core so it will be 1 transport policy for the customer which ALL clients (and credentials will use.

PR #361, opened 2020-02-18
add iot and mqtt facility codes

danewalton:


sdk/core/core/inc/az_result.h
```diff
@@ -19,6 +19,8 @@ enum {
   AZ_FACILITY_CORE = 0x1,
   AZ_FACILITY_JSON = 0x2,
   AZ_FACILITY_HTTP = 0x3,
+  AZ_FACILITY_IOT  = 0x4,
```
JeffreyRichter: This is good but can we make MQTT 4 and IOT 5 so that the 2 network protocols (HTTP & MQTT) are next to each other.

danewalton: Yup no problem. Updated and pushed

antkmsft:
https://github.com/Azure/azure-sdk-for-c/pull/377

PR #362, opened 2020-02-19
IoT Provisioning & Hub Public API Review

CIPop:
- MQTT State machine design spec.
- Public API include files (flat folder structure)
- Paho samples for provisioning and hub (work-in-progress, requires updates to compile and use current include files).

@JeffreyRichter @jspaith PTAL


sdk/iot/inc/az_iot_hub_properties.h
```diff
@@ -0,0 +1,27 @@
...
+
+typedef struct az_iot_hub_properties {
+    struct {
+        az_span property_string;   // "URIENCODED(name1=val1&name2=val2&...)"
+    } _internal;
+} az_iot_hub_properties;
+
+// TODO: Do we need an enumerator (requires state preserved)
+
+az_result az_iot_hub_properties_add(az_iot_hub_properties *properties, az_span name, az_span value);
```
hihigupt: >az_iot_hub_properties *properties [](start = 36, length = 33)

Do we need this new type, az_span is enough?

sdk/iot/inc/az_iot_hub_properties.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_span.h>
+#include <az_result.h>
+
+#ifndef _az_IOT_HUB_PROPERTIES_H
+#define _az_IOT_HUB_PROPERTIES_H
```
hihigupt: >_az_ [](start = 8, length = 4)

This need to be capital in all the files

sdk/iot/inc/az_iot_hub_methods.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
danewalton: `#include <stdint.h>`

sdk/iot/inc/az_iot_hub_methods.h
```diff
@@ -0,0 +1,39 @@
...
+    az_span connection_timeout; // NULL terminated
+} az_iot_method_response;
+
+typedef struct az_iot_method_response {
+    uint8_t status;
+    az_span request_id; // NULL terminated
+    az_span payload;
+} az_iot_method_response;
+
+az_result az_iot_methods_send_response(az_iot_hub_client* client, az_iot_method_response* response, az_iot_mqtt_publish *mqtt_pub);
```
danewalton: what is the difference between this and `az_iot_hub_methods_send_response`? They are oddly similarly named.

danewalton: additionally, if this doesn't actually send the response, the action verb on the api probably shouldn't be "send". Should it be something like `az_iot_methods_response_init`?

sdk/iot/inc/az_iot_hub_twin.h
```diff
@@ -0,0 +1,28 @@
...
+#include <_az_cfg_prefix.h>
+
+typedef struct az_iot_hub_twin_response {
+    az_span payload;
+    az_span request_id;
+    az_span etag;
+} az_iot_hub_twin_response;
+
+az_result az_iot_twin_get(az_iot_hub_client* client, az_span request_id, az_iot_mqtt_publish *mqtt_pub);
+az_result az_iot_twin_patch(az_iot_hub_client* client, az_span request_id, az_span etag, az_span, payload, az_iot_mqtt_pub *mqtt_pub);
```
danewalton: one too many commas on `az_span, payload`

danewalton: should `az_iot_mqtt_pub` be `az_iot_mqtt_publish`?

sdk/iot/inc/az_iot_hub_properties.h
```diff
@@ -0,0 +1,27 @@
...
+    struct {
+        az_span property_string;   // "URIENCODED(name1=val1&name2=val2&...)"
+    } _internal;
+} az_iot_hub_properties;
+
+// TODO: Do we need an enumerator (requires state preserved)
+
+az_result az_iot_hub_properties_add(az_iot_hub_properties *properties, az_span name, az_span value);
+az_result az_iot_hub_properties_read(az_iot_hub_properties *properties, az_span name, az_span value);
+az_result az_iot_hub_properties_update(az_iot_hub_properties *properties, az_span name, az_span value);
```
danewalton: do we need an update api? User could just remove and add again.

sdk/iot/inc/az_iot_hub_sas.h
```diff
@@ -0,0 +1,27 @@
...
+#include "az_iot_hub_client.h"
+
+#ifndef _az_IOT_SAS_H
+#define _az_IOT_SAS_H
+
+#include <_az_cfg_prefix.h>
+
+typedef struct az_iot_hub_sas {
+    struct {
+        az_span audience;
```
danewalton: should this be named audience? Is this a native term for sas keys? I searched online and nothing came up.

sdk/iot/inc/az_iot_hub_telemetry.h
```diff
@@ -0,0 +1,21 @@
...
+#include "az_iot_mqtt.h"
+#include "az_iot_hub_client.h"
+#include "az_iot_hub_properties.h"
+
+#ifndef _az_IOT_HUB_TELEMETRY_H
+#define _az_IOT_HUB_TELEMETRY_H
+
+#include <_az_cfg_prefix.h>
+
+az_result az_iot_sendtelemetry(az_iot_hub_client* client, az_span payload, az_iot_hub_properties* properties, az_iot_mqtt_publish *mqtt_pub);
```
danewalton: these two functions are named the same thing

sdk/iot/inc/az_iot_mqtt.h
```diff
@@ -0,0 +1,29 @@
...
+    az_span user;       // NULL terminated string.
+    az_span password;   // NULL terminated string (may be an empty string).
+} az_iot_mqtt_connect;
+
+typedef struct az_iot_topic {
+    az_span name;       // NULL terminated string.
+    uint8_t qos;
+} az_iot_topic;
+
+typedef struct az_iot_mqtt_publish {
```
danewalton: should this be generalized to `az_iot_mqtt_msg`? We could use this to parse responses into as well.

sdk/iot/inc/az_iot_provisioning_sas.h
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_span.h>
+#include <az_result.h>
+#include <time.h>
```
danewalton: [Link to Jeffrey's comment about using int for time](https://github.com/Azure/azure-sdk-for-c/pull/302#discussion_r365350923)

danewalton: [Link to http using `uint16_t` for timing](https://github.com/Azure/azure-sdk-for-c/blob/25f8a0228e5f250c02e389f19d88c064c93959c1/sdk/core/core/inc/az_http.h#L171)

sdk/iot/inc/az_iot_provisioning_sas.h
```diff
@@ -0,0 +1,28 @@
...
+        az_span key_name; // Can be empty.
+    } _internal;
+} az_iot_provisioning_sas;
+
+az_result az_iot_provisioning_sas_init(az_iot_provisioning_sas* sas, az_provisioning_client* client, az_span id_scope, az_span key_name);
+az_result az_iot_provisioning_sas_update(az_iot_provisioning_sas* sas, time_t expiration_time, az_span base64_hmac256_signature, az_iot_mqtt_connect* mqtt_connect);
+
+#include <_az_cfg_suffix.h>
+
+#endif //!_az_IOT_PROVISIONING_SAS_H
```
danewalton: new line at end of file

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+    } _internal;
+} az_iot_hub_sas;
+
+az_result az_iot_hub_sas_init(az_iot_hub_sas* sas, az_iot_hub_client* client, az_span hub_name, az_span key_name);
+az_result az_iot_hub_sas_update(az_iot_hub_sas* sas, time_t expiration_time, az_span base64_hmac256_signature, az_iot_mqtt_connect* mqtt_connect);
+```
+
+
+### Subscribe topic information
+Note: All services (Device Provisioning, IoT Hub) require either no subscription (Telemetry) or _a single_ subscription to a topic _filter_.
```
danewalton: I do not believe this is true. Twin has subscriptions to the following:
```
$iothub/twin/res/#
$iothub/twin/PATCH/properties/desired/#
```

sdk/iot/inc/az_iot_hub_properties.h
CIPop: The reason is an idea from @danewalton that was also confirmed by @JeffreyRichter : the properties is very similar to an URI query string `name1=value1&name2=value2&...&namen=valuen`: the type definition above (the internal portion) will change to accommodate the CRUD operations and potentially an enumerator.
We recommend that the property list will not normally change, especially during send and so the buffers of previous messages can be reused with minimal changes. (Also, we would recommend that the last tuple changes if necessary as it won't require any major string editing.)

For details on what I mean, see az_http_request's query parsing using query_start: https://github.com/Azure/azure-sdk-for-c/blob/25f8a0228e5f250c02e389f19d88c064c93959c1/sdk/core/core/inc/az_http.h#L44

@JeffreyRichter may have a better URI example.

sdk/iot/docs/iot_hub_paho_sample.c
```diff
@@ -0,0 +1,217 @@
+#include <stdio.h>
```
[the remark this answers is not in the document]
CIPop: Right: also, the sample will be placed in another folder and rewritten to use our include files.

sdk/iot/docs/iot_hub_paho_sample.c
```diff
@@ -0,0 +1,217 @@
...
+{
+    int i;
+    char* payloadptr;
+
+    printf("Message arrived\n");
+    printf("     topic: %s\n", topicName);
+    printf("   message: ");
+
+    payloadptr = message->payload;
+    for(i=0; i<message->payloadlen; i++)
```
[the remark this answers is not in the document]
CIPop: Yeah: sample's code is most of paho's code as-is with minimal changes. I forgot to say that I need to do major refactoring and rewrite of the samples.

sdk/iot/docs/iot_hub_paho_sample.c
```diff
@@ -0,0 +1,217 @@
...
+    // Proposed solutions:
+    // 1. use logging from the SDK
+    // 2. create an app function az_iot_handled (similar to az_succeded but that prints out errors)
+    if (az_succeeded(ret = az_iot_c2d_handle(pub_received, &c2d_request)))
+    {
+        // c2d_request contains .payload and .properties / count
+    }
+    else if (az_succeeded(ret = az_iot_methods_handle(pub_received, &method_request)))
+    {
+        printf("Method called: %s with payload_size: %d\n",
```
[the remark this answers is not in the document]
CIPop: I tried adding comments where they are. Spans containing strings will be null terminated in certain cases and not in other based on our documentation. This was the conclusion we've reached with @JeffreyRichter .

sdk/iot/docs/iot_hub_paho_sample.c
```diff
@@ -0,0 +1,217 @@
...
+    // TODO: retry
+}
+
+int main(int argc, char* argv[])
+{
+    MQTTClient client;
+    MQTTClient_connectOptions conn_opts = MQTTClient_connectOptions_initializer;
+    MQTTClient_message pubmsg = MQTTClient_message_initializer;
+    MQTTClient_deliveryToken token;
+    int rc;
```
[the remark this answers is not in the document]
CIPop: Copy/pasted from paho's sample.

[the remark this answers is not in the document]
danewalton: room for improvement from theirs :)

sdk/iot/inc/az_iot_hub_methods.h
CIPop: Wrong copy/paste.

CIPop: Deleted.

sdk/iot/inc/az_iot_hub_properties.h
JeffreyRichter: The C language spec states that undescore uppercase symbols are reserved for the language itself and so, to not conflict with this, we use lowercase for az

sdk/iot/docs/iot_hub_paho_sample.c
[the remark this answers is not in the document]
vhvb1989: `.clang-format` can help you to auto-format this.
I recommend setting IDE to auto-format code on saving

sdk/iot/inc/az_iot_hub_properties.h
CIPop: Based on offline discussion: will remove update and add an enumerator object.

sdk/iot/inc/az_iot_hub_sas.h
CIPop: That's what we named it in other SDKs. It's also internal - apps shouldn't use it.

sdk/iot/inc/az_iot_hub_telemetry.h
CIPop: Based on offline discussion: we'll likely need to change this API completely. Also, we can use `identity` instead of separating `device_id/module_id`

sdk/iot/inc/az_iot_mqtt.h
CIPop: This is already used for receiving: it's a pub message in that case as well.
Offline discussion: do we actually need this?

sdk/iot/inc/az_iot_provisioning_sas.h
CIPop: The links are for offset timing not for RTC. We need a very specific formatted RTC value (UnixTime) and that is obtainable through a well-defined and C99-required interface in time.h if a RTC / NTP system is available.
The lib will need to understand the system epoch then calculate the offset from the one expected by IoT Hub/DPS.

sdk/iot/docs/mqtt_state_machine.md
CIPop: You're right: I forgot to fix the doc after our discussion. There should be 2 separate API calls to get the PATCH vs the GET response for Twin. These are missing.

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+
+Assumptions and recommendations for the application:
+- Our API does not support unsubscribing from any of the previously subscribed topics. We assume that the device will only set-up topics that must be used.
+- It is recommeded that MQTT layer uses the `clean_session == FALSE` during MQTT CONNECT. This allows the application to avoid resubscribing.
+
+## API
+
+### MQTT messages
+
+#### MQTT CONNECT
```
CIPop: Offline discussion: proposal to remove this struct and have _get_user and _get_password instead.

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+    uint8_t qos;
+} az_iot_topic;
+```
+
+#### MQTT PUB
+
+`Note:` this applies for the MQTT PUB either sent or received.
+
+```C
+typedef struct az_iot_mqtt_pub {
```
CIPop: Offline discussion: proposal to remove this struct and have _get_topic instead. Payload will be managed by the application layer entirely:
- telemetry, method_response, twin_patch accept a payload
- all other request-based APIs need an empty payload

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+    } _internal;
+} az_provisioning_client;
+
+void az_provisioning_client_init(az_provisioning_client* client, az_span registration_id);
+
+// IoT Hub
+
+typedef struct az_iot_hub_client {
+    struct {
+        az_span identity;   // deviceId[/moduleId]
```
CIPop: Need to change: deviceId and moduleId need to be stored separately to allow topic names of form `devices/<deviceId>/modules/<moduleId>`

CIPop: Unclear if we need either of the client structs to simply pass the registration_id or device/module strings. We can have methods that accept these strings instead.

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+// Provisioning
+typedef struct az_iot_provisioning_sas {
+    struct {
+        az_span audience; // URLENC("<id_scope>/registrations/<registration_id>"). This is HMAC256-ed together with expiration_time.
+        az_span key_name; // Can be empty.
+    } _internal;
+} az_iot_provisioning_sas;
+
+az_result az_iot_provisioning_sas_init(az_iot_provisioning_sas* sas, az_provisioning_client* client, az_span id_scope, az_span key_name);
+az_result az_iot_provisioning_sas_update(az_iot_provisioning_sas* sas, time_t expiration_time, az_span base64_hmac256_signature, az_iot_mqtt_connect* mqtt_connect);
```
CIPop: Decision to switch from `time_t` to `uint32_t` that means Unix Epoch time (current time in seconds from 1-1-1970). It is the application's responsibility to compute the time.
Question for AZ SDKs: /cc @JeffreyRichter - what API should we use to convert this in to a string of format:
```
audience\n
unixtime\n
```
required by the SAS token builder.

sdk/iot/docs/mqtt_state_machine.md
```diff
@@ -0,0 +1,223 @@
...
+
+Two authentication schemes are currently supported: _X509 Client Certificate Authentication_ and _Shared Access Signature_ authentication. The API must support both cases as well without making assumptions of direct access to secrets (Hardware Security Modules are supported).
+
+```C
+// Provisioning
+az_result az_iot_provisioning_client_init(az_provisioning_client* client, az_span id_scope, az_iot_mqtt_connect* mqtt_connect);
+az_result az_iot_provisioning_get_connect_information(az_iot_hub_client* client, az_span id_scope, az_iot_mqtt_connect* mqtt_connect);
+
+// IoT Hub
+az_result az_iot_hub_client_init(az_iot_hub_client* client, az_iot_mqtt_connect* mqtt_connect);
```
CIPop: Missing APIs: 
`az_iot_*_connect_error_adapter` (converts from CONNACK error status into `az_iot_result`)

sdk/iot/inc/az_iot_provisioning_sas.h
CIPop: We took a decision to use a uint32_t instead of relying on time.h. We expect that the value represents the correct Unix epoch time.

sdk/iot/inc/az_iot_hub_sas.h
CIPop: From https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-tokens I think the correct name is "signature" so I changed to that.

sdk/iot/docs/iot_hub_flow.puml
```diff
@@ -0,0 +1,146 @@
+@startuml
```
danewalton: actually one comment on this docs directory. Would it be possible to put all the "resources" in their own directory and have the top level of `docs/` be only full fledged documentation? So only the readme's in the top level and everything else like pictures, raw `puml` files, etc in `docs/resources`. I fear that as we add more documentation users will have to comb through a lot of "junk" they don't care about before getting to the docs they do care about.

CIPop: Makes sense: as you probably noticed, there are other things there that shouldn't, like samples.

I propose:

```
\readme.md
\docs
    \resources
\samples
    \paho
\inc (flat)
\src (flat)
\test
```

@ewertons is this in-line with what you're proposing? I'd like to keep things alligned in this PR to make it easier for me to rebase after you merge.

danewalton: I like it. The top top level readme explaining the iot directory should go where it is as you just commented. Any supplemental documentation should go in `docs/` and any resources that are used to create/add graphics etc to documentation should go in `docs/resources`

gilbertw: I'm not opposed to this structure.  However it would be the only service with a service level docs folder.
https://azure.github.io/azure-sdk/policies_repostructure.html

Is there a reason for not placing all the documentation in the iot/README.md and not placing the samples in the samples folder?
  /iot/samples/....

CIPop: @gilbertw apologizes, I didn't know that spec existed.

> Is there a reason for not placing all the documentation in the iot/README.md and not placing the samples in the samples folder?

Won't fit: we will have one design doc for each service feature (telemetry, c2d, methods, twin, etc). The one I have posted here is the overall state machine and how-to guide.

From the spec: I don't understand the reason for doc\devdoc to be in root. Is that the place to put our design docs required for devs wanting to contribute? Why is this repo not having one?

> Is there a reason for not placing all the documentation in the iot/README.md and not placing the samples in the samples folder?

Samples will go in /iot/samples. Each app in it's own folder. We will have ports for various MQTT clients as well as platforms (why I plan to have a paho folder).
I will modify the current PR. This was sent out mainly to start the API design & review discussion.

sdk/iot/inc/az_iot_hub_methods.h
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_span.h>
+#include <az_result.h>
+#include "az_iot_mqtt.h"
+#include "az_iot_hub_client.h"
```
antkmsft: In our SDK, we include azure headers first, then put an empty line, and then include the standard library ones, then include _az_cfg. This way, they don't get mixed when auto-formatting sorts them.

Also, why you include other headers before the include guard, is this a merge mistake or intentional? We never had to do this, but maybe you have a good reason.

sdk/iot/inc/az_iot_hub_properties.h
```diff
@@ -0,0 +1,27 @@
...
+typedef struct az_iot_hub_properties {
+    struct {
+        az_span property_string;   // "URIENCODED(name1=val1&name2=val2&...)"
+    } _internal;
+} az_iot_hub_properties;
+
+// TODO: Do we need an enumerator (requires state preserved)
+
+az_result az_iot_hub_properties_add(az_iot_hub_properties *properties, az_span name, az_span value);
+az_result az_iot_hub_properties_read(az_iot_hub_properties *properties, az_span name, az_span value);
```
antkmsft: Almost everywhere, except for the logging logic where we explicitly want to ignore the result, we put `AZ_NODISCARD` before `az_result`, so that if you don't handle the errors, you get warnings/errors that you don't check the return value.

sdk/iot/inc/az_iot_provisioning_register.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef _az_IOT_PROVISIONING_REGISTER_H
+#define _az_IOT_PROVISIONING_REGISTER_H
+
+#include <_az_cfg_prefix.h>
+
+az_result az_provisioning_register_get_subscribe_topic(az_provisioning_client* client, az_iot_topic* mqtt_topic_filter);
+
+#include <_az_cfg_suffix.h>
+
+#endif //!_az_IOT_PROVISIONING_REGISTER_H
```
antkmsft:
```suggestion
#endif //!_az_IOT_PROVISIONING_REGISTER_H

```

sdk/iot/inc/az_iot_hub_methods.h
```diff
@@ -0,0 +1,39 @@
...
+
+az_result az_iot_methods_send_response(az_iot_hub_client* client, az_iot_method_response* response, az_iot_mqtt_publish *mqtt_pub);
+
+az_result az_iot_methods_handle(az_iot_hub_client* client, az_iot_mqtt_publish* pub_received, az_iot_method_request* method_request);
+
+az_result az_iot_hub_methods_send_response(az_iot_hub_client* client, az_iot_mqtt_publish* pub_received, az_iot_method_response* method_request);
+
+#include <_az_cfg_suffix.h>
+
+#endif //!_az_IOT_METHODS_H
```
antkmsft: nit: our #endif comments for include guards look as `#endif // _az_HEADER_H`

sdk/iot/inc/az_iot_hub_sas.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_span.h>
+#include <az_result.h>
+#include <time.h>
+#include "az_iot_mqtt.h"
```
antkmsft: We never use quote-style includes in public or internal header files, only in the private headers, and in sources (so, basically only in files that are in the `src` directory). Jeffrey has a strong opinion about this, that compilers are free to interpret quote-includes however they wish, and not necessary to start looking in the current folder, so you're not gaining anything from using quote includes, so our approach is to use angle bracket includes, unless you have to use the quoted ones, and, again, we don't have a single quote-style-include in our public headers.

antkmsft: Also, I mentioned this above, we'd
```
#include <az_iot_hub_client.h>
#include <az_iot_mqtt.h>
#include <az_span.h>
#include <az_result.h>

#include <time.h>

#include <_az_cfg_prefix.h>
```

sdk/iot/inc/az_iot_mqtt.h
```diff
@@ -0,0 +1,29 @@
...
+
+typedef struct az_iot_mqtt_connect {
+    az_span client_id;  // NULL terminated string.
+    az_span user;       // NULL terminated string.
+    az_span password;   // NULL terminated string (may be an empty string).
+} az_iot_mqtt_connect;
+
+typedef struct az_iot_topic {
+    az_span name;       // NULL terminated string.
+    uint8_t qos;
```
antkmsft: include <stdint.h>

sdk/iot/inc/az_iot_provisioning_register.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <az_span.h>
```
antkmsft: you don't need span.h here

sdk/iot/inc/az_iot_hub_methods.h
CIPop: Addressed in the new PR.

sdk/iot/inc/az_iot_hub_methods.h
CIPop: Umm... but that's logically wrong? The #if is for a negation - why do they end in something else?

sdk/iot/inc/az_iot_hub_sas.h
CIPop: Not relevant for the next commit as we've coalesced all .h into 2 files for the two services we support.
Makes sense and I'll keep that in mind.

sdk/iot/inc/az_iot_mqtt.h
CIPop: I missed this in the new PR. Will update.

CIPop:
> I strongly recommend to apply all the suggestions I listed in this PR (AZ_NODISCARD everywhere, include sections, quoted includes, include guard position), and let's talk if you disagree; but I won't prevent your PR from being checked in for any of these reasons.

@antkmsft please do point them out in the new PR if I missed any and do block the PR from merging until it complies with all rules.

> Also we now are working to cover all our public headers with doxygen comments, and I am not sure if your code is in our repo now, is it also considered public API, and if so, if lack of doxygen comments would block us from declaring being ready to public; But my guess is even in that case you still have some time, probably up to a couple of weeks from today, to fix it.

I've tried my best to document. If the standard documentation format isn't correct, please point out unique patterns only and I'll apply the changes to the entire document. Also do point out any tools I could use to automatically generate the expected doxygen formatting (if any available).

CIPop:
Obsoleted by #409

sdk/iot/inc/az_iot_hub_methods.h
antkmsft: well... I think it is rather #else that does the negation, but then in the endif would you rather put the negated comment for the #if section, or #else section, or for #elif one?. It is always going to be hard to fully describe the condition you're negating, especially with the cascade of #elifs

So I think the goal is to clearly mark the block to where #endif corresponds to , and just plainly repeating the condition does the job most effectively with minimum visual clutter.
You could see in the _az_cfg.h what I did with the cascades of endifs - for every next #endif , sort of trying to list the things that we already excluded - not necessary in the compilable way but in a way that is recognizable to the numan eye while bringing the least visual clutter.

Also, for what it's worth, for C++ clangformat, when you have nested namespaces right within each other, and choose to use use no indentation, clangformat is going to put '} // namespace azure' after the closing bracket, and it does not use any negation syntax (https://github.com/Azure/azure-sdk-for-cpp/blob/master/sdk/core/azure-core/inc/azure.hpp).

It is up to you, what appeals to your aesthetical taste, but currently that would be inconsistent to what we use everywhere else. I am fine to change it - if you send a PR changing all the endif comments I'll approve it; but again, I personally don't think it is logically consistent for #endifs, I don't think it is possible to always have logically perfect comment for the end of every block, and then I don't think it is worth the effort to pursue that consistency rather than simply visually marking the block you are ending, while bringing the minimum comment clutter. If you feel strong about using the negation, please go for it, because I realistically could read it either way and really the only thing that slightly inconveniences me is inconsistency when using/not using '!', not so much whether we use it or not.

PR #371, opened 2020-02-20
Change az_span_null() to AZ_SPAN_NULL

danewalton:
The function call to `az_span_null()` can't be used in a constant expression for static span arrays. Example:
```c
static az_span twin_sub_topics[] = (az_span[]){
    AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/#"),
    AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/#"),
    az_span_null()
};

//This however will work
static az_span twin_sub_topics[] = (az_span[]){
    AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/#"),
    AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/#"),
    AZ_SPAN_NULL
};
```

Thoughts?


danewalton:
I think @barcharcraz is probably the guy to ask about the C++ behavior. I'm definitely no expert there. 

As for C99, since we are in theory a C99 compliant sdk I think we should be okay. I'll leave others to offer their definitive opinion though.

antkmsft:
> As for C99, since we are in theory a C99 compliant sdk I think we should be okay.
I agree, but modern GCC and Clang would warn you when you use perfectly legal C99 syntax ( '(az_span){ 0 }'), and would require '-Wmissing-field-initializers' and '-Wmissing-braces'. The reason you don't see them in our code is because '_az_cfg.h' disables it in our code.

barcharcraz:
I think if it expands to a designated initializer initializing every field that should be fine in c and c++, except that c++ doesn't have compound literals, so I'm not sure about the cast syntax, and I don't think (but could be wrong) `az_span{.size = ....}` is valid C.

In any case any AZ_SPAN_NULL macro is going to have exactly the same problems on the cxx compat front as AZ_SPAN_LITERAL_FROM_STR so we're not exactly making the situation worse

danewalton:
I believe @JeffreyRichter agreed it is a worthwhile change in a meeting we had today.

JeffreyRichter:
Yes, I agree that it is good. We should also do this for null pair. Let's accept the pr

sdk/core/core/inc/az_span.h
```diff
@@ -55,7 +55,7 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
  * @brief Creates an empty span
  *
  */
-AZ_NODISCARD AZ_INLINE az_span az_span_null() { return (az_span){ 0 }; }
+#define AZ_SPAN_NULL (az_span){ 0 }
```
antkmsft: Since az_span is defined in the same file, I'd again, define AZ_SPAN_NULL as

```
#include <stddef.h>
#define AZ_SPAN_NULL (az_span){ .ptr = NULL, .length = 0, .capacity = 0 }
```

because should you use AZ_SPAN_NULL when `_az_cfg.h` is not included (or, not between the `_az_span_prefix.h` and `_suffix.h`), GCC and Clang are going to produce warnings (which they shouldn't, but it's how the things currently are).

In any case, the final decision is up to you, if you considered that, and still think we should keep `{ 0 }` I am fine with the change either way. What do you need to complete this change? If you don't have access to the "Complete Pull Request" button, let me know, I can complete it.

danewalton: updated to use this per-member init. I can't merge so if you're good with it feel free to merge when appropriate.

antkmsft: I just realized recently that you can't merge, - will do once CI/CD passes!
BTW, it looks like you need to update the branch.

antkmsft: @danewalton I just realized recently that you can't merge, - will do once CI/CD passes!
BTW, it looks like you need to update the branch.

danewalton: yup currently fixing merge conflicts. will push an update

sdk/core/core/inc/az_span.h
```diff
@@ -55,7 +55,7 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
  * @brief Creates an empty span
  *
  */
-AZ_NODISCARD AZ_INLINE az_span az_span_null() { return (az_span){ 0 }; }
+#define AZ_SPAN_NULL (az_span){ ._internal.ptr = NULL, ._internal.length = 0, ._internal.capacity = 0 }
```
antkmsft: Nit: you can initialize as (az_span){ ._internal = {.ptr = NULL, .length = 0, .capacity = 0 }}

antkmsft: And also I think you need to include <stddef.h> for NULL

sdk/transport_policies/curl/test/main.c
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
danewalton: @antkmsft were these supposed to be deleted? i think my merge conflict resolution brought it back

danewalton: removed

antkmsft: You're right, sdk/transport_policies/curl is now platform/http_client/curl. But we deleted the test .c for az_curl lib entirely.

sdk/core/core/inc/az_span.h
antkmsft: Thanks for applying my feedback!

antkmsft:
Boom, merged :)

PR #374, opened 2020-02-20
PAL Layer (and Curl) + minor changes

antkmsft:


sdk/core/core/inc/az_result.h
```diff
@@ -17,8 +17,9 @@
 
 enum {
   AZ_FACILITY_CORE = 0x1,
-  AZ_FACILITY_JSON = 0x2,
-  AZ_FACILITY_HTTP = 0x3,
+  AZ_FACILITY_PLATFORM = 0x2,
+  AZ_FACILITY_JSON = 0x3,
+  AZ_FACILITY_HTTP = 0x4,
```
JeffreyRichter: IOT and MQTT facilities should be here; did they get lost in a merge?

sdk/core/core/internal/az_config_internal.h
```diff
@@ -21,6 +21,7 @@ enum {
 enum {
   _az_TIME_SECONDS_PER_MINUTE = 60,
   _az_TIME_MILLISECONDS_PER_SECOND = 1000,
+  _az_TIME_MICROSECONDS_PER_MILLISECOND = 1000,
```
JeffreyRichter: Do we really need this? Especially if we're not going to have 64-bit integer resolution

sdk/keyvault/keyvault/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ target_link_libraries(az_keyvault PRIVATE az_core_internal)
 add_library (az::keyvault ALIAS az_keyvault)
 
 add_executable (az_keyvault_test test/main.c)
-target_link_libraries(az_keyvault_test PRIVATE az_keyvault az_core_internal)
+target_link_libraries(az_keyvault_test PUBLIC az_keyvault az_core_internal)
```
JeffreyRichter: I think we have az_core now; not az_core_internal

sdk/keyvault/keyvault/samples/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ set(CMAKE_C_STANDARD 99)
 
 # make keys client example
 add_executable (keys_client_example src/keys_client_example.c)
-target_link_libraries(keys_client_example PRIVATE az_keyvault az_core az_core_internal az_curl)
+
+target_link_libraries(keys_client_example PRIVATE az_keyvault az_core az_core_internal)
```
JeffreyRichter: OK, I think this just doesn't have the changes I made with Victor. Victor will fix this with his PR

sdk/core/core/inc/az_result.h
antkmsft: I saw their PR, it is not yet checked in, I think they will merge now

sdk/keyvault/keyvault/samples/CMakeLists.txt
antkmsft: My understanding is that we still need this - core is public headers, and you need core_internal whenever you need to use internal headers in your code.

sdk/core/core/internal/az_config_internal.h
antkmsft: I know, but let's get rid of the 64 bits, and update this as part of that PR.

sdk/keyvault/keyvault/CMakeLists.txt
antkmsft: My understanding is that we still need this - core is public headers, and you need core_internal whenever you need to use internal headers in your code.

PR #384, opened 2020-02-25
remove -Wdocumentation-unknown-command

danewalton:
`-Wdocumentation-unknown-command` flags the doxygen tag `@retval` as an invalid command when it is indeed a valid tag that we should be using in our documentation. We should remove it to unblock good docs.


danewalton:
@antkmsft you are right thanks for pointing that out. I have updated it accordingly.

PR #398, opened 2020-02-27
Headers count

gilbertw:
Add helper around the headers count
Add comments noting that headers is an span of az_pair(s)

### DIFF
--- a/sdk/core/core/inc/az_http.h
+++ b/sdk/core/core/inc/az_http.h
@@ -320,17 +320,6 @@ az_http_response_init(az_http_response* self, az_span http_response)
 }
 
 /**
- * @brief Returns the count of headers on the request
- *        Each header is an az_pair
- *
- */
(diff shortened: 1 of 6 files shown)


sdk/core/core/inc/az_http.h
```diff
@@ -33,6 +33,12 @@ typedef enum
  */
 typedef az_span az_http_method;
 
+/**
+ * @brief Define az_http_headers as an az_span so it is limited to that type and not to any az_span
+ *
+ */
+typedef az_span az_http_headers;
```
antkmsft: az_http_headers is only used internally, and is located in _internal struct - if you still want to go with the typedef, I think we should add an underscore `typedef az_span _az_http_headers;`

gilbertw: That's a good point, I'll switch to that

antkmsft: Are you also going to move the function to internal header?

PR #399, opened 2020-02-27
move log buffer size to user config

danewalton:
1024 could be a large size for RTOS thread stack usage. Any way we could move this to a user config?


JeffreyRichter:
The logs are owned by the customer. So, I think it's ok if the customer makes the buffer small. In other words, I don't think we should enforce a minimum size (after all, a customer can turn logging off entirely and this is the default).

Making the log buffer size a config option makes sense to me.

PR #409, opened 2020-02-29
IoT and Provisioning client APIs and documentation.

CIPop:
1. Include files containing IoT Hub and IoT Provisioning public API after the initial review sessions.
2. Readme.md with the MQTT state machine that application developers need to follow to use the IoT Core SDK.

Addresses (most) comments and replaces #362.


sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,525 @@
...
+ *
+ * @param client
+ * @param token_expiration_unix_time The time, in seconds, from the Unix epoch (1/1/1970).
+ * @param signature An empty span with sufficient capacity to hold the SAS signature.
+ * @param out_signature The output span containing the SAS signature.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_hub_client_sas_signature_get(
+    az_iot_hub_client const* client,
+    uint32_t token_expiration_unix_time,
```
CIPop: From the other PR: I'll add #include <stdint.h>

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,525 @@
...
+
+/**
+ * @brief Initializes the Telemetry or C2D properties.
+ *
+ * @param properties
+ * @param buffer Can either be an empty span or a span containing properly formatted properties.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result
+az_iot_hub_client_properties_init(az_iot_hub_client_properties* properties, az_span buffer);
```
hihigupt: >az_iot_hub_client_properties_init [](start = 0, length = 33)

Is it possible to add some flag to this API, which tells that buffer used under this structure does not starts with first property.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * @brief Attempts to parse a received message's topic.
+ *
+ * @param client
+ * @param received_topic A span containing the received topic.
+ * @param received_payload A span containing the received topic.
+ * @param out_response If the message is register-operation related, this will contain the
+ *                     #az_iot_provisioning_client_register_response.
+ * @return AZ_NODISCARD az_iot_provisioning_register_topic_payload_parse
```
danewalton: `#az_result`

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,525 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
danewalton: needs the `@file` tag described in provisioning header

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,351 @@
...
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic);
+
+/**
+ * @brief Gets the MQTT topic that must be used to submit a Register Status request.
+ * @note The payload of the MQTT publish message should be empty.
+ *
+ * @param client
+ * @param request_id The request id. Must match a received
+ * #az_iot_provisioning_client_register_response request_id.
```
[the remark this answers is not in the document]
CIPop: I thought it'll make it look better but now I think it actually looks worse. Will leave it as-is for now.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,351 @@
...
+AZ_NODISCARD bool az_iot_provisioning_client_is_retriable_status(
+    az_iot_provisioning_client_status status);
+
+/**
+ * @brief Gets the MQTT topic filter to subscribe to register responses.
+ *
+ * @param client
+ * @param mqtt_topic_filter An empty span with sufficient capacity to hold the MQTT topic filter.
+ * @param out_mqtt_topic_filter The output span containing the MQTT topic filter.
+ * @return AZ_NODISCARD az_iot_provisioning_client_register_subscribe_topic_filter_get
```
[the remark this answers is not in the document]
CIPop: The auto doxygen generator is not able to figure out the AZ_NODISCARD.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,351 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_IOT_PROVISIONING_CLIENT_H
+#define _az_IOT_PROVISIONING_CLIENT_H
+
+#include <az_result.h>
```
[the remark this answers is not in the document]
CIPop: I've also switched from `int` to actual sizes and included stdint.h

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -16,17 +23,9 @@
...
-   */
-  az_span module_id;
-
-  /**
-   * @brief The user-agent is a formatted string that will be used for Azure IoT usage statistics.
-   *
-   */
-  az_span user_agent;
+  az_span module_id; /** @brief The module name (if a module identity is used). */
+  az_span user_agent; /** @brief The user-agent is a formatted string that will be used for Azure
```
ericwol-msft: do we need to say that this is optional?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+ *
+ * @param[in] client
+ * @param[in] token_expiration_unix_time The time, in seconds, from the Unix epoch (1/1/1970).
+ * @param[in] signature An empty #az_span with sufficient capacity to hold the SAS signature.
+ * @param[out] out_signature The output #az_span containing the SAS signature.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_hub_client_sas_signature_get(
+    az_iot_hub_client const* client,
+    uint32_t token_expiration_unix_time,
```
jspaith: >token_expiration_unix_time [](start = 13, length = 26)

Nit: epoch time instead of unix time?  I try to avoid hard-coding product/trademarky stuff when possible.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+  {
+    az_span properties;
+    uint8_t* current_property;
+  } _internal;
+} az_iot_hub_client_properties;
+
+/**
+ * @brief Initializes the Telemetry or C2D properties.
+ *
+ * @param properties
```
jspaith: >param [](start = 4, length = 5)

Nit: param beyond this point are missing the [in]/[out] marker in doxygen.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+ */
+
+#ifndef _az_IOT_HUB_CLIENT_H
+#define _az_IOT_HUB_CLIENT_H
+
+#include <az_result.h>
+#include <az_span.h>
+#include <stdbool.h>
+#include <stdint.h>
+
```
hihigupt: Can we order headerfile in following order:

In dir/foo.cc or dir/foo_test.cc, whose main purpose is to implement or test the stuff in dir2/foo2.h, order your includes as follows:

dir2/foo2.h.
A blank line
C system headers (more precisely: headers in angle brackets with the .h extension), e.g. <unistd.h>, <stdlib.h>.
A blank line
C++ standard library headers (without file extension), e.g. <algorithm>, <cstddef>.
A blank line
Other libraries' .h files.
Your project's .h files.
Separate each non-empty group with one blank line.


excerpt:
[https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Yes: sorry about that - had to interrupt my work (why I've added "WIP - work in progress"). Will update all.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: We need to document the flow. According to @JeffreyRichter the way this is meant to be used is:
1. Application calls `az_iot_hub_client_options_default()`
2. Application makes changes within the `az_iot_hub_client_options` struct.

Filling in an uninitialized struct should not be supported.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+ * @param global_endpoint_hostname The Global device endpoint.
+ * @param id_scope The ID Scope.
+ * @param registration_id The Registration ID. This must match the client certificate name (CN part
+ *                        of the certificate subject).
+ * @param options A reference to an #az_iot_provisioning_client_options structure. Can be NULL.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_init(
+    az_iot_provisioning_client const* client,
+    az_span global_endpoint_hostname,
```
hihigupt: >global_endpoint_hostname [](start = 12, length = 24)

hostname?

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+ * @param client
+ * @param global_endpoint_hostname The Global device endpoint.
+ * @param id_scope The ID Scope.
+ * @param registration_id The Registration ID. This must match the client certificate name (CN part
+ *                        of the certificate subject).
+ * @param options A reference to an #az_iot_provisioning_client_options structure. Can be NULL.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_init(
+    az_iot_provisioning_client const* client,
```
hihigupt: >az_iot_provisioning_client const* client [](start = 4, length = 40)

Here you are initializing the client (write operation), but type is pointer to const  client ?

danewalton: yes the pointer asterisk should be on the struct not const:
[C Gibberish](https://www.cdecl.org/?q=int+const*+az_client)

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+    az_span* out_mqtt_topic_filter);
+
+/**
+ * @brief The registration operation state.
+ * @note This is returned only when the operation completed.
+ *
+ */
+typedef struct az_iot_provisioning_client_registration_state
+{
+  az_span assigned_hub_hostname; /** @brief Assigned Azure IoT Hub hostname. @note This is only
```
danewalton: Note on the comment style in my previous comment:
```
/**< ......  */
```
There should be an angle bracket after the second asterisk and then closing at the end. Also no need for the `@brief` tag.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+ */
+AZ_NODISCARD az_result az_iot_hub_client_telemetry_publish_topic_get(
+    az_iot_hub_client const* client,
+    az_iot_hub_client_properties const* properties,
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic);
+
+/**
+ *
+ * Cloud to device (C2D) APIs
```
jspaith: >Cloud to device (C2D) APIs [](start = 3, length = 26)

Should all of this live in one header file?  *In theory* the linker optimizer will be smart enough to cleanup up.  But if stuff is in separate .c/.h makes both MS and customer be more thoughtful RE what they're dragging it at any stage?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+#include <_az_cfg_prefix.h>
+
+/**
+ * @brief Azure IoT Hub Client options.
+ *
+ */
+typedef struct az_iot_hub_client_options
+{
+  az_span module_id; /** @brief The module name (if a module identity is used). */
+  az_span user_agent; /** @brief The user-agent is a formatted string that will be used for Azure
```
jspaith: >/** @brief The user-agent is a formatted string that will be used for Azure [](start = 22, length = 75)

Could we (or should we?) put the doxygen comments on a separate line as the actual variable itself?  To my eye this feels more readable.

sdk/iot/core/inc/az_iot_hub_client.h
danewalton: I think a lot of the original plans for v2 were shooting for single headers. uLib even started making a monolithic `<thing>_api.h` header file which has in it everything the user needs to get going. At that point they can choose to use whatever they want. Actually with that: any ideas on whether we should change the name to `az_iot_hub_api.h`? And provisioning to `az_iot_provisioning_api.h`? Pardon me for the continued name discussions. Feel free to dismiss if its too much

sdk/iot/core/inc/az_iot_hub_client.h
danewalton: The main problem I have with this is the following:
The `.clang_format` we currently have limits indentations to 2 spaces instead of four. With the documentation on it's own line, it becomes somewhat difficult as the vertical space the struct declaration takes up expands to track where the struct starts and ends. There isn't a big difference between the left justified alignment on the beginning and two spaces in for each member. Some members start to look like global variable declarations. This at least tries to keep it compact.

sdk/iot/core/inc/az_iot_hub_client.h
JeffreyRichter: Yes, Chris is right. This is the flow. This flow eases versioning in the future. That is, we can add new fields to the options structure; the *_default() function will know about those fields and initialize ALL Fields (even the new fields). The customer can override the fields that they knew about but leave fields they don't know about as they were initialized by the _default() function. We do NOT support the pattern of a 0-initialized options structure.

For things like retry policies; using 0 for max-retries or 0 for timeouts does not work. So the _default() function initializes these fields with good defaults. This pattern also allows a customer to see our default values before passing the options to a XxxClient init function.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+    az_span mqtt_topic_filter,
+    az_span* out_mqtt_topic_filter);
+
+/**
+ * @brief The Cloud To Device Request.
+ *
+ */
+typedef struct az_iot_hub_client_c2d_request
+{
+  az_iot_hub_client_properties
```
jspaith: >az_iot_hub_client_properties [](start = 2, length = 28)

Nit: properties variable is on a different line.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+  AZ_IOT_HUB_CLIENT_STATUS_PRECONDITION_FAILED = 412,
+  AZ_IOT_HUB_CLIENT_STATUS_REQUEST_TOO_LARGE = 413,
+  AZ_IOT_HUB_CLIENT_STATUS_UNSUPPORTED_TYPE = 415,
+  AZ_IOT_HUB_CLIENT_STATUS_THROTTLED = 429,
+  AZ_IOT_HUB_CLIENT_STATUS_CLIENT_CLOSED = 499,
+  AZ_IOT_HUB_CLIENT_STATUS_SERVER_ERROR = 500,
+  AZ_IOT_HUB_CLIENT_STATUS_BAD_GATEWAY = 502,
+  AZ_IOT_HUB_CLIENT_STATUS_SERVICE_UNAVAILABLE = 503,
+  AZ_IOT_HUB_CLIENT_STATUS_TIMEOUT = 504,
+} az_iot_hub_client_status;
```
jspaith: >az_iot_hub_client_status [](start = 2, length = 24)

What happens if we get a status code from IoThub that is not explicitly defined in this range?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+  AZ_IOT_HUB_CLIENT_STATUS_NOT_CONFLICT = 409,
+  AZ_IOT_HUB_CLIENT_STATUS_PRECONDITION_FAILED = 412,
+  AZ_IOT_HUB_CLIENT_STATUS_REQUEST_TOO_LARGE = 413,
+  AZ_IOT_HUB_CLIENT_STATUS_UNSUPPORTED_TYPE = 415,
+  AZ_IOT_HUB_CLIENT_STATUS_THROTTLED = 429,
+  AZ_IOT_HUB_CLIENT_STATUS_CLIENT_CLOSED = 499,
+  AZ_IOT_HUB_CLIENT_STATUS_SERVER_ERROR = 500,
+  AZ_IOT_HUB_CLIENT_STATUS_BAD_GATEWAY = 502,
+  AZ_IOT_HUB_CLIENT_STATUS_SERVICE_UNAVAILABLE = 503,
+  AZ_IOT_HUB_CLIENT_STATUS_TIMEOUT = 504,
```
jspaith: >AZ_IOT_HUB_CLIENT_STATUS_TIMEOUT [](start = 2, length = 32)

Bigger question I know, but what happens if we need to add a new status code here, in terms of versioning of enum?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+typedef struct az_iot_hub_client_twin_response
+{
+  az_iot_hub_client_twin_response_type response_type; /** @brief Twin response type. */
+  az_iot_hub_client_status status; /** @brief The operation status. */
+  az_span request_id; /** @brief Request ID matches the ID specified when issuing a Get or Patch
+                         command. */
+  az_span version; /** @brief The Twin object version.
+                    * @note This is only returned when
+                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_DESIRED_PROPERTIES or
+                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES */
```
jspaith: >AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES [](start = 37, length = 52)

I thought the $version was only stamped globally to the twin and on desired?  What is meaning of it for a reported?

jspaith: Ohh - or is this from mqtt topic itself, not the payload?  If so would be interested on naming.

---
In reply to: [387280421](https://github.com/Azure/azure-sdk-for-c/pull/409#discussion_r387280421) [](ancestors = 387280421)

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
+    az_iot_hub_client const* client,
+    az_span request_id,
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic);
+
+/**
+ * @brief Gets the MQTT topic that must be used to submit a Twin GET request.
```
jspaith: >Gets the MQTT topic that must be used to submit a Twin GET request. [](start = 10, length = 67)

Comment is a copy paste of line 468 for az_iot_hub_client_twin_get_publish_topic_get?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+{
+  az_iot_hub_client_twin_response_type response_type; /** @brief Twin response type. */
+  az_iot_hub_client_status status; /** @brief The operation status. */
+  az_span request_id; /** @brief Request ID matches the ID specified when issuing a Get or Patch
+                         command. */
+  az_span version; /** @brief The Twin object version.
+                    * @note This is only returned when
+                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_DESIRED_PROPERTIES or
+                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES */
+} az_iot_hub_client_twin_response;
```
jspaith: >az_iot_hub_client_twin_response [](start = 2, length = 31)

Is "response" the right word here (and other places we get the functions with "response" in function name).  It makes sense for the simple GET_TWIN/Response scenario.  But is this the same structure used when a delta arrives, correct?  This feels less like a response, which I think of 1-1 with request.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+ *
+ * @param client
+ * @param request_id The request id.
+ * @param if_match_version Can be either "*" to overwrite or the twin version to limit the reported
+ *                         properties patch to that version only.
+ * @param mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
+ * @param out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_hub_client_twin_patch_publish_topic_get(
```
jspaith: >az_iot_hub_client_twin_patch_publish_topic_get [](start = 23, length = 46)

Again if I'm misremembering set me straight, but don't we need a topic for the subscribe for deltas disttinct from what we have?

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -0,0 +1,505 @@
...
+    az_span* out_mqtt_topic);
+
+/**
+ * @brief Gets the MQTT topic that must be used to submit a Twin GET request.
+ * @note The payload of the MQTT publish message should contain a JSON document formatted according
+ *       to the Twin specification.
+ *
+ * @param client
+ * @param request_id The request id.
+ * @param if_match_version Can be either "*" to overwrite or the twin version to limit the reported
```
jspaith: >if_match_version [](start = 10, length = 16)

Assuming app doesn't want do blast with the "*", does app get this from az_iot_hub_client_twin_response or ??  Someplace need to doc how this gets correlated up.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+  az_iot_provisioning_client_status status; /** @brief The current request status.
+                                             * @note The response for the register operation is
+                                             * available through #registration_state only. */
+  az_span request_id; /** @brief Request ID for the register operation.
+                       * @note This must match all messages associated with the registration
+                       * operation. */
+  az_span registration_state; /** @brief An #az_span containing the state of the register operation.
+                               * @details This can be one of the following: 'unassigned',
+                               * 'assigning', 'assigned', 'failed', 'disabled'. */
+  uint32_t
```
jspaith: >uint32_t [](start = 2, length = 8)

Nit - this and some other vars here don't align.

sdk/iot/core/doc/resources/iot_hub_flow.puml
```diff
@@ -0,0 +1,120 @@
+@startuml
+
```
jspaith: Were you planning on shipping these to customer?  If yes, then we should talk with Eric (if not) because we'll need to have folks maintain it when changes happen.

My thought is sample code is easier to follow and something that we can put gates up to check if it's screwed up.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * @brief definition for Azure Device Provisioning Service device-client SDK.
+ */
+
+#ifndef _az_IOT_PROVISIONING_CLIENT_H
+#define _az_IOT_PROVISIONING_CLIENT_H
+
+#include <az_result.h>
+#include <az_span.h>
+#include <stdbool.h>
```
antkmsft: use a single empty line between az_ includes and the ones from stdlib, then auto-format (Ctrl+K+D in VS), and it is going to place az_span before the az_result. without an empty line, it would sort both our headers and stdlib headers alphabetically, placing <assert.h> before <az_span.h>.

sdk/iot/core/inc/az_iot_provisioning_client.h
```diff
@@ -0,0 +1,310 @@
...
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_PRECONDITION_FAILED = 412,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_REQUEST_TOO_LARGE = 413,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_UNSUPPORTED_TYPE = 415,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_THROTTLED = 429,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_CLIENT_CLOSED = 499,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_SERVER_ERROR = 500,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_BAD_GATEWAY = 502,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_SERVICE_UNAVAILABLE = 503,
+  AZ_IOT_PROVISIONING_CLIENT_STATUS_TIMEOUT = 504,
+} az_iot_provisioning_client_status;
```
antkmsft: In az_http.h, we have az_http_status_code, do you want to use that?

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: I'm trying to match the Azure Portal. I will change to `global_device_endpoint` : 
image

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: Good catch: replace all issue. I'll remove the pointer to const.

@danewalton I've interpreted what you said as having a const pointer to a non-const struct instead of the current (obviously wrong) pointer to a const struct. (Please don't write C in English ever again :)) ).  It sounds like a good idea to require the pointers to be const but it should be further discussed as it would require changes everywhere (i.e. why not have const pointers to const data?). You may want to open another issue.

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: Ah! Way better.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: @jspaith We've already discussed and decided offline on only 2 header files. This is in-line with AzRTOS as well as other RTOSes. @tawalke was actually debating if we should have only a single .h file for all provided services.

@danewalton we've also settled on having "client" within the name. I'll let Az SDKs comment if we should change.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: 1. I kind of agree with @jspaith _if_ we can move them outside of the struct. 
1. @danewalton made a comment above about using > instead of @brief ,etc. That should make things a bit more compact.

I'll defer to @danewalton since he's the doc expert. What do you think about using: https://stackoverflow.com/questions/7199151/use-doxygen-to-document-members-of-a-c-structure-outside-of-the-structure-defini ?

```
/** @struct foreignstruct
 *  @brief This structure blah blah blah...
 *  @var foreignstruct::a 
 *  Member 'a' contains...
 *  @var foreignstruct::b 
 *  Member 'b' contains...
 */
```

Azure SDKs - what's your take on this? I couldn't find any sample of a public struct member doxygen documentation.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Yes: that's the auto-clang format. It's because of the same line doc (related to above comment). I don't like it either.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: I thought about 2 scenarios:

1. Apps may choose to use the enums. In that case the app will likely not be able to handle the response and crash. That would be a service breaking change.
1. Apps will use the provided `az_iot_hub_client_is_success_status` and `az_iot_hub_client_is_retriable_status` APIs. Those should be implemented using ranges instead of actual codes and so the apps should work without any changes.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Why would that be a problem? It's additive and we don't support binary app-compat.
Existing apps will continue to compile, new apps can take a dependency.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Yes: this is from topic, _not_ payload. I think it's actually called "version" in the topic.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: I agree... I first had push vs req/resp separate but it complicates the programming model.
What would a better name be?

sdk/iot/core/doc/resources/iot_hub_flow.puml
CIPop: Yes: I do plan to have these maintained.

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: We brought that up in the review with @JeffreyRichter and the decision was that we want to separate the status codes. These are IoT Hub / Provisioning specific and come over proprietary protocol over MQTT. (I don't have any data about this but it feels better to consider the values subject to change in future versions of our services.)

sdk/iot/core/inc/az_iot_hub_client.h
jspaith: 1 - What happens if we go through a proxy (or a load balancer or a ... IoTHub doesn't control) at some point?  Could the proxy return an error code outside the error range in the enum list?
2 - Even if 1 isn't an issue, is service team onboard with fact that these errors are a strict contract and they can't add new ones?

jspaith: Or actually I may be thinking too HTTP in my comment above, since this is ultimately MQTT land and any sort of proxy error would be a web error at the client's transport layer, correct?

sdk/iot/core/inc/az_iot_hub_client.h
jspaith: I see.  Only problem if an app does like a switch() without a default, which they shououldn't do anyway.

sdk/iot/core/inc/az_iot_hub_client.h
jspaith: I'm struggling to come up with a better name.  How much did it complicate the model?  Did we ever have to move fields from a _req to a _resp, e.g., only because of different names? -- which would be bad.  Or is it just more structs?

sdk/iot/core/doc/resources/iot_hub_flow.puml
ericwol-msft: I would love to get to a point were we maintain things like this

jspaith: Cool - if you guys are tracking this, I'm have less concern it'll become abandonware.  Thanks

sdk/iot/core/inc/az_iot_provisioning_client.h
JeffreyRichter: Yes, these should be separate; it is an implementation coincidence that they match HTTP status codes. Also, it would not look right to compare status codes returned via MQTT with az_http_status_code.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Following AZ SDKs guidelines. I did insert a blank line per their guidance.

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: Actually FYI: mine didn't work. Changed all to /**< and tested with actual doxygen.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: I moved them to the way @danewalton recommended for now. Please open a separate issue if we want to change again.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Fixed now using /**<

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Right: the md file is setting clear expectations on us not supporting WebSocket at this layer.
That said, Edge is a reasonable alternative that would enable 1. 
Still my thinking is: apps that have code testing for a particular value probably only want to address that particular value. 
Most apps should rely on the two APIs I mention. Those should be implemented to cover the entire uint16_t range, not just 100-500s.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: The real complication is in adding one more handler for incoming messages: one handling request/responses and one handling twin patches.

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: There are 5 APIs in total here:
2x SUB topics - one for responses (from both GET and reported PATCH) and one for deltas (desired PATCH initiated from the server). (`az_iot_hub_client_twin_response_subscribe_topic_filter_get`, `az_iot_hub_client_twin_patch_subscribe_topic_filter_get` respectively)
1x for processing incoming messages `az_iot_hub_client_twin_received_topic_parse`
2x PUB topics - one for GET and one for PATCH. (`az_iot_hub_client_twin_get_publish_topic_get` and `az_iot_hub_client_twin_patch_publish_topic_get` respectively - again in need for some better names).

sdk/iot/core/inc/az_iot_hub_client.h
CIPop: Added a reference in the documentation.

sdk/iot/core/inc/az_iot_provisioning_client.h
CIPop: I'm using the clang auto-formatter. Since the names are long, there aren't a lot of ways to massage this better. Let's take it offline and find a nicer solution.

sdk/iot/core/doc/resources/iot_hub_flow.puml
CIPop: Looks like the latest doxygen can use puml. http://www.doxygen.nl/manual/commands.html#cmdstartuml

CIPop:
@JeffreyRichter @jspaith @ericwol-msft @joshfree @gilbertw this is blocked waiting on two of you signing off. It's unclear if you are waiting on any of the comments. If so, please reply at the bottom of the PR with what those issues are.

PR #413, opened 2020-03-03
Retry policy

antkmsft:
This PR replaces the previous one that was based on the old sources (https://github.com/Azure/azure-sdk-for-c/pull/320).
I think I've covered all the feedback from the previous one.


sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,228 @@
...
+  return (az_http_policy_retry_options){
+    .max_retries = 4,
+    .retry_delay_msec = 4 * _az_TIME_MILLISECONDS_PER_SECOND, // 4 seconds
+    .max_retry_delay_msec
+    = 2 * _az_TIME_SECONDS_PER_MINUTE * _az_TIME_MILLISECONDS_PER_SECOND, // 2 minutes
+    .status_codes = _default_status_codes,
+  };
+}
+
+AZ_NODISCARD bool az_http_policy_retry_options_validate(
```
JeffreyRichter: I don't think we need this function

sdk/core/core/internal/az_http_internal.h
```diff
@@ -170,6 +172,15 @@ az_http_request_set_query_parameter(_az_http_request* p_request, az_span name, a
 AZ_NODISCARD az_result
 az_http_request_append_header(_az_http_request* p_request, az_span key, az_span value);
 
+/**
+ * @brief Validate HTTP retry options.
+ *
+ * @param retry_options HTTP retry options to validate.
+ * @return true if return options are valid, false if invalid.
+ */
+AZ_NODISCARD bool az_http_policy_retry_options_validate(
```
JeffreyRichter: I don't think we need this function

sdk/core/core/internal/az_retry_internal.h
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_RETRY_INTERNAL_H
+#define _az_RETRY_INTERNAL_H
+
+#include <stdint.h>
+
+#include <_az_cfg_prefix.h>
+
+AZ_NODISCARD AZ_INLINE int32_t _az_retry_calc_delay(
```
JeffreyRichter: Yes, this looks perfect!

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,228 @@
...
+
+    az_pair header = { 0 };
+    while (az_http_response_get_next_header(ref_response, &header) == AZ_OK)
+    {
+      if (az_span_is_content_equal_ignoring_case(header.key, AZ_SPAN_FROM_STR("retry-after-ms"))
+          || az_span_is_content_equal_ignoring_case(
+              header.key, AZ_SPAN_FROM_STR("x-ms-retry-after-ms")))
+      {
+        // The value is in milliseconds.
+        int32_t const msec = _az_uint32_span_to_int32(header.value);
```
JeffreyRichter: retry-after headers return a number in seconds; not msec

sdk/core/core/src/az_http_policy_private.h
```diff
@@ -0,0 +1,30 @@
...
+typedef struct
+{
+  _az_http_pipeline_policy_fn const func;
+  struct
+  {
+    _az_http_policy* policies;
+    _az_http_request* ref_request;
+    az_http_response* ref_response;
+  } params;
+} az_http_policy_callback;
```
JeffreyRichter: I don't think we should need this structure at all. From looking at your code, it appears that you inserted a level of indirection between a pipeline policy and a policy and because of this, you introduced this structure. I think if you moved the policy code into the pipeline policy function, you could get rid of this structure entirely and simplify the code. We can review in-person together

sdk/keyvault/keyvault/src/az_keyvault_client.c
```diff
@@ -127,6 +131,11 @@ AZ_NODISCARD az_result az_keyvault_keys_client_init(
     }
   };
 
+  if (!az_http_policy_retry_options_validate(&self->_internal.options.retry))
```
JeffreyRichter: We don't need this

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
```diff
@@ -107,6 +113,11 @@ AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
     }
   };
 
+  if (!az_http_policy_retry_options_validate(&self->_internal.options.retry))
```
JeffreyRichter: We don't need this

sdk/core/core/internal/az_config_internal.h
```diff
@@ -12,7 +12,7 @@
 #include <_az_cfg_prefix.h>
 enum
 {
-  AZ_HTTP_REQUEST_HEADER_BUF_SIZE = AZ_HTTP_REQUEST_HEADER_COUNT * sizeof(az_pair),
+  _az_HTTP_REQUEST_HEADER_BUF_SIZE = AZ_HTTP_REQUEST_HEADER_COUNT * sizeof(az_pair),
```
JeffreyRichter: I think AZ_HTTP_REQUEST_HEADER_COUNT should not be defined at all in core. The number of headers required really depends on the service and, in fact, on a specific azure service operation. So, this is better defined by the client library code (not in core) as only the client library code knows what the number of headers it needs.

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,228 @@
...
+AZ_NODISCARD bool az_http_policy_retry_options_validate(
+    az_http_policy_retry_options const* retry_options)
+{
+  if (retry_options == NULL || retry_options->max_retries < 0
+      || retry_options->status_codes == NULL)
+  {
+    return false;
+  }
+
+  if (retry_options->retry_delay_msec < 0 || retry_options->max_retry_delay_msec < 0)
```
ahsonkhan: How is the test coverage for this new code/policy?

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -0,0 +1,228 @@
...
+        {
+          *retry_after_msec = msec;
+          return AZ_OK;
+        }
+      }
+      else if (az_span_is_content_equal_ignoring_case(header.key, AZ_SPAN_FROM_STR("Retry-After")))
+      {
+        // The vaule is either seconds or date.
+        int32_t const seconds = _az_uint32_span_to_int32(header.value);
+        if (seconds >= 0) // int32_t max == ~68 years
```
ahsonkhan: What do we do when seconds is negative because we have "invalid" header.value? It looks like we'd return `AZ_OK`. Is that what we want?

sdk/core/core/src/az_span.c
```diff
@@ -65,7 +65,7 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span self, uint64_t* out)
   }
 
   uint64_t value = 0;
-  for (int32_t i = 0; i < self_length; i++)
+  for (int32_t i = 0; i < self_length; ++i)
```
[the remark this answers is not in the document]
antkmsft: In C++, Sutter, Meyers in their books - all recommend prefix. The reason being you don't lose anything, but in order to implement an overloaded postfix form, as when using C++ iterators, you get to increment the internal stated but need to create a copy and return it, and that is extra cycles.

We should all stick to prefix.

I blame C++ language name for the postfix increment popularity. It should almost never be used. It is, by design, an operation with side effects. The dedicated CPU instruction for that - increment but get the previous value - existed natively only on the PDP-11.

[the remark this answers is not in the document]
ahsonkhan: Let's fix it across the board then, and keep it that way (maybe add to code style guidelines).
Any way to auto fix it with some linter/code formatter?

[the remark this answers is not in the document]
antkmsft: And check out the irony for the C++ language name. Increment, but return previous value! Almost like us who have to use C++11/14 when 17/20 is out :)

sdk/core/core/src/az_span.c
```diff
@@ -77,7 +77,49 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span self, uint64_t* out)
...
+  }
+
+  *out = value;
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_span_to_uint32(az_span self, uint32_t* out)
+{
+  int32_t self_length = az_span_length(self);
+  if (self_length <= 0)
```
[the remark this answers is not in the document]
antkmsft: Probably it will never really be, but technically it is possible. I would prefer to limit the code to work only when we know it can work correctly, if that comes at no cost. In this case, if we only check for `== 0`, we know the behavior for '0' and '> 0', but what happens when '< 0'? The behavior is rather undefined.
I don't think it would make a practical difference, so I could write it either way, but prefer <= 0.

sdk/keyvault/keyvault/src/az_keyvault_client.c
antkmsft: I am not objecting, but at a time I wrote checks in the policy implementation, and you told me that the clients should validate options on creation instead, so that when we are inside the policy, we know that max_attempts, delay and max_delay - all are non-negative, and delay is <= max_delay, so I made this function.

sdk/core/core/src/az_span.c
```diff
@@ -77,7 +77,49 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span self, uint64_t* out)
...
+  {
+    uint8_t result = az_span_ptr(self)[i];
+    if (!isdigit(result))
+    {
+      return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+    }
+    uint32_t const d = (uint32_t)result - '0';
+    if ((UINT32_MAX - d) / 10 < value)
+    {
+      return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
```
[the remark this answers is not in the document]
antkmsft: I don't wrote this code -I changed it from int64 to int32.
This check does not check for digit, but it is good that you brought the attention to it - so it looks like we would return parser error instead of arithmetic overflow when we parse and int that con't fit into uint32. I'll ask Jeffrey which one we want, it looks like now we would always get parser error, which wouldn't allow me to use int32_max when we parse uint64, because now the code can't distinguish between parsing some garbage string vs parsing some valid integer, just too big one.

sdk/core/core/src/az_http_policy_private.h
antkmsft: Let's do it! We need the callback - if we want the code for retry policy and for the logging policy to be in different files other than all being in az_http_policy.c; otherise they won't be able to call the next policy.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
antkmsft: I fixed all the instances. As far as guidelines... Maybe. It is a minor thing (but with one clear winner answer)

sdk/core/core/src/az_http_policy_retry.c
antkmsft: https://github.com/Azure/azure-sdk-for-c/issues/416

sdk/core/core/src/az_http_policy_retry.c
antkmsft: Yes - we would initialize retry_after_msec with "-1" indicating that we didn't manage to get the value from the "Retry-After" headers, which is still AZ_OK. In that case, we'll fall back to the default behavior which is to wait exponentially.

sdk/core/core/internal/az_config_internal.h
antkmsft: Let's change it to per-client then, during our meeting later today.

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
antkmsft: https://github.com/Azure/azure-sdk-for-c/pull/413#discussion_r387288780

sdk/core/core/internal/az_http_internal.h
antkmsft: https://github.com/Azure/azure-sdk-for-c/pull/413#discussion_r387288780

sdk/core/core/src/az_http_policy_retry.c
antkmsft: https://github.com/Azure/azure-sdk-for-c/pull/413#discussion_r387288780

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: Sounds good. Maybe `az_span_length` should guarantee the returned length has to be >= 0, by validating on creation of the span.

sdk/core/core/inc/az_http_transport.h
```diff
@@ -35,7 +35,7 @@ AZ_INLINE az_http_method az_http_method_patch() { return AZ_SPAN_FROM_STR("PATCH
  *     - `index` is out of range.
  */
 AZ_NODISCARD az_result
-az_http_request_get_header(_az_http_request* p_request, int32_t index, az_pair* out_result);
+az_http_request_get_header(_az_http_request const* p_request, int32_t index, az_pair* out_result);
```
vhvb1989: I think Jeff wanted to remove all `const` from parameters

JeffreyRichter: I did at one point but I'm OK if we have "const *" parameters

gilbertw: Its odd to do it randomly.  If we want const parameters we should do that everywhere they are const.  Lets not do that at this time however

sdk/core/core/inc/az_result.h
```diff
@@ -61,14 +61,16 @@ typedef enum az_result
 {
   AZ_OK = _az_RESULT_MAKE_SUCCESS(AZ_FACILITY_CORE, 0),
   AZ_CONTINUE = _az_RESULT_MAKE_SUCCESS(AZ_FACILITY_CORE, 1),
+  AZ_CANCELED = _az_RESULT_MAKE_ERROR(AZ_FACILITY_CORE, 0),
```
gilbertw: Add a comment to break up the errors from the successes
//Success results
AZ_OK
AZ_CONTINUE
//Error results
AZ_CANCELLED

sdk/core/core/inc/az_http_transport.h
CIPop: I did this everywhere in IoT based on @JeffreyRichter 's recommendation :).

sdk/core/core/inc/az_result.h
CIPop: Even more, I think it's important we document what AZ_OK vs AZ_CONTINUE mean.

sdk/core/core/internal/az_retry_internal.h
```diff
@@ -0,0 +1,25 @@
...
+    int16_t attempt,
+    int32_t retry_delay_msec,
```
CIPop: Shouldn't this be in seconds? Why would anyone care about ms retry resolution from an embedded client?
Then we can move to int16_t (I don't think anyone would want to wait 18hrs before a retry.)

Here's one retry model I designed (it is in ms but doesn't need to):

x axis is the retry count; y is the delay for that retry. Total time waited is the integral: the key thing is to limit the total # of retries to never exceed 1, max 2 per second (which would quickly drain battery).

image

CIPop: Even more for my argument to move to seconds: `Retry-After` is also in seconds.

gilbertw: @CIPop is your y axis representing seconds or msec?  Was your implementation well received/liked by clients?

sdk/core/core/src/az_http_policy.c
```diff
@@ -30,6 +28,14 @@ AZ_NODISCARD AZ_INLINE az_result az_http_pipeline_nextpolicy(
       &(p_policies[1]), p_policies[0]._internal.p_options, p_request, p_response);
 }
 
+static AZ_NODISCARD az_result _az_http_pipeline_nextpolicy(
```
gilbertw: inline

sdk/core/core/inc/az_http_transport.h
antkmsft: That constness is different.
It makes little sense to write:
`f(int const n)`
`f(char* const n)`

But does make sense, and we should use as often as possible:
`f(char const* n)`

sdk/core/core/internal/az_retry_internal.h
antkmsft: 'retry-after-ms' and 'x-ms-retry-after-ms' are in milliseconds.
I also see the problem, but currently our PAL layer operates on int64_t msec. (az_platform_clock_msec() and az_platform_sleep_msec() return it)

We have a plan to review this aspect, to make SDK usable on both platforms with only seconds resolution, and 32-bit clock, and on desktop-like platforms, such as windows and linux, with millisecond resolution.

Ideally, int32 one_hundred_millisecond unit should be good enough for all applications. Then, if a platform only supports 0.5 seconds, or even seconds - sleep() would sleep 1 second if you ask it to sleep for `1 * one_hundred_millisecond`. Then, int32_max is ~6.8 years.

But all this requires deeper thinking, and we will most likely change it soon. But not in the scope of this PR; taking care of it tactically in one place is not going to solve the problem, but will bring inconsistency.

sdk/core/core/src/az_http_policy.c
antkmsft: I removed it in the next iteration, but the whole goal of this specific helper was to generate static body for the inline function, so its address can be taken. As you can se all this function was doing - was to call the inline version. At the same time we didn't want to remove inline status from the az_http_pipeline_nextpolicy(), because that could potentially increase the stack size for no real reason, since pipelines do heavily rellay on a deep chain of nested calls of functions.

sdk/core/core/inc/az_http_transport.h
antkmsft: We should not be removing const in the `foo const*` cases, absolutely not. I am not going to revert this change, and if we overdid this in the past due to misunderstanding - that was a mistake, and we should review our code and add back constness to the objects that pointers are pointing to. Maybe I can find some automatic way of doing that - when I would add constness everywhere and then start removing it until the code compiles.
It would be good to combine that with applying our new naming conventions - `ref_`s and `out_`s. (and no `p_`s).

I am basically consider this PR feedback as 'done', and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

sdk/core/core/internal/az_retry_internal.h
antkmsft: I am consider this PR feedback as 'done' for the scope of this PR, and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

sdk/core/core/src/az_http_policy_retry.c
antkmsft: I am consider this PR feedback as 'done' for the scope of this PR, and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
antkmsft: I am consider this PR feedback as 'done' for the scope of this PR, and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
antkmsft: It is good you brought this into attention, Ahson. I changed the logic to remove the ARITHMETIC_OVERFLOW error logic that I added. Now, if you're parsing uint32, and the result can't fit into uint32, it is parsing error - the same one that you get if you try to parse `"foo"` as uint32. And the logic that relies on that is fine with that - I confirmed yesterday with Jeffrey. If we asked to wait for more than int32_max milliseconds... there's likely an error anyway, because that's ~24 days.

So basically I reverted this function to the previous behavior (I just added the function for uint32, the one we had before is for uint64, but the logic is exactly the same).

I am consider this PR feedback as 'done' for the scope of this PR, and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
antkmsft: We do validate it, and you should never end up with negative values, unless there's a bug in our code, or if user wrote the code that directly modifies values in the `_internal` struct (which they shouldn't do). But I would still leave the check as `<= 0`, because we don't go that of an extra mile to filter out the subset of input values that does not apply to this function - either `==` or `<=` is one assembly instruction.

I am consider this PR feedback as 'done' for the scope of this PR, and not going to do more commits,and I'm going to resolve this conversation. Please revert it or ping me if you disagree and think that I should not check this in.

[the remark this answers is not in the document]
antkmsft: But if you feel strong about it, I can change it. I just feel we're intentionally introducing undefined behavior where we could easy avoid it. There's an argument to explode as soon as possible, if something's wrong, but I don't think that proceeding with the negative value as normal is going to explode right away.

sdk/core/core/src/az_http_policy_retry.c
antkmsft: Done :)

sdk/core/core/src/az_http_policy_private.h
antkmsft: Done :)

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
antkmsft: Done :)

sdk/core/core/src/az_http_policy_retry.c
antkmsft: I consider this feedback as 'done' in the scope of this PR.
Please revert if you disagree.

As I said during the standup, I'll try to implement az_platform_test soon, and that will be a separate PR.

sdk/core/core/inc/az_result.h
antkmsft: Break up is done.

Regarding documenting AZ_CONTINUE:
Honestly, I am not fully sure what AZ_CONTINUE means, but none of the result codes are documented at this moment. Let's document all of them in the scope of another PR.

I am going to resolve this conversation and I consider feedback as 'done'. If you disagree and think I should cover it in the scope of this PR, please reopen the conversation or ping me.

sdk/core/core/internal/az_config_internal.h
antkmsft: Done :)

sdk/keyvault/keyvault/src/az_keyvault_client.c
antkmsft: Done :)

sdk/core/core/internal/az_http_internal.h
antkmsft: Done :)

antkmsft:
As we agreed yesterday, I am going to check this in only after Victor's preconditions PR (https://github.com/Azure/azure-sdk-for-c/pull/421) goes in, and I will rebase and check in after that.

sdk/core/core/internal/az_retry_internal.h
CIPop: Not that big of a deal I think. I'd like to spend some time on an actual 8bit platform before pushing for optimizations such as the ones I proposed.
Didn't know about retry-after-ms - that doesn't seem to be a HTTP standard header. If some of the servers use those then I see good reason to hold ms resolution.

@gilbertw it's in msec. The implementation was transformed over several iterations to avoid certain issues detected by clients such as the max # of retries / second (high CPU/fast battery consumption). 
Obviously retry-after is the best way since the service has a lot more knowledge of its state (e.g. throttled - retry tomorrow). 

When the network is partitioned though the retry policy must be tuned by the device manufacturer as a function of number of devices, time to re-connect all devices, network latency, max timeouts within the network infrastructure, etc. The retry model above is favoring a busy service to massively reconnect a high number of devices (the randomization around +/-5 sec) with low TCP/TLS impact. This is quite common in IoT when a large area becomes partitioned then is reconnected.

PR #419, opened 2020-03-03
add literal null span

danewalton:
needed for initialization of other structs. eg:
```
#define TEST_EMPTY_OPTIONS \
  { \
    .module_id = AZ_SPAN_LITERAL_NULL, .user_agent = AZ_SPAN_LITERAL_NULL \
  }
```

### DIFF
--- a/sdk/core/core/inc/az_span.h
+++ b/sdk/core/core/inc/az_span.h
@@ -55,10 +55,15 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
 /********************************  CONSTRUCTORS */
 
 /**
+ * @brief Creates an empty span literal
+ */
+#define AZ_SPAN_LITERAL_NULL { ._internal = {.ptr = NULL, .length = 0, .capacity = 0} }
+
+/**
  * @brief Creates an empty span
  *
  */
-#define AZ_SPAN_NULL (az_span){ ._internal = {.ptr = NULL, .length = 0, .capacity = 0} }
+#define AZ_SPAN_NULL (az_span) AZ_SPAN_LITERAL_NULL
 
 // A size of the string literal.
 // Details: to make sure that `S` is a `string literal`, we are appending `""`


gilbertw:
Why can't you use the AZ_SPAN_NULL.  Something like this:
`
#define AZ_HTTP_RESPONSE_EMPTY \
(az_http_repsonse){ ._internal = { \
     .http_repsonse = NULL, \
     .parser = { \
        .remaining = AZ_SPAN_NULL, \
        .next_kind = NULL, \
     }, \
    }, \
  }
`

danewalton:
> Why can't you use the AZ_SPAN_NULL. Something like this:
> `#define AZ_HTTP_RESPONSE_EMPTY \ (az_http_repsonse){ ._internal = { \ .http_repsonse = NULL, \ .parser = { \ .remaining = AZ_SPAN_NULL, \ .next_kind = NULL, \ }, \ }, \ }`

The compiler throws an error saying the member isn't constant. I assume because it is casted.

gilbertw:
That code compiles fine for me when I did it in az_http.h.
`
az_http_response_init(az_http_response* self, az_span http_response)
{
  (void)http_response;
  *self = AZ_HTTP_RESPONSE_EMPTY;
`

I did it a tweek to to it to confirm it inited as expected and ran the sample correctly.
`
#define AZ_HTTP_RESPONSE_EMPTY(S) \
(az_http_response){ ._internal = { \
     .http_response = S, \
     .parser = { \
        .remaining = AZ_SPAN_NULL, \
        .next_kind = AZ_HTTP_RESPONSE_KIND_STATUS_LINE, \
     }, \
    }, \
  }

az_http_response_init(az_http_response* self, az_span http_response)
{
  *self = AZ_HTTP_RESPONSE_EMPTY(http_response);

`

Can you share the error you are seeing.  It might be because you don't cast the #define my_type_null (my_type) { .......

danewalton:
@gilbertw 
The cast of an internal designator for a compound literal is not a constant expression. So the problem isn't that the "top level" initializer was casted, but that when you nest these macro's which have casts, it's a problem. So in your example:
```
#define AZ_HTTP_RESPONSE_EMPTY(S) \
  (az_http_response) \
  { \
    ._internal = {
        .http_response = S,
        .parser = {
            .remaining = AZ_SPAN_NULL,
            .next_kind = AZ_HTTP_RESPONSE_KIND_STATUS_LINE,
          },
      }, \
  }

az_http_response_init(az_http_response* self, az_span http_response)
{
*self = AZ_HTTP_RESPONSE_EMPTY(http_response);
```
it's fine because no assignment to a designator was casted. But notice how the `.http_response` wasn't casted. Try casting `S` to an `az_span` and it should throw an error. Thats the use case of the `AZ_SPAN_LITERAL_NULL`

danewalton:
[Example here with GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71713)

danewalton:
pushed in other PR

PR #426, opened 2020-03-05
Anonymous credential

gilbertw:
Add az_credential_anonymous
  -Anonymous credentials can be used with SAS urls
Rename the client_secret_credential to credential_client_secret
Refactor the headers allowing customers to only include the credential they are using
Update the blob client sample to expect SAS token URLs


sdk/core/core/inc/az_credential_client_secret.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+AZ_NODISCARD az_result az_credential_client_secret_init(
+    az_credential_client_secret* self,
+    az_span tenant_id,
+    az_span client_id,
+    az_span client_secret);
+
+#include <_az_cfg_suffix.h>
+
+#endif // _az_CREDENTIAL_H
```
antkmsft: // _az_CREDENTIAL_CLIENT_SECRET_H, should you go this way.

gilbertw:
Passing null implies sending an empty client secret credential.  That is not what we should imply.  Anonymous is stating the credential is anonymous.  That right now means its an empty credential, but that is an implementation detail that might mean something more than NULL in the future.

antkmsft:
> Passing null implies sending an empty client secret credential.
Not necessary NULL Client Secret credential - it means passing NULL credential, which is never useful.

> That right now means its an empty credential, but that is an implementation detail that might mean something more than NULL in the future.
NULL credential, in my opinion is never useful, and I can't imagine a context where it could mean anything meaning or useful, other than to indicate an intent to not authenticate (i.e. use anonymous credential). I can only think of null pointer reference, or anonymous.

> That is not what we should imply. Anonymous is stating the credential is anonymous.
Yes, but do customer want the knowledge about that detail? It is an extra mental concept they'll need to know about.

gilbertw:
There is mental concept for understanding what it means to pass NULL to be anonymous vs explicitly stating your are anonymous.  Either way you need to understand the anonymous (no auth tokens/callbacks) concept.
I did consider passing NULL vs an anonymous credential.  The reason I elected to use a value was partly driven by the fact other languages explicitly disallow a null credential.
`
            if (credentials == null)
            {
                throw new System.ArgumentNullException("credentials");
            }
`

gilbertw:
@antkmsft @JeffreyRichter Updated with changes.

PR #437, opened 2020-03-07
Dawalton/telemetry pr

danewalton:


sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,95 @@
...
+#include <stdint.h>
+
+#include "az_iot_hub_client.h"
+#include <az_contract_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+static const uint8_t telemetry_prop_delim = '?';
+static const uint8_t telemetry_prop_separator = '&';
+static const uint8_t telemetry_null_terminator = 0;
```
ericwol-msft: '\0'

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,95 @@
...
+    az_iot_hub_client_properties const* properties,
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic)
+{
+  AZ_CONTRACT_ARG_NOT_NULL(client);
+  AZ_CONTRACT_ARG_VALID_SPAN(mqtt_topic);
+  AZ_CONTRACT_ARG_NOT_NULL(out_mqtt_topic);
+
+  az_result result;
+
```
ericwol-msft: extra line needed?

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,95 @@
...
+  int32_t required_size = 0;
```
ericwol-msft: why set to 0 if the next line populates it?

danewalton: it was using += before so I had to set it. Will remove

sdk/iot/core/src/az_iot_telemetry.c
danewalton: other than easier readability no

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,95 @@
...
+  if (!az_span_is_empty(client->_internal.options.user_agent))
+  {
+    required_size += az_span_length(client->_internal.options.user_agent)
+        + sizeof(telemetry_prop_delim);
+  }
+
+  // Only build topic if the span has the capacity
+  if (az_span_capacity(mqtt_topic) < required_size)
+  {
+    result = AZ_ERROR_OUT_OF_MEMORY;
```
ericwol-msft: ick, is thee really no buffer to small error? This seams like the heap is out of memory

danewalton: I'm not married to either. In my head "overflow" didn't make exact sense because the buffer didn't actually "overflow". Nothing was put in in the first place. Instead there is insufficient memory so "out of memory" seemed more appropriate. I can change to either

ericwol-msft: everywhere else has early returns, so do it here as well?

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,95 @@
...
+
+    AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_topic, telemetry_topic_suffix, out_mqtt_topic));
+
+    if (properties != NULL)
+    {
+      AZ_RETURN_IF_FAILED(az_span_append_uint8(*out_mqtt_topic, telemetry_prop_delim, out_mqtt_topic));
+      AZ_RETURN_IF_FAILED(
+          az_span_append(*out_mqtt_topic, properties->_internal.properties, out_mqtt_topic));
+    }
+    if (az_span_length(client->_internal.options.user_agent) != 0)
```
ericwol-msft: why do you have extra line in some places, but here you don't. nit: I prefer a blank line between if statements. This way people don't quickly read it as an else block

sdk/iot/core/src/az_iot_telemetry.c
danewalton: done

sdk/iot/core/src/az_iot_telemetry.c
danewalton: that's fine with me. Just modeled it to how it makes sense in my head. Will space the `if`'s

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+#include <cmocka.h>
+
+#define TEST_CORRECT_TOPIC_PREFIX "devices/"
+#define TEST_CORRECT_TOPIC_SUFFIX "/messages/events/"
+#define TEST_CORRECT_TOPIC_MODULES_MID "/modules/"
+
+#define TEST_FQDN "myiothub.azure-devices.net"
+#define TEST_DEVICE_ID "my_device"
+#define TEST_MODULE_ID "my_module_id"
+#define TEST_USER_AGENT "api-version=2017-11-08-preview&DeviceClientType=iothubclient"
```
ericwol-msft: update the api version?

danewalton: Thats what it is in the current sdk. Directly copied over.

sdk/iot/core/tests/cmocka/main.c
```diff
@@ -24,9 +27,27 @@ void az_iot_sas_token_generate_sas_token_overflow_fails(void** state);
 void az_iot_sas_token_generate_succeeds(void** state);
 void az_iot_sas_token_generate_with_keyname_succeeds(void** state);
 
+/*
+ * Telemetry Unit Tests
```
ericwol-msft: how out of control is this file going to get?

danewalton: honestly could be pretty hairy. I have additional thoughts about validating that all tests that are written are actually put here in this main. Inevitably tests are going to be written and someone will make an honest mistake and forget to copy over to here, test will never run, and we could have a test go unchecked.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+          &g_test_valid_client_no_options, NULL, null_mqtt_topic, &null_mqtt_topic)
+      == AZ_ERROR_ARG);
+}
+
+void az_iot_hub_client_telemetry_publish_topic_get_NULL_out_mqtt_topic_fails(void** state)
+{
+  (void)state;
+  uint8_t mqtt_topic_buf[TEST_MQTT_SPAN_BUFFER_SIZE];
+
+  az_span mqtt_topic = az_span_init(mqtt_topic_buf, 0, sizeof(mqtt_topic_buf) / sizeof(mqtt_topic_buf[0]));
```
ericwol-msft: we don't have ARRAYSIZE() macro for this?

danewalton: Nope. We could think about doing that though.

ericwol-msft: we should bring this up in the next C99 meeting

CIPop:
@danewalton for future reference please properly name the PR and add clear description of the things you've changed (especially things that are not directly related to the title).

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
CIPop: +1 if we don't have it, please define one.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
+#include <az_iot_hub_client.h>
+#include <az_span.h>
+
+#include <stdio.h>
+
+#include <setjmp.h>
```
CIPop: Do we really want to take a dependency on something using setjmp for running tests on 8 and 16bit MCUs?

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+#define TEST_VALID_OPTIONS_USER_AGENT \
+  { \
+    .module_id = AZ_SPAN_LITERAL_NULL, .user_agent = AZ_SPAN_LITERAL_FROM_STR(TEST_USER_AGENT) \
+  }
+
+static const az_iot_hub_client_properties g_test_params
+    = { ._internal
+        = { .properties = AZ_SPAN_LITERAL_FROM_STR(TEST_PARAMS), .current_property = 0 } };
+
+static const char g_test_correct_topic_no_options_no_params[]
```
CIPop: These are extremely hard to read for a unit test. Instead you should define the actual strings.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+
+static const az_iot_hub_client g_test_valid_client_no_options
+    = { ._internal = { .iot_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_FQDN),
+                       .device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID),
+                       .options = TEST_EMPTY_OPTIONS } };
+static const az_iot_hub_client g_test_valid_client_with_options_both
+    = { ._internal = { .iot_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_FQDN),
+                       .device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID),
+                       .options = TEST_VALID_OPTIONS_BOTH } };
+static const az_iot_hub_client g_test_valid_client_with_options_module_id
```
CIPop: Same with these: it's very hard to follow what the test story is: instead of defining globals, you should allocate these on the stack. 
I am of the opinion that UTs will not reuse much code but instead provide clarity as a single can be read to understand the scenario being tested.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+    = { ._internal = { .iot_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_FQDN),
+                       .device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID),
+                       .options = TEST_VALID_OPTIONS_MODULE_ID } };
+static const az_iot_hub_client g_test_valid_client_with_options_user_agent
+    = { ._internal = { .iot_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_FQDN),
+                       .device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID),
+                       .options = TEST_VALID_OPTIONS_USER_AGENT } };
+
+void az_iot_hub_client_telemetry_publish_topic_get_NULL_client_fails(void** state)
```
CIPop: Test names shouldn't be in the same namespace as the public API.
There should be guidance from the SDKs team. If not, please follow their naming notation (`test_<unit>_`).

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -0,0 +1,283 @@
...
+{
+  (void)state;
```
CIPop: What's this state and why do we need to be bothered by it?

sdk/iot/core/tests/cmocka/main.c
```diff
@@ -24,9 +27,27 @@ void az_iot_sas_token_generate_sas_token_overflow_fails(void** state);
...
+ */
+void az_iot_hub_client_telemetry_publish_topic_get_NULL_client_fails(void** state);
```
CIPop: To me, it feels like these are too verbose and are not actually providing a lot of extra value. We should have a reduced set of tests that can cover all blocks and definitely not cover 2^n permutations of every single possible call an app can make (especially since the code is not that complex).
We could also either reduce the number of functions of provide include files with both the definitions and the lists.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
danewalton: not sure it's a cmocka concept. Can figure it out afterwards if necessary.

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,94 @@
...
+#include <stdint.h>
+
+#include "az_iot_hub_client.h"
+#include <az_contract_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+static const uint8_t telemetry_prop_delim = '?';
+static const uint8_t telemetry_prop_separator = '&';
+static const uint8_t telemetry_null_terminator = '\0';
```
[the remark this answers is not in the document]
danewalton: I think the default behavior we are going with is a null terminated topic which can then be passed along to the MQTT library (important for those which require it to be null terminated. It will be documented as such and honestly if someone wants that extra byte back, we will get a GH issue and we can change it then. I don't want to complicate the compilation process early on if realistically it is unnecessary.

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,94 @@
...
+// SPDX-License-Identifier: MIT
+
+#include <stdint.h>
+
+#include "az_iot_hub_client.h"
+#include <az_contract_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+static const uint8_t telemetry_prop_delim = '?';
```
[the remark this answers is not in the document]
danewalton: the `az_span` api to append uses `uint8_t` not char so will likely leave as is unless there is enthusiastic contention.

sdk/iot/core/README.md
```diff
@@ -10,7 +10,33 @@ TBD
...
+
+  //Initialize the client to then pass to the telemetry API
+  az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, &my_options);
+
+  //Allocate a span to put the telemetry topic
+  uint8_t telemetry_topic_buffer[64];
+  az_span topic_span = AZ_SPAN_FROM_BUFFER(telemetry_topic_buffer);
+
+  //Get the NULL terminated topic and put in topic_span to send the telemetry
+  az_iot_hub_client_telemetry_publish_topic_get(&client, NULL, topic_span, &topic_span);
```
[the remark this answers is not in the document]
danewalton: That can likely go in the code snippet for using properties.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -55,7 +55,7 @@ AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default();
 /**
  * @brief Initializes an Azure IoT Hub Client.
  *
- * @param[out] client
+ * @param[out] client The #az_iot_hub_client to use for this call
```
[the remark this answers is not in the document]
danewalton: The build system flags it as an undocumented param which is why I had to go in and add it to all of the calls. Won't pass the gate otherwise. And I don't see a reason why we shouldn't have at least the most mundane documentation for everything. In this case, it gives the user easy access to check the documentation for the `az_iot_hub_client` struct by providing a link right then and there if they want it. Otherwise, they have to do additional hunting or searching for the docs for this struct.

sdk/iot/core/README.md
```diff
@@ -10,7 +10,33 @@ TBD
...
+
+```C
+static az_iot_hub_client my_client;
+static az_span my_iothub_hostname = AZ_SPAN_LITERAL_FROM_STR("constoso.azure-devices.net");
+static az_span my_device_id = AZ_SPAN_LITERAL_FROM_STR("contoso_device");
+
+void my_telemetry_func()
+{
+  //Get the default options for the client
+  az_iot_hub_options my_options = az_iot_hub_client_options_default();
```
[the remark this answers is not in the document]
danewalton: agreed. removed the options but kept the client init

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,94 @@
...
+    required_size += az_span_length(client->_internal.options.user_agent)
+        + sizeof(telemetry_prop_delim);
+  }
+
+  // Only build topic if the span has the capacity
+  if (az_span_capacity(mqtt_topic) < required_size)
+  {
+    return AZ_ERROR_BUFFER_OVERFLOW;
+  }
+  else
```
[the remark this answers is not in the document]
danewalton: removed

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
danewalton: That's a discussion which was briefly had on last Friday which can be brought up separately if you'd like. I think for now we'll keep it.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
danewalton: noted. changed

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
danewalton: These remove the dependency on the client init calls. Primarily they don't exist yet so I need to allocated them here for now.

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
danewalton: I think I have my preferences here. I can change them if you feel strongly but this allows me to compose them easily and make sure copy paste errors don't happen.

sdk/iot/core/tests/cmocka/main.c
danewalton: I think I have just enough tests here. One test to check for each invalid param. Make sure that each optional param is covered in the successful case and in the case where there isn't enough room. And that if we have some combination of options each of those are covered (since size checks are done for each possible combo)

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,94 @@
...
+
+#include "az_iot_hub_client.h"
+#include <az_contract_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+static const uint8_t telemetry_prop_delim = '?';
+static const uint8_t telemetry_prop_separator = '&';
+static const uint8_t telemetry_null_terminator = '\0';
+static const az_span telemetry_topic_prefix = AZ_SPAN_LITERAL_FROM_STR("devices/");
```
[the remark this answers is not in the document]
danewalton: This causes as extra call to `strlen` which I'm not sure will always get compiled away. The inline function call to `az_span_from_str` doesn't always have to be inlined and at that point it likely won't. Extra CPU cycles vs two extra 32 bit int's for each span which should we go with?

sdk/iot/core/inc/az_iot_hub_client.h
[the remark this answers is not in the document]
CIPop: Makes sense: could you please update the provisioning one to ensure we use the same pattern?
Also: I think the other docs end in ".". Please make sure we're consistent.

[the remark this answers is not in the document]
danewalton: I will after this PR just to keep scopes.

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <stdint.h>
+
+#include "az_iot_hub_client.h"
+#include <az_precondition.h>
+#include <az_result.h>
+#include <az_span.h>
+
```
vhvb1989: #include <_az_cfg_prefix.h>

danewalton: ah thank you

danewalton: adding the suffix too

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,98 @@
...
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  // Required topic parts
+  int32_t required_size = az_span_length(telemetry_topic_prefix)
+      + az_span_length(client->_internal.device_id) + az_span_length(telemetry_topic_suffix)
```
vhvb1989: Wonder if we would like to have az_iot_hub_client_device_id(az_iot_hub_client self), so we don't access internal field here and instead we use function to get it.
No need to change,, just wondering

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,98 @@
...
+    required_size += az_span_length(client->_internal.options.module_id);
+  }
+
+  if (properties != NULL)
+  {
+    required_size
+        += az_span_length(properties->_internal.properties) + sizeof(telemetry_prop_delim);
+  }
+
+  if (az_span_ptr(client->_internal.options.user_agent) != NULL)
```
vhvb1989: is it an option to calculate this required_size when client is init?  and then update it every time it gets updated?
so then we don't have to calculate it every time this publish method is called?
Basically adding another field to the client to know about this at all time

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,98 @@
...
+  if (az_span_capacity(mqtt_topic) < required_size)
+  {
+    return AZ_ERROR_BUFFER_OVERFLOW;
+  }
+
+  // Build topic string
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, telemetry_topic_prefix, out_mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_topic, client->_internal.device_id, out_mqtt_topic));
+
+  if (az_span_length(client->_internal.options.module_id) != 0)
```
vhvb1989: what about creating a variable for `az_span_length(client->_internal.options.module_id)` since it was previously called on line 39.  
So, we might want to call it just once (unless it could be externally updated?)

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,98 @@
...
+  {
+    AZ_RETURN_IF_FAILED(
+        az_span_append(*out_mqtt_topic, telemetry_topic_modules_mid, out_mqtt_topic));
+    AZ_RETURN_IF_FAILED(
+        az_span_append(*out_mqtt_topic, client->_internal.options.module_id, out_mqtt_topic));
```
vhvb1989: I would also copy `client->_internal.options.module_id` to a local variable to avoid referencing pointer `client->`  multiple times

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  if (az_span_length(client->_internal.options.user_agent) != 0)
+  {
+    AZ_RETURN_IF_FAILED(
+        properties == NULL
+            ? az_span_append_uint8(*out_mqtt_topic, telemetry_prop_delim, out_mqtt_topic)
+            : az_span_append_uint8(*out_mqtt_topic, telemetry_prop_separator, out_mqtt_topic));
+    AZ_RETURN_IF_FAILED(
+        az_span_append(*out_mqtt_topic, client->_internal.options.user_agent, out_mqtt_topic));
```
vhvb1989: copy `client->_internal.options` to stack variable and use it instead of using ptr multiple times to access options

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
vhvb1989: yeah, this is a cmocka required arg.
It can be used to run tests sharing state. Might be useful when many tests requires the same setup

sdk/iot/core/src/az_iot_telemetry.c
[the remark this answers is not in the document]
CIPop: I keep missing that point and making the same mistake: you are correct: strlen(const) should be replaced with a const at compile time. The only extra here would be one int/string (caused by len and capacity being equal).  
We'll probably have < 10s of strings so 10s of bytes are not a big deal for me at this time.

sdk/iot/core/tests/cmocka/main.c
vhvb1989: move this test definitions and adding tests to list to a  `az_test_definitions.h`
let's keep this file just to call tests like
```c
int main(void) { return cmocka_run_group_tests_name("az_iot_core", tests, NULL, NULL); }
```
See how it is on az_core...   

Well, that's just an option to avoid this file to keep growing and growing..
for az_core, I am trying to split things a little bit and I add some comments about how to add tests

sdk/iot/core/src/az_iot_telemetry.c
danewalton: this will often be used only once by the developer. They will get their topic that they publish to and reuse that every time they send telemetry. So in the common case the required size will only be calculated once.

sdk/iot/core/src/az_iot_telemetry.c
danewalton: possibly. this will likely be changed if someone asks for that ability. but duly noted.

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,99 @@
...
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  az_span user_agent = client->_internal.options.user_agent;
+  az_span module_id = client->_internal.options.module_id;
+
+  // Required topic parts
+  int32_t required_size = az_span_length(telemetry_topic_prefix)
+      + az_span_length(client->_internal.device_id) + az_span_length(telemetry_topic_suffix)
+      + sizeof(telemetry_null_terminator);
```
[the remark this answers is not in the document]
CIPop: We've decided to always guarantee C strings given most of the MQTT libraries require strings (Paho, Mosquitto, etc).

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Only build topic if the span has the capacity
+  if (az_span_capacity(mqtt_topic) < required_size)
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY;
+  }
+
+  // Build topic string
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, telemetry_topic_prefix, out_mqtt_topic));
```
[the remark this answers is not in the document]
CIPop: This is a very good point: @JeffreyRichter looks like @danewalton 's implementation is doing some extra checks - should we rely on az_span itself here instead of us calculating?

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -0,0 +1,99 @@
...
+    az_iot_hub_client const* client,
+    az_iot_hub_client_properties const* properties,
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  az_span user_agent = client->_internal.options.user_agent;
```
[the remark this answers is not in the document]
CIPop: +1 (Or just use it directly from the client).

sdk/iot/core/src/az_iot_telemetry.c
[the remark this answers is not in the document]
danewalton: Originally I had the checks removed but the current requirement on the `az_span_append` is to not ignore the return value. We should likely coordinate with @JeffreyRichter to see if there's a way to remove these kinds of checks.

sdk/iot/core/README.md
```diff
@@ -10,7 +10,30 @@ TBD
...
+{
+  //Initialize the client to then pass to the telemetry API
+  az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, NULL);
+
+  //Allocate a span to put the telemetry topic
+  uint8_t telemetry_topic_buffer[64];
+  az_span topic_span = AZ_SPAN_FROM_BUFFER(telemetry_topic_buffer);
+
+  //Get the NULL terminated topic and put in topic_span to send the telemetry
+  az_iot_hub_client_telemetry_publish_topic_get(&client, NULL, topic_span, &topic_span);
```
TiejunMS: I'm not sure where to leave comments for already merged changes.
`client` should be replaced with `my_client`

danewalton: Good catch I’ll update it

danewalton: Updated [here with the hub client PR](https://github.com/Azure/azure-sdk-for-c/pull/455/commits/3efddf6c84ad9ef37fe4422c34ec99d0c8ad5939)

sdk/iot/core/src/az_iot_telemetry.c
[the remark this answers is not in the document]
CIPop: @danewalton what I'm referring to is the calculation of required size. I think your code should simply do the operations and if one fails, you will return the error.
i.e. all code doing `required_size` and the variable itself can be removed.

PR #440, opened 2020-03-10
Add null span precondition option boolean

danewalton:
@JeffreyRichter @vhvb1989 
Here's the PR for the change. I believe in the cases for core code NULL shouldn't be allowed so I edited them as such. Let me know if I am wrong.


sdk/core/core/inc/az_precondition.h
```diff
@@ -31,20 +31,23 @@ extern az_precondition_failed az_precondition_failed_callback;
 
 #define AZ_PRECONDITION_NOT_NULL(arg) AZ_PRECONDITION((arg != NULL))
 
-AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_length)
+AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_length, bool null_is_valid)
 {
   int32_t span_length = az_span_length(span);
   /* Valid Span is:
      Span length equals to 0 (Pointer can be NULL) and greater or equal to min_length
```
danewalton: update this to reflect reality

sdk/core/core/inc/az_precondition.h
```diff
@@ -31,20 +31,26 @@ extern az_precondition_failed az_precondition_failed_callback;
...
+     In the case of the pointer not being NULL, two conditions must be met:
+        - The length is greater than or equal to zero and the capactity is greater than or equal to the length.
+        - The length is greater than or equal to a user defined minimum value.
   */
   return (
-      (span_length == 0
-       || (az_span_ptr(span) != NULL && az_span_ptr(span) <= az_span_ptr(span) + span_length - 1))
-      && min_length <= span_length);
+      (null_is_valid && (az_span_ptr(span) == NULL) && (span_length == 0)
+       && (span_capacity == 0))
```
JeffreyRichter: Even if the the span's ptr is null, the length still has to be at least min_length so this additional test should be included here.

sdk/core/core/src/az_http_request.c
```diff
@@ -26,9 +26,9 @@ AZ_NODISCARD az_result az_http_request_init(
     az_span body)
 {
   AZ_PRECONDITION_NOT_NULL(p_hrb);
-  AZ_PRECONDITION_VALID_SPAN(method, 0);
-  AZ_PRECONDITION_VALID_SPAN(url, 0);
-  AZ_PRECONDITION_VALID_SPAN(headers_buffer, 0);
+  AZ_PRECONDITION_VALID_SPAN(method, 0, false);
```
JeffreyRichter: Now that I look at these again, the min_length for method & url should be 1; headers can be 0 - please fix

sdk/core/core/src/az_http_request.c
```diff
@@ -85,8 +85,8 @@ AZ_NODISCARD az_result
 az_http_request_set_query_parameter(_az_http_request* p_hrb, az_span name, az_span value)
 {
   AZ_PRECONDITION_NOT_NULL(p_hrb);
-  AZ_PRECONDITION_VALID_SPAN(name, 0);
-  AZ_PRECONDITION_VALID_SPAN(value, 0);
+  AZ_PRECONDITION_VALID_SPAN(name, 0, false);
```
JeffreyRichter: The min_length for name & value should be 1

sdk/core/core/src/az_http_request.c
```diff
@@ -119,8 +119,8 @@ AZ_NODISCARD az_result
 az_http_request_append_header(_az_http_request* p_hrb, az_span key, az_span value)
 {
   AZ_PRECONDITION_NOT_NULL(p_hrb);
-  AZ_PRECONDITION_VALID_SPAN(key, 0);
-  AZ_PRECONDITION_VALID_SPAN(value, 0);
+  AZ_PRECONDITION_VALID_SPAN(key, 0, false);
+  AZ_PRECONDITION_VALID_SPAN(value, 0, false);
```
JeffreyRichter: The min_length for key & value should be 1

sdk/core/core/src/az_span.c
```diff
@@ -139,8 +139,8 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span self, uint32_t* out)
  */
 AZ_NODISCARD az_result az_span_copy(az_span dst, az_span src, az_span* out)
 {
-  AZ_PRECONDITION_VALID_SPAN(dst, 0);
-  AZ_PRECONDITION_VALID_SPAN(src, 0);
+  AZ_PRECONDITION_VALID_SPAN(dst, 0, false);
+  AZ_PRECONDITION_VALID_SPAN(src, 0, false);
```
JeffreyRichter: NULL is allowed for both of these; change to true.
Copying to/from NULL is like a no-op. It's like having a null stream where you can stream to null to just consume data. Or like memcopy from a buffer with 0 bytes - this is OK

sdk/core/core/src/az_span.c
```diff
@@ -174,8 +174,8 @@ AZ_NODISCARD AZ_INLINE bool should_encode(uint8_t c)
 AZ_NODISCARD az_result az_span_copy_url_encode(az_span dst, az_span src, az_span* out)
 {
   AZ_PRECONDITION_NOT_NULL(out);
-  AZ_PRECONDITION_VALID_SPAN(dst, 0);
-  AZ_PRECONDITION_VALID_SPAN(src, 0);
+  AZ_PRECONDITION_VALID_SPAN(dst, 0, false);
+  AZ_PRECONDITION_VALID_SPAN(src, 0, false);
```
JeffreyRichter: should be true for both

sdk/core/core/src/az_span.c
```diff
@@ -259,7 +259,7 @@ void _az_span_swap(az_span a, az_span b)
  */
 AZ_NODISCARD az_result az_span_to_str(char* s, int32_t max_size, az_span span)
 {
-  AZ_PRECONDITION_VALID_SPAN(span, 0);
+  AZ_PRECONDITION_VALID_SPAN(span, 0, false);
```
JeffreyRichter: Should be true -- just creates an empty string

sdk/iot/core/src/az_iot_sas_token.c
```diff
@@ -25,9 +25,9 @@ az_result az_iot_sas_token_get_document(
     az_span document,
     az_span* out_document)
 {
-  AZ_PRECONDITION_VALID_SPAN(device_id, 1);
-  AZ_PRECONDITION_VALID_SPAN(iothub_fqdn, 1);
-  AZ_PRECONDITION_VALID_SPAN(document, 0);
+  AZ_PRECONDITION_VALID_SPAN(device_id, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(iothub_fqdn, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(document, 0, false);
```
JeffreyRichter: Are you sure document can't be NULL span?

danewalton: I don't believe so. The API attempts to append data to the span so writing to a NULL buffer shouldn't be allowed.

sdk/core/core/inc/az_precondition.h
danewalton: Updated and changed the comment description above of a valid span.

PR #443, opened 2020-03-11
Span documentation

gilbertw:
Add comments to span.h
Update parameter names
Update signatures for uniformity


gilbertw:
@JeffreyRichter here is the PR with your documentation changes and all the corresponding code changes

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +154,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
 
 /**
- * @brief convert all the span content to uint32
+ * @brief az_span_to_uint64 parses an az_span containing ASCII digits into a uint64 number
  *
- * @param self source span
- * @param out result span
- * @return AZ_NODISCARD az_span_to_uint32
+ * @param[in] span The az_span containing the ASCII digits to be parsed.
+ * @param[in] out The pointer to the variable that is to receive the number
```
ahsonkhan: CI failing because of docs mismatch. Should be `out_number`.
```text
/Users/runner/runners/2.165.1/work/1/s/sdk/core/core/inc/az_span.h:231:15: note: did you mean 'out_number'?
 * @param[in] out The pointer to the variable that is to receive the number
```

sdk/core/core/inc/az_span.h
```diff
@@ -84,30 +88,37 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
...
- * Creates an az_span literal which can be used to initialize a constant. For example
+ * @brief AZ_SPAN_LITERAL_FROM_BUFFER returns a literal az_span over an unitialized byte buffer. For
+ * example:
  *
- * uint8_t buffer[2 * 1024];
- * const az_span foo = AZ_SPAN_LITERAL_FROM_BUFFER(buffer);
+ * uint8_t buffer[1024];
+ * const az_span buf = AZ_SPAN_LITERAL_FROM_BUFFER(buffer);  // Len=0, Cap=1024
  */
 #define AZ_SPAN_LITERAL_FROM_BUFFER(BYTE_BUFFER) \
```
ahsonkhan: I am having trouble understanding the need for both `AZ_SPAN_LITERAL_FROM_BUFFER` and `AZ_SPAN_FROM_BUFFER`. Why not just have `AZ_SPAN_FROM_BUFFER`?

sdk/core/core/inc/az_span.h
```diff
@@ -118,13 +129,21 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * Creates a az_span literal which can be used to initialize a constant. For example
+ * @brief AZ_SPAN_LITERAL_FROM_INITIALIZED_BUFFER returns a literal az_span over an initialized byte
+ * buffer. For example:
  *
- * uint8_t buffer[] = {"Hello World"};
- * const az_span foo = AZ_SPAN_LITERAL_FROM_BUFFER(buffer);
+ * uint8_t buffer[] = { 1, 2, 3 };
+ * const az_span buf = AZ_SPAN_LITERAL_FROM_INITIALIZED_BUFFER(buffer); // Len=3, Cap=3
  */
 #define AZ_SPAN_LITERAL_FROM_INITIALIZED_BUFFER(BYTE_BUFFER) \
```
ahsonkhan: Can we remove this and only keep `AZ_SPAN_FROM_INITIALIZED_BUFFER`?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +154,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
     }, \
   }
 
+/**
+ * @brief AZ_SPAN_FROM_INITIALIZED_BUFFER returns an az_span expression over an inittialized byte
```
ahsonkhan: typo: inittialized

sdk/core/core/inc/az_span.h
```diff
@@ -4,7 +4,8 @@
 /**
  * @file az_span.h
  *
- * @brief az_span and utilities definition
+ * @brief An az_span represents a byte buffer and is used for string manipulations,
```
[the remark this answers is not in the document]
JeffreyRichter: I like the "contiguous" part. I worry that "dynamically-sized" could easily be misinterpreted as dynamically resizing the buffer itself and now the "view" into the buffer and it's important that customer not question our allocation-free mantra.

sdk/core/core/inc/az_span.h
```diff
@@ -35,44 +38,45 @@ typedef struct
 /********************************  SPAN GETTERS */
 
 /**
- * @brief convenient way to get the pointer from one span
+ * @brief az_span_ptr returns the span byte buffer's starting memory address
```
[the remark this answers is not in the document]
JeffreyRichter: As far as English grammar goes, the rule is to put a period after a complete sentence and to NOT put a period after a sentence fragment. Most bullets on PowerPoint slides are fragments and should therefore not have periods. I do think consistency is important here and, in my rush to do this yesterday, I'm sure I missed things. I started a convention of starting each brief the same way "<function name> returns ..." and I think we should maintain this convention. And yes, I think the briefs should be full sentences ending with a period. 

I think the parameter descriptions are sentence fragments and should not have periods.

sdk/core/core/inc/az_span.h
```diff
@@ -20,7 +21,9 @@
 #include <_az_cfg_prefix.h>
 
 /**
- * An immutable span of bytes (octets).
+ * az_span is a "view" over a byte buffer. It contains a pointer to the
+ * start of the byte buffer, the buffer's capacity; and the length (number of
+ * bytes) in use from the start of the buffer.
```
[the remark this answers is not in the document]
JeffreyRichter: I'm tempted to not do this because, on smaller MCUs, we may switch to 16-bit integers and we really haven't resolved the signed vs unsigned issue yet either. If we say something here, it is likely to be wrong later.

sdk/core/core/inc/az_span.h
```diff
@@ -35,44 +38,45 @@ typedef struct
...
- * @brief convenient way to get the pointer from one span
+ * @brief az_span_ptr returns the span byte buffer's starting memory address
  *
  */
 AZ_NODISCARD AZ_INLINE uint8_t* az_span_ptr(az_span span) { return span._internal.ptr; }
 
 /**
- * @brief convenient way to get the length from one span
+ * @brief az_span_length returns the span byte buffer's length (number of bytes in use from the
+ * buffer's start)
```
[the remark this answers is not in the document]
JeffreyRichter: Do you have an alternate suggestion for "in use" - I'm happy to change it but I struggled to come up with something much better. 

We can put (in bytes) for length; I just thought is was obvious enough.

sdk/core/core/inc/az_span.h
```diff
@@ -35,44 +38,45 @@ typedef struct
...
  *
  */
 AZ_NODISCARD AZ_INLINE int32_t az_span_length(az_span span) { return span._internal.length; }
 
 /**
- * @brief convenient way to get the capacity from one span
+ * @brief az_span_capacity returns the span byte buffer's capacity (in bytes)
```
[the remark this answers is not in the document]
JeffreyRichter: I struggled with this too. The span is not the byte buffer; a span refers to a subset of a byte buffer. I don't have an easy way to say this that just flows off the tongue. But, I'm open to suggestions. I don't think it is right to say "Do not access the memory beyond this capacity." because, when using a span and its functions, this is impossible. And also, a span could be to a subset of a byte buffer and the customer might know that accessing bytes past the span's capacity IS OK. 

Saying that length is always <= capacity is fine with me.

sdk/core/core/inc/az_span.h
```diff
@@ -35,44 +38,45 @@ typedef struct
...
-// A size of the string literal.
-// Details: to make sure that `S` is a `string literal`, we are appending `""`
-// to `S`.
+// Returns the size (in bytes) of a liternal string
+// Note: Concatenating "" to S produces a compiler error if S is not a literal string
 #define _az_STRING_LITERAL_LEN(S) (sizeof(S "") - 1)
 
 /**
- * Creates an az_span literal which can be used to initialize a constant. For example
+ * @brief The AZ_SPAN_LITERAL_FROM_STR macro returns a literal az_span over a liternal string. For
```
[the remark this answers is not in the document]
JeffreyRichter: This macro DOES return a literal az_span over a literal string. I think "literal" should appear twice. The macro makes a literal az_span (hence its name). 
Sure, we can say that the length & capacity are set to the number of bytes in the string not including the 0-terminating byte.

sdk/core/core/inc/az_span.h
```diff
@@ -84,30 +88,37 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
...
+ * example: `some_function(AZ_SPAN_FROM_STR("Hello world"));` where `void some_function(const
+ * az_span span);`
  */
 #define AZ_SPAN_FROM_STR(STRING_LITERAL) (az_span) AZ_SPAN_LITERAL_FROM_STR(STRING_LITERAL)
 
 /**
- * @brief Init span
+ * @brief az_span_init returns a span over a byte buffer.
  *
+ * @param[in] ptr The memory address of the 1st byte in the byte buffer
```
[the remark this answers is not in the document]
JeffreyRichter: We can change ptr to pointer.
I think I already change dst to destination and src to source. If you see any more of these, we should spell them out.

sdk/core/core/inc/az_span.h
```diff
@@ -84,30 +88,37 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
...
+ * @param[in] length The number of bytes initialized in the byte buffer
+ * @param[in] capacity The number of total bytes in the byte buffer
+ * @return az_span The az_span "view" over the byte buffer
  */
 AZ_NODISCARD az_span az_span_init(uint8_t* ptr, int32_t length, int32_t capacity);
```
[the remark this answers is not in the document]
JeffreyRichter: Yes, we do these preconditions now in the az_span.c file.
ptr can be null; but then length & capacity must be 0.

sdk/core/core/inc/az_span.h
```diff
@@ -84,30 +88,37 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
...
 
 /**
- * @brief Creates a span from zero terminated string
+ * @brief az_span_from_str returns an az_span from a 0-terminated array of bytes (chars)
```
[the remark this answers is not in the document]
JeffreyRichter: A span is a byte buffer. Bytes are 0-255. However, you can reinterpret bytes however you want. So, you can "cast" a byte to a char which is really a signed 8-bit integer. An unsigned char is an unsigned 8-bit integer which is the same as a byte.

All buffers are 0-terminated: You'll eventually hit a 0 or end of the address space and then wrap around and eventually hit a 0.

sdk/core/core/inc/az_span.h
```diff
@@ -84,30 +88,37 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_capacity(az_span span) { return span._int
...
  *
+ * @param[in] str The pointer to the 0-terminated array of bytes (chars)
+ * @return az_span An az_span over the byte buffer; length & capacity are set to the string's
+ * length.
  */
 AZ_NODISCARD az_span az_span_from_str(char* str);
 
 /**
- * Creates an az_span literal which can be used to initialize a constant. For example
+ * @brief AZ_SPAN_LITERAL_FROM_BUFFER returns a literal az_span over an unitialized byte buffer. For
```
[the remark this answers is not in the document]
JeffreyRichter: What makes it an az_span literal are the rules of C. Here's an article I found: http://nickdesaulniers.github.io/blog/2013/07/25/designated-initialization-with-pointers-in-c/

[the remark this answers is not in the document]
JeffreyRichter: The buffer doesn't have to be uninitialized. If it is, use this macro and the length will be 0.
If it isn't use the other macro and the length will be the buffer's length. Sure, we can say this macro sets the length to 0.

sdk/core/core/inc/az_span.h
JeffreyRichter: Both required due to a compiler issue. Statically initializing a struct (this is the definition of a 'literal struct') won't compile with the cast. 
Passing an initialized struct (created at runtime; so not a literal struct) won't compile if the ast is not present.

So, we need both macros. I'm unhappy about this too, believe me but we really had no choice.

sdk/core/core/inc/az_span.h
JeffreyRichter: No

sdk/core/core/inc/az_span.h
JeffreyRichter: Good catch

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: That's fair. How would you reconcile/mention the capacity concept in the API reference docs then?

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: Aren't we already encoding some of this behavior in the public surface area? For instance, the length getter returns an int32. Maybe, we shouldn't shy away from documenting the current behavior as is. If we were to change it while in preview, we can change the docs then too.

The only reason I can see us not documenting such detail is that if we are able to change it in subsequent releases without affecting the public surface area in any significant observable way. I am not sure if that's feasible.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: > And also, a span could be to a subset of a byte buffer and the customer might know that accessing bytes past the span's capacity IS OK.

Hmm... interesting. I understand that the caller can do this, but should they be? Isn't it considered potentially bad practice to go outside the bounds of the span? How do you reason about what the scope of the "original" span is to know it is safe to go outside of its capacity bounds?

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: > An unsigned char is an unsigned 8-bit integer which is the same as a byte.

Great, so maybe update the comment to explicitly state `(unsigned chars)` instead of just `(chars)`.

```suggestion
 * @brief az_span_from_str returns an az_span from a 0-terminated array of bytes (unsigned chars)
```

> You'll eventually hit a 0 or end of the address space and then wrap around and eventually hit a 0.

But could that happen after more than int32 max bytes? And if so, does our length count potentially overflow to negative?

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: I would have thought `literal` meant something fixed or constant and not related to initializers and compact initialization syntax.

For instance, from https://www.tutorialspoint.com/cprogramming/c_constants.htm
> Constants refer to fixed values that the program may not alter during its execution. These fixed values are also called literals.

Even in the doc example, there is nothing preventing me from changing the characteristics of the span/contents of the buffer regardless of whether it was created as a literal or not. Is it really that helpful to encapsulate how the caller is going to use/call the span-creation method with specific macros? I am trying to understand the nuance a bit more here :)

For example, is it wrong for me to write this code:
```c
uint8_t buffer[] = {"Hello World"};
const az_span foo = AZ_SPAN_FROM_BUFFER(buffer); // instead of AZ_SPAN_LITERAL_FROM_BUFFER
```

And if we do want to keep the literal concept, should we call it `compound literal`, or is that implied already?
https://en.cppreference.com/w/c/language/compound_literal

> The buffer doesn't have to be uninitialized. If it is, use this macro and the length will be 0.
If it isn't use the other macro and the length will be the buffer's length.

Thank you for the clarification and link. That was quite helpful. Part of my intent with posing such questions was to suggest that maybe those types of questions should be answered/clarified in the API doc.

sdk/core/core/inc/az_span.h
ahsonkhan: I see. Interesting. So, the difference between the two macros is that explicit cast. I am trying to understand this in the two cases you mentioned (especially with writing code at the call site).

When doing so, I get a compiler error. Any ideas what I am doing wrong?

```c
int Temp(uint8_t buff[])
{
  az_span span = AZ_SPAN_FROM_BUFFER(buff);

  az_span literal = AZ_SPAN_LITERAL_FROM_BUFFER(buff);

  return az_span_length(span) + az_span_length(literal);
}
```

```text
test_span.c(16): warning C6384: Dividing sizeof a pointer by another value. 
test_span.c(18): warning C6384: Dividing sizeof a pointer by another value. 
```

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
gilbertw: I can go either way on dst/src vs destination/source.  ptr though is idiomatic to C.

sdk/core/core/inc/az_context.h
```diff
@@ -20,69 +20,112 @@
 typedef struct az_context az_context;
 
 /**
- * @brief An az_context value in a node in a tree that represents expiration times and key/value
+ * @brief An az_context value is a node in a tree that represents expiration times and key/value
```
ahsonkhan: nit: Can we separate out the `az_context.h` related changes/docs into a separate PR, leaving this span focused? If it is not convenient/easy to do so, ignore it.

gilbertw: Context is small compared to span.  Would prefer to do it at the same time

gilbertw: I pulled it out.  I'll put it in a seperate PR

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: > ptr though is idiomatic to C.

I can see why that would be true for C++ (with types like unique_ptr, auto_ptr, etc.). I don't see a lot of `ptr` usage in stl (likely not looking at all the right places). Found just one:
https://github.com/microsoft/STL/blob/92508bed6387cbdae433fc86279bc446af6f1b1a/stl/inc/xcharconv.h#L40

Also see some hits for pointer:
https://en.cppreference.com/mwiki/index.php?title=Special%3ASearch&search=pointer

What else is considered idiomatic to C?

I was looking at some common apis like memmove, strcpy, etc.  for reference.
https://en.cppreference.com/w/cpp/string/byte/memmove
https://en.cppreference.com/w/c/string/byte/strcpy

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: > I worry that "dynamically-sized" could easily be misinterpreted as dynamically resizing the buffer itself and now the "view" into the buffer and it's important that customer not question our allocation-free mantra.

Thinking about this some more, I agree and the rationale makes sense. Just to clarify, the span's capacity cannot grow beyond what it was initialized with, correct?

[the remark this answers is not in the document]
JeffreyRichter: Correct. If you slice into the span, the capacity can shrink, but not grow

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: Actually, I've written more docs today and the parameters docs are turning out to be complete sentences. so I am putting periods at the end.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: We may change the types during the preview period. We were having debates on this and so it was in flux. The debates stopped and so this may just make it all the way to GA. Then we can doc the int types.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: It is common to create span over the large byte buffer. And then to create span slices with the buffer. So, the customer is working with multiple span instances simultaneously. The "proper" practice would be to throw away the slice span and go back to the original and then not access bytes beyond the out span's capacity. 

So, yes, you're right that they shouldn't do this but I don't think we should explicitly say so. In reality, the only way they could do this is by getting the span's pointer and then doing whatever they want to with it. Really, they should not get the pointer as that is effectively unsafe code. If they just stick with ALL our other function, then the manipulations are safe. I'd rather discourage calling az_span_ptr.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: It is idiomatic in C to have a char* point to a string; not an unsigned char*. So I don't want to make this change in docs or in the function signature. 

Our documented contract is that we accept a pointer to a C string; which, by definition means a 0-terminated string. If the customer violates the contract, then behavior is unpredictable. So, it doesn't matter if the "buffer" (which is not what our contract says it expects) is not 0-terminated.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: A literal is NOT a constant. A literal is variable whose value to set/initialized at compile time. The value of the variable may change over time.

The code you wrote won't compile if foo is a global but it will compile if foo is a local. 

Literal is not our concept. It is a C language concept just like an expression is a C language concept. We have no choice but to keep these. We want to allow the customer code (and our code) to create global, initialized spans. And, we we also want to be able to create local, initialized spans. Unfortunately, the C compilers require different syntax to accomplish both these tasks and so, we have 2 macros.

The API doc and the conceptual docs I handed off to you have example of the usage for literal and expression macros. Between the docs and the sample, I think customers should get it. Plus, the customers know C, then they should be familiar with this too. And, if they get it wrong, the compiler will complain and then the customer should be able to figure this out without too much difficulty.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: Sounds good. Thanks for providing such details and help talk through the design and doc nuances :)

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: Works for me, as long as we remain consistent amongst ourselves.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
-AZ_NODISCARD bool az_span_is_content_equal_ignoring_case(az_span a, az_span b);
+AZ_NODISCARD bool az_span_is_content_equal_ignoring_case(az_span span1, az_span span2);
 
 /**
- * @brief convert all the span content to uint64
+ * @brief az_span_to_str copies a source span containing a string (not 0-terminated) to a
+ destination char buffer and appends the 0-terminating byte.
+
+ * The buffer refered to by destination must have a max_size that is at least 1 byte bigger
+ than the source az_span
```
ahsonkhan:
```suggestion
 * The buffer referred to by \p destination must have a size that is at least 1 byte bigger
 than the \p source az_span.
```

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ *  the string in converts \p destination span to zero-terminated str. Content is copied to \p source
```
ahsonkhan: How am I supposed to read this sentence fragment. Re-word?
```text
the string in converts \p destination span to zero-terminated str.
```

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ buffer and then \0 is
+ * addeed at the end. Then out_result will be created out of buffer
```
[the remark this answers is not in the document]
ahsonkhan: Also, what is `out_result` referring to?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
  *
- * @param self source span
- * @param out result span
- * @return AZ_NODISCARD az_span_to_uint64
+ * @param[in] destination A pointer to a buffer where the string should be copied
+ * @param[in] destination_max_size The maximum size of the buffer refered to by destination
```
ahsonkhan:
```suggestion
 * @param[in] destination_max_size The maximum available space within the buffer referred to by destination.
```

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * @param[in] source The az_span containing the not-0-terminated string
```
ahsonkhan:
```suggestion
 * @param[in] source The az_span containing the non-0-terminated characters.
```

ahsonkhan:
I'll try to finish reviewing the rest, tomorrow. Thanks for your patience.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,297 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
  *
+ * @param[in] span The original az_span
+ * @param[in] low_index An index into the original az_span indicating where the returned az_span
+ * will start
+ * @param[in] high_index An index into the original az_span indicting where the returned az_span
+ * should stop. The byte at the high_index is NOT included in the returned az_span.
+ * @return An az_span (view) into a portion (from low_index to high_index - 1) of the original
+ * az_span.
  */
 AZ_NODISCARD az_span az_span_slice(az_span span, int32_t low_index, int32_t high_index);
```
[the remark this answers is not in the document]
JeffreyRichter: We modeled after Go' slices but also how would you get the count? Typically, you scan forward in the buffer for the next occurrence of some byte until you find its offset and then you slice from the start index to the found index (exclusive). If you make it a count, then the customer has to do the math which is more error prone. I think we addressed the more common use case in a better way.

I'm fine with start_index & end_index instead of low/high.

sdk/core/core/inc/az_span.h
JeffreyRichter: There is no such thing as non-0-terminated characters. You can have a non-0-terminated STRING of characters. Or a non-0-terminated string.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
ahsonkhan: Btw, not all the typos were fixed/feedback addressed (for example `addeed`). Was this intentional?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
  *
- * @param dst buffer where to copy span
- * @param src span to be copied
- * @param out span resulting after copy
- * @return AZ_NODISCARD az_span_move
+ * @param span The az_span containing the ASCII digits to be parsed.
+ * @param out_number The pointer to the variable that is to receive the number
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_PARSER_UNEXPECTED_CHAR if a non-ASCII digit
```
ahsonkhan: I think it should also return an overflow if we see a span with too many digits that it won't fit a `uint32`. Same applies in other cases.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
  */
-AZ_NODISCARD az_result az_span_copy(az_span dst, az_span src, az_span* out);
+AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
-AZ_NODISCARD az_result az_span_copy_url_encode(az_span dst, az_span src, az_span* out);
+/******************************  SPAN APPENDING */
 
 /**
- * @brief append az_span if there is enough capacity for it
+ * @brief az_span_append_uint8 appends the uint8 a to the destination starting at the
```
ahsonkhan: Typo.
```suggestion
 * @brief az_span_append_uint8 appends the uint8 /p byte to the destination starting at the
```

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * @param[in] byte The uint8 to append to the destination span
+ * @param[out] out_span A pointer to an az_span that receives the span referring to the
+ * destination span with its length increased by 1
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY if the destination is not big enough to
+ * contain the appended byte
  */
-AZ_NODISCARD az_result az_span_append(az_span self, az_span span, az_span* out);
+AZ_NODISCARD az_result az_span_append_uint8(az_span destination, uint8_t byte, az_span* out_span);
```
ahsonkhan: This API shape is confusing to me and the `out_span` parameter description doesn't help :(
What's `out_span` supposed to contain and how is it used?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * @return AZ_NODISCARD az_span_move
+ * @param span The az_span containing the ASCII digits to be parsed.
+ * @param out_number The pointer to the variable that is to receive the number
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_PARSER_UNEXPECTED_CHAR if a non-ASCII digit
+ * is found within the span.
  */
-AZ_NODISCARD az_result az_span_copy(az_span dst, az_span src, az_span* out);
+AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
```
ahsonkhan: Why don't we have other number sizes, including signed/unsigned flavors?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * @return AZ_NODISCARD az_span_to_uint32
+ * @param[in] span The az_span containing the ASCII digits to be parsed.
+ * @param[in] out_number The pointer to the variable that is to receive the number
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_PARSER_UNEXPECTED_CHAR if a non-ASCII digit is found within the span.
  */
-AZ_NODISCARD az_result az_span_to_uint32(az_span self, uint32_t* out);
+
+AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
```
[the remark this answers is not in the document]
JeffreyRichter: Yes, I like source

sdk/core/core/inc/az_span.h
JeffreyRichter: Rick is looking in to this but we may decide to return an overflowed value. This is idiomatic for C (and many other languages): math just overflows. There is a code size and perf cost with validating this and it is very, very unlikely to actually happen so I'm tempted to just let it overflow.

sdk/core/core/inc/az_span.h
JeffreyRichter: This is a very common pattern for us. The destination span is passed by value with its pointer, length, & capacity. The returned span has the same pointer & capacity but the length is different based on what got appended. This allows all span instance to be "immutable" which reduces the chance of memory corruption/race conditions occurring. The end result is a much more predictable code base. And, if we ever do have a bug, it should be much simpler to locate to fix.

sdk/core/core/inc/az_span.h
JeffreyRichter: We have only what we need. We don't add arbitrary functions in anticipation of customer needs. Code size is at a premium. Also, we don't want to document, test, & support functions that we don't need and we don't know if a customer needs either.

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
gilbertw: Not intentional, There was lots of feedback on these docs.  I'm reviewing all of them again and will apply to a new PR anything that was missed.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * @brief Returns a sub span of the given span.
+ * @brief az_span_slice returns a new az_span which is a sub-span of the specified span.
  *
+ * @param[in] span The original az_span
+ * @param[in] low_index An index into the original az_span indicating where the returned az_span
+ * will start
+ * @param[in] high_index An index into the original az_span indicating where the returned az_span
+ * should stop. The byte at the high_index is NOT included in the returned az_span.
+ * @return An az_span (view) into a portion (from low_index to high_index - 1) of the original
+ * az_span.
```
ahsonkhan: We should mention what the "special" sentinel values mean like -1. Looks like a high_index of -1 means up to end (capacity and not length). We should document that and also that unlike high_index, low_index = -1 doesn't have any meaning (should it?).

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * hold the source's length
+ */
+AZ_NODISCARD az_result
+az_span_copy_url_encode(az_span destination, az_span source, az_span* out_span);
+
+/******************************  SPAN PAIR  */
+
+/**
+ * An az_pair represents a key/value pair of az_span instances.
+ *  This is typically used for HTTP query parameters and headers.
```
ahsonkhan: nit: indentation
```suggestion
 * This is typically used for HTTP query parameters and headers.
```

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * @brief                  Appends one uint8_t value to the #az_span if there is enough capacity for
- * it.
- * @param[in]  span        #az_span where to append the char.
- * @param[in]  c           Value to be appended.
- * @param[out] out_span    #az_span returned as result of the concatenation.
- * @return                 An #az_result value indicating the result of the operation.
- *                         #AZ_OK                      If no errors occurs.
- *                         #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY   If there is not enough capacity left to append the `c`.
+ * @brief The AZ_PAIR_NULL macro returns an az_pair instance whose key and value fields are
+ * initialized to AZ_SPAN_NULL
```
ahsonkhan: @danieljurek - is there a concept of "See also" or "refer to X" in doxygen and if so, how can we have links/pointers to other types/APIs in the doc comments?

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
- * @param[out] out_span   Pointer where to store the resulting #az_span.
- * @return                An #az_result value indicating the result of the operation.
- *                        #AZ_OK                      If no errors occurs.
- *                        #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY   If there is not enough capacity left to append the `n`.
+ * @brief az_pair_from_str returns an az_pair with its key and value fields initialized to span's
+ * over the specified key and value 0-terminated string parameters.
+ *
+ * @param[in] key A string representing the key
+ * @param[in] value A string representing the key's value
+ * @return  An az_pair with the field initialized to the az_span instances over the passed-in strings
```
ahsonkhan: Should we mention what the length/capacity of each span will be within the pair?

sdk/core/core/src/az_span.c
```diff
@@ -44,43 +44,39 @@ AZ_NODISCARD az_span az_span_slice(az_span span, int32_t low_index, int32_t high
...
-/**
- * ASCII lower case.
- */
 AZ_NODISCARD AZ_INLINE uint8_t az_ascii_lower(uint8_t value)
 {
   return 'A' <= value && value <= 'Z' ? value + AZ_ASCII_LOWER_DIF : value;
 }
 
-AZ_NODISCARD bool az_span_is_content_equal_ignoring_case(az_span a, az_span b)
+AZ_NODISCARD bool az_span_is_content_equal_ignoring_case(az_span span1, az_span span2)
```
ahsonkhan: Is there any tooling we can leverage to make sure the source and header match exactly (including the parameter names)? Otherwise, there is nothing auditing things to make sure they don't go out of sync (other than manual inspection/validation in code review).

sdk/core/core/src/az_span.c
```diff
@@ -226,74 +211,55 @@ AZ_INLINE void _az_uint8_swap(uint8_t* a, uint8_t* b)
 }
 
 /**
- * @brief move the content from span @b a to span @b b and viceverse.
+ * @brief move the content from span @span1 to span @span2 and viceverse.
```
ahsonkhan:
```suggestion
 * @brief move the content from span @span1 to span @span2 and vice versa.
```

ahsonkhan: Also, should this documentation live on the `az_span_private.h` header file instead of the source?

sdk/core/core/src/az_span.c
```diff
@@ -226,74 +211,55 @@ AZ_INLINE void _az_uint8_swap(uint8_t* a, uint8_t* b)
...
- * a = 221
- * b = 11
+ * span1 = 221
+ * span2 = 11
  *
  * @param a source/destination span
  * @param b destination/source span
  */
-void _az_span_swap(az_span a, az_span b)
+void _az_span_swap(az_span span1, az_span span2)
```
ahsonkhan: Do we need this method? Doesn't look like we have any callers.

sdk/core/core/src/az_span.c
```diff
@@ -453,23 +419,26 @@ static AZ_NODISCARD az_result _az_span_builder_append_uint64(az_span* self, uint
   return az_span_append(*self, az_span_init(&value_to_append, 1, 1), self);
 }
 
-AZ_NODISCARD az_result az_span_append_uint64(az_span* self, uint64_t n)
+AZ_NODISCARD az_result
+az_span_append_u64toa(az_span destination, uint64_t source, az_span* out_span)
```
ahsonkhan: Why is this signature split between two lines? Is there a length cut-off that we use as part of the coding style guidelines?

sdk/core/core/inc/az_span.h
ahsonkhan: Alright. Maybe we can reword the `out_span` description then. I'll see if I can propose something that I would find helpful as a reader.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * digits
+ * @param[out] out_span A pointer to an az_span that receives the span referring to the
+ * destination span with its length updated
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY if the /p destination is not big enough to
+ * contain the appended bytes
+ */
+AZ_NODISCARD az_result
+az_span_append_i32toa(az_span destination, int32_t source, az_span* out_span);
```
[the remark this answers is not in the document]
JeffreyRichter: Rick can decide - not a big deal to me either way

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * digits
+ * @param[out] out_span A pointer to an az_span that receives the span referring to the
+ * destination span with its length updated
+ * @return An #az_result value indicating the result of the operation:
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY if the destination is not big enough to
+ * contain the appended bytes
+ */
+AZ_NODISCARD az_result
+az_span_append_u32toa(az_span destination, uint32_t source, az_span* out_span);
```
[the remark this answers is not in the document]
JeffreyRichter: They are named differently because they do different things. *_uint8 actaully appends a single byte as is to the span.  *toa functions convert the number to ascii and then append it to the span; the number is NOT appended as-is.

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
 
-AZ_NODISCARD AZ_INLINE az_pair az_pair_from_str(char* key, char* value)
+/**
+ * @brief az_span_set sets all the bytes of the destination span (up to its capacity) to \p
+ * fill.
+ *
+ * @param[in] destination The span whose bytes will be set to \p fill
+ * @param[in] fill The byte to be replicated within the destination span
+ */
+AZ_INLINE void az_span_set(az_span destination, uint8_t fill)
```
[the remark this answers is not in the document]
JeffreyRichter: We went back and forth on this one a few times. I think IoT might have wanted it; did you look to see if they are using it for anything? If not, then we should kill this API. 
Your questions are reasonable and I'd prefer not to think them through and just kill the API instead. 
The key client sample uses this to "erase" a secret so that it doesn't stay in memory long and to reset the buffer to a known state. So, if IOT doesn't use this, then lets kill the API entirely. Keyvault can just call memset directly.

Good catch!

sdk/core/core/inc/az_span.h
```diff
@@ -135,161 +158,306 @@ AZ_NODISCARD az_span az_span_from_str(char* str);
...
+ * @param[in] source The span containing the bytes to copy to the destination
+ * @param[out] out_span A pointer to an az_span that receives the span referring to the destination
+ * span with its length updated
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if successful
+ *          #AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY if the /p destination is not big enough to hold the
+ * source's length
  */
-AZ_INLINE void az_span_set(az_span span, uint8_t fill)
+AZ_NODISCARD az_result az_span_copy(az_span destination, az_span source, az_span* out_span);
```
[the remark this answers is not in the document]
JeffreyRichter: Yes, @gilbertw can you please add this to the docs

sdk/core/core/inc/az_span.h
JeffreyRichter: I don't think this is necessary as this is the only thing that would make any sense.

sdk/core/core/src/az_span.c
JeffreyRichter: Not that the C team is aware of; we've asked them about this a few months ago.

sdk/core/core/src/az_span.c
JeffreyRichter: Yes. We should document the internal headers for our partners/client library engineers. This is just lower priority

sdk/core/core/src/az_span.c
JeffreyRichter: Great - if we have no caller then we should get rid of it. I know we have dead code in the product. We're hoping that code coverage will make this obvious and then we can get rid of stuff. Good catch

sdk/core/core/src/az_span.c
JeffreyRichter: Yes there is and we have some tools that reformat the code to a bunch of code style guidelines we have set up. Ask Victor, Anton, or Rick about them.

gilbertw: We use Clang formating of 100 characters.  The exact length was rather hotly contested but at the end the consensus was 100 was the standard we settled on in the guidelines
Take a look in the root folder ".clang-format" for our auto format settings

sdk/core/core/src/az_span.c
gilbertw: removed

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
gilbertw: @JeffreyRichter This is a large change.  We use parameters named "span" in a lot of places.  Also that might mean we want other places to be source1 source2
`AZ_NODISCARD AZ_INLINE bool az_span_is_content_equal(az_span span1, az_span span2)`

What about out_span?  I don't have a better way to describe an appended span.  Thus we will still have span on some parameters i.e. out_span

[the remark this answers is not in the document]
JeffreyRichter: That's fine, we can leave it as is.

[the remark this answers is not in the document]
ahsonkhan: I think its fine to keep using `span1`, `span2` where it makes sense and same with `out_span`. For the examples you shared, those parameter names have useful meaning. I am not sure that just because we have use of `span1`, `span2`, `out_span`, we should leave `span` as is. To me, those points are a bit unrelated. `span1`/`span2` isn't necessarily redundant, `span` is. In the places where `source` is more meaningful (when we have a single source span), we should use it. That said, I don't feel too strongly about it, and we can leave it as is if that's the consensus :)

PR #446, opened 2020-03-11
Adding Azure Embedded IoT SDK retry recommendations

CIPop:
- Adding retry documentation.
- Adding MQTT link.
- Documenting MQTT Will message.

- Unifying service status, error and retry timing in `az_iot_common.h`.


sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
         }
     }
 ```
+
+### Retrying Operations
+
+The IoT Core SDK documents and provides APIs for the error policy and timing policies separately. We are also supplying optional APIs for error classification and retry timing.
```
ericwol-msft: The IoT device library also supply's optional APIs for error classification and retry timing.

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+    {
+        // fail
+    }
+}
+```
+
+##### Provisioning Service Errors
+
+Each Provisioning operation will return a status encoded as `az_iot_provisioning_client_status`.
+After the register operation is complete the `az_iot_provisioning_client_registration_state registration_information` field will contain the `az_iot_provisioning_client_status error_code` that represents the overall status of the registration operation.
```
ericwol-msft: After the register operation is complete, the `az_iot_provisioning_client_registration_state registration_information` field will

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+// random_ms is a random value between 0 and max_jitter_msec
+// The previous operation took operation_msec
+
+int32_t total_delay_msec =  random_msec + _az_retry_calc_delay(attempt, min_retry_delay_msec, max_retry_delay_msec);
+int32_t delay_msec = max(operation_msec - total_delay_msec, 0); 
+```
+
+_Note 1_: The network stack may have used more time than the recommended delay before timing out. (e.g. The operation timed out after 2 minutes while the delay between operations is 1 second). In this case there is no need to delay the next operation.
+
+_Note 2_: To determine the parameters of the exponential with back-off retry strategy we recommend modeling the network characteristics (including failure-modes). Compare the results with defined SLAs for device connectivity (e.g. 1M devices must be connected in under 30 minutes) and with the available IoT Azure scale (especially consider _throttling_, _quotas_ and maximum _requests/connects per second_).
```
ericwol-msft: _Note 2_: To determine the parameters of the exponential with back-off retry strategy, we recommend modeling the network

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+- Authentication errors should not be retried.
+- Communication-related errors other than ones security-related should be considered retriable.
+
+Both IoT Hub and Provisioning services will use `MQTT CONNACK` as described in Section 3.2.2.3 of the [MQTT v3 specification](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Table_3.1_-).
+
+##### IoT Hub Service Errors
+
+APIs using `az_iot_hub_client_status` report service-side errors to the client through the IoT Hub protocols. (At the time of writing, only Twin responses may return errors.)
+
+The following APIs may be used to determine if the status indicates error and if the error should be retried:
```
ericwol-msft: The following APIs may be used to determine if the status indicates an error and if the error should be retried:

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+}
+```
+
+#### Suggested Retry Strategy
+
+Combining the functions above we recommend the following flow:
+
+![image](resources/iot_retry_flow.png "IoT MQTT Retry Flow")
+
+When devices are using IoT Hub without Provisioning Service we recommend attempting to rotate the IoT Credentials (SAS Token or X509 Certificate) on authentication issues.
```
ericwol-msft: When devices are using IoT Hub without Provisioning Service, we recommend attempting to rotate the IoT Credentials (SAS Token or

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Agreed - I will remove the "provides APIs" from the first sentence and replace the second with your proposal.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Ah: "the operation". Right!

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+
+Network timeouts and the MQTT keep-alive interval should be configured considering tradeoffs between how fast network issues are detected vs traffic overheads. [This document](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#default-keep-alive-timeout) describes the recommended keep-alive timeouts as well as the minimum idle timeout supported by Azure IoT services.
+
+For connectivity issues at all layers (TCP, TLS, MQTT) as well as cases where there is no `retry-after` sent by the service, we suggest using an exponential back-off with random jitter function. `_az_retry_calc_delay` is available in Azure SDK Core:
+
+```C
+// random_ms is a random value between 0 and max_jitter_msec
+// The previous operation took operation_msec
+
+int32_t total_delay_msec =  random_msec + _az_retry_calc_delay(attempt, min_retry_delay_msec, max_retry_delay_msec);
```
CIPop: @JeffreyRichter @antkmsft Please review this portion: 
1. `_az_retry_calc_delay` doesn't include a random jitter. This is fine as the application may have platform-specific ways of implementing that. Should it though be included in the calculation (to have a strong requirement for a random int)?
2. The API above is "internal" and I'm recommending it is used by app-developers. We do expect it to be mostly used by convenience layers but the Embedded devices IoT SDK should be usable as-is.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Reformulated.

jspaith:
> Device Provisioning and IoT Hub service protocols require additional state management on top of the MQTT protocol. The Azure IoT SDK for C provides a common programming model layered on an MQTT client selected by the application developer.

Per offline discussion, I think this is "embedded devices C SDK".  We should be consistent with whatever it is throughout docs.

---
Refers to: sdk/iot/core/doc/mqtt_state_machine.md:5 in 715bce6. [](commit_id = 715bce676c499eb420d07fcb9544db66b42fc05b, deletion_comment = False)

jspaith:
> 1. Parse service errors and output an uniform error object model. 

Do we care about adhering to markdown lint?  If so, opening this in Visual Studio Code will find places its unhappy about like trailing space.

---
Refers to: sdk/iot/core/doc/mqtt_state_machine.md:10 in 715bce6. [](commit_id = 715bce676c499eb420d07fcb9544db66b42fc05b, deletion_comment = True)

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+
+### Retrying Operations
+
+Retrying operations requires understanding two aspects: error evaluation (did the operation fail, should the operation be retried) and retry timing (how long to delay before retrying the operation). The IoT Embedded Devices SDK is supplying optional APIs for error classification and retry timing.
+
+#### Error policy
+
+The SDK will not handle protocol-level (WebSocket, MQTT, TLS or TCP) errors. The application-developer is expected to classify and handle errors the following way:
+
+- Authentication errors should not be retried.
```
danewalton: Pardon my ignorance here...Should this have some kind of qualifier about "retrying with the same parameters/state/etc"
Or is that assumed?

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+    {
+        // fail
+    }
+}
+```
+
+##### Provisioning Service Errors
+
+Each Provisioning operation will return a status encoded as `az_iot_provisioning_client_status`.
+After the register operation is complete, the `az_iot_provisioning_client_registration_state registration_information` field will contain the `az_iot_provisioning_client_status error_code` that represents the overall status of the registration operation.
```
danewalton: nit: could `az_iot_provisioning_client_status error_code` be changed to "... contain the `error_code` of type `az_iot_provisioning_client_status`". Reading quickly it might read as one type since a space and underscore can mesh.

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
         }
     }
 ```
+
+### Retrying Operations
+
+Retrying operations requires understanding two aspects: error evaluation (did the operation fail, should the operation be retried) and retry timing (how long to delay before retrying the operation). The IoT Embedded Devices SDK is supplying optional APIs for error classification and retry timing.
```
jspaith: >IoT Embedded Devices SDK [](start = 203, length = 24)

Naming convention on IoT Embedded Devices SDK?

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -28,12 +30,12 @@ The following aspects need to be handled by the application or convenience layer
 ## Design Decisions
 Porting requirements:
 - The target platform C99 compiler is generating reentrant code. The target platform supports a stack of several kB.
-- Our SDK relies on types such as `uint8_t` existing. If a platform doesn't already have them defined, they should be added to the application code using `typedef` statements.
+- The SDK relies on types such as `uint8_t` existing. If a platform doesn't already have them defined, they should be added to the application code using `typedef` statements.
 
 The SDK is provided only for the MQTT protocol. Support for WebSocket/WebProxy tunneling as well as TLS security is not handled by the SDK.
```
jspaith: >Support for WebSocket/WebProxy tunneling as well as TLS security is not handled by the SDK [](start = 48, length = 90)

I get that "not handled" is correct, but I'm not sure if someone new to our system will properly parse it.  They could think "this won't work in 1st place"

I think "not handled" could alternaly be said "out of scope" or "bring your own TLS/WebProxy" etc., correct?

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+
+Network timeouts and the MQTT keep-alive interval should be configured considering tradeoffs between how fast network issues are detected vs traffic overheads. [This document](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#default-keep-alive-timeout) describes the recommended keep-alive timeouts as well as the minimum idle timeout supported by Azure IoT services.
+
+For connectivity issues at all layers (TCP, TLS, MQTT) as well as cases where there is no `retry-after` sent by the service, we suggest using an exponential back-off with random jitter function. `_az_retry_calc_delay` is available in Azure SDK Core:
+
+```C
+// random_ms is a random value between 0 and max_jitter_msec
+// The previous operation took operation_msec
+
+int32_t total_delay_msec =  random_msec + _az_retry_calc_delay(attempt, min_retry_delay_msec, max_retry_delay_msec);
```
jspaith: >_az_retry_calc_delay [](start = 42, length = 20)

_az_retry_calc_delay is in an internal header but we're pointing customers at it.

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+        // fail
+    }
+}
+```
+
+#### Retry timing
+
+Network timeouts and the MQTT keep-alive interval should be configured considering tradeoffs between how fast network issues are detected vs traffic overheads. [This document](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#default-keep-alive-timeout) describes the recommended keep-alive timeouts as well as the minimum idle timeout supported by Azure IoT services.
+
+For connectivity issues at all layers (TCP, TLS, MQTT) as well as cases where there is no `retry-after` sent by the service, we suggest using an exponential back-off with random jitter function. `_az_retry_calc_delay` is available in Azure SDK Core:
```
danewalton: can `_az_retry_calc_delay` be linked to?

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+// random_ms is a random value between 0 and max_jitter_msec
+// The previous operation took operation_msec
+
+int32_t total_delay_msec =  random_msec + _az_retry_calc_delay(attempt, min_retry_delay_msec, max_retry_delay_msec);
+int32_t delay_msec = max(operation_msec - total_delay_msec, 0); 
+```
+
+_Note 1_: The network stack may have used more time than the recommended delay before timing out. (e.g. The operation timed out after 2 minutes while the delay between operations is 1 second). In this case there is no need to delay the next operation.
+
+_Note 2_: To determine the parameters of the exponential with back-off retry strategy, we recommend modeling the network characteristics (including failure-modes). Compare the results with defined SLAs for device connectivity (e.g. 1M devices must be connected in under 30 minutes) and with the available IoT Azure scale (especially consider _throttling_, _quotas_ and maximum _requests/connects per second_).
```
jspaith: >.g. 1M devices must be connected in under 30 minutes [](start = 228, length = 52)

SLA's can change over time and without notice.  Can we point to a permalink for the 1M devices / 30M.

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+
+In the absence of modeling, we recommend the following default:
```
jspaith: Nit: the term modeling is used all over the place in PnP.  I get "network modeling" is also a thing, but if we could not use "monitoring" here maybe better.  

Also I think it should be "defaults" and not "default" since it's per value.

Maybe something like "We recommend the following defaults unless your network has more specific requirements."

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+```
+
+#### Retry timing
+
+Network timeouts and the MQTT keep-alive interval should be configured considering tradeoffs between how fast network issues are detected vs traffic overheads. [This document](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#default-keep-alive-timeout) describes the recommended keep-alive timeouts as well as the minimum idle timeout supported by Azure IoT services.
+
+For connectivity issues at all layers (TCP, TLS, MQTT) as well as cases where there is no `retry-after` sent by the service, we suggest using an exponential back-off with random jitter function. `_az_retry_calc_delay` is available in Azure SDK Core:
+
+```C
+// random_ms is a random value between 0 and max_jitter_msec
```
jspaith: Consider adding " that the application calculates" or even some pseoudo-code that it's clear APP writes.  I get it's bordering on over-communication, but what SDK does and app does is more nuanced in this world.  And especially because we recommend a default for max_jitter_msec below, I worry that a dev could search in vain for our helper to determine this unless we're clear (or unless we move jitter into the API below per your comment to Jeffrey.

sdk/iot/core/doc/mqtt_state_machine.md
```diff
@@ -136,3 +138,120 @@ _Example:_
...
+
+#### Suggested Retry Strategy
+
+Combining the functions above we recommend the following flow:
+
+![image](resources/iot_retry_flow.png "IoT MQTT Retry Flow")
+
+When devices are using IoT Hub without Provisioning Service, we recommend attempting to rotate the IoT Credentials (SAS Token or X509 Certificate) on authentication issues.
+
+_Note:_ Authentication issues observed in the following cases do not require credentials to be rotated and require further user actions:
```
jspaith: >user [](start = 123, length = 4)

Nit: I think we should avoid word "user" here since it's unclear if this is end user or the "user of our SDK, aka dev".

sdk/iot/core/doc/mqtt_state_machine.md
JeffreyRichter: Terminology: There is an one & only one "Azure Embedded C SDK" which consists of a bunch of "client libraries". You can say "SDK" here because you are speaking on behalf of the entire Azure Embedded C SDK. If you want to talk about your "client Library", then you should use the term "IoT Embedded Devices Client Library"

sdk/iot/core/doc/mqtt_state_machine.md
JeffreyRichter: I'm not sure what IoT Core SDK refers to. If this is my team's core then this is "Azure C SDK's Core library" as it is not IoT-specific. If this is not my team's core, then this is not an SDK at all; but is instead a Client Library.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: @JeffreyRichter @jspaith there are multiple comments regarding naming - we're still deciding on that offline. Once it's finalized we'll make the update.
Also, I believe the change will also be made to the folder structure (right now it's iot/core, why I named this "IoT Core SDK" initially).

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Right, please keep that discussion in https://github.com/Azure/azure-sdk-for-c/pull/446#pullrequestreview-373235235

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: I don't link to anything else so I wouldn't here either. We should have that consistent throughout the Azure SDK. 
I do like the idea - do you want to open an Azure SDK bug to track that work?

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: This is the _application's_ SLA. For example (fictitious: ) 

> If a large-scale Internet outage is impacting >5M devices, Contoso would like the recovery time to be under 1hr to meet their mobile phone app's SLA of 99% monthly up-time. Assuming the outage gets resolved for all 5M at the same time, the devices should retry at a rate of no more than x  RPS against n units of IoT Hub S2 within the affected region. This would incur offline battery consumption of z mAh. 
To achieve this, the parameters for the timing retry policy should be ...

There is no permalink for an actual system since most of the system characteristics vary substantially between devices and usage.
I can and will add a permalink to the service SLAs which need to be factored in.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Model used in PnP is very different from the context in which modeling is used here. (https://en.wikipedia.org/wiki/Modeling_and_simulation)

See the disambiguation here https://en.wikipedia.org/wiki/Conceptual_model_(computer_science):
> Conceptual modeling should not be confused with other modeling disciplines such as data modelling, logical modelling and physical modelling. 


I don't agree with the reasoning that if PnP uses the noun form we can't use the verb here.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: > I think "not handled" could alternaly be said "out of scope" or "bring your own TLS/WebProxy" etc., correct?

@jspaith I agree - I'll reformulate to state that the app-developer must handle those items instead of saying we're "not supporting". I now see how that can be interpreted the wrong way.

> Terminology: There is an one & only one "Azure Embedded C SDK" [...]

@JeffreyRichter I'll add this and you to an internal discussion about naming. I will make the changes once we've settled on a name.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: I will modify this to call out "operations" need to be retried. I can also change "should be considered retriable" to "may be retried".
The retry itself means different parameters will be used at least at some level in the stack so I don't want to call that out. State has already changed (operation failed).

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Let's discuss this in https://github.com/Azure/azure-sdk-for-c/pull/446#discussion_r391888842

CIPop:
@jspaith I can't comment on https://github.com/Azure/azure-sdk-for-c/pull/446#issuecomment-598269638 and https://github.com/Azure/azure-sdk-for-c/pull/446#issuecomment-598271105 for some reason (GitHub bug?).

I didn't know about markdownlint. I'll fix all issues.
Let's keep naming in the discussion above.

sdk/iot/core/doc/mqtt_state_machine.md
JeffreyRichter: If you expect customers code to call this (under ANY circumstance [like not using one of your convenience layers), then this function should be public, documents, tested, fully supported, and NOT start with an underscore. 

I don't think that jitter is important. We have a strong goal of keeping the PAL functions to a minimum. Everything the customer has to do to bring up the SDK on a new device is a barrier to entry so the fewer functions the better. How bout we ignore jitter for now and then address it IF it becomes a problem later,

CIPop: I do expect customer code to call this since there is nothing else similar to it. Should I:

1. Propose a change in SDKs to make this public
1. Make our own az_iot specific public function (in which case I'll likely add a mandatory jitter "random" arg).

JeffreyRichter: Yes, make this public in your client library. You can add the jitter param.

sdk/iot/core/doc/mqtt_state_machine.md
CIPop: Removed since @jspaith @JeffreyRichter and I decided to unify error and retry.

PR #450, opened 2020-03-12
Explicitly have correct test values and clang format

danewalton:
After continuing with other tests for other components, I like better @CIPop idea about having the correct strings to compare to explicitly instead of macros. Makes others and more importantly me be more careful about what the correct "thing" should be.

### DIFF
--- a/sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
+++ b/sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
@@ -23,7 +23,7 @@
 
 #define TEST_BAD_SPAN \
   { \
-    ._internal = { .ptr = (uint8_t*)0x7FFFFFFF, .length = -1, .capacity = 10  }, \
+    ._internal = { .ptr = (uint8_t*)0x7FFFFFFF, .length = -1, .capacity = 10 }, \
   }
 #define TEST_EMPTY_OPTIONS \
   { \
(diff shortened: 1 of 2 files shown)


CIPop:
Since you're changing things - could you also please also check that all our files have the mandatory license header?

danewalton:
> 
> 
> Since you're changing things - could you also please also check that all our files have the mandatory license header?

Yes good point will check

sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -87,7 +83,8 @@ void test_az_iot_hub_client_telemetry_publish_topic_get_NULL_client_fails(void**
 
   uint8_t mqtt_topic_buf[TEST_MQTT_SPAN_BUFFER_SIZE];
 
-  az_span mqtt_topic = az_span_init(mqtt_topic_buf, 0, sizeof(mqtt_topic_buf) / sizeof(mqtt_topic_buf[0]));
+  az_span mqtt_topic
+      = az_span_init(mqtt_topic_buf, 0, sizeof(mqtt_topic_buf) / sizeof(mqtt_topic_buf[0]));
```
CIPop: It would be good to have a [_countof(array)](https://stackoverflow.com/questions/4415524/common-array-length-macro-for-c). 
@barcharcraz, @antkmsft /cc @JeffreyRichter should we define our own or add to core internal? Does something like that exist already and we've missed it?

danewalton: Just as a note I have a list of things to bring up with @JeffreyRichter and friends. Feedback from azrtos team and others. Likely will be for Friday.

PR #451, opened 2020-03-12
Improve az_log interface

antkmsft:


sdk/core/core/src/az_http_policy_logging.c
```diff
@@ -56,7 +56,7 @@ static az_result _az_http_policy_logging_append_http_request_msg(
 
   if (request == NULL)
   {
-    AZ_RETURN_IF_FAILED(az_span_append(*ref_log_msg, AZ_SPAN_FROM_STR("NULL"), ref_log_msg));
+    AZ_RETURN_IF_FAILED(az_span_append(*ref_log_msg, AZ_SPAN_FROM_STR("NULL\0"), ref_log_msg));
```
JeffreyRichter: Is the \0 here to 0-terminate the message?
If so, I wouldn't do this. Instead, I keep all the logging code as it was, and then, just before calling the callback function, I'd append a 0 to the end of the span. If there is no room, then I'd just force the last byte of the span to be 0 and then call the callback.

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -51,7 +51,7 @@ AZ_INLINE az_result _az_http_policy_retry_append_http_retry_msg(
       az_span_append(*ref_log_msg, AZ_SPAN_FROM_STR(" will be made in "), ref_log_msg));
 
   AZ_RETURN_IF_FAILED(az_span_append_i32toa(*ref_log_msg, delay_msec, ref_log_msg));
-  AZ_RETURN_IF_FAILED(az_span_append(*ref_log_msg, AZ_SPAN_FROM_STR("ms."), ref_log_msg));
+  AZ_RETURN_IF_FAILED(az_span_append(*ref_log_msg, AZ_SPAN_FROM_STR("ms.\0"), ref_log_msg));
```
JeffreyRichter: Same thing here

sdk/core/core/src/az_http_policy_retry.c
```diff
@@ -61,7 +61,9 @@ AZ_INLINE void _az_http_policy_retry_log(int16_t attempt, int32_t delay_msec)
   uint8_t log_msg_buf[AZ_LOG_MSG_BUF_SIZE] = { 0 };
   az_span log_msg = AZ_SPAN_FROM_BUFFER(log_msg_buf);
   (void)_az_http_policy_retry_append_http_retry_msg(attempt, delay_msec, &log_msg);
-  az_log_write(AZ_LOG_HTTP_RETRY, log_msg);
+
+  // The span is going to b xero terminated, we wanto to exclude that from the length.
```
JeffreyRichter: typos: be 0-terminated.
But, see my comments earlier

sdk/core/core/src/az_log.c
```diff
@@ -10,62 +10,55 @@
...
 
-void az_log_write(az_log_classification classification, az_span message)
+void az_log_write(az_log_classification classification, char const* message, int32_t message_length)
 {
-  // TODO: thread safety
-  if (_az_log_listener != NULL && az_log_should_write(classification))
+  if (_az_log_message_callback != NULL && az_log_should_write(classification))
   {
-    (*_az_log_listener)(classification, message);
+    _az_log_message_callback(classification, message, message_length);
```
JeffreyRichter: This function should take the span like it used to.
Then, in the IF, this is where you put the code to 0-terminate the span and then convert it to a char * in order to call _az_log_message_callback.

sdk/core/core/src/az_http_policy_logging.c
antkmsft: Done.

sdk/core/core/src/az_http_policy_retry.c
antkmsft: Done.

sdk/core/core/src/az_http_policy_retry.c
antkmsft: Done. (N/A)

sdk/core/core/src/az_log.c
antkmsft: Please see the new iteration, I could've overengineered it a bit for the sake of safety, but that's logging code, so I hope we can do a few extra "if" statements.
 
But the significant thing is that now AZ_SPAN_LITERAL_FROM_STR()'s span capacity (not length!) now includes the zero terminator. Which is not untrue.

I am doing my nest effort to prevent AV when logging function gets invoked with the string literal that resides in the read-only memory. It is now guaranteed safe to call it with a literal defined with AZ_SPAN_LITERAL_FROM_STR(). Realistically, I think it will cover the reasonable cases - for dynamical content, you have modifiable buffer, for static messages you have string literals.

sdk/core/core/src/az_log.c
```diff
@@ -10,62 +10,71 @@
...
-  if (_az_log_listener != NULL && az_log_should_write(classification))
+  if (_az_log_message_callback != NULL && az_log_should_write(classification))
   {
-    (*_az_log_listener)(classification, message);
+    int32_t const length = az_span_length(message);
+    int32_t const capacity = az_span_capacity(message);
+    // Do nothing if span is empty or if truncating it with zero would result in empty span.
+    if (length > 0 && capacity > 1)
+    {
+      int32_t const z_pos = length < capacity ? length + 1 : capacity - 1;
```
JeffreyRichter: I think you have a bug here. If length < capacity, then adding 1 to length could set it equal to capacity. And then buf[capacity] =0 (line 46) writes outside the buffer.

I would have just done this:
zero_pos = (length < capacity) ? length : capacity -1;
if (buf[zero_pos != 0) { buf[zero_pos] = 0; }

sdk/core/core/src/az_log.c
```diff
@@ -10,62 +10,71 @@
...
+      char* const buf = (char*)az_span_ptr(message);
+
+      // Don't write 0 if it's already there, for instance when the code logs a string literal span,
+      // don't try to modify it as there is a chance it resides in .text.
+      if (buf[z_pos] != 0)
+      {
+        buf[z_pos] = 0;
+      }
+
+      _az_log_message_callback(classification, buf, z_pos - 1);
```
JeffreyRichter: I don't think you want -1 here. The z_pos is the length.

sdk/core/core/src/az_span.c
```diff
@@ -29,16 +29,16 @@ AZ_NODISCARD az_span az_span_from_str(char* str)
...
+  return az_span_init((uint8_t*)str, length, length + 1);
 }
 
 AZ_NODISCARD az_span az_span_slice(az_span span, int32_t low_index, int32_t high_index)
 {
   AZ_PRECONDITION_RANGE(-1, high_index, az_span_capacity(span));
   AZ_PRECONDITION(high_index == -1 || (high_index >= 0 && low_index <= high_index));
 
-  int32_t capacity = az_span_capacity(span);
+  int32_t const capacity = az_span_length(span);
```
JeffreyRichter: Why did you change this to length? We want capacity here.

sdk/platform/http_client/curl/src/az_curl.c
```diff
@@ -340,7 +340,7 @@ static int32_t _az_http_client_curl_upload_read_callback(
   // Update the userdata span
   //  ptr will point to remaining data to be copied
   //  length and capacity are set to the size of the remaining content
-  *upload_content = az_span_slice(*upload_content, dst_buffer_size, -1);
+  *upload_content = az_span_slice(*upload_content, dst_buffer_size, az_span_length(*upload_content));
```
JeffreyRichter: This makes me nervous. -1 should be usable here. Did your changes break the meaning of -1?

sdk/core/core/src/az_log.c
antkmsft: This is why I attempted to change `AZ_SPAN_FROM_STR()` to have capacity that is greater than length by 1. Because the compiler does place the 0 at the end of the string literal, it is there.
But it looks like doing that does break `az_span_slice()`, and so far it is hard for me to get it working in all contexts where it is used with `-1`.
I think I should revert the code pretty much to what I've had before. I would extract adding 0 to a buffer logic into a separate inline function, so that when we update the code that forms the message to be printed, there is less chance to miss/misplace the 0.
I do not feel comfortable appending zero to the span too high in the chain of calls. Yes, it requires only updating code in one place, but now `az_log_wrinte(AZ_SPAN_FROM_STR("Hello"))` would be AV on Windows, and is a UB on microcontrollers, where string would sit in a read-only memory.

sdk/core/core/src/az_span.c
antkmsft: I will undo this change, sorry that it got here. I was trying to see how hard would it be to change `AZ_SPAN_FROM_STR()` in a way that for `"Hello"` it has `length = 5` but `capacity = 6`.

sdk/platform/http_client/curl/src/az_curl.c
antkmsft: I will undo this change. It is not that I wanted to change the meaning of `-1`, I just wanted the spans that were created from string literals to have one extra byte of capacity, for the 0 character that is always there for the string literals.

sdk/core/core/src/az_log.c
JeffreyRichter: Maybe the best answer is to put it all back the way it was and pass the span to the customer with the not-0-terminated string and the length.

antkmsft: I would keep the `const char*` and `length`, but I would put the code that appends 0 in the same scope where message buffer is created. I would extract the code tat appends 0 to a separate inline internal function that all logging code can reuse.
Requiring `const char*` and `length` in `az_log_write()` + having a zero-terminating function in the logging internal header would also signal to the internal code what is the expected format, and it can use string literals if needed, without trying to modify a read-only memory.

sdk/platform/http_client/curl/src/az_curl.c
antkmsft: Undone.

sdk/core/core/src/az_span.c
antkmsft: Undone.

sdk/core/core/src/az_log.c
antkmsft: No longer applies.

sdk/core/core/src/az_log.c
antkmsft: I've achieved what I am describing above in a different way.

sdk/core/core/src/az_log.c
antkmsft: @JeffreyRichter , how do you like the current version I just pushed? We still keep `conat chars*` but we also don't try to write `0` to just any buffer. `az_log_write()` would work correctly with the read-only buffer, given that the message is 0-terminated.

PR #455, opened 2020-03-12
Add IoT Hub Client Functionality

danewalton:
This includes three api's
`az_iot_hub_client_init()`
`az_iot_hub_client_user_name_get()`
`az_iot_hub_client_id_get()`
and associated testing.

If anyone would like to compare the resulting user name and client id to the documented requirements, links to those docs are here:
[Username and Client ID without Modules](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-device)
[Username and Client ID with Modules](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-module)

### DIFF
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -17,7 +17,22 @@ TBD
 To use IoT Hub connectivity, the first action by a developer should be to initialize the
 client with the `az_iot_hub_client_init()` API. Once that is initialized, you may use the
 `az_iot_hub_client_user_name_get()` and `az_iot_hub_client_client_id_get()` to get the
-user name and client id to establish a connection with IoT Hub. An example use case is below.
+user name and client id to establish a connection with IoT Hub. 
+
+The user name will be of the following form:
+```C
+//[Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
+//[Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
+```
+
+The client id will be of the following form:
+```C
+//[Format without module id] {device_id}
+//[Format with module id] {device_id}/{module_id}
+```
+
+
+An example use case is below.
 
 ```C
 //FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -12,11 +12,54 @@ TBD
 
 ## Examples
 
+### IoT Hub Client
+
+To use IoT Hub connectivity, the first action by a developer should be to initialize the
+client with the `az_iot_hub_client_init()` API. Once that is initialized, you may use the
+`az_iot_hub_client_user_name_get()` and `az_iot_hub_client_client_id_get()` to get the
+user name and client id to establish a connection with IoT Hub. An example use case is below.
+
+```C
+//FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
+
+az_iot_hub_client my_client;
+static az_span my_iothub_hostname = AZ_SPAN_LITERAL_FROM_STR("constoso.azure-devices.net");
+static az_span my_device_id = AZ_SPAN_LITERAL_FROM_STR("contoso_device");
+
+//Make sure to size the buffer to fit the user name (100 is an example)
+static uint8_t my_mqtt_user_name_buffer[100];
+static az_span my_mqtt_user_name = AZ_SPAN_LITERAL_FROM_BUFFER(my_mqtt_user_name_buffer);
+
+//Make sure to size the buffer to fit the client id (16 is an example)
+static uint8_t my_mqtt_client_id_buffer[16];
+static az_span my_mqtt_client_id = AZ_SPAN_LITERAL_FROM_BUFFER(my_mqtt_client_id_buffer);
+
+int main()
+{
+  //Get the default IoT Hub options
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+
+  //Initialize the client with hostname, device id, and options
+  az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, &options);
+
+  //Get the MQTT user name to connect
+  az_iot_hub_client_user_name_get(&my_client, my_mqtt_user_name, &my_mqtt_user_name)
+
+  //Get the MQTT client id to connect
+  az_iot_hub_client_client_id_get(&my_client, my_mqtt_client_id, &my_mqtt_client_id);
+
+  //At this point you are free to use my_mqtt_client_id and my_mqtt_user_name to connect using
+  //your MQTT client.
+}
+```
+
 ### Telemetry
 
 Telemetry functionality can be achieved by sending a user payload to a specific topic. In order to get the appropriate topic to which to send, use the `az_iot_hub_client_telemetry_publish_topic_get()` API. An example use case is below.
 
 ```C
+//FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
+
 static az_iot_hub_client my_client;
 static az_span my_iothub_hostname = AZ_SPAN_LITERAL_FROM_STR("constoso.azure-devices.net");
 static az_span my_device_id = AZ_SPAN_LITERAL_FROM_STR("contoso_device");
@@ -30,7 +73,8 @@ void my_telemetry_func()
   //Optionally, the size can include space for a null terminator.
   //Here, the buffer is zero initialized to null terminate the topic.
   uint8_t telemetry_topic_buffer[64] = { 0 };
-  az_span topic_span = AZ_SPAN_FROM_BUFFER(telemetry_topic_buffer);
+  az_span topic_span = az_span_init(telemetry_topic_buffer, 0,
+                sizeof(telemetry_topic_buffer) / sizeof(telemetry_topic_buffer[0]));
 
   //Get the NULL terminated topic and put in topic_span to send the telemetry
   az_iot_hub_client_telemetry_publish_topic_get(&client, NULL, topic_span, &topic_span);
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -26,8 +26,10 @@ void my_telemetry_func()
   //Initialize the client to then pass to the telemetry API
   az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, NULL);
 
-  //Allocate a span to put the telemetry topic
-  uint8_t telemetry_topic_buffer[64];
+  //Allocate a span with capacity large enough to put the telemetry topic.
+  //Optionally, the size can include space for a null terminator.
+  //Here, the buffer is zero initialized to null terminate the topic.
+  uint8_t telemetry_topic_buffer[64] = { 0 };
   az_span topic_span = AZ_SPAN_FROM_BUFFER(telemetry_topic_buffer);
 
   //Get the NULL terminated topic and put in topic_span to send the telemetry
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -19,19 +19,6 @@ client with the `az_iot_hub_client_init()` API. Once that is initialized, you ma
 `az_iot_hub_client_user_name_get()` and `az_iot_hub_client_client_id_get()` to get the
 user name and client id to establish a connection with IoT Hub. 
 
-The user name will be of the following form:
-```C
-//[Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
-//[Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
-```
-
-The client id will be of the following form:
-```C
-//[Format without module id] {device_id}
-//[Format with module id] {device_id}/{module_id}
-```
-
-
 An example use case is below.
 
 ```C
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -92,7 +92,7 @@ void my_telemetry_func()
                 sizeof(telemetry_topic_buffer) / sizeof(telemetry_topic_buffer[0]));
 
   //Get the NULL terminated topic and put in topic_span to send the telemetry
-  az_iot_hub_client_telemetry_publish_topic_get(&client, NULL, topic_span, &topic_span);
+  az_iot_hub_client_telemetry_publish_topic_get(&my_client, NULL, topic_span, &topic_span);
 }
 ```
 
--- a/sdk/iot/core/inc/az_iot_hub_client.h
+++ b/sdk/iot/core/inc/az_iot_hub_client.h
@@ -69,6 +69,10 @@ AZ_NODISCARD az_result az_iot_hub_client_init(
 
 /**
  * @brief Gets the MQTT user name.
+ * 
+ * The user name will be of the following format:
+ * [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
+ * [Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT user name.
@@ -82,6 +86,10 @@ AZ_NODISCARD az_result az_iot_hub_client_user_name_get(
 
 /**
  * @brief Gets the MQTT client id.
+ * 
+ * The client id will be of the following format:
+ * [Format without module id] {device_id}
+ * [Format with module id] {device_id}/{module_id}
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] mqtt_client_id An empty #az_span with sufficient capacity to hold the MQTT client id.
--- a/sdk/iot/core/inc/az_iot_hub_client.h
+++ b/sdk/iot/core/inc/az_iot_hub_client.h
@@ -71,8 +71,8 @@ AZ_NODISCARD az_result az_iot_hub_client_init(
  * @brief Gets the MQTT user name.
  * 
  * The user name will be of the following format:
- * [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
- * [Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
+ * [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30&{user_agent}
+ * [Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30&{user_agent}
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT user name.
--- a/sdk/iot/core/inc/az_iot_hub_client.h
+++ b/sdk/iot/core/inc/az_iot_hub_client.h
@@ -222,6 +222,10 @@ az_iot_hub_client_properties_next(az_iot_hub_client_properties* properties, az_p
 /**
  * @brief Gets the MQTT topic that must be used for device to cloud telemetry messages.
  * @note Telemetry MQTT Publish messages must have QoS At Least Once (1).
+ * 
+ * Should the user want a null terminated topic string, they may allocate a buffer large enough
+ * to fit the topic plus a null terminator. They must set the last byte themselves or zero initialize
+ * the buffer.
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] properties An optional #az_iot_hub_client_properties object (can be NULL).
--- a/sdk/iot/core/src/az_iot_hub_client.c
+++ b/sdk/iot/core/src/az_iot_hub_client.c
@@ -1,12 +1,128 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+#include <stdint.h>
+
 #include "az_iot_hub_client.h"
+#include <az_precondition.h>
+#include <az_result.h>
 #include <az_span.h>
 
 #include <_az_cfg.h>
 
+static const uint8_t hub_client_forward_slash = '/';
+static const uint8_t hub_client_param_separator = '&';
+static const uint8_t hub_client_null_terminate = '\0';
+
+static const az_span hub_client_api_version = AZ_SPAN_LITERAL_FROM_STR("?api-version=2018-06-30");
+
 AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default()
 {
   return (az_iot_hub_client_options){ .module_id = AZ_SPAN_NULL, .user_agent = AZ_SPAN_NULL };
+}
+
+AZ_NODISCARD az_result az_iot_hub_client_init(
+    az_iot_hub_client* client,
+    az_span iot_hub_hostname,
+    az_span device_id,
+    az_iot_hub_client_options const* options)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(iot_hub_hostname, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(device_id, 1, false);
+  AZ_PRECONDITION_NOT_NULL(options);
+
+  client->_internal.iot_hub_hostname = iot_hub_hostname;
+  client->_internal.device_id = device_id;
+  client->_internal.options.module_id = options->module_id;
+  client->_internal.options.user_agent = options->user_agent;
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_iot_hub_client_user_name_get(
+    az_iot_hub_client const* client,
+    az_span mqtt_user_name,
+    az_span* out_mqtt_user_name)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_user_name, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_user_name);
+
+  const az_span* module_id = &(client->_internal.options.module_id);
+  const az_span* user_agent = &(client->_internal.options.user_agent);
+
+  if (az_span_capacity(mqtt_user_name)
+      < (int32_t)(
+            az_span_length(client->_internal.iot_hub_hostname) + sizeof(hub_client_forward_slash)
+            + az_span_length(client->_internal.device_id) + sizeof(hub_client_forward_slash)
+            + az_span_length(*module_id)
+            + (az_span_length(*module_id) != 0 ? sizeof(hub_client_forward_slash) : 0)
+            + az_span_length(*user_agent)
+            + (az_span_length(*user_agent) != 0 ? sizeof(hub_client_param_separator) : 0)
+            + az_span_length(hub_client_api_version) + sizeof(hub_client_null_terminate)))
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY;
+  }
+
+  AZ_RETURN_IF_FAILED(
+      az_span_append(mqtt_user_name, client->_internal.iot_hub_hostname, out_mqtt_user_name));
+  AZ_RETURN_IF_FAILED(
+      az_span_append_uint8(*out_mqtt_user_name, hub_client_forward_slash, out_mqtt_user_name));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_user_name, client->_internal.device_id, out_mqtt_user_name));
+  AZ_RETURN_IF_FAILED(
+      az_span_append_uint8(*out_mqtt_user_name, hub_client_forward_slash, out_mqtt_user_name));
+  if (az_span_length(*module_id) > 0)
+  {
+    AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_user_name, *module_id, out_mqtt_user_name));
+    AZ_RETURN_IF_FAILED(
+        az_span_append_uint8(*out_mqtt_user_name, hub_client_forward_slash, out_mqtt_user_name));
+  }
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_user_name, hub_client_api_version, out_mqtt_user_name));
+  if (az_span_length(*user_agent) > 0)
+  {
+    AZ_RETURN_IF_FAILED(
+        az_span_append_uint8(*out_mqtt_user_name, hub_client_param_separator, out_mqtt_user_name));
+    AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_user_name, *user_agent, out_mqtt_user_name));
+  }
+  AZ_RETURN_IF_FAILED(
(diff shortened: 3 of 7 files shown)


sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,128 @@
...
+
+  const az_span* module_id = &(client->_internal.options.module_id);
+  const az_span* user_agent = &(client->_internal.options.user_agent);
+
+  if (az_span_capacity(mqtt_user_name)
+      < (int32_t)(
+            az_span_length(client->_internal.iot_hub_hostname) + sizeof(hub_client_forward_slash)
+            + az_span_length(client->_internal.device_id) + sizeof(hub_client_forward_slash)
+            + az_span_length(*module_id)
+            + (az_span_length(*module_id) != 0 ? sizeof(hub_client_forward_slash) : 0)
```
danewalton: make these greater than instead

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,128 @@
...
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(iot_hub_hostname, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(device_id, 1, false);
+  AZ_PRECONDITION_NOT_NULL(options);
+
+  client->_internal.iot_hub_hostname = iot_hub_hostname;
+  client->_internal.device_id = device_id;
+  client->_internal.options.module_id = options->module_id;
+  client->_internal.options.user_agent = options->user_agent;
```
hihigupt: how about memcpy of this struct? as option contains all the data in itself.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,128 @@
...
+  }
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_user_name, hub_client_api_version, out_mqtt_user_name));
+  if (az_span_length(*user_agent) > 0)
+  {
+    AZ_RETURN_IF_FAILED(
+        az_span_append_uint8(*out_mqtt_user_name, hub_client_param_separator, out_mqtt_user_name));
+    AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_user_name, *user_agent, out_mqtt_user_name));
+  }
+  AZ_RETURN_IF_FAILED(
```
hihigupt: can we add vertical spacing between blocks of code?

danewalton: yup adding spacing between if's and other blocks. thanks

sdk/iot/core/src/az_iot_hub_client.c
danewalton: just for options?

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,128 @@
...
+      az_span_append_uint8(*out_mqtt_user_name, hub_client_null_terminate, out_mqtt_user_name));
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_iot_hub_client_id_get(
```
hihigupt: Can we have some comments that show sample output of this impl?
Atleast the format that it will emit.

hihigupt: I mean all the API's impl

---
In reply to: [392373089](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r392373089) [](ancestors = 392373089)

danewalton: whoops yea will add to the README with the telemetry snippet.

hihigupt: That will be great for customer, what about developers who are looking at the impl? It is important to have comments on *c file also.

---
In reply to: [392373835](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r392373835) [](ancestors = 392373835)

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,133 @@
...
+AZ_NODISCARD az_result az_iot_hub_client_init(
+    az_iot_hub_client* client,
+    az_span iot_hub_hostname,
+    az_span device_id,
+    az_iot_hub_client_options const* options)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(iot_hub_hostname, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(device_id, 1, false);
+  AZ_PRECONDITION_NOT_NULL(options);
```
danewalton: @CIPop Should `NULL` be allowed here or should the behavior be that the user either uses a custom set (which could be all options `AZ_SPAN_NULL`) or uses default. But in either case options isn't `NULL`.

danewalton: [Current header says it can](https://github.com/Azure/azure-sdk-for-c/blob/95ff0e13dcaf5a6d70014037b1e652fc73b2ef74/sdk/iot/core/inc/az_iot_hub_client.h#L61) but in the storage SDK `NULL` isn't allowed. [Link here](https://github.com/Azure/azure-sdk-for-c/blob/95ff0e13dcaf5a6d70014037b1e652fc73b2ef74/sdk/storage/blobs/src/az_storage_blobs_blob_client.c#L61)

sdk/iot/core/src/az_iot_hub_client.c
danewalton: Would adding something like this be sufficient?
```C
// [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
// [Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
```

danewalton: added above the functions in the `.c`'s and in the doxygen

sdk/iot/core/src/az_iot_hub_client.c
danewalton: pushed change to use `memcpy` for options

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+
+  client->_internal.iot_hub_hostname = iot_hub_hostname;
+  client->_internal.device_id = device_id;
+  memcpy(&client->_internal.options, options, sizeof(az_iot_hub_client_options));
+
+  return AZ_OK;
+}
+
+//[Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
```
hihigupt: >//[Format without module id]  [](start = 0, length = 29)

Is it required to format comments in doxygen in *.c also?

danewalton: No only header files are parsed.

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
```diff
@@ -21,3 +36,175 @@ void test_az_iot_hub_client_get_default_options_succeed(void** state)
...
+
+void test_az_iot_hub_client_init_custom_options_succeed(void** state)
+{
+  (void)state;
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options;
+  options.module_id = AZ_SPAN_FROM_STR(TEST_MODULE_ID);
+  options.user_agent = AZ_SPAN_FROM_STR(TEST_USER_AGENT);
+  assert_true(
```
hihigupt: >assert_true [](start = 2, length = 11)

Assert_int_equal, helps to show what was the return value of function, when assert failed.

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
```diff
@@ -21,3 +36,175 @@ void test_az_iot_hub_client_get_default_options_succeed(void** state)
...
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
+
+  assert_string_equal(TEST_DEVICE_ID_STR, az_span_ptr(client._internal.device_id));
+  assert_string_equal(TEST_DEVICE_HOSTNAME_STR, az_span_ptr(client._internal.iot_hub_hostname));
+  assert_string_equal(TEST_MODULE_ID, az_span_ptr(client._internal.options.module_id));
+  assert_string_equal(TEST_USER_AGENT, az_span_ptr(client._internal.options.user_agent));
+}
+
+void test_az_iot_hub_client_user_name_get_succeed(void** state)
```
hihigupt: >test_az_iot_hub_client_user_name_get_succeed [](start = 5, length = 44)

Can we have bound check unit tests also ?

when we give memory with pattern in it and expect pattern is not erased outsize of capacity given to function?

hihigupt:
> #include <stdint.h>

Nit: can you please group all std header and then sort them?

---
Refers to: sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c:12 in fd34a76. [](commit_id = fd34a767220c67e3baea0eb5091d34935c0031e1, deletion_comment = False)

sdk/iot/core/tests/cmocka/main.c
```diff
@@ -57,6 +57,16 @@ void test_az_iot_hub_client_telemetry_publish_topic_get_with_options_user_agent_
...
+void test_az_iot_hub_client_init_succeed(void** state);
+void test_az_iot_hub_client_init_custom_options_succeed(void** state);
+void test_az_iot_hub_client_user_name_get_succeed(void** state);
+void test_az_iot_hub_client_user_name_get_small_buffer_fail(void** state);
+void test_az_iot_hub_client_user_name_get_user_options_succeed(void** state);
+void test_az_iot_hub_client_user_name_get_user_options_small_buffer_fail(void** state);
+void test_az_iot_hub_client_id_get_succeed(void** state);
+void test_az_iot_hub_client_id_get_small_buffer_fail(void** state);
+void test_az_iot_hub_client_id_module_get_succeed(void** state);
+void test_az_iot_hub_client_id_module_get_small_buffer_fail(void** state);
```
hihigupt: how about sorting them case on name of the function?

danewalton: I can bring this up in the meeting today but the policy on unit testing framework (grouping, where functions are declared etc) will be discussed today.

sdk/iot/core/tests/cmocka/main.c
```diff
@@ -104,6 +114,16 @@ int main()
...
+    cmocka_unit_test(test_az_iot_hub_client_init_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_init_custom_options_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_user_name_get_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_user_name_get_small_buffer_fail),
+    cmocka_unit_test(test_az_iot_hub_client_user_name_get_user_options_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_user_name_get_user_options_small_buffer_fail),
+    cmocka_unit_test(test_az_iot_hub_client_id_get_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_id_get_small_buffer_fail),
+    cmocka_unit_test(test_az_iot_hub_client_id_module_get_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_id_module_get_small_buffer_fail),
```
hihigupt: this is becoming unmanageable

Can we create a variable for each group? and then pass to group runner?

danewalton: Will be discussed in the SDK meeting today. I am continuing with the way the IoT SDK has written unit tests to not block and will reformat the tests as decided on by the SDK at large. Agreed though.

hihigupt:
Can you also add RTOS team into the PR review.

danewalton:
> 
> 
> Can you also add RTOS team into the PR review.

@hihigupt yup will send to them

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
danewalton: good call. I'll make a PR updating the other code too.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+    az_iot_hub_client const* client,
+    az_span mqtt_user_name,
+    az_span* out_mqtt_user_name)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_user_name, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_user_name);
+
+  const az_span* module_id = &(client->_internal.options.module_id);
+  const az_span* user_agent = &(client->_internal.options.user_agent);
```
ericwol-msft: const az_span * const

danewalton: will update in telemetry too in another PR. along with the suggestion by Himanshu with `assert_int_equal`

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+      < (int32_t)(
+            az_span_length(client->_internal.iot_hub_hostname) + sizeof(hub_client_forward_slash)
+            + az_span_length(client->_internal.device_id) + sizeof(hub_client_forward_slash)
+            + az_span_length(*module_id)
+            + (az_span_length(*module_id) > 0 ? sizeof(hub_client_forward_slash) : 0)
+            + az_span_length(*user_agent)
+            + (az_span_length(*user_agent) > 0 ? sizeof(hub_client_param_separator) : 0)
+            + az_span_length(hub_client_api_version) + sizeof(hub_client_null_terminate)))
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY;
```
ericwol-msft: why do we need to do this? Why can az_span_append fail with this error code? I think this was something that Jeff talked about

danewalton: The behavior should be the same without it as discussed at scrum. The overhead is going through memcpy's even though the end result could be a failure.

danewalton: So in the case that we keep this, the following calls would be converted to:
```
assert(az_span_append(....));
```
to eliminate the additional check. Otherwise, if we don't check for the correct size, we would keep the check on each return value with `AZ_RETURN_IF_FAILED()`

sdk/iot/core/README.md
```diff
@@ -12,11 +12,54 @@ TBD
...
+az_iot_hub_client my_client;
+static az_span my_iothub_hostname = AZ_SPAN_LITERAL_FROM_STR("constoso.azure-devices.net");
+static az_span my_device_id = AZ_SPAN_LITERAL_FROM_STR("contoso_device");
+
+//Make sure to size the buffer to fit the user name (100 is an example)
+static uint8_t my_mqtt_user_name_buffer[100];
+static az_span my_mqtt_user_name = AZ_SPAN_LITERAL_FROM_BUFFER(my_mqtt_user_name_buffer);
+
+//Make sure to size the buffer to fit the client id (16 is an example)
+static uint8_t my_mqtt_client_id_buffer[16];
```
TiejunMS: Assume user has no knowledge about how `user_name` and `client_id` is built, shall we tell users how to find out the proper size for these buffers in comments?

danewalton: Yes I was thinking exactly that. I added it above the functions in the `.c` and the header file for doxygen. I can add it here too though.

sdk/iot/core/README.md
```diff
@@ -28,7 +71,8 @@ void my_telemetry_func()
 
   //Allocate a span to put the telemetry topic
   uint8_t telemetry_topic_buffer[64];
-  az_span topic_span = AZ_SPAN_FROM_BUFFER(telemetry_topic_buffer);
+  az_span topic_span = az_span_init(telemetry_topic_buffer, 0,
```
TiejunMS: It looks like the same before/after this change. Why don't we use `AZ_SPAN_FROM_BUFFER` directly?

danewalton: On Linux it would be fine but it won't compile on Windows MSVC. [Example here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=291075&view=logs&j=e05b29e1-8ab3-5e14-836c-84b55b312f66&t=879b0a1c-774d-5a15-eb9c-e1b448929f69&l=83) at the gate where it fails.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+#include <az_result.h>
 #include <az_span.h>
 
 #include <_az_cfg.h>
 
+static const uint8_t hub_client_forward_slash = '/';
+static const uint8_t hub_client_param_separator = '&';
+static const uint8_t hub_client_null_terminate = '\0';
+
+static const az_span hub_client_api_version = AZ_SPAN_LITERAL_FROM_STR("?api-version=2018-06-30");
```
TiejunMS: Is there a convenient way to upgrade API version? Once the library is built, user application is not able to change it. For example, PnP is not supported by this version. But if user application decides to implement some features for PnP itself, `az_iot_hub_client_user_name_get` can not be used.

danewalton: We have had discussions about moving this and all versions for the sdk to a header file but it sounds here like you would want to be able to change it in other ways?

TiejunMS: I'm not sure if it is a reasonable request. Since the api-version is not a constant value, it would be flexible if user application is able to change it at run time, or maybe at compiler time.

sdk/iot/core/README.md
danewalton: added here in the docs too

sdk/iot/core/src/az_iot_hub_client.c
danewalton: I can get further information about this but I believe the API version is essentially established at compile time since the functionality which the SDK is capable of won't change at runtime.
We didn't have this capability in the current SDK but maybe it's something we can think about. I'll leave that to @ericwol-msft and @jspaith maybe to chime in on.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+    az_iot_hub_client_options const* options)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(iot_hub_hostname, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(device_id, 1, false);
+  AZ_PRECONDITION_NOT_NULL(options);
+
+  client->_internal.iot_hub_hostname = iot_hub_hostname;
+  client->_internal.device_id = device_id;
+  memcpy(&client->_internal.options, options, sizeof(az_iot_hub_client_options));
```
TiejunMS: If we are not going to add more members to `az_iot_hub_client_options`, we can expand `memcpy` to two assignments for both code size and performance.

danewalton: This is not guaranteed to stay at two options but for now that's what we have. I can revert back for now if assignments are better than `memcpy`

sdk/iot/core/src/az_iot_hub_client.c
ericwol-msft: I think we should allow the user to modify the version. The reason is that the user is now more in control of talking to the service. This will allow them to implement new service features without SDK support.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,137 @@
...
+//[Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
+//[Format with module id] {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30
+AZ_NODISCARD az_result az_iot_hub_client_user_name_get(
+    az_iot_hub_client const* client,
+    az_span mqtt_user_name,
+    az_span* out_mqtt_user_name)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_user_name, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_user_name);
```
hihigupt: why are we are not allowing NULL out? If user does not want out span he should be allowed, atleast that is how sdk-core span related apis are designed. Am I missing anything in this API?

hihigupt: same comment for all API impl

---
In reply to: [392637338](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r392637338) [](ancestors = 392637338)

danewalton: I'm not sure which API's use that dynamic but ones like this
https://github.com/Azure/azure-sdk-for-c/blob/34c8b78dd70d07c93e8fb63b3e865351da223a5e/sdk/core/core/src/az_span.c#L160-L165
or
https://github.com/Azure/azure-sdk-for-c/blob/34c8b78dd70d07c93e8fb63b3e865351da223a5e/sdk/core/core/src/az_span.c#L223-L225
require the `out_xxx` param to not be `NULL`. If you see some that don't require that maybe we can talk to the Azure SDK for C team to rectify inconsistencies.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,138 @@
...
+
+  if (az_span_length(*module_id) > 0)
+  {
+    AZ_RETURN_IF_FAILED(
+        az_span_append_uint8(*out_mqtt_client_id, hub_client_forward_slash, out_mqtt_client_id));
+    AZ_RETURN_IF_FAILED(az_span_append(*out_mqtt_client_id, *module_id, out_mqtt_client_id));
+  }
+
+  AZ_RETURN_IF_FAILED(
+      az_span_append_uint8(*out_mqtt_client_id, hub_client_null_terminate, out_mqtt_client_id));
```
hihigupt: Can we only leak out variable only when we are almost done with function? Something like this?

AZ_RETURN_IF_FAILED(
      az_span_append(mqtt_client_id, client->_internal.device_id, mqtt_client_id));

  if (az_span_length(*module_id) > 0)
  {
    AZ_RETURN_IF_FAILED(
        az_span_append_uint8(mqtt_client_id, hub_client_forward_slash, &mqtt_client_id));
    AZ_RETURN_IF_FAILED(az_span_append(mqtt_client_id *module_id, &mqtt_client_id));
  }

  AZ_RETURN_IF_FAILED(
      az_span_append_uint8(mqtt_client_id hub_client_null_terminate, &mqtt_client_id));

if (out_mqtt_client_id != NULL)
{
   memcpy((void*)out_mqtt_client_id, &mqtt_client_id, sizeof(az_span));
}

sdk/iot/core/src/az_iot_hub_client.c
danewalton: talked with Jeffrey and the output span should be valid. so we will check for `NULL`.

sdk/iot/core/src/az_iot_hub_client.c
danewalton: pushed a change to update the output span only on success. however the output span will have a precondition to not be `NULL`

sdk/iot/core/src/az_iot_hub_client.c
danewalton: Noted for the future but for now will resolve

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -69,6 +69,10 @@ AZ_NODISCARD az_result az_iot_hub_client_init(
 
 /**
  * @brief Gets the MQTT user name.
+ * 
+ * The user name will be of the following format:
+ * [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30
```
jspaith: >2018-06-30 [](start = 72, length = 10)

I don't think we should hard-code this in comment.  We want to minimize of places we change this when we rev.

Also -- this is pretty old, is it really latest from hub that's universally deployed?

jspaith: >-version=2018-06-30 [](start = 63, length = 19)

If user specifies user_agent in the options, isn't this appended here too?

sdk/iot/core/src/az_iot_hub_client.c
jspaith: Do we need this comment here in 1st place?  Isn't the comment in the .h for this function enough?

sdk/iot/core/inc/az_iot_hub_client.h
danewalton: This is straight from the documentation online [here](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-device). I think it might be useful from the user perspective so that they know what to expect. If you think I should put it in one place only instead of in multiple places across the code base I'm open to that. (So would remove from the readme and `.c` file)

danewalton: Yes the user params get appended after the api-version. I can put a note after so that user params are accounted for?
```suggestion
 * [Format without module id] {iothubhostname}/{device_id}/?api-version=2018-06-30&{user_params}

```

jspaith: That user documentation isn't great as it shouldn't do this either, because it locks customers onto a specific (very out of date) version.  This is really a variable just like {device_id}, except it comes from us.  I'm not sure how we define globals but I do think it should come from a .h, or historically it would at least.

---
In reply to: [393369267](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r393369267) [](ancestors = 393369267)

jspaith: +1

---
In reply to: [393369684](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r393369684) [](ancestors = 393369684)

danewalton: Ah I see your point. I should have linked this my bad.
https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r392545310
This has been noted and will be updated once we have a clear understanding about how versions/configs will be done. This API version will likely be a config in a header. For now Eric has said we will circle back and update. I do have a list of things to track for improvements. Let me know if that's alright with you.

jspaith: Yeah provided you're tracking an improvement, I don't consider my feedback to be a checkin blocker.

danewalton: updated

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
```diff
@@ -4,15 +4,29 @@
...
 #include <stddef.h>
 #include <stdint.h>
+#include <stdio.h>
 
 #include <cmocka.h>
 
+#define TEST_DEVICE_ID_STR "my_device"
+#define TEST_DEVICE_HOSTNAME_STR "myiothub.azure-devices.net"
+#define TEST_MODULE_ID "my_module_id"
+#define TEST_USER_AGENT "os=azrtos"
```
hihigupt: suggestion: convert all #define string to const char*

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
```diff
@@ -22,11 +36,267 @@ void test_az_iot_hub_client_get_default_options_succeed(void** state)
...
+{
+  (void)state;
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_int_equal(
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options), AZ_OK);
+
+  assert_string_equal(TEST_DEVICE_ID_STR, az_span_ptr(client._internal.device_id));
+  assert_string_equal(TEST_DEVICE_HOSTNAME_STR, az_span_ptr(client._internal.iot_hub_hostname));
```
hihigupt: You missed length check

hihigupt: same for all tests

---
In reply to: [393964290](https://github.com/Azure/azure-sdk-for-c/pull/455#discussion_r393964290) [](ancestors = 393964290)

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -1,12 +1,111 @@
...
+    AZ_RETURN_IF_FAILED(az_span_append(mqtt_user_name, *user_agent, &mqtt_user_name));
+  }
+
+  *out_mqtt_user_name = mqtt_user_name;
+
+  return AZ_OK;
+}
+
+//[Format without module id] {device_id}
+//[Format with module id] {device_id}/{module_id}
```
hihigupt: sorry, but lets remove this as header file should be good enough.

danewalton: woops yea removed the other one. will remove

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
```diff
@@ -22,11 +36,267 @@ void test_az_iot_hub_client_get_default_options_succeed(void** state)
...
+{
+  (void)state;
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_int_equal(
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options), AZ_OK);
+
+  uint8_t test_span_buffer[sizeof(test_correct_user_name)] = { 0 };
+  az_span test_span = az_span_init(test_span_buffer, 0, _az_COUNTOF(test_span_buffer));
```
hihigupt: >_az_COUNTOF [](start = 56, length = 11)

lets use either sizeof or _az_COUNTOF

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
danewalton: I had made a previous comment here
https://github.com/Azure/azure-sdk-for-c/pull/471#discussion_r393354970
detailing why I haven't. If you think it might then I can. But in this case I would be testing the implementation of `az_span` not hub client.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
```diff
@@ -4,12 +4,11 @@
 #include <az_iot_sas_token.h>
 #include <az_span.h>
 
-#include <stdio.h>
-
 #include <setjmp.h>
 #include <stdarg.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <stdio.h>
```
hihigupt: Which API in your test is using stdio?

sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c
danewalton: resolving in another PR

sdk/iot/core/src/az_iot_hub_client.c
CIPop: Please follow the guidelines as other SDKs may still need to be updated to follow the latest docs.
https://azure.github.io/azure-sdk/clang_design.html#functions-with-many-parameters

CIPop: NULL should work: the implementation will assume (call) defaults.

PR #457, opened 2020-03-13
Add _az_countof() macro

danewalton:
Add a macro to determine the number of elements in a buffer. Added in the recommended spot by @JeffreyRichter.

### DIFF
--- a/sdk/core/core/inc/_az_cfg.h
+++ b/sdk/core/core/inc/_az_cfg.h
@@ -83,4 +83,7 @@
 #define AZ_NODISCARD
 #endif // _MSC_VER
 
+// Get the number of elements in an array
+#define _az_countof(array) (sizeof(array) / sizeof(array[0]))
+
 #endif // _az_CFG_H
--- a/sdk/core/core/inc/_az_cfg.h
+++ b/sdk/core/core/inc/_az_cfg.h
@@ -84,6 +84,6 @@
 #endif // _MSC_VER
 
 // Get the number of elements in an array
-#define _az_countof(array) (sizeof(array) / sizeof(array[0]))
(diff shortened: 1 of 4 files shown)


sdk/core/core/inc/_az_cfg.h
```diff
@@ -83,4 +83,7 @@
 #define AZ_NODISCARD
 #endif // _MSC_VER
 
+// Get the number of elements in an array
+#define _az_countof(array) (sizeof(array) / sizeof(array[0]))
```
antkmsft: PLease rename the macro to `_az_COUNTOF`. Macros are uppercase, except for the _`az_` prefix which we can't do.

sdk/core/core/test/cmocka/test_log.c
```diff
@@ -189,7 +189,7 @@ void test_az_log(void** state)
     // our code attempts to log a classification that's not in the customer's allow list.
     az_log_classification const classifications[] = { AZ_LOG_HTTP_REQUEST };
     az_log_set_classifications(
-        classifications, sizeof(classifications) / sizeof(classifications[0]));
+        classifications, _az_countof(classifications));
```
antkmsft: If you want, you can ignore updating this one place. I have a PR in progress that replaces that with a single array without size, but with a terminating element.

sdk/core/core/inc/_az_cfg.h
danewalton: done

sdk/core/core/test/cmocka/test_log.c
danewalton: I did a find all replace all so no big deal. I can remove if it makes your merge easier but otherwise I can leave it.

PR #458, opened 2020-03-13
Reduce assembly size for az_ascii_lower

ahsonkhan:
Given code size is a priority here, we save 2 instructions/16 bytes along with some run-time performance improvements.

Also, improving test coverage.


sdk/core/core/test/cmocka/test_span.c
```diff
@@ -20,6 +21,52 @@
   assert_true(az_span_append_uint8(buffer, 'a', NULL) == AZ_ERROR_ARG);
 } */
 
+void az_single_char_ascii_lower_test()
+{
+  for (uint8_t i = 0; i <= SCHAR_MAX; i++)
```
antkmsft: nit: prefix ++

sdk/core/core/src/az_span.c
```diff
@@ -46,7 +46,13 @@ AZ_NODISCARD az_span az_span_slice(az_span span, int32_t low_index, int32_t high
 
 AZ_NODISCARD AZ_INLINE uint8_t az_ascii_lower(uint8_t value)
 {
-  return 'A' <= value && value <= 'Z' ? value + AZ_ASCII_LOWER_DIF : value;
+  // This is equivalent to the following but with fewer conditions.
+  // return 'A' <= value && value <= 'Z' ? value + AZ_ASCII_LOWER_DIF : value;
+  if ((uint8_t)(int8_t)(value - 'A') <= ('Z' - 'A'))
+  {
+    value += AZ_ASCII_LOWER_DIF;
```
antkmsft: We are probably not consistent on this yet, but I freaked out a bit when I saw the name `AZ_ASCII_` as if it is a part of our public interface. It is not, which is good news - it is defined right in this .c file and is only visible there, but can you please rename it to `_az_ASCII_LOWER_DIF`?

sdk/core/core/src/az_span.c
```diff
@@ -46,7 +46,13 @@ AZ_NODISCARD az_span az_span_slice(az_span span, int32_t low_index, int32_t high
 
 AZ_NODISCARD AZ_INLINE uint8_t az_ascii_lower(uint8_t value)
```
antkmsft: It looks like the function is internal, local to this file. I think we should name it as `_az_tolower()` i.e. starting with underscore.But without the underscore after `to`: we do `to_something` in other places, but here we would make a reference to the stdlib function `tolower()` which does the similar job. And given that we are referencing `tolower()`, I don;t think `_ascii_` is needed in the name.

ahsonkhan:
> Did you use https://godbolt.org/ ? :)

Yep :)

sdk/core/core/test/cmocka/test_span.c
ahsonkhan: I will have to adjust to this, and unlearn this habit :p

PR #462, opened 2020-03-13
Unit test rework

danewalton:
@barcharcraz 
@vhvb1989 
As discussed in today's meeting, we need to have a unified way of structuring unit tests. I have here a proposed solution which could work and I would like your guys' opinion.
Note: `cmocka_run_group_tests_name()` returns the number of failed tests so the final returned result via the `+=` will be the total number of failed tests.

This is the output with `ctest --verbose`
MicrosoftTeams-image

And this is the output with `ctest -N`
MicrosoftTeams-image (1)


sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
```diff
@@ -249,10 +244,31 @@ void az_iot_sas_token_generate_with_keyname_succeeds(void** state)
...
   assert_true(az_succeeded(az_iot_sas_token_generate(
       iothub_fqdn, device_id, signature, expiry_time_secs, key_name, sas_token, &sas_token)));
   assert_true(strncmp(expected_sas_token, (char*)raw_sas_token, az_span_length(sas_token)) == 0);
 }
+
+int test_iot_sas_token()
+{
+  const struct CMUnitTest tests[] = {
+    // SAS Token
+    /*cmocka_unit_test(az_iot_sas_token_get_document_NULL_document_fails),
```
ahsonkhan: Why do we have commented out test code? Should we just remove these?

danewalton: No because we would like to have these tests but must remove them for now due to precondition changes. See #467 for tracking of it.

PR #464, opened 2020-03-16
Fix build error when compiling with NO_PRECONDITION_CHECKING

antkmsft:


ahsonkhan:
@antkmsft - how did you turn off precondition? Did you define it via the command line?

Would this work?
```text
cmake -DNO_PRECONDITION_CHECKING=ON ..
```

antkmsft:
@ahsonkhan , I think just `-DNO_PRECONDITION_CHECKING` would've worked, myt in my case I just searched the sources for `NO_PRECONDITION_CHECKING`, there was 1 occurrence, I inverthed the `ifdef` condition and tried to compile.

ahsonkhan:
>  I think just `-DNO_PRECONDITION_CHECKING` would've worked

That results in an error:
```text
> cmake -DNO_PRECONDITION_CHECKING ..

Parse error in command line argument: -DNO_PRECONDITION_CHECKING
Should be: VAR:type=value
CMake Error: No cmake script provided.
CMake Error: Problem processing arguments. Aborting.
```

>  I just searched the sources for `NO_PRECONDITION_CHECKING`, there was 1 occurrence, I inverted the `ifdef` condition and tried to compile.

I will do that for now. Thanks.

PR #471, opened 2020-03-16
remove null terminator guarantee from telemetry

danewalton:
After talking with Himanshu we can reach a compromise with the null terminator discussion. In this way, users may get a topic with a null terminator if they would like. Otherwise, they can get one with only the size necessary.

### DIFF
--- a/sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
+++ b/sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
@@ -121,7 +121,7 @@ void test_az_iot_hub_client_telemetry_publish_topic_get_no_options_no_params_suc
 {
   (void)state;
 
-  uint8_t mqtt_topic_buf[sizeof(g_test_correct_topic_no_options_no_params)];
+  uint8_t mqtt_topic_buf[sizeof(g_test_correct_topic_no_options_no_params) - 1];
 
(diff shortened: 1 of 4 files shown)


sdk/iot/core/tests/cmocka/az_iot_telemetry_tests.c
```diff
@@ -121,55 +121,82 @@ void test_az_iot_hub_client_telemetry_publish_topic_get_no_options_no_params_suc
...
   az_span mqtt_topic = az_span_init(mqtt_topic_buf, 0, _az_COUNTOF(mqtt_topic_buf));
 
   assert_true(
       az_iot_hub_client_telemetry_publish_topic_get(
           &g_test_valid_client_no_options, NULL, mqtt_topic, &mqtt_topic)
       == AZ_OK);
+  assert_memory_equal(
+      g_test_correct_topic_no_options_no_params,
+      (char*)az_span_ptr(mqtt_topic),
+      sizeof(g_test_correct_topic_no_options_no_params) - 1);
```
hihigupt: span produces 2 things :
1. ptr having data
2. length

can we explicitly check both of them with expected value?
here we are only checking data, what about length

danewalton: So I can but I added that test to the one where the buffer/span is not the same size as the resulting topic. If the size of the buffer is the same size as requested topic, the only way for it to not be the same is if the az_span API's fail no?

PR #474, opened 2020-03-16
Remove the request id policy

gilbertw:
The policy can be added once pal uuid support is provided


sdk/core/core/src/az_http_policy.c
```diff
@@ -49,11 +49,11 @@ AZ_NODISCARD az_result az_http_pipeline_policy_uniquerequestid(
   (void)p_options;
 
   // TODO - add a UUID create implementation
-  az_span const uniqueid = AZ_SPAN_LITERAL_FROM_STR("123e4567-e89b-12d3-a456-426655440000");
+  //az_span const uniqueid = AZ_SPAN_LITERAL_FROM_STR("123e4567-e89b-12d3-a456-426655440000");
```
ahsonkhan: nit: Can we remove the code and file an issue linking back to this PR for the TODO rather than commenting it out? We always have code commit history, but haven't commented out code makes it seem like someone accidentally left "dead" code around. Alternatively, explicitly state why this code is commented out, so it doesn't look like "real" C code.

sdk/core/core/src/az_http_policy.c
```diff
@@ -49,11 +49,11 @@ AZ_NODISCARD az_result az_http_pipeline_policy_uniquerequestid(
...
 
   // Append the Unique GUID into the headers
   //  x-ms-client-request-id
-  AZ_RETURN_IF_FAILED(az_http_request_append_header(p_request, AZ_MS_CLIENT_REQUESTID, uniqueid));
+  //AZ_RETURN_IF_FAILED(az_http_request_append_header(p_request, AZ_MS_CLIENT_REQUESTID, uniqueid));
```
ahsonkhan: CI failing with `error: ‘AZ_MS_CLIENT_REQUESTID’ defined but not used`, remove it?
```text
/home/vsts/work/1/s/sdk/core/core/src/az_http_policy.c:12:22: error: ‘AZ_MS_CLIENT_REQUESTID’ defined but not used [-Werror=unused-const-variable=]
 static const az_span AZ_MS_CLIENT_REQUESTID = AZ_SPAN_LITERAL_FROM_STR("x-ms-client-request-id");

```

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
```diff
@@ -77,12 +77,12 @@ AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
                 .p_options= &self->_internal.options._internal.api_version,
               },
             },
-            {
+            /*{
```
ahsonkhan: Same **nit** here. Remove it rather than commenting it out.

PR #477, opened 2020-03-17
IoT Improvements

danewalton:
Some updates as rec'd by @hihigupt and I think @ericwol-msft .
- Uses `assert_int_equal()` for returned values so on failure will show what the two compared values were
- Uses `assert_memory_equal()` to compare non null terminated strings to output compared values on failure.
- Add `const az_span* const` where appropriate.

### DIFF
--- a/sdk/iot/core/src/az_iot_telemetry.c
+++ b/sdk/iot/core/src/az_iot_telemetry.c
@@ -27,8 +27,8 @@ AZ_NODISCARD az_result az_iot_hub_client_telemetry_publish_topic_get(
   AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
   AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
 
-  const az_span* user_agent = &(client->_internal.options.user_agent);
-  const az_span* module_id = &(client->_internal.options.module_id);
+  const az_span* const user_agent = &(client->_internal.options.user_agent);
+  const az_span* const module_id = &(client->_internal.options.module_id);
 
   // Required topic parts
   int32_t required_size = az_span_length(telemetry_topic_prefix)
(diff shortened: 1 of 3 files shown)


ahsonkhan:
> - Uses `assert_int_equal()` for returned values so on failure will show what the two compared values were

Nice. I noticed this recently myself and the output in the failure case is way more useful :)

sdk/iot/core/src/az_iot_telemetry.c
```diff
@@ -27,8 +27,8 @@ AZ_NODISCARD az_result az_iot_hub_client_telemetry_publish_topic_get(
   AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
   AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
 
-  const az_span* user_agent = &(client->_internal.options.user_agent);
-  const az_span* module_id = &(client->_internal.options.module_id);
+  const az_span* const user_agent = &(client->_internal.options.user_agent);
+  const az_span* const module_id = &(client->_internal.options.module_id);
```
ahsonkhan: This is probably a naive question, but how did you know where to apply the `const` to?

danewalton: No questions are too naive :)

Here is essentially how it breaks down. I'll use char* as an example:
```C
char* const x
```
is a constant pointer to a char. So you can change the char at `*x` but not the value of `x`.

```C
const char* x
```
is a pointer to a constant char. So you can change the pointer but not the content at `*x`.

```C
const char* const x
```
is a constant pointer to a constant char. So you can't change the value of the pointer nor the content at `*x`.

[This website](https://cdecl.org) is really great to understand pretty much any declaration in C. I included links to my examples.

https://cdecl.org/?q=const+char*+const+x
https://cdecl.org/?q=char*+const+x
https://cdecl.org/?q=const+char*+x

PR #482, opened 2020-03-17
Victor/core/json

vhvb1989:


sdk/core/core/inc/az_json.h
```diff
@@ -128,71 +180,183 @@ typedef struct
...
+ * AZ_ERROR_BUFFER_OVERFLOW if the buffer is too small.
+ */
 AZ_NODISCARD az_result
-az_json_builder_append_object(az_json_builder* self, az_span name, az_json_token token);
-
-AZ_NODISCARD az_result az_json_builder_append_object_close(az_json_builder* self);
+az_json_builder_append_token(az_json_builder* json_builder, az_json_token token);
 
+/*
+ * @brief az_json_builder_append_object appends a JSON to the JSON buffer.
```
ahsonkhan:
```suggestion
 * @brief az_json_builder_append_object appends a JSON object with it's name to the buffer.
```

sdk/core/core/inc/az_json.h
```diff
@@ -128,71 +180,183 @@ typedef struct
...
+ * @param json_builder A pointer to an az_json_builder instance containing the buffer to append the
+ * object to.
+ * @param name A span containing the JSON object's name.
+ * @param token A pointer to the az_json_token to append as the object's value.
+ * @return AZ_OK if the token was appended successfully.<br>
+ * AZ_ERROR_BUFFER_OVERFLOW if the buffer is too small.
+ */
 AZ_NODISCARD az_result
-az_json_builder_append_array_item(az_json_builder* self, az_json_token token);
+az_json_builder_append_object(az_json_builder* json_builder, az_span name, az_json_token token);
```
ahsonkhan: What happens if someone calls this method when inside a JSON array (which could result in invalid JSON)?
```json
[ 1, 2, "foo":{}, 4]
```

sdk/core/core/inc/az_json.h
```diff
@@ -128,71 +180,183 @@ typedef struct
...
+AZ_NODISCARD az_result
+az_json_builder_append_array_item(az_json_builder* json_builder, az_json_token token);
 
 /************************************ JSON PARSER ******************/
 
-typedef uint64_t az_json_stack;
+typedef uint64_t _az_json_stack;
+
+/*
+ * @brief An az_json_parser return the JSON tokens contained within a JSON object buffer.
```
ahsonkhan:
```suggestion
 * @brief An az_json_parser returns the JSON tokens contained within a JSON buffer.
```

sdk/core/core/inc/az_json.h
```diff
@@ -128,71 +180,183 @@ typedef struct
...
   struct
   {
     az_span reader;
-    az_json_stack stack;
+    _az_json_stack stack;
   } _internal;
 } az_json_parser;
 
+/*
+ * @brief An az_json_token_member represents a JSON element's name and token.
```
ahsonkhan:
```suggestion
 * @brief An az_json_token_member represents a JSON element's name and value.
```

sdk/core/core/inc/az_json.h
```diff
@@ -128,71 +180,183 @@ typedef struct
...
 {
   az_span name;
   az_json_token token;
 } az_json_token_member;
 
-AZ_NODISCARD az_result az_json_parser_init(az_json_parser* self, az_span json_buffer);
-
-AZ_NODISCARD az_result az_json_parser_parse_token(az_json_parser* self, az_json_token* out_token);
+/*
+ * @brief az_json_parser_init initializes an az_json_parser to parse the JSON object contained in a
```
ahsonkhan:
```suggestion
 * @brief az_json_parser_init initializes an az_json_parser to parse the JSON payload contained within the passed in buffer.
```

sdk/core/core/src/az_json_pointer.c
```diff
@@ -67,11 +67,14 @@ AZ_NODISCARD static az_result az_span_reader_read_json_pointer_char(az_span* sel
   }
 }
 
+/**
+ * Returns a next reference token in the JSON pointer. The JSON pointer parser is @var
```
ahsonkhan: What do you mean by `The JSON pointer parser is @var az_span_reader.`?
Also, what's the purpose of `@var`?

sdk/core/core/src/az_json_string_private.h
```diff
@@ -36,6 +37,14 @@ AZ_NODISCARD az_result _az_span_reader_read_json_pointer_token(az_span* self, az
  */
 AZ_NODISCARD az_result _az_span_reader_read_json_pointer_token_char(az_span* self, uint32_t* out);
 
+AZ_NODISCARD AZ_INLINE az_json_token az_json_token_span(az_span span)
```
ahsonkhan: Since this is private, start with `_`.
```suggestion
AZ_NODISCARD AZ_INLINE az_json_token _az_json_token_span(az_span span)
```

Also, where is this method called/used?

sdk/core/core/inc/az_json.h
```diff
@@ -46,79 +54,123 @@ typedef struct
...
 
-/**
- * Copies a boolean value to @var out from the given JSON value.
+/*
+ * @brief az_json_token_get_boolean returns the JSON token's boolean.
  *
- * If the JSON value is not a boolean then the function returns an error.
+ * @param token A pointer to an az_json_token instance.
+ * @param out_value A pointer to a variable to receive the value.
+ * @return AZ_OK if the boolean is returned.<br>
```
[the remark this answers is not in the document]
ahsonkhan: Looks like we do need the explicit `<br>` tag for new line. And the `#` makes it a link to the enum value, so let's keep that:
image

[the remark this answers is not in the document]
JeffreyRichter: OK

sdk/core/core/inc/az_json.h
JeffreyRichter: Well, it could be an array of arrays in which case it would be valid JSON.
We can't protect against this. Someone can always just write a string with invalid JSON and not use our APIs at all.

sdk/core/core/inc/az_json.h
```diff
@@ -23,21 +24,28 @@
...
+  AZ_JSON_TOKEN_BOOLEAN,
+  AZ_JSON_TOKEN_NUMBER,
+  AZ_JSON_TOKEN_STRING,
+  AZ_JSON_TOKEN_OBJECT_START,
+  AZ_JSON_TOKEN_OBJECT_END,
+  AZ_JSON_TOKEN_ARRAY_START,
+  AZ_JSON_TOKEN_ARRAY_END,
+  // AZ_JSON_TOKEN_SPAN represents a token consisting of nested JSON; the JSON parser never returns
+  // a token of this type.
+  AZ_JSON_TOKEN_SPAN,
```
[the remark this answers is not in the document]
JeffreyRichter: The explicit #s served no value. It implied some importance which does not exist.
It doesn't matter if we add new symbols; we are source code compatible (not binary) and so ALL the code must be recompiled.

sdk/core/core/src/az_json_pointer.c
JeffreyRichter: This text is just bad and should be fixed; it refers to an az_span_reader which we don't have any more - this is not public docs.

sdk/core/core/src/az_json_string_private.h
JeffreyRichter: @vhvb1989 should look at this

sdk/core/core/inc/az_json.h
ahsonkhan: Of course, but since I am using a JSON builder, shouldn't that help validate such things? Otherwise, we are reducing part of the benefit of such an API, no?

JeffreyRichter: How do we validate if an array has array elements to allow or disallow this?

ahsonkhan: Based on the current state of the builder. If the current state/token of the builder is "writing inside an array", then any calls to `az_json_builder_append_object` should fail with an error since writing a property name within an array is invalid JSON.

Similarly, if the current state is "writing inside an object", calling anything that doesn't write a a propertyname first show fail.

> How do we validate if an array has array elements to allow or disallow this?

I don't see how arrays of arrays is related here. Also I don't see how whether an array has a single element, or is empty, or has array elements in it, is related to validation. You only need to know about the state of the inner most container (if it is array or object).

To do this, you would need some sort of stack as part of the `az_json_builder` state.

**So, here are a couple of invalid cases:**
- writing a single JSON value inside a JSON object without accompanying key name
```json
{"just a value with no key"}
```
- writing a key-value pair inside a JSON array instead of just a value
```json
["key":"value ]
```

sdk/core/core/src/az_json_string_private.h
vhvb1989: to be removed. No need to change here

sdk/core/core/inc/az_json.h
```diff
@@ -46,79 +54,123 @@ typedef struct
...
   return (az_json_token){
     .kind = AZ_JSON_TOKEN_STRING,
-    .value.string = value,
+    ._internal.string = value,
   };
 }
 
-AZ_NODISCARD AZ_INLINE az_json_token az_json_token_span(az_span span)
+/*
+ * @brief az_json_token_object_start returns a az_json_token representing the start of an object.
```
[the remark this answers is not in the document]
vhvb1989: fixed

sdk/core/core/inc/az_json.h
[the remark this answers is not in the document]
vhvb1989: the link should be created without requiring the `#`  ..  Doxygen makes automatically a link of any known symbol

sdk/core/core/inc/az_json.h
vhvb1989: @ahsonkhan , calling `az_json_builder_append_object()` inside an array would add brackets first and would end up as
```json
{ "array": [1, 2, { "obj": value}, 3] }
``` 
I will add an extra test for this so you can see it.
There is no a way of adding `:` directly with our parser.

The only way we can end up with a bad Json is by using append_array_item() without appending an array_start. This would generate some invalid thing like
```json
{ array_element }
```
Or calling append_token directly without making sure object or array was started.

However, if we add validation for this, we would be adding more code and size to something that looks like as a nice-to-have thing

sdk/core/core/inc/az_json.h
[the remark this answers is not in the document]
ahsonkhan: I checked. Without the `#`, the docs don't show AZ_OK as a link but rather just plain text.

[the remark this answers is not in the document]
antkmsft: I think we should be using `@retval` that is designed for this: http://www.doxygen.nl/manual/commands.html#cmdretval

sdk/core/core/inc/az_json.h
ahsonkhan: > calling `az_json_builder_append_object()` inside an array would add brackets first

I might be reading the code wrong, but that doesn't look like its true.

This test is passing and producing invalid JSON.

```c
uint8_t array1[200];
az_json_builder builder1 = { 0 };

TEST_EXPECT_SUCCESS(az_json_builder_init(&builder1, AZ_SPAN_FROM_BUFFER(array1)));

TEST_EXPECT_SUCCESS(az_json_builder_append_token(&builder1, az_json_token_array()));

TEST_EXPECT_SUCCESS(az_json_builder_append_object(
    &builder1, AZ_SPAN_FROM_STR("int-max"), az_json_token_number(9007199254740991ull)));

TEST_EXPECT_SUCCESS(az_json_builder_append_array_close(&builder1));

    builder1._internal.json,
    AZ_SPAN_FROM_STR( //
        "["
        "\"int-max\":9007199254740991"
        "]")));
```

PR #483, opened 2020-03-17
Add property init and append

danewalton:
Adds the following API's:
- `az_iot_hub_client_properties_init()`
- `az_iot_hub_client_properties_append()`

### DIFF
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -55,6 +55,43 @@ int main()
 }
 ```
 
+### Properties
+
+Included in the IoT SDK are helper functions to form and manage properties for IoT Hub services. Functionality starts by using the `az_iot_hub_client_properties_init()` API. The user is free to intitialize using an empty but appropriately sized span to later append properties or an already populated span containing a properly formated property buffer. "Properly formatted" properties follow the form `{key}={value}&{key}={value}`.
+
+Below is an example use case of appending properties.
+
+```C
+//FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
+void my_property_func()
+{
+  //Allocate a span to put the properties
+  uint8_t property_buffer[64];
+  az_span property_span = az_span_init(property_buffer, 0, sizeof(property_buffer));
+  
+  //Initialize the property struct with the span
+  az_iot_hub_client_properties props;
+  az_iot_hub_client_properties_init(&props, property_span);
+  //Append properties
+  az_iot_hub_client_properties_append(&props, AZ_SPAN_FROM_STR("key"), AZ_SPAN_FROM_STR("value"));
+  //At this point, you are free to pass the `props` to other API's with property parameters.
+}
+```
+
+Below is an example use case of initializing an already populated property span.
+
+```C
+//FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
+static az_span my_prop_span = AZ_SPAN_LITERAL_FROM_STR("my_device=contoso&my_key=my_value");
+void my_property_func()
+{
+  //Initialize the property struct with the span
+  az_iot_hub_client_properties props;
+  az_iot_hub_client_properties_init(&props, my_prop_span);
+  //At this point, you are free to pass the `props` to other API's with property parameters.
+}
+```
+
 ### Telemetry
 
 Telemetry functionality can be achieved by sending a user payload to a specific topic. In order to get the appropriate topic to which to send, use the `az_iot_hub_client_telemetry_publish_topic_get()` API. An example use case is below.
--- a/sdk/iot/core/README.md
+++ b/sdk/iot/core/README.md
@@ -57,7 +57,7 @@ int main()
 
 ### Properties
 
-Included in the IoT SDK are helper functions to form and manage properties for IoT Hub services. Functionality starts by using the `az_iot_hub_client_properties_init()` API. The user is free to intitialize using an empty but appropriately sized span to later append properties or an already populated span containing a properly formated property buffer. "Properly formatted" properties follow the form `{key}={value}&{key}={value}`.
+Included in the IoT SDK are helper functions to form and manage properties for IoT Hub services. Implementation starts by using the `az_iot_hub_client_properties_init()` API. The user is free to intitialize using an empty, but appropriately sized, span to later append properties or an already populated span containing a properly formated property buffer. "Properly formatted" properties follow the form `{key}={value}&{key}={value}`.
 
 Below is an example use case of appending properties.
 
@@ -74,7 +74,7 @@ void my_property_func()
   az_iot_hub_client_properties_init(&props, property_span);
   //Append properties
   az_iot_hub_client_properties_append(&props, AZ_SPAN_FROM_STR("key"), AZ_SPAN_FROM_STR("value"));
-  //At this point, you are free to pass the `props` to other API's with property parameters.
+  //At this point, you are able to pass the `props` to other API's with property parameters.
 }
 ```
 
@@ -88,7 +88,7 @@ void my_property_func()
   //Initialize the property struct with the span
   az_iot_hub_client_properties props;
   az_iot_hub_client_properties_init(&props, my_prop_span);
-  //At this point, you are free to pass the `props` to other API's with property parameters.
+  //At this point, you are able to pass the `props` to other API's with property parameters.
 }
 ```
 
--- a/sdk/iot/core/inc/az_iot_hub_client.h
+++ b/sdk/iot/core/inc/az_iot_hub_client.h
@@ -176,8 +176,8 @@ typedef struct az_iot_hub_client_properties
  * @brief Initializes the Telemetry or C2D properties.
  *
  * @param[in] properties The #az_iot_hub_client_properties to initialize
- * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted
- *                   properties with the following format: {key}={value}&{key}={value}.
+ * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted,
+ *                   RFC 2396-encoded properties with the following format: {key}={value}&{key}={value}.
  * @return #az_result
  */
 AZ_NODISCARD az_result
--- a/sdk/iot/core/inc/az_iot_hub_client.h
+++ b/sdk/iot/core/inc/az_iot_hub_client.h
@@ -177,7 +177,7 @@ typedef struct az_iot_hub_client_properties
  *
  * @param[in] properties The #az_iot_hub_client_properties to initialize
  * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted
- *                   properties starting with the '?' character.
+ *                   properties with the following format: {key}={value}&{key}={value}.
  * @return #az_result
  */
 AZ_NODISCARD az_result
--- a/sdk/iot/core/src/az_iot_hub_client.c
+++ b/sdk/iot/core/src/az_iot_hub_client.c
@@ -127,8 +127,8 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_append(
     az_span value)
 {
   AZ_PRECONDITION_NOT_NULL(properties);
-  AZ_PRECONDITION_VALID_SPAN(name, 0, false);
-  AZ_PRECONDITION_VALID_SPAN(value, 0, false);
+  AZ_PRECONDITION_VALID_SPAN(name, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(value, 1, false);
 
   az_span prop_span = properties->_internal.properties;
 
--- a/sdk/iot/core/src/az_iot_hub_client.c
+++ b/sdk/iot/core/src/az_iot_hub_client.c
@@ -12,6 +12,7 @@
 
 static const uint8_t hub_client_forward_slash = '/';
 static const uint8_t hub_client_param_separator = '&';
+static const uint8_t hub_client_param_equals = '=';
 
 static const az_span hub_client_api_version = AZ_SPAN_LITERAL_FROM_STR("?api-version=2018-06-30");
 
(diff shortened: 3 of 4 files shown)


sdk/iot/core/README.md
```diff
@@ -55,6 +55,43 @@ int main()
 }
 ```
 
+### Properties
+
+Included in the IoT SDK are helper functions to form and manage properties for IoT Hub services. Functionality starts by using the `az_iot_hub_client_properties_init()` API. The user is free to intitialize using an empty but appropriately sized span to later append properties or an already populated span containing a properly formated property buffer. "Properly formatted" properties follow the form `{key}={value}&{key}={value}`.
```
ericwol-msft: Implementation starts by using the

ericwol-msft: The user is free to intitialize using an empty, but appropriately sized span, to later

sdk/iot/core/README.md
```diff
@@ -55,6 +55,43 @@ int main()
...
+  //Allocate a span to put the properties
+  uint8_t property_buffer[64];
+  az_span property_span = az_span_init(property_buffer, 0, sizeof(property_buffer));
+  
+  //Initialize the property struct with the span
+  az_iot_hub_client_properties props;
+  az_iot_hub_client_properties_init(&props, property_span);
+  //Append properties
+  az_iot_hub_client_properties_append(&props, AZ_SPAN_FROM_STR("key"), AZ_SPAN_FROM_STR("value"));
+  //At this point, you are free to pass the `props` to other API's with property parameters.
```
ericwol-msft: free -> able

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -107,3 +108,40 @@ AZ_NODISCARD az_result az_iot_hub_client_id_get(
...
 }
+
+AZ_NODISCARD az_result
+az_iot_hub_client_properties_init(az_iot_hub_client_properties* properties, az_span buffer)
+{
+  AZ_PRECONDITION_NOT_NULL(properties);
+  AZ_PRECONDITION_VALID_SPAN(buffer, 0, false);
+
+  properties->_internal.properties = buffer;
+  properties->_internal.current_property = az_span_ptr(buffer);
```
ericwol-msft: why do you do this as it's already in _internal.properties?

danewalton: setting the `current_property`?

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -107,3 +108,40 @@ AZ_NODISCARD az_result az_iot_hub_client_id_get(
...
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_iot_hub_client_properties_append(
+    az_iot_hub_client_properties* properties,
+    az_span name,
+    az_span value)
+{
+  AZ_PRECONDITION_NOT_NULL(properties);
+  AZ_PRECONDITION_VALID_SPAN(name, 0, false);
```
ericwol-msft: min length should be 1 for name and value

danewalton: ah thank you

sdk/iot/core/src/az_iot_hub_client.c
ericwol-msft: y

ericwol-msft: I only see it used in a unit test, or am I blind?

danewalton: This could then get passed on to the parse functions and be used to parse through props. If the `properties` struct comes in and that isn't explicitly set, it could have some garbage value and the parse would attempt to iterate through garbage. If we explicitly set it, it can then be parsed with `az_iot_hub_client_properties_next()`. 

I believe the intended behavior also is that if someone is iterating through props and wants to reset, then call this function again and it acts as a reset.

sdk/iot/core/README.md
```diff
@@ -55,6 +55,43 @@ int main()
...
+  //Initialize the property struct with the span
+  az_iot_hub_client_properties props;
+  az_iot_hub_client_properties_init(&props, property_span);
+  //Append properties
+  az_iot_hub_client_properties_append(&props, AZ_SPAN_FROM_STR("key"), AZ_SPAN_FROM_STR("value"));
+  //At this point, you are able to pass the `props` to other API's with property parameters.
+}
+```
+
+Below is an example use case of initializing an already populated property span.
```
TiejunMS: Shall we mention that the properties must be RFC 2396-encoded as per https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#sending-device-to-cloud-messages?

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -107,3 +108,40 @@ AZ_NODISCARD az_result az_iot_hub_client_id_get(
...
+  AZ_PRECONDITION_VALID_SPAN(value, 1, false);
+
+  az_span prop_span = properties->_internal.properties;
+
+  if (az_span_length(prop_span) > 0)
+  {
+    AZ_RETURN_IF_FAILED(az_span_append_uint8(prop_span, hub_client_param_separator, &prop_span));
+  }
+
+  AZ_RETURN_IF_FAILED(az_span_append(prop_span, name, &prop_span));
```
TiejunMS: Do we expected the `name` and `value` already RFC 2396-encoded as per https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#sending-device-to-cloud-messages?

sdk/iot/core/README.md
danewalton: I can mention it in the doxygen header docs.

danewalton: Good point btw

sdk/iot/core/src/az_iot_hub_client.c
danewalton: As per my current understanding yes. @ericwol-msft can confirm.

sdk/iot/core/README.md
danewalton: [Updated](https://github.com/Azure/azure-sdk-for-c/pull/483/commits/020e7b3b9f2840c55fa6378cba14897c0b0983ae)

sdk/iot/core/src/az_iot_hub_client.c
TiejunMS: Should convenience layer do the encoding or we reply on user application?
Should we check validity the encoding before sending telemetry to cloud?
What is the expected result from user application's perspective when the encoding is incorrect?

danewalton: 1) I don't think that should be a baked in functionality to this API. That would mean someone that knows they have it encoded correctly would have to go through the same API and it would then do similar checks/validations unnecessarily.
2) We should not validate. Let the service do the grunt work of validating. Less cycles on the device and less code size.
3) Expect the service to fail the request/operation.

TiejunMS: When the service fails the request, how does user application know what is going on? Correct me if I was wrong. The service would disconnect from device at TCP layer if the property encoding is incorrect. There is no mechanism at MQTT layer to return error status on publish. So what user application would see is, the connection is disconnected. Should user application continue with a clean session or not (whether to retransmit telemetry data)?

danewalton: I think I stand to be corrected.
Based on the MQTT spec I don't think it's required to be RFC2396 encoded. Should be UTF-8 RFC3629.
[Source here](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.pdf#%5B%7B%22num%22%3A127%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C105%2C546%2C0%5D)

To back this, I tried using MQTT Explorer and sent a message with properties that weren't URL encoded and everything seemed to work. Now I don't know if it is doing underlying manipulation but it does seem to work.

TiejunMS: You are correct. I double verified escape values in topic. Only "/" is not supported by service and service will disconnect immediately. But why the link I pasted above says the property should be url-encoded? Is it optional or mandatory?

danewalton: I don't know but thank you for bringing this up. I'll maybe reach out to someone from the service team to get "the truth".

sdk/iot/core/src/az_iot_hub_client.c
ericwol-msft: ok

ericwol-msft: rename it iterator?

sdk/iot/core/src/az_iot_hub_client.c
danewalton: Regardless of what the ultimate decision is (encode or not), it won't be baked into this API. Will take a note but resolve for now.

sdk/iot/core/src/az_iot_hub_client.c
danewalton: I'd like to issue a new PR to name change to keep scope.

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -107,3 +108,40 @@ AZ_NODISCARD az_result az_iot_hub_client_id_get(
...
+  properties->_internal.current_property = az_span_ptr(buffer);
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_iot_hub_client_properties_append(
+    az_iot_hub_client_properties* properties,
+    az_span name,
+    az_span value)
+{
```
hihigupt: how about exposing system properties keynames from core?

sdk/iot/core/src/az_iot_hub_client.c
```diff
@@ -107,3 +108,40 @@ AZ_NODISCARD az_result az_iot_hub_client_id_get(
...
+  AZ_PRECONDITION_NOT_NULL(properties);
+  AZ_PRECONDITION_VALID_SPAN(name, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(value, 1, false);
+
+  az_span prop_span = properties->_internal.properties;
```
hihigupt: >az_span prop_span = properties->_internal.properties [](start = 2, length = 52)

is it possible to use pointer?

danewalton: If I make it a pointer I can't have the dynamic we had talked about where you only set the "output" span once everything is successful. In this way, if you try to append a key and value and it would exceed the span, previously appended k/v won't get ruined. The unit tests validate that dynamic [here](https://github.com/Azure/azure-sdk-for-c/blob/f1f7b859b19718500d29cccc96023b5188d4c193/sdk/iot/core/tests/cmocka/az_iot_hub_client_tests.c#L377-L384).

sdk/iot/core/src/az_iot_hub_client.c
danewalton: As in which are available? Whats the context here

hihigupt: like content type, 
correlation-id
content-encoding
message-id

danewalton: Yea we definitely can have them as needed. I'd like to add them on an as needed basis as we roll features out.

sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -176,8 +176,8 @@ typedef struct az_iot_hub_client_properties
  * @brief Initializes the Telemetry or C2D properties.
  *
  * @param[in] properties The #az_iot_hub_client_properties to initialize
- * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted
- *                   properties starting with the '?' character.
+ * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted,
+ *                   RFC 2396-encoded properties with the following format: {key}={value}&{key}={value}.
```
CIPop: That RFC is obsolete. https://tools.ietf.org/html/rfc3986 instead. 

We need to understand exactly what _part_ of the RFC is required here.
Different parts of the URI require different types of encoding and usually most characters should not be percent-encoded.

The buffer must start with '?' either way - not sure why that part is removed.

sdk/iot/core/README.md
```diff
@@ -55,6 +55,43 @@ int main()
...
+
+```C
+//FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
+static az_span my_prop_span = AZ_SPAN_LITERAL_FROM_STR("my_device=contoso&my_key=my_value");
```
CIPop: I don't think this example is that relevant. The idea for supporting the initializer was to allow parsing and changes to an existing topic.
For that you would need to 
1. Have a buffer with topic+properties.
2. Find the topic len (w/o the properties)
3. Initialize the properties struct at the correct location in the buffer.

The example could be to add another property or update an existing one.

Scenario: Application generates buffers with topics and wants to use a property for tracing. It should be able to modify the tracing ID without rewriting the topic and properties.

sdk/iot/core/src/az_iot_hub_client.c
CIPop: Right: init initializes the enumeration.
We had the naming discussion already and we didn't want to call it "iterator".

sdk/iot/core/src/az_iot_hub_client.c
CIPop: It was a design decision to not encode. If Unicode support is required we would need to use wchar_t, allow higher Uniocode code-points, etc.
This is still possible today as the application developer can bake them in the application. If they change dynamically then they will need to obtain a proper URI-encoder which is non-trivial and may be quite large.

BTW: we expect the same issues with the host-name part of the URI. @TiejunMS does NX support IDN such as these for DNS resolution today: https://www.w3.org/2003/Talks/0425-duerst-idniri/slide12-0.html?

TiejunMS: We don't encode domain name in DNS.
Add @hihigupt , in this case, I don't think we should encode properties in convenience layer either. What do you think?

PR #485, opened 2020-03-18
Update IoT header to require options not be NULL

danewalton:
Update to align with [this comment](https://github.com/Azure/azure-sdk-for-c/pull/455#pullrequestreview-374495384) in a previous PR.


sdk/iot/core/inc/az_iot_hub_client.h
```diff
@@ -58,7 +58,7 @@ AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default();
  * @param[out] client The #az_iot_hub_client to use for this call.
  * @param[in] iot_hub_hostname The IoT Hub Hostname.
  * @param[in] device_id The Device ID.
- * @param[in] options A reference to an #az_iot_hub_client_options structure. Can be NULL.
+ * @param[in] options A reference to an #az_iot_hub_client_options structure. Cannot be `NULL`.
```
ericwol-msft: I don't think we need the last sentence. how about

 The #az_iot_hub_client_options to use for this call.

CIPop: @jspaith, @JeffreyRichter and I talked about this during a review meeting and I remember the pattern was that NULL can be passed. In that case, internally the defaults will be called.

https://azure.github.io/azure-sdk/clang_design.html#functions-with-many-parameters

danewalton: I think the documentation should detail the expectation for the parameters. Otherwise users have to look at the implementation in `.c` files to understand what the preconditions are for functions.

danewalton:
> I don't think this is correct.

I think the guideline might want to be changed. In the case that someone doesn't want to use the default, they can allocate their own options and pass it. Internal to the function which accepts options, there then isn't an `if` case which pulls in the default function and the default struct (adding to code size).
I'm not sure how deep code optimizations go but if the onus is on the developer to either pass a default option struct or their own, code size can more easily be reduced.

CIPop:
> I think the guideline might want to be changed. In the case that someone doesn't want to use the default, they can allocate their own options and pass it. 

I don't see why that would be better than the current constructor (or factory if you look at options) design pattern.
1. If you allow the struct to be allocated, adding a new member would be a breaking compilation change or run-time overflow. Keeping the options factory allows us, the lib maintainers, to safely add new properties _and their defaults_.
2. If you _don't_ copy the values (just pass in a pointer) you are correct that the code would be faster and smaller as the options object can be reused between clients, less copy operations, etc. It will also have threading implications.

danewalton:
Closing this and updating to allow `NULL` for options. Thanks for the correcting me @CIPop

Issue #468, opened 2020-03-16
[BUG] Re-enable az_iot_sas_token negative unit tests


ewertons:
**Describe the bug**
The az_iot_sas_token negative unit tests were disabled by commit https://github.com/Azure/azure-sdk-for-c/commit/34c8b78dd70d07c93e8fb63b3e865351da223a5e
They must be re-enabled.

***Exception or Stack Trace***
Add the exception log and stack trace if available

**To Reproduce**

***Code Snippet***
Add the code snippet that causes the issue.

**Expected behavior**

**Screenshots**

**Setup (please complete the following information):**
 - OS: Any
 - IDE : Any
 - Version of the Library used: https://github.com/Azure/azure-sdk-for-c/commit/34c8b78dd70d07c93e8fb63b3e865351da223a5e

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Bug Description Added
- [ ] Repro Steps Added
- [x] Setup information Added

ewertons:
This issue depends on fixing https://github.com/Azure/azure-sdk-for-c/issues/467

ewertons:
Fixed.

PR #486, opened 2020-03-18
Rewrite az_iot_sas_token unit tests to validate precondition checks

ewertons:
This addresses github issue https://github.com/Azure/azure-sdk-for-c/issues/468
According to discussion with @JeffreyRichter devs are welcome to add unit tests to verify
that preconditions are verified in functions. These changes implement a method of testing such
preconditions.


ewertons:
@JeffreyRichter , please review as well. Thanks.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
```diff
@@ -19,7 +22,38 @@
...
+// This block defines the resources needed to verify precondition checking.
+// The macro assert_precondition_checked(func) alone shall be used to assert if a precondition is verified.
+// If a precondition is not verified within a function the result will be either:
+// - A crash in the test, as the the target function will continue to execute with an invalid argument, or
+// - An assert failure indicating a precondition was not tested.
+// Notice that:
+// - If assert_precondition_checked(func) is used, the module must include <setjmp.h>;
+// - If a function has two precondition checks and both are supposed to fail on a given test, assert_precondition_checked(func)
+//   is unable to distinguish which precondition has failed first. Testing precondition checking separately is advised.
+// - Tests using assert_precondition_checked(func) must not be run in parallel (!).
```
JeffreyRichter: Fyi: you could make parallel work if you used thread local storage or a map of thread ID to jmp_buf.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
```diff
@@ -19,7 +22,38 @@
...
+  longjmp(g_jmp_buf, 0);
+} 
+
+#define assert_precondition_checked(fn) \
+  precondition_test_count = 0; \
+  (void)setjmp(g_jmp_buf); \
+  if (precondition_test_count == 0) { \
+    fn; \
+  } \
+  assert_int_equal(1, precondition_test_count);
```
danewalton: Because this doesn't return from the function, if two preconditions fail would this evaluate to false? I guess that's a feature not a bug.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
```diff
@@ -19,7 +22,38 @@
...
+// - An assert failure indicating a precondition was not tested.
+// Notice that:
+// - If assert_precondition_checked(func) is used, the module must include <setjmp.h>;
+// - If a function has two precondition checks and both are supposed to fail on a given test, assert_precondition_checked(func)
+//   is unable to distinguish which precondition has failed first. Testing precondition checking separately is advised.
+// - Tests using assert_precondition_checked(func) must not be run in parallel (!).
+
+static jmp_buf g_jmp_buf; 
+static unsigned int precondition_test_count = 0;
+static void az_precondition_test_failed_fn()
```
danewalton: Is there any way we can have this code in a .c or headers so it doesn't have to be copied over for each new test file?

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
ewertons: True. I'll make that a note in the documentation above.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
ewertons: This mechanism kicks in on the first precondition that fails, no matter which and how many would fail on a given function call.
If two preconditions are to fail, only the first one will actually be checked.

sdk/iot/core/tests/cmocka/az_iot_sas_token_tests.c
ewertons: Makes sense. @JeffreyRichter , do you think this solution is good enough to go into the core SDK? And if yes, where would you recommend it to go?

JeffreyRichter: Not in the SDK but it can go here in this folder: https://github.com/Azure/azure-sdk-for-c/tree/master/sdk/core/core/test/cmocka

ewertons: Moved these macros into az_precondition_internal.h

JeffreyRichter: No, az_precondition_internal.h is part of the SDK. and not in the folder I suggested.

sdk/iot/core/tests/cmocka/az_precondition_testing.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
CIPop: @ewertons this file can't be placed in the `iot` folder as it doesn't follow the naming guidelines.
Is it common? Should it go to `core` instead?

PR #491, opened 2020-03-19
Dev guide changes

CIPop:
- Fixes for dependency installation
- Ubuntu 18.04 guide for prerequisites
- Some hints for code coverage


CONTRIBUTING.md
```diff
@@ -98,7 +93,42 @@ Right after opening project, Visual Studio will read cmake files and generate ca
...
+```
+
+> Note: On macOS, `.\bootstrap-vcpkg` may fail if your version of the C++ toolchain is not new enough to support vcpkg. To resolve
+this, vcpkg recommends installing `gcc@6` from Homebrew (`brew install gcc@6`), then re-run the bootstrapping script.
+
+#### Debian
+
+Alternatively, for Ubuntu 18.04 you can use:
+
+`sudo apt install build-essential cmake libcmocka-dev libcmocka0 gcovr lcov doxygen curl libcurl4-openssl-dev libssl-dev`
```
ahsonkhan: Are all these required? I recently got the repo building on Ubuntu (using WSL) and I didn't need all these (that said, I only built/test the sources & docs, not samples). Did you find these to be needed in your experience? For example, isn't `build-essential` already there out of the box (including gcc/make)?

CIPop: Depends on the box and what you've installed. I don't consider WSL since it's designed with devs in mind.

In my experience Ubuntu Server LTS, Raspbian-Lite and a Debian I used on an iMX6 device did not have them. I copied these from IOT C-SDK: https://github.com/Azure/azure-iot-sdk-c/blob/master/build_all/linux/setup.sh.

What I'm not 100% sure of are the ones starting with `curl` since that's only required for certain types of builds.

CIPop: Also see other libs having the same requirements: https://github.com/eclipse/paho.mqtt.c#build-requirements--compilation-using-cmake

`apt-get install build-essential gcc make cmake cmake-gui cmake-curses-gui`

CONTRIBUTING.md
```diff
@@ -98,7 +93,42 @@ Right after opening project, Visual Studio will read cmake files and generate ca
 
 ### Linux / Mac
 
-VCPKG can also be used in Linux to avoid installing libraries globally. Follow instructions [here](#vcpkg) to use VCPKG in Linux.
+#### VCPKG
+
+VCPKG can be used to download packages sources, headers and build libraries for whatever TRIPLET is set up (platform/arq).
```
ahsonkhan: arq -> architecture?

CONTRIBUTING.md
```diff
@@ -98,7 +93,42 @@ Right after opening project, Visual Studio will read cmake files and generate ca
...
+
+Follow next steps to install VCPKG and have it linked to cmake
+
+```bash
+# Clone vcpgk:
+git clone https://github.com/Microsoft/vcpkg.git
+# (consider this path as PATH_TO_VCPKG)
+cd vcpkg
+# build vcpkg (remove .bat on Linux/Mac)
+./bootstrap-vcpkg.sh
```
ahsonkhan: comment and instructions mismatch. there is no `.bat` here. Change the command, or update comment to say use `.bat` on Windows.

CONTRIBUTING.md
```diff
@@ -48,7 +48,7 @@ Thank you for your interest in contributing to Azure SDK for C.
 
 Project contains files to work on Windows, Mac or Linux based OS.
 
-### VCPKG
+### Windows
 
 vcpkg is the easiest way to have dependencies installed. It downloads packages sources, headers and build libraries for whatever TRIPLET is set up (platform/arq).
```
ahsonkhan: We now seem to be repeating this paragraph twice. Can we avoid that by restructuring things?

CIPop: I'm repeating the paragraph because:
1. It's not currently clear how to get this running on Linux (suggestions are not complete)
2. We should always keep Windows vs Linux vs Mac separate as much as possible. I don't have Mac separated completely because it doesn't work today at all.

CIPop: I agree with the restructuring - let's keep that in a separate PR. I need to add some things to VCPkg for Paho samples.
Based on our discussion, we need to have:
1. Something for people wanting to _use_ the SDK - how to use VCPKG or other things to build and run samples.
2. Something for people wanting to contribute (this document). For these, _maybe_ we need to split into devs for a particular client since we have very different requirements.

PR #492, opened 2020-03-19
Fix all parameter name mismatch between header and source files.

ahsonkhan:
Fixing warnings from `clang-tidy` related to `[readability-inconsistent-declaration-parameter-name]`

The goal here is consistency. I always deferred to the parameter names in the header file rather than what's in source.

If a parameter name is incorrect in either place, please let me know.


sdk/core/core/src/az_aad.c
```diff
@@ -84,13 +84,13 @@ AZ_NODISCARD az_result _az_aad_build_body(
...
 }
 
-AZ_NODISCARD az_result _az_aad_request_token(_az_http_request* ref_request, _az_token* out_token)
+AZ_NODISCARD az_result _az_aad_request_token(_az_http_request* request, _az_token* out_token)
 {
   // FIXME: If you uncomment the line below, we'll start getting HTTP 400 Bad Request instead of 200
   // OK. I suspect, it is because there's a bug in the code that adds headers. Could be something
   // else, of course.
   /*AZ_RETURN_IF_FAILED(az_http_request_append_header(
-      ref_request,
```
antkmsft: I think we should use the `ref_request` naming

sdk/core/core/src/az_http_policy.c
```diff
@@ -83,26 +83,26 @@ _az_apply_credential(_az_credential* credential, _az_http_request* ref_request)
 }
 
 AZ_NODISCARD az_result az_http_pipeline_policy_credential(
-    _az_http_policy* policies,
-    void* options,
-    _az_http_request* ref_request,
-    az_http_response* out_response)
```
antkmsft: We should be using `out_response` naming.

sdk/core/core/src/az_http_policy_logging.c
```diff
@@ -158,29 +158,29 @@ void _az_http_policy_logging_log_http_response(
...
   if (az_log_should_write(AZ_LOG_HTTP_REQUEST))
   {
-    _az_http_policy_logging_log_http_request(ref_request);
+    _az_http_policy_logging_log_http_request(p_request);
   }
 
   if (!az_log_should_write(AZ_LOG_HTTP_RESPONSE))
   {
     // If no logging is needed, do not even measure the response time.
-    return az_http_pipeline_nextpolicy(policies, ref_request, ref_response);
```
antkmsft: `policies, ref_request, ref_response`is the ideal naming. We don't use the `p_` convention anomore, we should change it to the other way.
```
somestruct const* options // means it is read-only and we will not modify it
somestruct* ref_somestruct //means we read it and (may) modify it
somestruct* out_somestruct // means we don't read it, the customer can provide an uninitialized struct, and we will fill it in as a returning result
``

File: sdk/core/core/src/az_http_policy_retry.c [hunk 4]
```diff
@@ -141,30 +141,30 @@ AZ_INLINE AZ_NODISCARD az_result _az_http_policy_retry_get_retry_after(
 }
 
 AZ_NODISCARD az_result az_http_pipeline_policy_retry(
-    _az_http_policy* policies,
-    void* options,
```
antkmsft: And I think Jeffrey wanted to name it options rather than data, check with him.

sdk/core/core/src/az_http_request.c
```diff
@@ -20,22 +20,22 @@ AZ_NODISCARD az_result _az_is_question_mark(az_span slice)
 }
 
 AZ_NODISCARD az_result az_http_request_init(
-    _az_http_request* p_hrb,
```
antkmsft: It is good we no longer use `hrb` naming. At a time, the struct was named `http_request_builder` so this is where `hrb` comes from; but now it is just `http_request`, so the correct name would be `request`, and in this specific function we should name it as `out_request` because we don't read it, but fill it in as an output parameter.

sdk/core/core/src/az_http_request.c
```diff
@@ -54,39 +54,39 @@ AZ_NODISCARD az_result az_http_request_init(
   return AZ_OK;
 }
 
-AZ_NODISCARD az_result az_http_request_append_path(_az_http_request* p_hrb, az_span path)
+AZ_NODISCARD az_result az_http_request_append_path(_az_http_request* p_request, az_span path)
```
antkmsft: `p_request` is better than `p_hrb`, but ideally this instance should be named `ref_request`

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
```diff
@@ -53,19 +53,19 @@ AZ_NODISCARD az_storage_blobs_blob_client_options az_storage_blobs_blob_client_o
 }
 
 AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
-    az_storage_blobs_blob_client* self,
+    az_storage_blobs_blob_client* client,
```
antkmsft: Let's check across the team, if we still use `self` (which is basically "`this`" but does not use a reserved keyword if compiled by a C++ compiler), or named parameters such as `client`. I am fine either way as long as we are consistent. If we decide to not use `self`, let's replace all other instances. My opinion is that `client` is more readable and easier to for the customers, and not using `self` is easier for us when the function does not conceptually aligns to `self`'s "method" in the OOP sense.

JeffreyRichter: I agree, I really have come not to like 'self'. When I am reading a function's code, I have no idea what 'self' refers to and since we use it a lot, it has different meaning in different places and it's really hard to manage understand the code base. I'd like to abolish 'self'

sdk/core/core/src/az_aad.c
JeffreyRichter: I'm fine with this but we need to go through the whole code base and fix 'ref_' everywhere.
I think I fixed a lot of 'out_'s so I think we're OK with this.

sdk/core/core/src/az_http_policy.c
JeffreyRichter: I agree

sdk/core/core/src/az_http_policy_logging.c
JeffreyRichter: Yes, I like these conventions - please apply them

sdk/core/core/src/az_http_policy_retry.c
JeffreyRichter: Yes, policy's have 'options'; not 'data'

sdk/core/core/src/az_http_request.c
JeffreyRichter: Yes, I agree

sdk/core/core/src/az_http_request.c
JeffreyRichter: Yes, I agree

sdk/storage/blobs/src/az_storage_blobs_blob_client.c
gilbertw: I prefer client.

antkmsft: Ahson, can you please check the code for the instances of using `self` and update them to better names depending on the context?

gilbertw: Nit, I'd prefer that cleanup be a separate PR.

ahsonkhan: > Nit, I'd prefer that cleanup be a separate PR.

I agree.

ahsonkhan:
Thanks for the feedback. My intent with this PR was mainly to synchronize the header/source files and not change parameter names. Though I agree with most of the suggestions, I think we should fix them all up and come with with parameter naming patterns/guidelines together and then apply them across the board.

PR #504, opened 2020-03-20
move the prop span to a public field [DO NOT MERGE]

danewalton:


danewalton:
> 
> 
> There is no description on why this public API is changed - why would that internal property be exposed?
> Also, changing the public API will require review from Azure SDKs and the API board.

My apologies this was supposed to be a draft. Will mark as do not merge.

PR #505, opened 2020-03-20
Initial checkin of embedded IoT PnP

jspaith:
Initial checkin of PnP header file.  Currently contains:
* `az_iot_pnp_client` structure 
* MQTT connection primitives (like user name, sas, etc.)
* PnP telemetry
* PnP commands

NOTES:
* Assuming for now `content_type` and `encoding_type` are (a) baked on a per-telemetry message level and (b) required at message send time, not earlier.  This is a different discussion we should deal with pre-GA, but not for now header reflects what we've got.
* This is mirroring the az_iot_hub* closely.  There's been some incoming changes today in this though that I'm not trying to keep up with in real-time.  As the iot_hub settles, this keeping comments inline will be something we need to remember but hopfully easier given less churn.


CIPop:
@jspaith sorry for adding even more work on this. Please see: https://github.com/Azure/azure-sdk-for-c/pull/506.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * @brief Azure IoT PnP Client.
+ *
+ */
+typedef struct az_iot_pnp_client
+{
+  struct
+  {
+    az_iot_hub_client iot_hub_client;
+	az_span root_interface_name;
```
danewalton: nit: formatting

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+    az_span iot_hub_hostname,
+    az_span device_id,
+    az_span root_interface_name,
+    az_iot_pnp_client_options const* options);
+
+/**
+ * @brief Gets the MQTT user name.
+ *
+ * The user name will be of the following format:
+ * {iothubhostname}/{device_id}/?api-version=2018-06-30&{user_agent}?digital-twin-model-id={root_interface_name}
```
danewalton: should the character after the `{user_agent}` be a `&` instead of a `?`

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+AZ_NODISCARD az_result az_iot_pnp_client_user_name_get(
+    az_iot_pnp_client const* client,
+    az_span mqtt_user_name,
+    az_span* out_mqtt_user_name);
+
+/**
+ * @brief Gets the MQTT client id.
+ *
+ * The client id will be of the following format:
+ * {device_id}
```
danewalton: just to be clear here there is no concept of modules with pnp correct?

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+ * @param[in] mqtt_client_id An empty #az_span with sufficient capacity to hold the MQTT client id.
+ * @param[out] out_mqtt_client_id The output #az_span containing the MQTT client id.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_id_get(
+    az_iot_pnp_client const* client,
+    az_span mqtt_client_id,
+    az_span* out_mqtt_client_id);
+
+/**
```
danewalton: is this going to get pulled in as doxygen too? If you just want it to be a C comment remove the second asterisk

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+ * @note Telemetry MQTT Publish messages must have QoS At Least Once (1).
+ * @note This topic can also be used to set the MQTT Will message in the Connect message.
+ *
+ * Should the user want a null terminated topic string, they may allocate a buffer large enough
+ * to fit the topic plus a null terminator. They must set the last byte themselves or zero
+ * initialize the buffer.
+ *
+ * @param[in] client The #az_iot_pnp_client to use for this call.
+ * @param[in] component_name An #az_span specifying the component name to publish telemetry on.
+ * @param[in] content_type An #az_span specifying the content type of the message payload.
```
danewalton: will the available options for this be documented somewhere or is it "anything goes"

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+ * @param[in] encoding_type An #az_span specifying the encoding type of the message payload.
```
danewalton: same here: available options?

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+ * @param[in] client The #az_iot_pnp_client to use for this call.
+ * @param[in] base64_hmac_sha256_signature The Base64 encoded value of the HMAC-SHA256(signature,
+ *                                         SharedAccessKey). The signature is obtained by using
+ *                                         #az_iot_pnp_client_sas_signature_get.
+ * @param[in] key_name The Shared Access Key Name (Policy Name). This is optional. For security
+ *                     reasons we recommend using one key per device instead of using a global
+ *                     policy key.
+ * @param[in] mqtt_password An empty #az_span with sufficient capacity to hold the MQTT password.
+ * @param[out] out_mqtt_password The output #az_span containing the MQTT password.
+ * @return #az_result.
```
danewalton: nit: there are several places where `#az_result` is followed by a `.` this should probably be removed

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * @brief A command request received from the PnP service.
+ *
+ */
+typedef struct az_iot_pnp_client_command_request
+{
+  az_span request_id; /**< The request id.
+                       * @note The application must match the method request and method response. */
+  az_span component; /**< The component the command is targeted towards. */
+  az_span command_name; /**< The name of command to invoke */
```
danewalton: nit: just to maintain sentence structure of other docs, add `the` in front of command

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * @brief Attempts to parse a received message's topic.
+ *
+ * @param[in] client The #az_iot_pnp_client to use for this call.
+ * @param[in] received_topic An #az_span containing the received topic.
+ * @param[in] received_payload An #az_span containing the received payload.
+ * @param[out] out_request If the message is a command request, this will contain the
+ *                         #az_iot_pnp_client_command_request.
+ * @return #az_result
```
danewalton: I think this probably goes for the hub docs as well but there should be documentation here about what to expect in the case that it IS a command request and in the case it IS NOT a command request. Is there a particular `az_result` that is returned when it is or is not? Is there something about the makeup of the returned `az_iot_pnp_client_command_request`?

Because if I'm a dev here, I could expect that `AZ_OK` is returned if it is a command OR if it was just successfully parsed and the notification that it is a command comes through the `out_request`. If that makes sense.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Correct.  When we add modules, I'll add details analogous to iothub layer.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: There's rules in practice about what various backends can be expected to support (basically utf-8 and application/json).  I'm putting notes of what's currently supported inline just to steer folks correctly.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Yup, adding here too.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Yup!

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: fixing

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Agreed.  I think there should not be the period, because just \#az_result by itself is not a complete sentence.  We need consistency at least.

I see some other places like az_span.h and az_json.h where we do indeed have a `.`, but these are complete sentences.

@CIPop az_iot_hub_client.h is itself inconsistent RE whether there's a period or not after \#az_result.  I'm removing periods from the pnp header unless you strongly object.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: fixed.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Agreed, that merits special consideration in docs.  @CIPop  - I poked around IoT headers and I couldn't find this error code.  Is it declared yet?

sdk/iot/pnp/inc/az_iot_pnp_client.h
danewalton: I can send a PR to remove from other docs as well once the new directory structure is pushed.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+
+/**
+ * @brief Azure IoT PnP command response options.
+ *
+ */
+typedef struct az_iot_pnp_client_command_response_options
+{
+  struct
+  {
+    int _not_used;
```
CIPop: I see these left with every operation. I think they make the public API a bit harder to understand/read. Could we also have versions of these APIs without the parameter that must be `NULL` today every single call? 
We could maybe postpone adding these options until we really need them?

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+ * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
+ * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_telemetry_publish_topic_get(
+    az_iot_pnp_client const* client,
+    az_span component_name,
+    az_span content_type,
+    az_span encoding_type,
+    az_iot_pnp_client_telemetry_options const* options,
```
CIPop: Same here and everywhere else (if you decide to keep), it should probably be the last argument.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+ * @param[in] device_id The Device ID.
+ * @param[in] root_interface_name The root interface of the #az_iot_pnp_client.
+ * @param[in] options A reference to an #az_iot_pnp_client_options structure. Can be NULL.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_init(
+    az_iot_pnp_client* client,
+    az_span iot_hub_hostname,
+    az_span device_id,
+    az_span root_interface_name,
```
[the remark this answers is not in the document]
JeffreyRichter: Maybe these spans should be copied into byte buffers that are in the client to make them immutable and the customer doesn't have to keep the spans around. Although it's not a problem if the span are over literal strings.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+ * @brief Azure IoT PnP telemetry options.
+ *
+ */
+typedef struct az_iot_pnp_client_telemetry_options
+{
+  struct
+  {
+    // OPEN: I can't find samples in Azure SDK modeling this.  Closest is
+    // sdk\storage\blobs\inc\az_storage_blobs.h where it has 'az_span option" in
+    // az_storage_blobs_blob_download_options that's ignored in practice.
```
JeffreyRichter: Yes, this is fine.

sdk/iot/pnp/inc/az_iot_pnp_client.h
JeffreyRichter: Tough call on this one. I feel that output parameters should be last so maybe just before out_mqtt_topic?
Although I know it';s nice to have mqtt_topic next to out_mqtt_topic.
I don't have an answer. I'm open to suggestions.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @param[in] client The #az_iot_pnp_client to use for this call.
+ * @param[in] options A reference to an #az_iot_pnp_client_command_subscription_options structure.
+ * Can be NULL.
+ * @param[in] mqtt_topic_filter An empty #az_span with sufficient capacity to hold the MQTT topic
+ *                              filter.
+ * @param[out] out_mqtt_topic_filter The output #az_span containing the MQTT topic filter.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_commands_subscribe_topic_filter_get(
```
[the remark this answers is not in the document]
JeffreyRichter: Yes, many of these functions do not match our guidelines; these should be fixed.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+    az_iot_pnp_client const* client,
+    az_iot_pnp_client_command_subscription_options const* options,
```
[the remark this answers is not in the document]
JeffreyRichter: Or, 2nd to last?

sdk/iot/pnp/inc/az_iot_pnp_client.h
JeffreyRichter: I think KEY az_result values should be specified here. See https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/inc/az_json.h and search for az_json_token_get_boolean to see an example.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
jspaith: Fixed, never dynamic.  

The root_interface defines what the device is in the 1st place, e.g. "Constoso-Air-Conditioner" and reflects modeled DTDL.  The only time it's used *today* is on the initial MQTT CONNECT packet in building user name.  However, this could potentially be different in a future binding so I don't want to have this passed in the `get_username` stuff.  And it's fundamental to what the device is it seems to make sense in the pnp_client itself.

sdk/iot/pnp/inc/az_iot_pnp_client.h
JeffreyRichter: If you have no options today, how about this:
1. remove these *_options structs from the public API
2. remove the function that returns the default *_options struct instances.
3. modify the functions that takes these structs to take a void* called 'reserved' and document that NULL must be passed.

This will prevent customers from creating these things and calling functions that do nothing and force them to use the simpler code. Later, you can introduce these things publicly when you're ready.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+ * be NULL.
+ * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
+ * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_commands_response_publish_topic_get(
+    az_iot_pnp_client const* client,
+    az_span request_id,
+    uint16_t status,
+    az_iot_pnp_client_command_response_options const* options,
```
JeffreyRichter: This is he content type stuff, right?
If so, I think this should be on the client and not passed to each of these functions.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
jspaith: @JeffreyRichter - the vast majority of the time these will be string literals, since it'll be hard-coded into the application itself.  It knows it's always a "Contoso-....-whatever", this being dynamically built would be surprising.  So I think the simpler mechanism in there today is OK.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+                           Only AZ_SPAN_FROM_STR("utf-8") is currently supported.
+ * @param[in] options A reference to an #az_iot_pnp_client_telemetry_options structure. Can be NULL.
+ * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
+ * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @return #az_result
+ */
+AZ_NODISCARD az_result az_iot_pnp_client_telemetry_publish_topic_get(
+    az_iot_pnp_client const* client,
+    az_span component_name,
+    az_span content_type,
```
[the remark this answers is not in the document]
jspaith: ;tl-dr - I'm going to do what you suggest

Details:
I *do* think these should be optional and NOT required.  New thread started offline to make sure there's no surprises there but pushing this as plan.

Here's what I'm thinking generally:
* Move to options to the `client` since overwhelming majority if not all of time it'll be set.  And certainly for GA that will be the case RE only one type.
* If(!) it turns out convenience layer needs to mix/match JSON and some other binding -- and there's no clear scenario for that yet! - then we'd extend out the telemetry_options field to let it over-ride the "global".  So we're not architected into a corner here, but are designing for common case.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
jspaith: Prior art in SDK suggests it should be 2nd to last, and out params of `out_mqtt_...` come last.  E.G. `az_storage_blobs_blob_upload` and `az_storage_blobs_blob_upload`.  

@CIPop I plan on making it 2nd to last based on this; if you think this is wrong in general could you please start up a new thread with stake-holders so we get on same page?

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
jspaith: Fixed here & other places.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: OK - agree void* is cleaner, will fix.  @JeffreyRichter should this guidance be put into the general API guidelines someplace so others don't re-invent a different pattern?

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Mooted because we're going to the pvReserved (uhm, reserved.  Still want to call it pvResorved per Windows/Hungarian days...)

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: Sounds good @danewalton.  I've fixed PnP's so resolving at my level.

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,334 @@
...
+
+/**
+ * @brief Azure IoT PnP telemetry options.
+ *
+ */
+typedef struct az_iot_pnp_client_telemetry_options
+{
+  struct
+  {
+    // OPEN: I can't find samples in Azure SDK modeling this.  Closest is
```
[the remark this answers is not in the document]
jspaith: Yes, agree the TODO pattern in comment is correct way had it gone through to a merge to master.  I'd hoped to resolve pre-merge so I didn't want a GitHub bug on non-merged code; in future I can just put in PR comments or offline.

In any case his is mooted per the `reserved` discussion.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
jspaith: Ugh, missed other discussion thread when I typed above.  Let's continue any discussion at https://github.com/Azure/azure-sdk-for-c/pull/505#discussion_r396810033, and I'm resolving this one.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: No - the command_response options is currently not set to anything, and will be changed to a reserved per other discussions.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: I personally like the ordering being (client_handle, in-required1, ... in-required<n>, optional, &out_param).  That's how I've seen it in other projects at least.  It's unfortunate that the out_mqtt_topic gets split from its buddy though, but I don't like optional in middle of in-required*, either.

I'm going to go with this *for now*.  As this is really easy to change, if/when SDK guidance is different here I'll follow convention.

sdk/iot/pnp/inc/az_iot_pnp_client.h
jspaith: One final question - I want to explicitly document that `reserved` must be NULL and put a pre-condition check on it in the implementation.  This way we can be sure all callers are setting it NULL if we ever allow non-NULL.  Allowing this to be anything would run risk of breaking anything for whatever reason was passing stuff in here.

Any concerns?

JeffreyRichter:
This sounds so good to me. One thing: true guidance should say: if there are multiple VB reserved parameters they should be called reserved, reserved2, reserved3, etc.

Or, something like this. I know it's very unlikely to happen so the ugliness is not a big deal but we should just address it now.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
JeffreyRichter: ok

sdk/iot/pnp/inc/az_iot_pnp_client.h
```diff
@@ -0,0 +1,261 @@
...
+
+/**
+ * @brief Azure IoT PnP Client options.
+ *
+ */
+typedef struct az_iot_pnp_client_options
+{
+  az_span user_agent; /**< The user-agent is a formatted string that will be used for Azure IoT
+                         usage statistics. */
+  az_span content_type; /**< Content type specified in MQTT topic of telemetry messages */
```
CIPop: :+1:

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
CIPop: > The root_interface defines what the device is in the 1st place, e.g. "Constoso-Air-Conditioner" and reflects modeled DTDL.

It really makes no sense to me to have it here - the Contoso-Air-Conditioner cannot suddenly become the Contoso-High-Performance-Computer overnight. The association between a device and it's type should be done one time only: at provisioning.
Allowing this to change from the device side, dynamically, presents all sorts of issues.

PR #507, opened 2020-03-23
[WIP]-feature/az_json-support-chaining-objects

amih90:
I failed to use `az_json` to build nested json, for example:
`{"foo":[{"bar":true}]}`

added test scenario that shows how I tried to do it:
- Created the internal object B - `{"bar":true}`
- Created the outer object A and the array
- Append the internal object B in the array in A.

**Problem**
Using `az_json_append_span` which changed to private lately result with escaped slashes, i.e. 
`{"foo":[\{\\"bar\\":true\}]}`
and using `az_json_append_string` result with object as string
`{"foo":["{"bar":true}"]}`

**Work in Progress**
- Support `AZ_JSON_TOKEN_OBJECT` and `az_json_append_object` (it changed recently from 'az_json_append_object()' to 'az_json_append_object_start()'

**TODO**
- Support also `AZ_JSON_TOKEN_ARRAY` and `az_json_append_array` or to support array as an object?


sdk/core/core/test/cmocka/test_json_builder.c
```diff
@@ -150,4 +150,57 @@ void test_json_builder(void** state)
...
+    TEST_EXPECT_SUCCESS(az_json_builder_init(&builder, AZ_SPAN_FROM_BUFFER(array)));
+
+    // 0___________________________________________________________________________________________________1
+    // 0_________1_________2_________3_________4_________5_________6_________7_________8_________9_________0
+    // 01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456
+    // {"foo":[{"bar":true}]}
+    TEST_EXPECT_SUCCESS(az_json_builder_append_token(&builder, az_json_token_object_start()));
+    {
+      TEST_EXPECT_SUCCESS(az_json_builder_append_object(&builder, AZ_SPAN_FROM_STR("foo"), az_json_token_array_start()));
+      TEST_EXPECT_SUCCESS(az_json_builder_append_array_item(&builder, az_json_token_object(nested_object_builder._internal.json)));
```
amih90: az_span_append(dest=builder._internal.json, src=nested_object_builder._internal.json)?

PR #513, opened 2020-03-24
az_iot_hub_client_c2d_received_topic_parse

ewertons:
The code of this function is missing the URL-decoding part, which is being discussed by the teams.


sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+    token->_internal.capacity = span._internal.length;
+
+    out_span->_internal.ptr = NULL;
+    out_span->_internal.length = 0;
+    out_span->_internal.capacity = 0;
+  }
+
+  return AZ_OK;
+}
+
```
danewalton: nit: three new lines here

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,55 @@
...
+    az_span received_topic,
+    az_iot_hub_client_c2d_request* out_request)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(received_topic, 1, false);
+  AZ_PRECONDITION_NOT_NULL(out_request);
+
+  az_span token;
+  AZ_RETURN_IF_FAILED(az_span_token_next(received_topic, c2d_topic_suffix, &token, &received_topic));
+  AZ_RETURN_IF_FAILED(az_span_token_next(received_topic, c2d_topic_suffix, &token, &received_topic));
```
danewalton: is there a reason this is getting run through twice?

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+AZ_NODISCARD az_result az_span_token_next(az_span span, az_span delim, az_span* token, az_span* out_span)
+{
+  AZ_PRECONDITION_VALID_SPAN(span, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(delim, 1, false);
+  AZ_PRECONDITION_NOT_NULL(token);
+  AZ_PRECONDITION_NOT_NULL(out_span);
+
+  az_span instance;
+
+  token->_internal.ptr = span._internal.ptr;
```
JeffreyRichter: We prefer calling az_span_init in the if & else; this makes he code more maintainable and can catch some programmer errors.

JeffreyRichter: token is an out parameter and so should be named out_token then.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+
+  if (az_span_find(span, delim, &instance) == AZ_OK)
+  {
+    uint8_t* instance_ptr = az_span_ptr(instance);
+    int32_t instance_length = az_span_length(instance);
+
+    token->_internal.length = (int32_t)(instance._internal.ptr - span._internal.ptr);
+    token->_internal.capacity = token->_internal.length;
+
+    out_span->_internal.ptr = instance._internal.ptr + instance._internal.length;
```
JeffreyRichter: use az_span_init for out_token & out_span. I think out_span should be renamed: what does this span represent?

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+    out_span->_internal.capacity = (int32_t)(span._internal.capacity - instance_length - (instance_ptr - span._internal.ptr));
+  }
+  else
+  {
+    token->_internal.length = span._internal.length;
+    token->_internal.capacity = span._internal.length;
+
+    out_span->_internal.ptr = NULL;
+    out_span->_internal.length = 0;
+    out_span->_internal.capacity = 0;
```
JeffreyRichter: use az_span_init for out_token; use AZ_SPAN_NULL for out_span

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,55 @@
...
+#include <az_precondition_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+#include <_az_cfg.h>
+
+static const uint8_t c2d_null_terminator = '\0';
+static const az_span c2d_topic_prefix = AZ_SPAN_LITERAL_FROM_STR("devices/");
+static const az_span c2d_topic_suffix = AZ_SPAN_LITERAL_FROM_STR("/messages/devicebound/");
+static const az_span hash_tag = AZ_SPAN_LITERAL_FROM_STR("#");
```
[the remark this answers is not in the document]
ewertons: Done.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,55 @@
...
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic_filter);
+
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic_filter, c2d_topic_prefix, out_mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_topic_filter, client->_internal.device_id, out_mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_topic_filter, c2d_topic_suffix, out_mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_topic_filter, hash_tag, out_mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(
```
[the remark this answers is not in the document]
ewertons: Removed.

sdk/core/core/src/az_span.c
ewertons: Trimmed.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
ewertons: Yes, this is a tokenizer. The first run gets you the first token up to <c2d_topic_suffix>, the second one gives you the part after that (the properties we are looking for).

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,10 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
  */
 AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
+AZ_NODISCARD az_result az_span_find(az_span span, az_span target, az_span* out_span);
```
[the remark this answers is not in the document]
ewertons: Absolutely. This is a preliminary code, t hat was on my radar.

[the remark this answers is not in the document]
ewertons: Changed the param name. Added documentation.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,10 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+
+AZ_NODISCARD az_result az_span_token_next(az_span span, az_span delim, az_span* token, az_span* out_span);
```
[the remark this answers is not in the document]
ewertons: Agree for all comments, except about using uint8_t only for the delimiter. It would then need to be null terminated to search for it in the "source". 
Documentation is added now.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,55 @@
...
+    az_span mqtt_topic_filter,
+    az_span* out_mqtt_topic_filter)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic_filter, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic_filter);
+
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic_filter, c2d_topic_prefix, out_mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(*out_mqtt_topic_filter, client->_internal.device_id, out_mqtt_topic_filter));
```
hihigupt: Please update out only once it is ready to escape from function without error.

Please look at telemetry [https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/hub/src/az_iot_hub_client_telemetry.c](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/hub/src/az_iot_hub_client_telemetry.c)

sdk/core/core/inc/az_span.h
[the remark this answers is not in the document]
JeffreyRichter: I don't understand the null-terminated commented. Are you looking for a single character delimiter? You know the end due to the span's length.

sdk/core/core/src/az_span.c
ewertons: Done.

sdk/core/core/src/az_span.c
ewertons: Done.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+        {
+            break;
+        }
+      }
+
+      if (j == target_length)
+      {
+        out_span->_internal.ptr = span_ptr + i;
+        out_span->_internal.length = target_length;
+        out_span->_internal.capacity = target_length;
```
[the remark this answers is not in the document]
ewertons: Done.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,82 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+  int32_t span_length = az_span_length(span);
+  uint8_t* target_ptr = az_span_ptr(target);
+  int32_t target_length = az_span_length(target);
+
+  for (int i = 0; i < span_length; i++)
+  {
+    if (span_ptr[i] == target_ptr[0])
+    {
+      int j;
+      for (j = 1; j < target_length; j++)
```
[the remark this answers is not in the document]
ewertons: Good catch. Missed a "&& (i + j) < source_length" as second condition in that for loop.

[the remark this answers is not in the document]
ewertons: Line 158 got replaced by an az_span_init. Please recheck after I send an update to the PR. Thanks.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
ewertons: Done.

ewertons:
@JeffreyRichter , @danewalton , @hihigupt , comments addressed.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,244 @@
...
+#include <cmocka.h>
+#include <az_test_precondition.h>
+
+#define TEST_DEVICE_ID_STR "my_device"
+#define TEST_DEVICE_HOSTNAME_STR "myiothub.azure-devices.net"
+#define TEST_MODULE_ID "my_module_id"
+
+static const az_span test_device_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_HOSTNAME_STR);
+static const az_span test_device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID_STR);
+static uint8_t g_test_correct_subscribe_topic[] = {'d','e','v','i','c','e','s','/','m','y','_','d','e','v','i','c','e','/','m','e','s','s','a','g','e','s','/','d','e','v','i','c','e','b','o','u','n','d','/', '#'};
```
danewalton: Might be easier but I have these correct ones as `char[]` and then use the `assert_memory_equal` function to make sure they are equal.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -0,0 +1,244 @@
...
+  assert_true(
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
+
+  assert_true(
+      az_iot_hub_client_c2d_subscribe_topic_filter_get(&client, mqtt_sub_topic, &mqtt_sub_topic)
+      == AZ_OK);
+
+  az_span expected = az_span_init(
+    g_test_correct_subscribe_topic, 
+    sizeof(g_test_correct_subscribe_topic)/sizeof(g_test_correct_subscribe_topic[0]),
```
danewalton: you can use the `_az_COUNTOF()` macro for these

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
  */
 AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
+/**
+ * @brief az_span_find searches for `target` in `source`, returning an az_span if it finds it.
+ *
+ * @param[in] source The az_span containing the ASCII digits to be searched on.
```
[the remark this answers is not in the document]
ewertons: Fixed

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
ewertons: Changed

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
  */
 AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
+/**
+ * @brief az_span_find searches for `target` in `source`, returning an #az_span if it finds it.
```
JeffreyRichter: returning an #az_span within the source if it finds it.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ *
+ * @param[in] source The #az_span containing the ASCII digits to be searched on.
+ * @param[in] target The #az_span containing the token to be searched in `source`.
+ * @param[out] out_span The #az_span pointing to the first occurrence of `target` in `source`, if it is found.
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if `target` is found in `source`
+ *          #AZ_ERROR_EOF if `target` is not found in `source`
```
JeffreyRichter: AZ_ERROR_ITEM_NOT_FOUND may be a better error for not found. Maybe we should remove "ITEM" from this symbol and just call it AZ_ERROR_NOT_FOUND. Thoughts?

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ * is found within the span.
+ */
+AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out_span);
+
+/**
+ * @brief az_span_token_next is a string tokenizer for az_span.
+ *
+ * @param[in] source The #az_span containing the ASCII digits to be searched on.
```
JeffreyRichter: Why ASCII digits? This can just be bytes, right?

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ * @param[in] delimiter The #az_span containing the delimiter to "split" `source` into tokens.
+ * @param[out] out_token The #az_span pointing to the token delimited by the beginning of `source` 
+ *                       up to the first occurrence of `delimiter`, or the end of `source`
```
JeffreyRichter: I think you should add that the delimiter itself is not included in out_token

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+
+/**
+ * @brief az_span_token_next is a string tokenizer for az_span.
+ *
+ * @param[in] source The #az_span containing the ASCII digits to be searched on.
+ * @param[in] delimiter The #az_span containing the delimiter to "split" `source` into tokens.
+ * @param[out] out_token The #az_span pointing to the token delimited by the beginning of `source` 
+ *                       up to the first occurrence of `delimiter`, or the end of `source` 
+ *                       if `delimiter` is not found.
+ * @param[out] out_source The #az_span pointing to the next token in `source`, starting after the occurrence of `delimiter`.
```
JeffreyRichter: Don't you want this to be the remainder of the source (all bytes after the delimiter)? If so, I'd call it out_remainder.
This way, I can call this function in a loop passing in for source the previous result in out_remainder.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ *                        If the position after `delimiter` is the end of `source`, `out_source` is set to a NULL/empty #az_span.
```
JeffreyRichter: I don't think you should do this. If there are no more tokens, then don't return AZ_OK, return AZ_ERROR_ITEM_NOT_FOUND instead. This way, I can loop over all the tokens until I get AZ_ERROR_ITEM_NOT_FOUND. I shouldn't have to check for a special sentinel value in out_remainder.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,34 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ * @param[out] out_span The #az_span pointing to the first occurrence of `target` in `source`, if it is found.
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if `target` is found in `source`
+ *          #AZ_ERROR_EOF if `target` is not found in `source`
+ * is found within the span.
+ */
+AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out_span);
+
+/**
+ * @brief az_span_token_next is a string tokenizer for az_span.
```
JeffreyRichter: Consider renaming this to az_span_token because it finds a token.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,78 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+  AZ_PRECONDITION_VALID_SPAN(source, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(target, 1, false);
+  AZ_PRECONDITION_NOT_NULL(out_span);
+
+  uint8_t* source_ptr = az_span_ptr(source);
+  int32_t source_length = az_span_length(source);
+  uint8_t* target_ptr = az_span_ptr(target);
+  int32_t target_length = az_span_length(target);
+
+  for (int i = 0; i < source_length; i++)
```
JeffreyRichter: I'd still love to see some comments in here explaining how this code is trying to accomplish its task. Imagine that someone else has to maintain this and there may be a bug or they might need to add a feature to it. Briefly explain the algorithm you're using to solve this problem.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,78 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+      if (j == target_length)
+      {
+        *out_span = az_span_init(source_ptr + i, target_length, target_length);
+
+        return AZ_OK;
+      }
+    }
+  }
+
+  return AZ_ERROR_EOF;
```
JeffreyRichter: AZ_ERROR_ITEM_NOT_FOUND

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,78 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+}
+
+AZ_NODISCARD az_result az_span_token_next(az_span source, az_span delimiter, az_span* out_token, az_span* out_span)
+{
+  AZ_PRECONDITION_VALID_SPAN(source, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(delimiter, 1, false);
+  AZ_PRECONDITION_NOT_NULL(out_token);
+  AZ_PRECONDITION_NOT_NULL(out_span);
+
+  az_span instance;
```
JeffreyRichter: nit: I'd move this to just before the if statement. It is good practice to not allow uninitialized variables to live long in the code. That is declare them right next to initialization. Someone else working on this code may attempt to use instance between lines 175 & 180 resulting in unpredictable behavior. Declaring just before initialization decreases code brittleness.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,78 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+  uint8_t* source_ptr = az_span_ptr(source);
+  int32_t source_length = az_span_length(source);
+  int32_t source_capacity = az_span_capacity(source);
+
+  if (az_span_find(source, delimiter, &instance) == AZ_OK)
+  {
+    uint8_t* instance_ptr = az_span_ptr(instance);
+    int32_t instance_length = az_span_length(instance);
```
JeffreyRichter: Move this just before initializing *out_span. It is good code hygiene to declare variables immediately before their use.
If we deleted or rewrote how out_span got initialized, then it would be easy to forget about this line which would be dead code now.

sdk/core/core/inc/az_span.h
ewertons: Correct. Fixed.

sdk/core/core/inc/az_span.h
ewertons: That's absolutely correct. I'll add that.

sdk/core/core/inc/az_span.h
ewertons: That's exactly how this function works right now.
I'll change out_source to out_remainder.

JeffreyRichter: Your doc says az_span points to the next "token" in source but it's not the next token; it is the full remainder of the source (which may contain multiple more tokens in it). This is why I'm suggesting the parameter change and a doc change.

sdk/core/core/src/az_span.c
ewertons: Sure.

sdk/core/core/src/az_span.c
ewertons: We had issues in the past of compilers that do not handle well variables that are not defined on the top. The only example that comes to mind right now is Windows-CE (yeah, I know...).
But I'll follow your recommendation on this.

sdk/core/core/src/az_span.c
ewertons: Yes, much better. Changed.

sdk/core/core/inc/az_span.h
ewertons: Changed.

sdk/core/core/inc/az_span.h
ewertons: You mean rename az_span_token_next to az_span_token ?
Yeah, no problem. Done.

sdk/core/core/src/az_span.c
ewertons: No problem at all. Done.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
 
+/**
+ * @brief az_span_find searches for `target` in `source`, returning an #az_span within `source` if it finds it.
+ *
+ * @param[in] source The #az_span with the content to be searched on.
+ * @param[in] target The #az_span containing the token to be searched in `source`.
+ * @param[out] out_span The #az_span pointing to the first occurrence of `target` in `source`, if it is found.
+ * @return An #az_result value indicating the result of the operation.
+ *          #AZ_OK if `target` is found in `source`
+ *          #AZ_ERROR_ITEM_NOT_FOUND if `target` is not found in `source`
```
ahsonkhan: Is it really an "error" condition to not find the `target` in the `source`? To me, not finding it is normal part of querying the source span. Maybe we can drop `ERROR_` and just have `AZ_ITEM_NOT_FOUND`.

We should probably also document that, if not found `out_span` will be `NULL`.

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
  */
 AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
+/**
+ * @brief az_span_find searches for `target` in `source`, returning an #az_span within `source` if it finds it.
+ *
+ * @param[in] source The #az_span with the content to be searched on.
+ * @param[in] target The #az_span containing the token to be searched in `source`.
+ * @param[out] out_span The #az_span pointing to the first occurrence of `target` in `source`, if it is found.
```
ahsonkhan: How does using back ticks look like in the generated docs? Elsewhere I have seen folks use `\p` to refer to the parameter.

http://www.doxygen.nl/manual/commands.html#cmdp

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ */
+AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out_span);
+
+/**
+ * @brief az_span_token is a string tokenizer for az_span.
+ *
+ * @param[in] source The #az_span with the content to be searched on. It must be a non-empty #az_span.
+ * @param[in] delimiter The #az_span containing the delimiter to "split" `source` into tokens.  It must be a non-empty #az_span.
+ * @param[out] out_remainder The #az_span pointing to the remaining bytes in `source`, starting after the occurrence of `delimiter`.
+ *                        If the position after `delimiter` is the end of `source`, `out_remainder` is set to a NULL/empty #az_span.
```
ahsonkhan: nit: Remove `NULL`. Just state `empty #az_span`. A NULL span is a bit confusing since it's a struct.

Also, why is this line so deeply indented?
```suggestion
 *                        If the position after `delimiter` is the end of `source`, `out_remainder` is set to an empty #az_span.
```

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ * @return The #az_span pointing to the token delimited by the beginning of `source` up to the first occurrence 
+ *         of (but not including the) `delimiter`, or the end of `source` if `delimiter` is not found.
```
ahsonkhan:
```suggestion
 *         of (but not including the) `delimiter`, or the end of `source` if the `delimiter` is not found.
```

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
...
+ *         If `source` or `delimiter` is empty, AZ_SPAN_NULL is returned instead.
```
ahsonkhan: You stated above that both source and delimiter must be non-empty.

sdk/core/core/inc/az_span.h
ahsonkhan: What happens to `out_remainder` if delimiter is not found? Worth documenting.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+ *    - the loop has reached the end of `source` (and there are still remaing bytes of `target` to be checked).
+ */
+  AZ_PRECONDITION_VALID_SPAN(source, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(target, 1, false);
+  AZ_PRECONDITION_NOT_NULL(out_span);
+
+  uint8_t* source_ptr = az_span_ptr(source);
+  int32_t source_length = az_span_length(source);
+  uint8_t* target_ptr = az_span_ptr(target);
+  int32_t target_length = az_span_length(target);
```
ahsonkhan: We should consider adding "early outs" for things like target length is larger than source length.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+
+  // This loop traverses `source` position by position (step 1.)
+  for (int i = 0; i < source_length; i++)
+  {
+    // This is the check done in step 1. above.
+    if (source_ptr[i] == target_ptr[0])
```
ahsonkhan: Similarly, we may want an early out if the remaining source length is too small for the target.

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+  }
+
+  // If the function hasn't returned before, all positions 
+  // of `source` have been evaluated but `target` could not be found.
+  return AZ_ERROR_ITEM_NOT_FOUND;
+}
+
+AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder)
+{
+  AZ_PRECONDITION_VALID_SPAN(delimiter, 1, false);
```
ahsonkhan: Missing precondition on source (or update comment to allow for empty source).

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+    int32_t source_capacity = az_span_capacity(source);
+
+    az_span instance;
+    if (az_span_find(source, delimiter, &instance) == AZ_OK)
+    {
+      uint8_t* instance_ptr = az_span_ptr(instance);
+      int32_t instance_length = az_span_length(instance);
+
+      *out_remainder = az_span_init(
+        instance_ptr + instance_length,
```
ahsonkhan: Should this be `instance_ptr + delimiter length`?

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+      int32_t instance_length = az_span_length(instance);
+
+      *out_remainder = az_span_init(
+        instance_ptr + instance_length, 
+        (int32_t)(source_length - instance_length - (instance_ptr - source_ptr)),
+        (int32_t)(source_capacity - instance_length - (instance_ptr - source_ptr)));
+
+      return az_span_init(
+        source_ptr, 
+        (int32_t)(instance_ptr - source_ptr),
```
ahsonkhan: Is this returning what we documented?
```text
The #az_span pointing to the token delimited by the beginning of `source` up to the first occurrence of (but not including the) `delimiter`.
```

sdk/core/core/src/az_span.c
```diff
@@ -129,6 +129,106 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
...
+            break;
+        }
+      }
+ 
+      if (j == target_length)
+      {
+        // All bytes in `target` have been checked and matched the corresponding bytes in `source` (from the start point `i`),
+        // so this is indeed an instance of `target` in that position of `source` (step 4.).
+        // Here we create a span in `source` to map the occurrence of `target` in `source` (pardon the redundancy of words, done for explicit clarity).
+        *out_span = az_span_init(source_ptr + i, target_length, target_length);
```
ahsonkhan: I don't know if it makes sense for `out_span` to have `length == target_length`. Otherwise, this is essentially a clone of `target`.

I think it should basically be a slice of the source starting at the index the target contents are found (i.e. keep the  source length/capacity).

So, imagine this as a test case (where _ indicates empty capacity):
- source: `[some source abc to look through_ _ _]`
- target: `[abc]` OR [abc_ _] (because target capacity doesn't matter here, just the length)
- out_span: `[abc to look through_ _ _]` (i.e. slice of source from the index we found target to its end)

sdk/core/core/inc/az_span.h
```diff
@@ -260,6 +260,32 @@ AZ_NODISCARD az_result az_span_to_uint64(az_span span, uint64_t* out_number);
  */
 AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
 
+/**
+ * @brief az_span_find searches for `target` in `source`, returning an #az_span within `source` if it finds it.
+ *
+ * @param[in] source The #az_span with the content to be searched on.
+ * @param[in] target The #az_span containing the token to be searched in `source`.
```
ahsonkhan: We should mention that the target length is all that matters when searching within the source and not the target capacity.

sdk/core/core/inc/az_span.h
danewalton: It renders the text as `code`. So exactly like markdown. I personally prefer it over the doxygen tags because if you only have one work to render in `code` then \p works fine. But for multiple word you need to use `<tt>my text </tt>`. So backticks is consistent whether it be single or multi word.

sdk/core/core/inc/az_span.h
danewalton: `AZ_SPAN_NULL` is maybe what to have instead. Or “empty”

sdk/core/core/inc/az_span.h
JeffreyRichter: By definition, an error is when the function doesn't say it will do (as indicated by the function name). So, if az_span_find fails to find, then it is an error - by definition.
It is also a convention that if an error is returned, then the customer should ignore the out parameter. We can set it to null but the "contract" is that the customer should not look at it. If we doc that we set it to null, this is a contract which is in opposition to our convention.

JeffreyRichter: By definition, an error is when a function fails to do what it says it will do (as indicated by the function name). So, if "find" fails to "find" then it is an error.

Also, our convention is that the customer should not look at out parameters if the function returns an error. Documenting that, on error, out_span will be null ties us to a contract that should not matter to the customer and is in violation of our convention. 

We could change our convention to always set output parameters to 0/null on error but it is odd to say this if customers should not look at it if an error is returned.

sdk/core/core/inc/az_span.h
ahsonkhan: I see. However, `\p` is meant for parameters, which can't be multiple words. I think either is fine, but we should be consistent across the board.

The fact that the same file/type has two different formats just highlights this inconsistency to the reader.

Here are some questions that might help decide which way to  go:
1) `\p` barely formats the parameters differently from regular text (the difference is hard to tell). Is that a good thing or do we want the back ticks which highlight it more prominently?
2) What do folks reading docs generated from doxygen expect and are familiar with?
3) This isn't as important, but since we ship sources, what format do folks generally use in their own projects (i.e. what's common in C projects out there that do use doxygen)?

sdk/core/core/inc/az_span.h
ahsonkhan: I was looking at APIs like C++ string::find and it returns [npos](http://www.cplusplus.com/reference/string/string/npos/) when the target isn't found, rather than throwing an exception (similarly .NET's IndexOf returns -1 rather than throwing an error).
http://www.cplusplus.com/reference/string/string/find/

Since C doesn't have exceptions, I read "ERROR" containing az_result codes as capturing those exception conditions. For example, C++ string::push_back throws an exception, which if we were to implement here would be an error code:
http://www.cplusplus.com/reference/string/string/push_back/

JeffreyRichter: .NET doesn't follow its own guidelines and I can point you to many examples that do not follow the rules. IndexOf is just one example.  Stream's ReadByte is another example.
C/C++ examples are also not ideal. The problem is that the way of thinking has changed over the years and many examples exist which retain the old way of thinking for backward compatibility reasons. 

Similarly, lots of people say exceptions should only be used in exceptional circumstances or don't use exceptions for code flow but these also don't make sense. What is a clear definition for "exceptional circumstance" and, the whole point of exceptions is to alter code flow similar to goto, continue, break, return, and similar language constructs. Why is break or return OK to prematurely exit from code flow but an exception is not OK? 

Some people would say perf but this is language, compiler, runtime, and OS specific. Exceptions are not bad perf in many combinations of these. Furthermore, we should not focus on premature perf optimizations. So, the "modern" thinking is that an error is when a function can't do what it says it will do and, when this happens, the function reports error via an error or exception - this is an easy to follow rule. 

Also, contract violations do NOT fall into this category: a function is allowed to behave unpredictably in response to contract violations.

sdk/core/core/src/az_span.c
ewertons: delimiter_length is equal to instance_length.

PR #516, opened 2020-03-25
add twin topic support

danewalton:
Adds the topic get functionality for twin (pub and sub). Parsing responses will come later after #513.

### DIFF
--- a/sdk/iot/hub/inc/az_iot_hub_client.h
+++ b/sdk/iot/hub/inc/az_iot_hub_client.h
@@ -460,10 +460,6 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] request_id The request id.
- * @param[in] if_match_version Can be either "*" to overwrite or the twin version to limit the
- *                             reported properties patch to that version only. The version is
- *                             available from the JSON document or from
- *                             #az_iot_hub_client_twin_response::version.
  * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
  * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
  * @return #az_result
@@ -471,7 +467,6 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
 AZ_NODISCARD az_result az_iot_hub_client_twin_patch_publish_topic_get(
     az_iot_hub_client const* client,
     az_span request_id,
-    az_span if_match_version,
     az_span mqtt_topic,
     az_span* out_mqtt_topic);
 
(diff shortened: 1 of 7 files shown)


sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -460,18 +460,13 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] request_id The request id.
- * @param[in] if_match_version Can be either "*" to overwrite or the twin version to limit the
```
CIPop: Are these not part of the topic? Can they only be accessed via the payload?

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_twin.c
```diff
@@ -0,0 +1,437 @@
...
+
+static void test_az_iot_hub_client_twin_response_subscribe_topic_filter_get_null_client_fails(
+    void** state)
+{
+  (void)state;
+
+  uint8_t test_span_buffer[_az_COUNTOF(test_correct_twin_response_topic_filter) - 1];
+  az_span test_span = az_span_init(test_span_buffer, 0, _az_COUNTOF(test_span_buffer));
+
+  assert_precondition_checked(
```
CIPop: This name is confusing - maybe assert_precondition_failed? (Not part of this PR though it's the first time I see it.)

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_twin.c
```diff
@@ -0,0 +1,437 @@
...
+
+static void test_az_iot_hub_client_twin_response_subscribe_topic_filter_get_succeed(void** state)
+{
+  (void)state;
+
+  az_iot_hub_client client;
+  assert_int_equal(
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, NULL), AZ_OK);
+
+  uint8_t test_span_buffer[_az_COUNTOF(test_correct_twin_response_topic_filter) - 1];
```
CIPop: I'd actually initialize this with fixed space (128, 256) to:

1. demonstrate the way we need app-developers to think about the buffers.
2. check to see the buffer is not overrun

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: I confirmed with Ewerton this is not for MQTT or AMQP. When I did a search in the current SDK it only came up in HTTP.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_twin.c
danewalton: Yea we could have a separate PR for changing if we want for sure.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_twin.c
danewalton: I kinda see your point about demonstrating use case but I'm not so sure that unit tests is the place where people should go to understand how to use API's.
As for the buffer overrun part I can see that too.
I would like to issue another PR that checks the span length for all tests. @hihigupt had brought that up previously and I think it's an sdk test improvement. Is it okay to push to another PR focusing solely on that?

PR #519, opened 2020-03-26
Create az_iot_pnp client init & user APIs

jspaith:
Implement set of core functions for az_iot_pnp library:
* `az_result az_iot_pnp_client_init`
* `az_iot_pnp_client_get_user_name`

And inline a few more:
* `az_result az_iot_pnp_client_get_id`
* `az_result az_iot_pnp_client_get_sas_signature`
* `az_result az_iot_pnp_client_get_sas_password`

These functions for the most part (or entirely in inline case) build on top of az_iot_hub layer.


sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -0,0 +1,74 @@
...
+
+AZ_NODISCARD az_result az_iot_pnp_client_init(
+    az_iot_pnp_client* client,
+    az_span iot_hub_hostname,
+    az_span device_id,
+    az_span root_interface_name,
+    az_iot_pnp_client_options const* options)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(root_interface_name, 1, false);
```
danewalton: real real nit here: but I prefer having the preconditions in order of the function declaration (just makes it easier to eyeball and check going down the list). so in this case iot_hub_hostname -> device_id -> root-interface_name. Feel free to leave as is if that's too much

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
```diff
@@ -0,0 +1,284 @@
...
+
+static const char test_correct_pnp_user_name_with_user_agent[]
+    = "myiothub.azure-devices.net/my_device/?api-version=2018-06-30&" TEST_USER_AGENT
+      "&digital-twin-model-id=" TEST_ROOT_INTERFACE_NAME;
+
+#ifndef NO_PRECONDITION_CHECKING
+
+enable_precondition_check_tests()
+
+    static void test_az_iot_pnp_client_init_NULL_client_fails(void** state)
```
danewalton: I'd like to see if there's a way we can fix this. I've ran into this issue making my own tests and the clang format doesn't like that `enable_precondition_check_tests()` doesn't have a semicolon so the next function gets indented too much. Not for this PR but I'd like to hopefully find a solution for it.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
```diff
@@ -0,0 +1,284 @@
...
+  (void)state;
+
+  az_iot_pnp_client client;
+  assert_int_equal(
+      az_iot_pnp_client_init(
+          &client, test_device_hostname, test_device_id, test_root_interface_name, NULL),
+      AZ_OK);
+
+  assert_non_null(az_span_ptr(client._internal.iot_hub_client._internal.iot_hub_hostname));
+  assert_string_equal(TEST_ROOT_INTERFACE_NAME, az_span_ptr(client._internal.root_interface_name));
```
danewalton: `assert_string_equal` shouldn't be used with any of our client members due to the fact that we don't guarantee null termination. A span doesn't guarantee null termination and if someone initializes the client with a non-null terminated span then they can have issues here. [I have a PR here](https://github.com/Azure/azure-sdk-for-c/pull/514) which changes that for other tests.
TLDR: you can use `assert_memory_equal()` here instead.

danewalton: elsewhere as well. there probably shouldn't be any `assert_string_equal` in this whole test file.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
```diff
@@ -0,0 +1,284 @@
...
+{
+  (void)state;
+
+  az_iot_pnp_client client;
+  assert_int_equal(
+      az_iot_pnp_client_init(
+          &client, test_device_hostname, test_device_id, test_root_interface_name, NULL),
+      AZ_OK);
+
+  uint8_t test_span_buffer[sizeof(test_correct_pnp_user_name) - 1];
```
danewalton: just for consistency all of the `sizeof()` calls in this file can probably be switched to `_az_COUNTOF()`.
[Macro found here](https://github.com/Azure/azure-sdk-for-c/blob/35629203f7dab143a65e55760f10ef211f204e1d/sdk/core/core/inc/_az_cfg.h#L92)

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -0,0 +1,74 @@
...
+#include <_az_cfg.h>
+
+static const uint8_t pnp_client_param_separator = '&';
+static const uint8_t pnp_client_param_equals = '=';
+
+static const az_span pnp_model_id = AZ_SPAN_LITERAL_FROM_STR("digital-twin-model-id");
+
+AZ_NODISCARD az_iot_pnp_client_options az_iot_pnp_client_options_default()
+{
+  az_iot_pnp_client_options options;
```
JeffreyRichter: We usually use literal init syntax for this. Example: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/storage/blobs/src/az_storage_blobs_blob_client.c#L34

sdk/iot/pnp/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+# TODO: Follow up with team about the linking/including strategy for SDK-clients on top of az_core
+target_link_libraries(${TARGET_NAME} PRIVATE az_core az_iot_hub)
+
+add_library (az::iot::pnp ALIAS ${TARGET_NAME})
+
+# set coverage excluding for az_core and az_iot_hub. Don't show coverage outside pnp
+set(COV_EXCLUDE
+    ${CMAKE_SOURCE_DIR}/sdk/core/core/inc/*
+    ${CMAKE_SOURCE_DIR}/sdk/core/core/internal/*
+	${CMAKE_SOURCE_DIR}/sdk/iot/hub/inc/*
```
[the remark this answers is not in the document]
jspaith: Fixed.

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -0,0 +1,74 @@
...
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_user_name, pnp_model_id, &mqtt_user_name));
+  AZ_RETURN_IF_FAILED(
+      az_span_append_uint8(mqtt_user_name, pnp_client_param_equals, &mqtt_user_name));
+  AZ_RETURN_IF_FAILED(
+      az_span_append(mqtt_user_name, client->_internal.root_interface_name, &mqtt_user_name));
+
+  *out_mqtt_user_name = mqtt_user_name;
+
+  return AZ_OK;
+}
```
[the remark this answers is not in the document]
jspaith: @danewalton  - is newline at end of file a new convention?  Poking around what's in master now at like https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/hub/src/az_iot_hub_client_c2d.c & friends I don't see anything else doing this.

If that's the convention I'm fine following though a clang format rule would help here, too.

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <stdint.h>
```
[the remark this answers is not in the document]
jspaith: Fixed.

sdk/iot/pnp/src/az_iot_pnp_client.c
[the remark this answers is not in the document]
danewalton: I don't know if it is an official one but I know at least it's burned in my head from the Stuben days (thanks Marcos). Also if you give a review in a PR on Github it has a little red circle with an x and pointing to the last line meaning "you don't have a new line at the end". So maybe it's a github convention too.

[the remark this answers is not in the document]
jspaith: I played around with our current clang rules and they're fine with either no extra line or 1 extra line.  (If you put >1 closing line in, it'll collapse to 1 but that's not in play on this one).

A random sampling (!! - only 3 files) of other core SDK logic here shows nothing else adds an extra line, either.

If you feel strongly on this we should start a side conversation with broader team.  For now I'm going to follow existing convention.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
jspaith: Any reason we can't just stick a semicolon here :)?  I tried it and clang doesn't do the indent thing (good catch btw).

I have seen certain extra-extra-strict compilations from gcc, where if there was like an "int foo;;" then it'd complain on semicolon number 2.  So maybe that'd discourage an unneeded ; after the "{...}" block the macro expands to??

FWIW I'm not going to hold checkin on this, but will put in as-is since it's a bigger problem you're tracking.

sdk/iot/pnp/src/az_iot_pnp_client.c
jspaith: Completely agree - fixing.

sdk/iot/pnp/src/az_iot_pnp_client.c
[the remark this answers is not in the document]
danewalton: yea no problem at all obviously not blocking. we can bring it up another time

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
jspaith: Fixing.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
jspaith: Agreed - the as_string() tests aren't needed; just in my memcmp() tests.  Though I'm going to check the length of span is also set right just in case.  (Shouldn't matter if memory is equal, but I want to be safe.)  Telemetry tests could potentially add this check, too...

danewalton: yup funny enough exactly what I'm doing now. Modifying all the other tests to ensure:
1. Length of the span is correct.
2. Span capacity is the untouched
3. Content of the span is as it should
4. The span wasn't written past its length (ie set buffer to all `0xFF` and makes sure index of `length` is `0xFF`)

sdk/iot/pnp/src/az_iot_pnp_client.c
jspaith: Changed.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client.c
jspaith: My next question was on overflow!, happy you're on it.

A few thoughts:
* First, is there a tool that can help here so we can just get this either statically or run-time a la Valgrind.  (Valgrind I don't think will work since it ties into heap, not stack.  https://stackoverflow.com/questions/5813444/how-to-debug-stack-overwriting-errors-with-valgrind though leads to https://github.com/google/sanitizers/wiki/AddressSanitizer).

* Or if we want to be able to run on arbitrary setup with no dependencies, we should think about building up the 0xFF thingy as generic.  I'm totally hand-wavy here ,but something where you told the size of buffer and it created the bufferLen+1, put the 0xFF in end, but then when you did a span off it ignored the 0xFF as a valid byte?

danewalton: Well Chris mentioned having a generic number to use instead of having it based on the length of the bytes we are expecting. So in my update the spans are all 128.
Essentially I'm updating all the tests to have this at the "setup faze"
```c
  uint8_t test_span_buf[TEST_SPAN_BUFFER_SIZE];
  memset(test_span_buf, 0xFF, _az_COUNTOF(test_span_buf));
  az_span test_span = az_span_init(test_span_buf, 0, _az_COUNTOF(test_span_buf));
```
and then this for check:
```c
  assert_memory_equal(
      test_correct_user_name_with_module_id,
      az_span_ptr(test_span),
      sizeof(test_correct_user_name_with_module_id) - 1);
  assert_int_equal(
      az_span_length(test_span), _az_COUNTOF(test_correct_user_name_with_module_id) - 1);
  assert_int_equal(az_span_capacity(test_span), TEST_SPAN_BUFFER_SIZE);
  assert_int_equal(test_span_buf[az_span_length(test_span)], 0xFF);
```

PR #520, opened 2020-03-26
More fixes to documentation

gilbertw:


sdk/core/core/inc/az_span.h
```diff
@@ -431,8 +439,8 @@ AZ_INLINE void az_span_set(az_span destination, uint8_t fill)
 AZ_NODISCARD az_result az_span_copy(az_span destination, az_span source, az_span* out_span);
 
 /**
- * @brief az_span_copy_url_encode copies a URL in the source span to the destination span by
- * url-encoding the source span's characters
+ * @brief az_span_copy_url_encode Copies a URL in the source span to the destination span by
```
JeffreyRichter: Is this function actually used anywhere?
If not, we should delete this.
If we do, what I think would be more useful is an append (not copy) function that appends a span with url-encoding.

ahsonkhan: It's used in aad, in an internal `_az_span_append_with_url_encode` method, which is used a few times in building the body (`_az_aad_build_body`), which follows what you are suggesting @JeffreyRichter  :)
https://github.com/Azure/azure-sdk-for-c/blob/b33ca37648e75c14a02e94257d85af7371249f26/sdk/core/core/src/az_aad.c#L38-L39

So, it makes sense to move that from aad into span and make it public and replace this one. Let's address this separately from the docs update PR.

sdk/core/core/inc/az_span.h
```diff
@@ -118,12 +125,13 @@ AZ_NODISCARD az_span az_span_init(uint8_t* ptr, int32_t length, int32_t capacity
...
  * @return az_span An az_span over the byte buffer; length & capacity are set to the string's
- * length.
+ * length not including the \0 terminator.
  */
 AZ_NODISCARD az_span az_span_from_str(char* str);
 
 /**
- * @brief AZ_SPAN_LITERAL_FROM_BUFFER returns a literal az_span over an uninitialized byte buffer.
+ * @brief AZ_SPAN_LITERAL_FROM_BUFFER returns a literal az_span over a byte buffer.
+ * The length of the resulting az_span is set to 0.
```
ahsonkhan: What is capacity set to?

sdk/core/core/inc/az_span.h
```diff
@@ -35,8 +35,8 @@ typedef struct
   struct
   {
     uint8_t* ptr;
-    int32_t length;
-    int32_t capacity;
+    int32_t length; //< length must be >= 0
```
ahsonkhan: Add an extra backslash.

That's what is causing CI failures:
```text
/Users/runner/runners/2.165.2/work/1/s/sdk/core/core/inc/az_span.h:38:21: error: not a Doxygen trailing comment [-Werror,-Wdocumentation]
    int32_t length; //< length must be >= 0
                    ^~~
                    ///<

```

PR #526, opened 2020-03-27
Update property docs to clarify encoding

danewalton:
Updates the docs to clarify that props functionality will not handle encoding.

### DIFF
--- a/sdk/iot/hub/inc/az_iot_hub_client.h
+++ b/sdk/iot/hub/inc/az_iot_hub_client.h
@@ -185,8 +185,9 @@ typedef struct az_iot_hub_client_properties
  *       avoid using the previously mentioned characters.
  *
  * @param[in] properties The #az_iot_hub_client_properties to initialize
- * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted,
- *                   properties with the following format: {key}={value}&{key}={value}.
+ * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted
+ *                   (with above mentioned characters encoded if applicable) properties with the 
+ *                   following format: {key}={value}&{key}={value}.
  * @return #az_result
  */
 AZ_NODISCARD az_result
--- a/sdk/iot/hub/inc/az_iot_hub_client.h
+++ b/sdk/iot/hub/inc/az_iot_hub_client.h
@@ -176,10 +176,17 @@ typedef struct az_iot_hub_client_properties
 
 /**
  * @brief Initializes the Telemetry or C2D properties.
+ * 
+ * @note The properties functionality will not encode properties. In order to support
+ *       the following characters: 
+ *          `/ % # &`
+ *       the onus is on the user to properly url encode them (RFC 2396). Only these characters
+ *       would have to be encoded. If you would like to avoid the need to encode the keys/values, 
+ *       avoid using the previously mentioned characters.
  *
  * @param[in] properties The #az_iot_hub_client_properties to initialize
  * @param[in] buffer Can either be an empty #az_span or an #az_span containing properly formatted,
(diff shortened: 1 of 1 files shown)


sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -176,17 +176,32 @@ typedef struct az_iot_hub_client_properties
 
 /**
  * @brief Initializes the Telemetry or C2D properties.
+ * 
+ * @note The properties functionality will not encode properties. In order to support
```
ericwol-msft: >>The properties functionality will not encode properties.
This does not make sense, please reword

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -176,17 +176,32 @@ typedef struct az_iot_hub_client_properties
 
 /**
  * @brief Initializes the Telemetry or C2D properties.
+ * 
+ * @note The properties functionality will not encode properties. In order to support
+ *       the following characters: 
+ *          `/ % # &`
+ *       the onus is on the user to properly url encode them (RFC 2396). Only these characters
```
ericwol-msft: the onus is on the application to properly url encode them (RFC 2396). Only these characters

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -176,17 +176,32 @@ typedef struct az_iot_hub_client_properties
 
 /**
  * @brief Initializes the Telemetry or C2D properties.
+ * 
+ * @note The properties functionality will not encode properties. In order to support
+ *       the following characters: 
+ *          `/ % # &`
+ *       the onus is on the user to properly url encode them (RFC 2396). Only these characters
+ *       would have to be encoded. If you would like to avoid the need to encode the keys/values, 
+ *       avoid using the previously mentioned characters.
```
ericwol-msft: avoid using these characters in your property names and values.

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: updated

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -176,17 +176,32 @@ typedef struct az_iot_hub_client_properties
 
 /**
  * @brief Initializes the Telemetry or C2D properties.
+ * 
+ * @note The properties init API will not encode properties. In order to support
+ *       the following characters: 
+ *          `/ % # &`
+ *       the onus is on the application to properly url encode them (RFC 2396). Only these characters
```
CIPop: Please rephrase to say "The application must encode the following characters." - What you wrote fails to translate in my native language via automatic translators.
Also, the RFC you quote is obsolete and url-encode is incorrect.
You could either:
1. Say "Percent-encode only the following characters as described in RFC 3986, section 2.2"
2. Replace the following characters with the following encoding (and for each of the 4 characters give the replacement formula).

I prefer 2 because the rules in the RFC require other characters to be percent encoded depending on the location within the URI.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -176,17 +176,32 @@ typedef struct az_iot_hub_client_properties
...
  */
 AZ_NODISCARD az_result
 az_iot_hub_client_properties_init(az_iot_hub_client_properties* properties, az_span buffer);
 
 /**
  * @brief Appends a key-value property to the list of properties.
+ * 
+ * @note The properties append API will not encode properties. In order to support
+ *       the following characters:
+ *          `/ % # &`
```
CIPop: Same 2 comments as above.

PR #530, opened 2020-03-30
Added MQTT connection/message defines

ericwol-msft:


sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -19,6 +19,18 @@
 
 #include <_az_cfg_prefix.h>
 
+
+/**
+ * @brief Azure IoT service MQTT properties for telemetry publish messages.
```
danewalton: should there be a clarification here that these are MQTT bit fields instead of being our proprietary option values.

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -18,6 +18,17 @@
...
 #include <_az_cfg_prefix.h>
 
+/**
+ * @brief Azure IoT service MQTT connection properties.
+ *
+ */
+enum
+{
+  AZ_CLIENT_DEFAULT_MQTT_CONNECT_PORT = 8883,
+  AZ_CLIENT_DEFAULT_MQTT_CONNECT_CLEAN_SESSION = 0x20,
```
CIPop: Clean session is a boolean in most MQTT libraries. I'm not sure about the value which appears to be particular to an MQTT implementation.
(3.1.2.3 Connect Flags of the [MQTT Spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html)

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -18,6 +18,17 @@
 
 #include <_az_cfg_prefix.h>
 
+/**
+ * @brief Azure IoT service MQTT connection properties.
+ *
+ */
+enum
+{
+  AZ_CLIENT_DEFAULT_MQTT_CONNECT_PORT = 8883,
```
CIPop: Naming: I don't think this should be the "Default" port. Instead it could be the "port" since it cannot change.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -19,6 +19,18 @@
...
 
+
+/**
+ * @brief Azure IoT service MQTT bit field properties for telemetry publish messages.
+ *
+ */
+enum
+{
+  AZ_HUB_CLIENT_DEFAULT_MQTT_TELEMETRY_QOS = 0,
+  AZ_HUB_CLIENT_DEFAULT_MQTT_TELEMETRY_DUPLICATE = 0,
```
CIPop: I'm not sure about documenting the Duplicate =  0 given that is an MQTT implementation detail: first message must have DUP set to 0. A retry should change that to 1.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -19,6 +19,18 @@
...
 #include <_az_cfg_prefix.h>
 
+
+/**
+ * @brief Azure IoT service MQTT bit field properties for telemetry publish messages.
+ *
+ */
+enum
+{
+  AZ_HUB_CLIENT_DEFAULT_MQTT_TELEMETRY_QOS = 0,
```
CIPop: Telemetry QoS is 1, not 0. (At least according to https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-device-sdks)

PR #534, opened 2020-03-31
Add helpers to make testing result spans easier

jspaith:
Checking az_span for correctness today can be somewhat tedious.  Involves memcmp, length checks, and ideally a buffer overflow check as well.

Consolidate this into two helpers: 
* `az_span_for_test_init` - lightweight, test only wrapper for `az_span_init` that sets the buffer to 0xFF prior to return.
* `az_span_for_test_verify` checks that the span is legal, including against buffer overflow.

Enable on Azure IoT telemetry as basic validation of header itself.


sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,68 @@
...
+ */
+AZ_INLINE void az_span_for_test_verify(
+    az_span result_span,
+    const void* const buffer_expected,
+    int32_t length_expected,
+    int32_t capacity_expected)
+{
+  assert_int_equal(az_span_length(result_span), length_expected);
+  assert_int_equal(az_span_capacity(result_span), capacity_expected);
+  assert_memory_equal(az_span_ptr(result_span), buffer_expected, length_expected);
```
CIPop: Nit: memory equal displays something very ugly if these are not equal. In general, our tests work with strings and it would look much better if we had string/len comparison and UTF8 display of differences.

danewalton: yea i was going to suggest the same thing but it might be something that I might do in my own time to issue a PR to Cmocka to get string comparison's with lengths. Because since here we don't have guaranteed null termination :/

jspaith: I'm all for easier to read logs from tests.  But I don't know how to take this feedback with tools we have from cmocka currently and fact we can't assume null terminated.

I'm very open to ideas!, but in meantime plan on going ahead with this checkin.  Given the exsiting tests were also doing assert_memory_equal we're not any worse than before.

CIPop: There are a few ways:
1. Implement our own diff system and find out how to print using cmocka.
2. If the memory is _not_ equal, insert a NULL in the end and use the cmocka string assert.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,68 @@
...
+
+  for (int32_t i = az_span_length(result_span); i < az_span_capacity(result_span); i++)
+  {
+    assert_true(*(uint8_t*)(az_span_ptr(result_span) + i) == 0xff);
```
ahsonkhan, line 64: nit: Can we have the value we are comparing against be a parameter (wishing we had default parameters in C)? What happens if I want to use this helper method in tests where I didn't use the other one to get the init span? The data between length and capacity would be 0.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param[in] ptr The memory address of the 1st byte in the byte buffer.
+ * @param[in] length The number of bytes initialized in the byte buffer.
+ * @param[in] capacity The number of total bytes in the byte buffer.
+ * @return az_span The "view" over the byte buffer, with the buffer filled with 0xFF.
+ */
+AZ_NODISCARD AZ_INLINE az_span az_span_for_test_init(uint8_t* ptr, int32_t length, int32_t capacity)
+{
+  az_span new_span = az_span_init(ptr, length, capacity);
+  az_span_set(new_span, 0xff);
```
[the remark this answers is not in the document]
CIPop: Nit: Can we actually set this to 0xCC (Assembly `INT3`) instead? 
Would immediately stop in a debugger if the filler somehow gets executed on x86.

sdk/core/core/test/inc/az_test_span.h
jspaith: This is good feedback, but beyond scope of this PR.  By having code in a central place though we'll make it easier to innovate over this in time.  For now I'm going to punt this.

sdk/core/core/test/inc/az_test_span.h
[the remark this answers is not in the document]
jspaith: Yup!  Changing.

sdk/core/core/test/inc/az_test_span.h
jspaith: We can't rely on the initial buffer to be all 0's though, unless we force test to do this, since the real az_span_init() doesn't touch the buffer itself.  

If we're forcing test to change to do a memset(0) on their buffer, I'd just make them change to use the new `az_span_for_test_init` in the 1st place :).

ahsonkhan: That's a fair point.

> We can't rely on the initial buffer to be all 0's though

Are arrays in C zero-initialized by default?

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+#include <az_span.h>
+
+#include <stdio.h>
+
+#include <_az_cfg_prefix.h>
+
+
+/**
+ * @brief az_span_for_test_init returns a span over a byte buffer, directly invoking
+ * #az_span_init.  The buffer is initialized with 0xFF to help tests check against buffer overflow.
```
ahsonkhan, line 30: comment/source mismatch. Should be 0xCC now.

sdk/core/core/test/inc/az_test_span.h
jspaith: No.  You need to explicitly either do a memset() to 0 or else in the variable initialization time set it to = {0}.  

Also by putting it through the test_init function an added bonus is we can potential add additional logic to init routine in future and all tests using it would get it "free".

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+    int32_t length_expected,
+    int32_t capacity_expected)
+{
+  assert_int_equal(az_span_length(result_span), length_expected);
+  assert_int_equal(az_span_capacity(result_span), capacity_expected);
+  assert_memory_equal(az_span_ptr(result_span), (size_t)buffer_expected, (size_t)length_expected);
+
+  for (int32_t i = az_span_length(result_span); i < az_span_capacity(result_span); i++)
+  {
+    assert_true(*(uint8_t*)(az_span_ptr(result_span) + i) == 0xcc);
```
ahsonkhan, line 68: nit: Should we make 0xcc a constant/define in this file since it is reused twice in this file and we don't want it to get out of sync? Also, it creates a place for us to document in source why we chose 0xcc rather than something else.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+#include <cmocka.h>
+
+#include <az_result.h>
+#include <az_span.h>
+
+#include <stdio.h>
+
+#include <_az_cfg_prefix.h>
+
+
```
ahsonkhan, line 27: nit: Remove Extra white space.

sdk/core/core/test/inc/az_test_span.h
jspaith: Ugh - checkin, just missed this.  I will fix this up on coming checkin.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+#include <setjmp.h>
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <cmocka.h>
+
+#include <az_result.h>
+#include <az_span.h>
+
+#include <stdio.h>
```
ahsonkhan, line 23: nit: move this include with the group above (with stdint.h, etc.). Not sure why this one in particular was separated out.

Also, do we need it?

sdk/core/core/test/inc/az_test_span.h
jspaith: I'd pondered that.  The issue here is that because this is all inline and C doesn't have name spacing, I would be introducing this constant everywhere so I just wanted to do the dup.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+
+#ifndef _az_SPAN_TESTING_H
+#define _az_SPAN_TESTING_H
+
+#include <assert.h>
+#include <setjmp.h>
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <cmocka.h>
```
ahsonkhan, line 18: Are all these includes necessary? I don't see much use of them in this file.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+
+
+/**
+ * @brief az_span_for_test_init returns a span over a byte buffer, directly invoking
+ * #az_span_init.  The buffer is initialized with 0xFF to help tests check against buffer overflow.
+ *
+ * @param[in] ptr The memory address of the 1st byte in the byte buffer.
+ * @param[in] length The number of bytes initialized in the byte buffer.
+ * @param[in] capacity The number of total bytes in the byte buffer.
+ * @return az_span The "view" over the byte buffer, with the buffer filled with 0xFF.
```
ahsonkhan, line 35: Same comment mismatch.

sdk/core/core/test/inc/az_test_span.h
```diff
@@ -0,0 +1,74 @@
...
+    az_span result_span,
+    const void* const buffer_expected,
+    int32_t length_expected,
+    int32_t capacity_expected)
+{
+  assert_int_equal(az_span_length(result_span), length_expected);
+  assert_int_equal(az_span_capacity(result_span), capacity_expected);
+  assert_memory_equal(az_span_ptr(result_span), (size_t)buffer_expected, (size_t)length_expected);
+
+  for (int32_t i = az_span_length(result_span); i < az_span_capacity(result_span); i++)
```
ahsonkhan, line 66: Is it guaranteed that `az_span_ptr(result_span) + az_span_capacity(result_span)` will be 0?

sdk/core/core/test/inc/az_test_span.h
ahsonkhan: Makes sense. Can we still capture @CIPop feedback to make it 0xcc in a comment in one of the two places just so the rationale of why we picked this lives in the code? Otherwise, I would have just used 0xFF and it's good to know that 0xCC has more benefit.

PR #537, opened 2020-04-01
Add support for auto bumping package versions upon releases

damirault:
Adopted from https://github.com/Azure/azure-sdk-for-cpp/blob/master/eng/Update-PkgVersion.ps1
Update existing version files to reference 'preview' instead of 'dev' to match conventions in other languages.

Part of #481


sdk/core/core/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Release History
 
-## 1.0.0-dev.1 (Unreleased)
+## 1.0.0-preview.1 (Unreleased)
 
 * Testing. Validating automation.
```
danieljurek, line 5: Delete the text under the header so we don't confuse people when it's time for release.

sdk/storage/blobs/CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 # Release History
 
-## 1.0.0-dev.1 (Unreleased)
+## 1.0.0-preview.1 (Unreleased)
 
 * Testing. Validating automation.
```
danieljurek, line 5: Delete the text under the header so we don't confuse people when it's time for release.

sdk/core/core/CHANGELOG.md
```diff
@@ -1,5 +1,4 @@
 # Release History
 
-## 1.0.0-dev.1 (Unreleased)
+## 1.0.0-preview.1 (Unreleased)
```
ahsonkhan, line 3: Do we publish the "release" artifacts anywhere? And if so, do we need to go and delete the 1.0.0-dev.1 artifacts to avoid sorting issues? I would imagine dev.1 always appears "newer" than preview.1 if sorted alphabetically (or subsequent preview.2, ... preview.X releases).

damirault: I don't believe we have published anything yet. @danieljurek is there any old docs that need deleting?

ahsonkhan: Maybe https://github.com/Azure/azure-sdk-for-c/releases?
image

danieljurek: Just the release is published. Docs and index have not been published yet. Releases are sorted chronologically by commit date (not by tag) so these will drop off as newer releases come out.

PR #541, opened 2020-04-02
Redesign az_span_find

ewertons:
According to discussion with Azure SDK team, the new signature shall be:
int32_t az_span_find(az_span source, az_span target)
Following similar pattern used by other find() implementations in C++, C#.


sdk/core/core/src/az_span.c
```diff
@@ -129,7 +129,9 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number)
   return AZ_OK;
 }
 
-AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out_span)
+#define AZ_SPAN_FIND_TARGET_NOT_FOUND -1
```
JeffreyRichter, line 132: We always makes these enums, not defines.
But, do we want to make this public? If so, then the docs should use the symbol; not -1.
If not, then I'd just define this as a const int inside the az_span_find function.

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -250,42 +250,32 @@ static void az_span_find_beginning_success()
...
 }
 
 static void az_span_find_not_found_fail()
 {
   az_span span = AZ_SPAN_FROM_STR("abcdefgabcdefg");
   az_span target = AZ_SPAN_FROM_STR("abd");
-  az_span instance;
 
-  assert_true(az_span_find(span, target, &instance) == AZ_ERROR_ITEM_NOT_FOUND);
+  assert_int_equal(az_span_find(span, target), -1);
```
JeffreyRichter, line 278: Or this is the symbol. I think I'm tempted to just leave it as -1 and not define the enum symbol

danewalton: i think i agree with the -1 as a number. I believe that's common with other languages too

sdk/core/core/src/az_span.c
```diff
@@ -176,55 +177,41 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
   else
   {
-    uint8_t* source_ptr = az_span_ptr(source);
-    int32_t source_length = az_span_length(source);
-    int32_t source_capacity = az_span_capacity(source);
+    int32_t index = az_span_find(source, delimiter);
 
-    az_span instance;
-    if (az_span_find(source, delimiter, &instance) == AZ_OK)
+    if (index != -1)
```
ericwol-msft, line 204: if (index != AZ_SPAN_FIND_TARGET_NOT_FOUND)

sdk/core/core/src/az_span.c
ericwol-msft, line 132: Does this goes into the .h file?

sdk/core/core/test/cmocka/test_span.c
ewertons: We can do that. Let's keep it simple. I'll change it to "-1"

sdk/core/core/src/az_span.c
ewertons: Let's keep it simple then. I'll define a const inside az_span_find then.

sdk/core/core/src/az_span.c
ewertons: We are getting rid of that define.

sdk/core/core/src/az_span.c
ewertons: No, it will be removed. The public API will return -1, and that's what should be checked by developers.

ewertons: Done.

sdk/core/core/src/az_span.c
```diff
@@ -176,55 +175,43 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
 
   // If the function hasn't returned before, all positions 
   // of `source` have been evaluated but `target` could not be found.
-  return AZ_ERROR_ITEM_NOT_FOUND;
+  const int32_t TARGET_NOT_FOUND = -1;
+
+  return TARGET_NOT_FOUND;
 }
 
 AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder)
```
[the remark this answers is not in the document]
ewertons: If we don't have it in the core SDK, we will need to move it to the iot SDK. 
We need that functionality in IoT SDK.

sdk/core/core/src/az_span.c
```diff
@@ -146,21 +146,20 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
  */
   AZ_PRECONDITION_VALID_SPAN(source, 1, false);
```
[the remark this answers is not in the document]
ewertons: I'm fine with that. I'll make those changes.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: Sounds good. I will look at current IoT usages of this to understand the scenario :)

sdk/core/core/src/az_span.c
```diff
@@ -176,55 +175,43 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-      *out_remainder = az_span_init(
-        instance_ptr + instance_length, 
-        (int32_t)(source_length - instance_length - (instance_ptr - source_ptr)),
-        (int32_t)(source_capacity - instance_length - (instance_ptr - source_ptr)));
-
-      return az_span_init(
-        source_ptr, 
-        (int32_t)(instance_ptr - source_ptr), 
-        (int32_t)(instance_ptr - source_ptr));
+      *out_remainder = az_span_slice(source, index + az_span_length(delimiter), az_span_length(source));
```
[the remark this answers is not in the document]
danewalton: First off I don't think so. I thought it should be the capacity of the source too but that would essentially be an overflow. 

This might be worth a separate discussion actually. Why does slice allow to extend to the end of capacity? Shouldn't it be `az_span_length()`?

sdk/core/core/src/az_span.c
```diff
@@ -176,55 +175,43 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
+
+      return az_span_slice(source, 0, index);
     }
     else
     {
       *out_remainder = AZ_SPAN_NULL;
```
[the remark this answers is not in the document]
ewertons: I'm not sure if the idea of flipping will give the expected functionality we need.

The current design works like this:
The user can keep invoking az_span_token for as long as `source` is not empty (AZ_SPAN_NULL). 
If source is empty, there is not token to be found, so AZ_SPAN_NULL is returned. That is the result that the calling application uses to identify if it should stop calling az_span_token.

On the other hand, if I try to tokenize an array with a sequence that is not present in the array, the whole array is returned as token right now. That is the one point I'm not sure about yet. Maybe it should return empty... the more I'm thinking about it, more I think it should actually return AZ_SPAN_NULL if the delimiter is not in `source`. What do you think, @ahsonkhan ?

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ewertons: I would say the capacity should be preserved somehow, don't you think?
For the usage we have for az_span_token, only the ptr and length are relevant though.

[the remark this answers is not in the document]
danewalton: In this case I think it should be kept as `az_span_length()`. Example here:

```c
buf[50];
az_span example = az_span_init(buf, 0, sizeof(buf));
topic_get(&example);
//Results in az/iot/hub/az_iot_hub_blah/?key=value
//Length = 37 , Capacity = 50

//Issue it again with same span but no props
topic_get(&example);
//Results in "az/iot/hub/az_iot_hub_blah/"
//Length = 27 , Capacity = 50

az_span_find(&example, "blah/");
//Returns the index but the returned "remainder" will include 
// the old props if using capacity which we don't want.
```

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ewertons: Hey, sorry to walk back on this, but I thinking more about it... the logic discussed in the links relates to strings. I agree any string contains empty strings. But would any byte sequence contain an empty byte sequence (AZ_SPAN_NULL)?
I'm thinking about the value provided by the function. Maybe the results make more sense like
- az_span_find(empty, empty) = 0 // since both are empty, ok target is in source.
- az_span_find(non-empty, empty) = -1 // I'm not sure it is useful to return 0, since AZ_SPAN_NULL would not be part of a non-empty sequence.
- az_span_find(non-empty, non-empty) = n (where n >=0)
- az_span_find(empty, non-empty) = -1

What do you think?

If there is a preference to adhere to how c++ does, I'm fine. We will then have to handle checking if source or target on the calling code to achieve the desired behavior.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: > //Returns the index but the returned "remainder" will include 
// the old props if using capacity which we don't want.

I like the approach of reasoning through the behavior by looking at the example usage. Can you update your example to use `az_span_toke`n, though, because `find` now doesn't have a `remainder` and the caller  can slice however they wish (keep capacity, or keep length, depending on  what they want to do next with the span).

It is not clear that having the remainder capacity contain the previous data is a bad thing. It's the same underlying buffer, and you may want to append stuff to it (to overwrite whatever was there before).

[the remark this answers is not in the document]
danewalton: Thanks for the correction. Updated the code snippet above to use `token`

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: So, I experimented with a couple of things:
https://godbolt.org/z/sn6sZY

If you re-wrote the `az_iot_hub_client_c2d_received_topic_parse` method in terms of the new `az_span_find`, the method size will shrink (assuming all span methods are inlined for simplicity).
- gcc (-O1): 127 to 114 instructions
- clang (-O3): 108 to 84 instructions

Granted, it means a bit more work for you (the caller of `az_span_find` instead of `az_span_token`). I will leave it up to you to decide if its worth the benefit.

https://github.com/Azure/azure-sdk-for-c/blob/35629203f7dab143a65e55760f10ef211f204e1d/sdk/iot/hub/src/az_iot_hub_client_c2d.c#L39-L54

```C
AZ_NODISCARD az_result az_iot_hub_client_c2d_received_topic_parse(
    az_span received_topic,
    az_iot_hub_client_c2d_request* out_request)
{
  AZ_PRECONDITION_VALID_SPAN(received_topic, 1, false);
  AZ_PRECONDITION_NOT_NULL(out_request);

  az_span token;
  token = az_span_slice(received_topic, az_span_find(received_topic, c2d_topic_suffix) + az_span_length(c2d_topic_suffix), -1);
  token = az_span_slice(token, 0, az_span_find(token, c2d_topic_suffix));
  AZ_RETURN_IF_FAILED(az_iot_hub_client_properties_init(&out_request->properties, token));

  return AZ_OK;
}
```

One thing I am not sure about is validation/error checking in the topic parsing between the repeated calls to `az_span_token`. It doesn't look like we ever check if the target string was found or not and assume it will be. Is that expected or should we check the result between the calls to avoid passing in potentially `null` source or -1 for the index?

I was also thinking of mirroring the signature of `az_span_find` and have `az_span_token` return an `int32_t` as well (rather than `az-span`), but that didn't really give great results (not sure why yet, I would have thought doing less work would be good):
```C
AZ_NODISCARD int32_t az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
```

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: > If source is empty, there is not token to be found, so AZ_SPAN_NULL is returned. That is the result that the calling application uses to identify if it should stop calling az_span_token.

The same can be done with the remainder instead of with the returned span, so I don't see now that scenario helps us decide one way or another.

> That is the one point I'm not sure about yet. Maybe it should return empty... the more I'm thinking about it, more I think it should actually return AZ_SPAN_NULL if the delimiter is not in `source`.

In either case, one has to be null, and the other has to be the whole source span (either remainder is null, result is the input OR vice versa). I suppose what we pick depends on how you view the "search" happening and what your mental model is of what the remainder means.

I will have to think through a couple of usage patterns, to form an intuition of what I expect the right behavior to be and get back to you (if you have and can share sample code snippets of what we expect callers to do, that would help).

For now, to unblock this PR, I would be fine with either option, so feel free to pick what you think works best, @ewertons. **Leaving it as is works for me too (least code churn when we don't have a strong opinion one way or another).** And we can defer changing it to the future if we find some insight.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ewertons: Yes, you are absolutely correct, there is validation we need to add in that function.
We didn''t add yet because we are still defining how az_span_find and az_span_token will look like. But it will get added.
So can the function be implemented with az_span_find only, yes, of course it can.
We will verify about code size afterwards, thanks for running that test. I'm not sure we will be able to call the functions inline like that, since we will need to add that validation I spoke above.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: I think maintaining the capacity here makes most sense to me (so if I were to pick one for this PR, I would chose that). Rationale: TODO :)

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ewertons: Would you like me to move az_span_token out of the core SDK for now? Then if its use is proven in the IoT SDK, we can try bringing it back to the core SDK.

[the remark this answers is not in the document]
JeffreyRichter: Yes

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
danewalton: To make sure we're on the same page here, the change should NOT be made. 
It should stay as 
```c
*out_remainder = az_span_slice(
    source, index + az_span_length(delimiter), az_span_length(source));
```

The slice API already maintains the capacity here:
https://github.com/Azure/azure-sdk-for-c/blob/88baf33becd27795facc78457ef0e2f070acbf96/sdk/core/core/src/az_span.c#L42-L45

Specifically the last param to `init` being `capacity - start_index`

sdk/iot/core/src/az_iot_core.c
```diff
@@ -10,3 +10,30 @@
...
+  {
+    return AZ_SPAN_NULL;
+  }
+  else
+  {
+    int32_t index = az_span_find(source, delimiter);
+
+    if (index != -1)
+    {
+      *out_remainder = az_span_slice(source, index + az_span_length(delimiter), az_span_length(source));
```
danewalton, line 28: in the docs it says:
> If the position after `delimiter` is the end of `source`, `out_remainder` is set to a NULL/empty #az_span.

Is that true here? Shouldn't there be a check for `index + length(delim) == length(source)`

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
  */
-  AZ_PRECONDITION_VALID_SPAN(source, 1, false);
-  AZ_PRECONDITION_VALID_SPAN(target, 1, false);
-  AZ_PRECONDITION_NOT_NULL(out_span);
 
-  uint8_t* source_ptr = az_span_ptr(source);
   int32_t source_length = az_span_length(source);
-  uint8_t* target_ptr = az_span_ptr(target);
   int32_t target_length = az_span_length(target);
+  const int32_t TARGET_NOT_FOUND = -1;
```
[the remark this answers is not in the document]
danewalton: https://azure.github.io/azure-sdk/clang_design.html#functions
Yea bottom of this section. Third bullet point.

sdk/iot/core/tests/cmocka/test_az_iot_core.c
```diff
@@ -0,0 +1,81 @@
...
+  // token: ""
+  token = az_span_token(span, delim, &out_span);
+  assert_non_null(az_span_ptr(token));
+  assert_true(az_span_length(token) == 0);
+  assert_true(az_span_ptr(out_span) == (az_span_ptr(span) + az_span_length(delim)));
+  assert_true(az_span_length(out_span) == (az_span_length(span) - az_span_length(delim)));
+  assert_true(az_span_capacity(out_span) == (az_span_capacity(span) - az_span_capacity(delim)));
+
+  // token: "defg" (span+3)
+  span = out_span;
```
ahsonkhan, line 36: The comment seems out of place.
Isn't `span = out_span = defgabcdefgabcdefg` here (i.e. remainder)?

sdk/iot/core/tests/cmocka/test_az_iot_core.c
```diff
@@ -0,0 +1,81 @@
...
+
+  token = az_span_token(span, delim, &out_span);
+  assert_true(az_span_ptr(token) == az_span_ptr(span));
+  assert_true(az_span_length(token) == 4);
+  assert_true(az_span_capacity(token) == az_span_capacity(span));
+  assert_true(az_span_ptr(out_span) == (az_span_ptr(span) + az_span_length(token) + az_span_length(delim)));
+  assert_true(az_span_length(out_span) == (az_span_length(span) - az_span_length(token) - az_span_length(delim)));
+  assert_true(az_span_capacity(out_span) == (az_span_capacity(span) - az_span_length(token) - az_span_length(delim)));
+
+  // token: "defg" (span+10)
```
ahsonkhan, line 46: Same here, should this be:
`span  = out_span = defgabcdefg`

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -93,6 +93,22 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
...
+ * @param[in] source The #az_span with the content to be searched on. It must be a non-empty
+ * #az_span.
+ * @param[in] delimiter The #az_span containing the delimiter to "split" `source` into tokens.  It
+ * must be a non-empty #az_span.
+ * @param[out] out_remainder The #az_span pointing to the remaining bytes in `source`, starting
+ * after the occurrence of `delimiter`. If the position after `delimiter` is the end of `source`,
+ * `out_remainder` is set to an empty #az_span.
+ * @return The #az_span pointing to the token delimited by the beginning of `source` up to the first
+ * occurrence of (but not including the) `delimiter`, or the end of `source` if `delimiter` is not
+ * found. If `source` or `delimiter` is empty, AZ_SPAN_NULL is returned instead.
```
ahsonkhan, line 108: Comment says `delimiter` can be empty, but `AZ_PRECONDITION_VALID_SPAN` says it must be at least length 1. Which is it?

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
   {
     uint8_t* source_ptr = az_span_ptr(source);
-    int32_t source_length = az_span_length(source);
-    int32_t source_capacity = az_span_capacity(source);
+    uint8_t* target_ptr = az_span_ptr(target);
 
-    az_span instance;
-    if (az_span_find(source, delimiter, &instance) == AZ_OK)
+    // This loop traverses `source` position by position (step 1.)
+    for (int32_t i = 0; i < source_length; i++)
```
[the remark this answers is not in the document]
ewertons: instead of having an extra memcpy at the end, I will iterate up to this:
  for (int32_t i = 0; i <= (source_length - target_length + 1); i++)

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-  return AZ_ERROR_ITEM_NOT_FOUND;
-}
-
-AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder)
-{
-  AZ_PRECONDITION_VALID_SPAN(delimiter, 1, false);
-  AZ_PRECONDITION_NOT_NULL(out_remainder);
-
-  if (az_span_is_content_equal(source, AZ_SPAN_NULL))
+  else if (source_length == 0)
```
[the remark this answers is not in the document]
ewertons: This is a good improvement. Thanks.

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-    {
-      *out_remainder = AZ_SPAN_NULL;
+      // This is the check done in step 1. above.
+      if (source_ptr[i] == target_ptr[0])
+      {
+        // The condition in step 2. has been satisfied.
+        int32_t j;
+        // This is the loop defined in step 3.
+        // The loop must be broken if it reaches the ends of `target` (step 3.) OR `source` (step 5.).
+        for (j = 1; j < target_length && (i + j) < source_length; j++)
```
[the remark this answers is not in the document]
ewertons: Correct. The list you put below shall cover it.

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-  AZ_PRECONDITION_VALID_SPAN(delimiter, 1, false);
-  AZ_PRECONDITION_NOT_NULL(out_remainder);
-
-  if (az_span_is_content_equal(source, AZ_SPAN_NULL))
+  else if (source_length == 0) 
   {
-    return AZ_SPAN_NULL;
+    return target_not_found;
   }
   else
```
vhvb1989, line 160: what about another condition like 
```c
else if (target_length > source_length) {
  return not_found; // target is greater than source. Can't be found within
}
```

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-    else
-    {
-      *out_remainder = AZ_SPAN_NULL;
+      // This is the check done in step 1. above.
+      if (source_ptr[i] == target_ptr[0])
+      {
+        // The condition in step 2. has been satisfied.
+        int32_t j;
+        // This is the loop defined in step 3.
+        // The loop must be broken if it reaches the ends of `target` (step 3.) OR `source` (step 5.).
```
vhvb1989, line 174: if you changed the above loop to loop until source_length - target_length, then we no longer need to check if we have reached the end of source here and it is enough to loop all target_length being safety

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -250,97 +250,42 @@ static void az_span_find_beginning_success()
 {
   az_span span = AZ_SPAN_FROM_STR("abcdefgabcdefg");
   az_span target = AZ_SPAN_FROM_STR("abc");
-  az_span instance;
 
-  assert_return_code(az_span_find(span, target, &instance), AZ_OK);
-  assert_true(az_span_length(target) == az_span_length(instance));
-  assert_true(az_span_ptr(span) == az_span_ptr(instance));
+  assert_int_equal(az_span_find(span, target), 0);
```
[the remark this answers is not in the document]
ewertons: Case 3 is the same as 2. Not added.
Case 4 is a subcase already covered by 2, but added that one.
These cases would apply if we did black-box testing, but added 7, 8 as well.
Others have been added.

sdk/iot/core/tests/cmocka/test_az_iot_core.c
ewertons: Not really, the comment defines that this block is expecting the token "defg". It is not about the span or out_span. We can talk offline to clarify.

sdk/iot/core/tests/cmocka/test_az_iot_core.c
ewertons: Same as above.

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-        (int32_t)(instance_ptr - source_ptr), 
-        (int32_t)(instance_ptr - source_ptr));
-    }
-    else
-    {
-      *out_remainder = AZ_SPAN_NULL;
+      // This is the check done in step 1. above.
+      if (source_ptr[i] == target_ptr[0])
+      {
+        // The condition in step 2. has been satisfied.
```
vhvb1989, line 171: what about calling az_span_is_equal(...) here, instead of below loop? 
We even have an az_span_is_equal_ignore_case()... (or something like that) in case we don't care about lower or upper case.

sdk/iot/core/inc/az_iot_core.h
ewertons: The return should be "...If `source` is empty, AZ_SPAN_NULL...".
Good catch.

sdk/core/core/src/az_span.c
```diff
@@ -144,89 +144,57 @@ AZ_NODISCARD az_result az_span_find(az_span source, az_span target, az_span* out
...
-  return AZ_ERROR_ITEM_NOT_FOUND;
-}
-
-AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder)
-{
-  AZ_PRECONDITION_VALID_SPAN(delimiter, 1, false);
-  AZ_PRECONDITION_NOT_NULL(out_remainder);
-
-  if (az_span_is_content_equal(source, AZ_SPAN_NULL))
+  else if (source_length == 0 || source_length < target_length)
```
ahsonkhan, line 156: Sorry, I just realized these conditions can be merged given target_length is greater than 0 at this point (because of the check above). And if `source_length == 0` when `target_length > 0`, `source_length < target_length` is already true.

Please validate, but I believe this should work:
```suggestion
  else if (source_length < target_length) 
```

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -288,6 +296,56 @@ static void az_span_find_error_cases_fail()
...
+
+static void az_span_find_target_overlap_continuation_of_source_fails()
+{
+  az_span span = AZ_SPAN_FROM_STR("abcd");
+  az_span target = AZ_SPAN_FROM_STR("cde");
+
+  assert_int_equal(az_span_find(span, target), -1);
+}
+
+static void az_span_find_target_more_chars_than_preffix_of_source_fails()
```
ahsonkhan, line 315: nit: typo
```suggestion
static void az_span_find_target_more_chars_than_prefix_of_source_fails()
```

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -288,6 +296,56 @@ static void az_span_find_error_cases_fail()
...
+{
+  az_span span = AZ_SPAN_FROM_STR("abcd");
+  az_span target = AZ_SPAN_FROM_STR("zab");
+
+  assert_int_equal(az_span_find(span, target), -1);
+}
+
+static void az_span_find_overlaping_target_success()
```
ahsonkhan, line 323: nit: Since we are doing overlapping tests, can we add one where the source and target both come as a slice of a larger buffer and that the find fails (returns -1)?

```c
  az_span span = AZ_SPAN_FROM_STR("abcdefghij");
  az_span source = az_span_slice(span, 1, 4);
  az_span target = az_span_slice(span, 6, 9);
  assert_int_equal(az_span_find(source, target), -1);
  assert_int_equal(az_span_find(source, az_span_slice(span, 1, 5)), -1);
  assert_int_equal(az_span_find(source, az_span_slice(span, 2, 4)), -1);
```

Also typo:
```suggestion
static void az_span_find_overlapping_target_success()
```

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -288,6 +296,56 @@ static void az_span_find_error_cases_fail()
...
+
+  assert_int_equal(az_span_find(span, target), 4);
+}
+
+static void az_span_find_capacity_checks_success()
+{
+  uint8_t* buffer = (uint8_t*)"aaaa";
+
+  assert_int_equal(az_span_find(az_span_init(buffer, 2, 4), az_span_init(buffer, 2, 3)), 0);
+  assert_int_equal(az_span_find(az_span_init(buffer, 2, 3), az_span_init(buffer, 2, 4)), 0);
```
ahsonkhan, line 344: Hmm... why is this returning 0 (instead of -1)? What am I getting wrong?
source: "a"
target: "aa"

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -288,6 +296,56 @@ static void az_span_find_error_cases_fail()
...
+  assert_int_equal(az_span_find(az_span_init(buffer, 2, 3), az_span_init(buffer, 0, 2)), 0);
```
ahsonkhan, line 345: Same issue here, I think.

sdk/core/core/src/az_span.c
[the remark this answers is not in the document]
ahsonkhan: I was going to mention that the bounds condition seemed wrong, or at least checking one too many bytes (`i <= (source_length - target_length + 1)`), but what you have now makes sense :)

sdk/core/core/src/az_span.c
ewertons: That is covered already. (line 156)

sdk/core/core/src/az_span.c
ewertons: Yes that is covered already as well. Did you get the latest iteration of the PR?

sdk/core/core/src/az_span.c
ewertons: I would say it's not as performant, because we only need to check one uint8_t at a time instead of the whole `target` every time with memcpy.

sdk/core/core/src/az_span.c
ewertons: Absolutely correct. Fixed.

sdk/core/core/test/cmocka/test_span.c
ewertons: Fixed.

sdk/core/core/test/cmocka/test_span.c
ewertons: Yes, added. 
Note that "assert_int_equal(az_span_find(source, az_span_slice(span, 2, 4)), -1);" is not true, the `target` would be contained in `source`. The correct would be to expect return equal 1.

And typo fixed.

sdk/core/core/test/cmocka/test_span.c
ewertons: Az_span_init is: az_span_init(ptr, length, capacity).

So az_span_init(buffer, 2, 3) will produce "aa" (source), with capacity of 3.
And az_span_init(buffer, 2, 4) will produce "aa" (target), with capacity of 4.
So target will still be contained in source, at position 0 of source.

sdk/core/core/test/cmocka/test_span.c
ewertons: Az_span_init is: az_span_init(ptr, length, capacity).

So az_span_init(buffer, 2, 3) will produce "aa" (source), with capacity of 3.
And az_span_init(buffer, 0, 2) will produce "" (target), with capacity of 2.
So target will still be contained in source, at position 0 of source, according to your comment [here](https://github.com/Azure/azure-sdk-for-c/pull/541#discussion_r402583720):

"Here is what I expect the results to be:
az_span_find(empty, empty) = 0
**az_span_find(non-empty, empty) = 0**
az_span_find(non-empty, non-empty) = 0
az_span_find(empty, non-empty) = -1"

sdk/core/core/test/cmocka/test_span.c
ahsonkhan: Oh right, lol, didn't realize you were calling `az_span_init` (and not slice). Ty :)

sdk/core/core/test/cmocka/test_span.c
ahsonkhan: Yep, resolving.

sdk/core/core/test/cmocka/test_span.c
ahsonkhan: > Note that "assert_int_equal(az_span_find(source, az_span_slice(span, 2, 4)), -1);" is not true, the `target` would be contained in `source`.

You are right, that's what I meant.

sdk/core/core/test/cmocka/test_span.c
```diff
@@ -250,97 +250,110 @@ static void az_span_find_beginning_success()
...
+  assert_int_equal(az_span_find(az_span_init(buffer, 0, 2), az_span_init(buffer, 0, 2)), 0);
+}
+
+static void az_span_find_overlapping_checks_success()
```
ahsonkhan, line 349: nit: Should we merge this test with `az_span_find_overlapping_target_success`. They are very similar in intent.

Also, given we are now also testing the "failure" condition of returning -1, maybe rename the test with the merge and remove `success` from it.

PR #543, opened 2020-04-02
Add PnP telemetry

jspaith:
Add implementation for `az_iot_pnp_client_telemetry_get_publish_topic` along with unit tests.


sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
```diff
@@ -0,0 +1,84 @@
...
+#include <_az_cfg.h>
+
+static const uint8_t pnp_telemetry_param_separator = '&';
+static const uint8_t pnp_telemetry_param_equals = '=';
+
+static const az_span pnp_telemetry_component_name_param = AZ_SPAN_LITERAL_FROM_STR("%24.ifname");
+static const az_span pnp_telemetry_content_type_param = AZ_SPAN_LITERAL_FROM_STR("%24.ct");
+static const az_span pnp_telemetry_content_encoding_param = AZ_SPAN_LITERAL_FROM_STR("%24.ce");
+
+static az_result az_add_telemetry_property(
```
danewalton, line 21: I think there is guidance against prefixing static functions with `az_`. Only the public ones should.
https://azure.github.io/azure-sdk/clang_design.html#functions

sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
```diff
@@ -0,0 +1,84 @@
...
+  AZ_PRECONDITION_VALID_SPAN(component_name, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
+  AZ_PRECONDITION_IS_NULL(reserved);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  AZ_RETURN_IF_FAILED(az_iot_hub_client_telemetry_publish_topic_get(
+      &client->_internal.iot_hub_client, NULL, mqtt_topic, &mqtt_topic));
+
+  AZ_RETURN_IF_FAILED(az_add_telemetry_property(
+      mqtt_topic, pnp_telemetry_component_name_param, component_name, false, &mqtt_topic));
```
danewalton, line 59: Correct me if I am wrong here, but setting the `separator` boolean here to false assumes that the `telemetry_public_topic_get()` API didn't append any properties correct? Because if it did and this prop was added, wouldn't there would be a missing `&`?

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client_telemetry.c
```diff
@@ -0,0 +1,367 @@
...
+
+#define TEST_DEVICE_ID_STR "my_device"
+#define TEST_DEVICE_HOSTNAME_STR "myiothub.azure-devices.net"
+#define TEST_ROOT_INTERFACE_NAME "my_root_interface_name"
+#define TEST_CONTENT_TYPE "my_content_type"
+#define TEST_CONTENT_ENCODING "my_content_encoding"
+#define TEST_COMPONENT_NAME "my_component_name"
+
+static const az_span test_device_id = AZ_SPAN_LITERAL_FROM_STR("my_device");
+static const az_span test_device_hostname = AZ_SPAN_LITERAL_FROM_STR("myiothub.azure-devices.net");
```
danewalton, line 29: Nit: any reason why you didn't use the macro's here?

sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
jspaith: Fixed.

sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
jspaith: Yes.  Because I'm passing `properties` as NULL in my call to the topic_get, it's not going to append anything.  I have to make this assumption to get the topic built up correctly.

If on the off-shot you added additional properties for whatever reason (which I don't think you can given the protocol), at a minimum the UT for PnP telemetry would start failing and at that point we'd make required changes.

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client_telemetry.c
jspaith: There is indeed a reason.  Which is I didn't know about them :).  Could you point me to macros you had in mind?

danewalton: Oh I meant the macro's for the strings. All the other `test` spans are using the macro's from above. Like `TEST_ROOT_INTERFACE`, `TEST_CONTENT_TYPE`, etc. So in these two cases it would be to use `TEST_DEVICE_ID_STR` and `TEST_DEVICE_HOSTNAME_STR`

sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
danewalton: From the `pnp_init()`: 
https://github.com/Azure/azure-sdk-for-c/blob/59d2c6ca8e7ffb10999485bd4bf40992686b817f/sdk/iot/pnp/src/az_iot_pnp_client.c#L42-L46

I think if the `user_agent` is appended that could screw this up?

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client_telemetry.c
jspaith: Yeah, why wasn't I consistent there?  Fixing.

sdk/iot/pnp/src/az_iot_pnp_client_telemetry.c
jspaith: Great catch!  By the way, @danewalton the IoTHub telemetry tests are also missing a test for `user_agent`.

danewalton: I think IIRC telemetry doesn't have it. Oh wait yea it's only on the connect that the user agent is put there. Let me find the PR in which we removed it.

danewalton: https://github.com/Azure/azure-sdk-for-c/pull/487
So in this case I think you're actually okay. My apologies I forgot this was removed in telemetry

jspaith: No need to apologize almost got me, too.  Closing out conversation based on this...

sdk/iot/pnp/tests/cmocka/test_az_iot_pnp_client_telemetry.c
```diff
@@ -0,0 +1,367 @@
...
+
+  assert_int_equal(
+      az_iot_pnp_client_telemetry_get_publish_topic(
+          &client, test_component_name, mqtt_topic, NULL, &mqtt_topic),
+      AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY);
+}
+
+static void
+test_az_iot_pnp_client_telemetry_publish_topic_get_content_encoding_with_small_buffer_fails(
+    void** state)
```
danewalton, line 292: just as a note: going forward each function doesn't need this `void** state`. Leaving it empty is valid for a variadic parameter. So future tests can just be `..._fails()` and can remove the `(void)state`. For now you can leave it because I will issue a PR removing all of them with regex.

PR #546, opened 2020-04-03
remove state from ut test functions

danewalton:


ericwol-msft:
I think you need to also fix the PNP unit tests as well

ericwol-msft:
John checked these in at around 3pm

danewalton:
> 
> 
> John checked these in at around 3pm

Those are updated as well. I'll have to look into if the Windows builds can be fixed but they might not let this go in.

Issue #545, opened 2020-04-03
Add Environment helpers


gilbertw:
Add helpers for using spans with environment variables

ahsonkhan:
There isn't a simple portable way to get environment variables. That makes it difficult to provide support for this within the SDK.

Within samples, folks can call `getenv` and copy it into their own buffer to avoid concerns with overriding data. Here's an example of that:
https://github.com/Azure/azure-sdk-for-c/blob/6a966901e467747d66e5705a3e25848b9ef45be8/sdk/samples/iot/hub/src/paho_iot_hub_telemetry_example.c#L191-L210

PR #547, opened 2020-04-04
IoT Provisioning client send

CIPop:
- Implementation for az_iot_provisioning_client
- Implementation for az_iot_provisioning_client_register PUB functions
- Some UTs for the added implementations
- Adding Paho-MQTT sample, prerequisites and build infrastructure

### DIFF
--- a/sdk/core/core/inc/_az_cfg.h
+++ b/sdk/core/core/inc/_az_cfg.h
@@ -91,4 +91,8 @@
 // Get the number of elements in an array
 #define _az_COUNTOF(array) (sizeof(array) / sizeof(array[0]))
 
+#ifndef UNUSED
+#define UNUSED(x) (void)(x)
+#endif
+
 #endif // _az_CFG_H
--- a/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
+++ b/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
@@ -77,8 +77,8 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
  * @brief Gets the MQTT user name.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT username.
- * @param[out] out_mqtt_user_name The output #az_span containing the MQTT username.
+ * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT user name.
+ * @param[out] out_mqtt_user_name The output #az_span containing the MQTT user name.
  * @return #az_result
  */
 AZ_NODISCARD az_result az_iot_provisioning_client_user_name_get(
@@ -247,7 +247,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_publish_topic_get(
  * @note The payload of the MQTT publish message should be empty.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] register_response The received response.
+ * @param[in] register_response The received #az_iot_provisioning_client_register_response response.
  * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
  * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
  * @return #az_result
--- a/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
+++ b/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
@@ -10,9 +10,9 @@
 #ifndef _az_IOT_PROVISIONING_CLIENT_H
 #define _az_IOT_PROVISIONING_CLIENT_H
 
+#include <az_iot_core.h>
 #include <az_result.h>
 #include <az_span.h>
-#include <az_iot_core.h>
 
 #include <stdbool.h>
 #include <stdint.h>
@@ -53,7 +53,7 @@ typedef struct az_iot_provisioning_client
  *
  * @return #az_iot_provisioning_client_options.
  */
-az_iot_provisioning_client_options az_iot_provisioning_client_options_default();
+AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_options_default();
 
 /**
  * @brief Initializes an Azure IoT Provisioning Client.
@@ -198,8 +198,8 @@ typedef struct az_iot_provisioning_client_registration_state
 typedef struct az_iot_provisioning_client_register_response
 {
   az_iot_status status; /**< The current request status.
-                                             * @note The response for the register operation is
-                                             * available through #registration_state only. */
+                         * @note The response for the register operation is
+                         * available through #registration_state only. */
   az_span operation_id; /**< Operation ID of the register operation. */
   az_span registration_state; /**< An #az_span containing the state of the register operation.
                                * @details This can be one of the following: 'unassigned',
@@ -247,15 +247,14 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_publish_topic_get(
  * @note The payload of the MQTT publish message should be empty.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] request_id The request id. Must match a received
- *                       #az_iot_provisioning_client_register_response request_id.
+ * @param[in] register_response The received response.
  * @param[in] mqtt_topic An empty #az_span with sufficient capacity to hold the MQTT topic.
  * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
  * @return #az_result
  */
 AZ_NODISCARD az_result az_iot_provisioning_client_get_operation_status_publish_topic_get(
     az_iot_provisioning_client const* client,
-    az_iot_provisioning_client_register_response register_response,
+    az_iot_provisioning_client_register_response const* register_response,
     az_span mqtt_topic,
     az_span* out_mqtt_topic);
 
--- a/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
+++ b/sdk/iot/provisioning/inc/az_iot_provisioning_client.h
@@ -19,6 +19,8 @@
 
 #include <_az_cfg_prefix.h>
 
+#define AZ_IOT_PROVISIONING_SERVICE_VERSION "2019-03-31"
+
 /**
  * @brief Azure IoT Provisioning Client options.
  *
@@ -69,14 +71,14 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
     az_span global_device_endpoint,
     az_span id_scope,
     az_span registration_id,
-    az_iot_provisioning_client_options* options);
+    az_iot_provisioning_client_options const* options);
 
 /**
- * @brief Gets the MQTT client id.
+ * @brief Gets the MQTT user name.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT client id.
- * @param[out] out_mqtt_user_name The output #az_span containing the MQTT client id.
+ * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT username.
+ * @param[out] out_mqtt_user_name The output #az_span containing the MQTT username.
  * @return #az_result
  */
 AZ_NODISCARD az_result az_iot_provisioning_client_user_name_get(
@@ -198,9 +200,7 @@ typedef struct az_iot_provisioning_client_register_response
   az_iot_status status; /**< The current request status.
                                              * @note The response for the register operation is
                                              * available through #registration_state only. */
-  az_span request_id; /**< Request ID for the register operation.
-                       * @note This must match all messages associated with the registration
-                       * operation. */
+  az_span operation_id; /**< Operation ID of the register operation. */
   az_span registration_state; /**< An #az_span containing the state of the register operation.
                                * @details This can be one of the following: 'unassigned',
                                * 'assigning', 'assigned', 'failed', 'disabled'. */
@@ -255,7 +255,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_publish_topic_get(
  */
 AZ_NODISCARD az_result az_iot_provisioning_client_get_operation_status_publish_topic_get(
     az_iot_provisioning_client const* client,
-    az_span request_id,
+    az_iot_provisioning_client_register_response register_response,
     az_span mqtt_topic,
     az_span* out_mqtt_topic);
 
--- a/sdk/iot/provisioning/samples/README.md
+++ b/sdk/iot/provisioning/samples/README.md
@@ -1,0 +1,60 @@
+---
+page_type: sample
+languages:
+  - c
+products:
+  - azure
+  - azure-iot
+  - azure-iot-provisioning
+urlFragment: iot-provisioning-samples
+---
+
+# Azure IoT Device Provisioning Service Sample
+
+This document explains samples for the Azure Embedded C SDK IoT Provisioning Client and how to use them.
+
+## Key concepts
+
+Key concepts are explained in detail [here][SDK_README_KEY_CONCEPTS].
+
+## Samples for Azure IoT Provisioning Client APIs
+
+This document describes how to use samples and what is done in each sample.
+
+## Getting started
+
+Getting started explained in detail [here][SDK_README_GETTING_STARTED].
+
+## Examples
+
+   Following section document various examples.
+
+1. [IoT Provisioning using Paho][samples_paho]: Register a device using the IoT Provisioning client and the [Eclipse Paho MQTT C client][Eclipse_Paho].
+
+## Troubleshooting
+
+When interacting with the Azure Device Provisioning Service using this C client library, errors are documented within the [MQTT State Machine][error_codes] requests.
+
+## Next steps
+
+Start using the IoT Provisioning Client in your solutions. Our SDK details could be found at [SDK README][IOT_CLIENT_README].
+
+### Additional Documentation
+
+For extensive documentation on Azure Device Provisioning Service, see the [API reference documentation][iot_provisioning_mqtt].
+
+## Contributing
+
+This project welcomes contributions and suggestions. Find [more contributing][SDK_README_CONTRIBUTING] details here.
+
+<!-- LINKS -->
+[IOT_CLIENT_README]: ../../README.md
+[SDK_README_CONTRIBUTING]:../../README.md#contributing
+[SDK_README_GETTING_STARTED]: ../../README.md#getting-started
+[SDK_README_KEY_CONCEPTS]: ../../README.md#key-concepts
+[samples_paho]: src/paho_iot_provisioning_example.c
+[iot_provisioning_mqtt]: https://docs.microsoft.com/en-us/azure/iot-dps/iot-dps-mqtt-support
+[error_codes]: ../../doc/mqtt_state_machine.md#IoT-Service-Errors
+[Eclipse_Paho]: https://www.eclipse.org/paho/clients/c/
+
+![Impressions](https://azure-sdk-impressions.azurewebsites.net/api/impressions/azure-sdk-for-c%2Fsdk%2Fiot%2Fprovisioning%2Fsamples%2FREADME.png)
--- a/sdk/iot/provisioning/samples/README.md
+++ b/sdk/iot/provisioning/samples/README.md
@@ -37,7 +37,7 @@ When interacting with the Azure Device Provisioning Service using this C client 
 
 ## Next steps
 
-Start using the IoT Provisioning Client in your solutions. Our SDK details could be found at [SDK README][IOT_CLIENT_README].
+Start using the IoT Provisioning Client in your solutions. Our SDK details can be found at [SDK README][IOT_CLIENT_README].
 
 ### Additional Documentation
 
--- a/sdk/iot/provisioning/samples/src/generate_certificate.sh
+++ b/sdk/iot/provisioning/samples/src/generate_certificate.sh
@@ -1,0 +1,13 @@
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
+openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
+openssl x509 -text -noout -fingerprint -in device_ec_cert.pem
+
+rm device_cert_store.pem
+cat device_ec_cert.pem >> device_cert_store.pem
+cat device_ec_key.pem >> device_cert_store.pem 
+
+echo Sample certificate generated:
+echo Upload device_ec_cert.pem to Device Provisioning Service.
+echo Use the device_cert_store.pem file within the sample:
+echo    export AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(pwd)/device_cert_store.pem
+
--- a/sdk/iot/provisioning/samples/src/generate_certificate.sh
+++ b/sdk/iot/provisioning/samples/src/generate_certificate.sh
@@ -9,6 +9,10 @@ openssl x509 -text -noout -fingerprint -in device_ec_cert.pem
 rm device_cert_store.pem
 cat device_ec_cert.pem device_ec_key.pem >> device_cert_store.pem
 
+echo It is recommended to use the OpenSSL Trusted CA store configured on your system.
+echo If required (for example on Windows), download the Baltimore PEM CA from https://www.digicert.com/digicert-root-certificates.htm to the current folder.
+echo    export AZ_IOT_DEVICE_X509_TRUST_PEM_FILE=$(pwd)/BaltimoreCyberTrustRoot.crt.pem
+
 echo Sample certificate generated:
 echo Upload device_ec_cert.pem to Device Provisioning Service.
 echo Use the device_cert_store.pem file within the sample:
--- a/sdk/iot/provisioning/samples/src/generate_certificate.sh
+++ b/sdk/iot/provisioning/samples/src/generate_certificate.sh
@@ -1,13 +1,15 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
 openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
 openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
 openssl x509 -text -noout -fingerprint -in device_ec_cert.pem
 
 rm device_cert_store.pem
-cat device_ec_cert.pem >> device_cert_store.pem
-cat device_ec_key.pem >> device_cert_store.pem 
+cat device_ec_cert.pem device_ec_key.pem >> device_cert_store.pem
 
 echo Sample certificate generated:
 echo Upload device_ec_cert.pem to Device Provisioning Service.
 echo Use the device_cert_store.pem file within the sample:
 echo    export AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(pwd)/device_cert_store.pem
-
--- a/sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
+++ b/sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
@@ -1,0 +1,304 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include <paho-mqtt/MQTTClient.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <az_iot_provisioning_client.h>
+#include <az_result.h>
+#include <az_span.h>
+
+// Service information
+#define GLOBAL_PROVISIONING_ENDPOINT_DEFAULT "ssl://global.azure-devices-provisioning.net:8883"
+#define GLOBAL_PROVISIONING_ENDPOINT_ENV "AZ_IOT_GLOBAL_PROVISIONING_ENDPOINT"
+#define ID_SCOPE_ENV "AZ_IOT_ID_SCOPE"
+
+// Device information
+#define REGISTRATION_ID_ENV "AZ_IOT_REGISTRATION_ID"
+// AZ_IOT_DEVICE_X509_CERT_PEM_FILE is the path to a PEM file containing the device certificate and
+// key as well as any intermediate certificates chaining to an uploaded group certificate.
+
+#define DEVICE_X509_CERT_PEM_FILE "AZ_IOT_DEVICE_X509_CERT_PEM_FILE"
+
+static char global_provisioning_endpoint[256] = { 0 };
+static char id_scope[16] = { 0 };
+static char registration_id[256] = { 0 };
+static char x509_cert_pem_file[256] = { 0 };
+
+static az_iot_provisioning_client provisioning_client;
+static MQTTClient mqtt_client;
+
+static az_result read_configuration_entry(
+    const char* name,
+    const char* env_name,
+    char* default_value,
+    bool hide_value,
+    az_span buffer,
+    az_span* out_value)
+{
+  printf("%s = ", name);
+  char* env = getenv(env_name);
+
+  if (env != NULL)
+  {
+    printf("%s\n", hide_value ? "***" : env);
+    *out_value = az_span_init(az_span_ptr(buffer), 0, az_span_capacity(buffer));
+    AZ_RETURN_IF_FAILED(az_span_copy(*out_value, az_span_from_str(env), out_value));
+  }
+  else if (default_value != NULL)
+  {
+    printf("%s\n", default_value);
+    AZ_RETURN_IF_FAILED(az_span_copy(*out_value, az_span_from_str(default_value), out_value));
+  }
+  else
+  {
+    printf("(missing) Please set the %s environment variable.\n", env_name);
+    return AZ_ERROR_ARG;
+  }
+
+  return AZ_OK;
+}
+
+static az_result read_configuration()
+{
+  az_span endpoint_span = AZ_SPAN_LITERAL_FROM_BUFFER(global_provisioning_endpoint);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "Global Device Endpoint",
+      GLOBAL_PROVISIONING_ENDPOINT_ENV,
+      GLOBAL_PROVISIONING_ENDPOINT_DEFAULT,
+      false,
+      endpoint_span,
+      &endpoint_span));
+
+  az_span id_scope_span = AZ_SPAN_LITERAL_FROM_BUFFER(id_scope);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "ID_Scope", ID_SCOPE_ENV, NULL, false, id_scope_span, &id_scope_span));
+
+  az_span registration_id_span = AZ_SPAN_LITERAL_FROM_BUFFER(registration_id);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "Registration ID",
+      REGISTRATION_ID_ENV,
+      NULL,
+      false,
+      registration_id_span,
+      &registration_id_span));
+
+  az_span temp = AZ_SPAN_LITERAL_FROM_BUFFER(x509_cert_pem_file);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "X509 Certificate PEM Store File", DEVICE_X509_CERT_PEM_FILE, NULL, false, temp, &temp));
+
+  AZ_RETURN_IF_FAILED(az_iot_provisioning_client_init(
+      &provisioning_client, endpoint_span, id_scope_span, registration_id_span, NULL));
+
+  return AZ_OK;
+}
+
+static int on_received(void* context, char* topicName, int topicLen, MQTTClient_message* message)
+{
+  UNUSED(context);
+  UNUSED(topicLen);
+
+  int i;
+  char* payloadptr;
+
+  printf("Message arrived\n");
+  printf("     topic: %s\n", topicName);
+  printf("   message: ");
+
+  payloadptr = message->payload;
+  for (i = 0; i < message->payloadlen; i++)
+  {
+    putchar(*payloadptr++);
+  }
+  putchar('\n');
+  MQTTClient_freeMessage(&message);
+  MQTTClient_free(topicName);
+  return 1;
+}
+
+static int connect()
+{
+  int rc;
+
+  MQTTClient_SSLOptions mqtt_ssl_options = MQTTClient_SSLOptions_initializer;
+  MQTTClient_connectOptions mqtt_connect_options = MQTTClient_connectOptions_initializer;
+
+  char username[128];
+  az_span username_span = AZ_SPAN_LITERAL_FROM_BUFFER(username);
+  if ((rc = az_iot_provisioning_client_user_name_get(
+           &provisioning_client, username_span, &username_span))
+      != AZ_OK)
+  {
+    printf("Failed to get MQTT username, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = az_span_append_uint8(username_span, '\0', &username_span)) != AZ_OK)
+  {
+    printf("Failed to get MQTT username, return code %d\n", rc);
+    return rc;
+  }
+
+  mqtt_connect_options.username = (char*)az_span_ptr(username_span);
+  mqtt_connect_options.password = NULL; // Using X509 Client Certificate authentication.
+
+  mqtt_ssl_options.keyStore = (char*)x509_cert_pem_file;
+  mqtt_connect_options.ssl = &mqtt_ssl_options;
+
+  if ((rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options)) != MQTTCLIENT_SUCCESS)
+  {
+    printf("Failed to connect, return code %d\n", rc);
+    return rc;
+  }
+
+  return 0;
+}
+
+static int subscribe()
+{
+  int rc;
+
+  char topic_filter[128];
+  az_span topic_filter_span = AZ_SPAN_LITERAL_FROM_BUFFER(topic_filter);
+  if ((rc = az_iot_provisioning_client_register_subscribe_topic_filter_get(
+           &provisioning_client, topic_filter_span, &topic_filter_span))
+      != AZ_OK)
+
+  {
+    printf("Failed to get MQTT SUB topic filter, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = az_span_append_uint8(topic_filter_span, '\0', &topic_filter_span)) != AZ_OK)
+  {
+    printf("Failed to get MQTT SUB topic filter, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = MQTTClient_subscribe(mqtt_client, topic_filter, 1)) != MQTTCLIENT_SUCCESS)
+  {
+    printf("Failed to subscribe, return code %d\n", rc);
+    return rc;
+  }
+
+  return 0;
+}
+
+static int register_device()
+{
+  int rc;
+  MQTTClient_message pubmsg = MQTTClient_message_initializer;
+
+  char topic[128];
+  az_span topic_span = AZ_SPAN_LITERAL_FROM_BUFFER(topic);
+
+  if ((rc = az_iot_provisioning_client_register_publish_topic_get(
+           &provisioning_client, topic_span, &topic_span))
+      != AZ_OK)
+
+  {
+    printf("Failed to get MQTT PUB register topic, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = az_span_append_uint8(topic_span, '\0', &topic_span)) != AZ_OK)
+  {
+    printf("Failed to get MQTT PUB register topic, return code %d\n", rc);
+    return rc;
+  }
+
+  pubmsg.payload = NULL; // empty payload.
+  pubmsg.payloadlen = 0;
+  pubmsg.qos = 1;
+  pubmsg.retained = 0;
+
+  if ((rc = MQTTClient_publishMessage(mqtt_client, topic, &pubmsg, NULL)) != MQTTCLIENT_SUCCESS)
+  {
+    printf("Failed to publish register request, return code %d\n", rc);
+    return rc;
+  }
+
+  return 0;
+}
+
+int main()
+{
+  int rc;
+
+  if ((rc = read_configuration()) != AZ_OK)
+  {
+    printf("Failed to read configuration from environment variables, return code %d\n", rc);
+    return rc;
+  }
+
+  char client_id[128];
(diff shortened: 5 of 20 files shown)


sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+
+static char global_provisioning_endpoint[256] = { 0 };
+static char id_scope[16] = { 0 };
+static char registration_id[256] = { 0 };
+static char x509_cert_pem_file[256] = { 0 };
+
+static az_iot_provisioning_client provisioning_client;
+static MQTTClient mqtt_client;
+
+static az_result read_configuration_entry(
```
CIPop, line 33: @gilbertw @JeffreyRichter proposed solution for #545

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+}
+
+static int connect()
+{
+  int rc;
+
+  MQTTClient_SSLOptions mqtt_ssl_options = MQTTClient_SSLOptions_initializer;
+  MQTTClient_connectOptions mqtt_connect_options = MQTTClient_connectOptions_initializer;
+
+  char username[128];
```
CIPop, line 128: @gilbertw @JeffreyRichter @barcharcraz  The pattern we've discussed today.

@danewalton and I worked on this a bit and have some ideas:
- making `char username[128] = {0}` would not be ideal as that requires a memset on the stack memory.
- making `username` global may be a viable idea when RAM is not an issue but, on a small device (e.g. 6k of RAM), that would take out the amount of RAM for the entire execution of the application instead of just when it's needed (during reconnect).

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -12,9 +12,153 @@
...
+
+static void test_az_iot_provisioning_client_default_options_get_connect_info_succeed()
+{
+  az_iot_provisioning_client client;
+  az_result ret = az_iot_provisioning_client_init(
+      &client, test_global_device_endpoint, test_id_scope, test_registration_id, NULL);
+  assert_int_equal(AZ_OK, ret);
+
+  uint8_t client_id_buffer[128];
+  az_span client_id = AZ_SPAN_LITERAL_FROM_BUFFER(client_id_buffer);
```
CIPop, line 41: @gilbertw @barcharcraz @JeffreyRichter this is actually not accepted by L4 warnings on the Microsoft Compiler:

```
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(41,1): error C2220: the following warning is treated as an error [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(41,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'client_id_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(40): message : see declaration of 'client_id_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(41,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(51,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'username_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(50): message : see declaration of 'username_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(51,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(76,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'username_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(75): message : see declaration of 'username_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(76,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(94,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(93): message : see declaration of 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(94,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(112,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(111): message : see declaration of 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(112,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(129,1): warning C4221: nonstandard extension used: 'ptr': cannot be initialized using address of automatic variable 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(128): message : see declaration of 'topic_buffer' [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
D:\a\1\s\sdk\iot\provisioning\tests\cmocka\test_az_iot_provisioning_client.c(129,1): warning C4204: nonstandard extension used: non-constant aggregate initializer [D:\a\1\s\build\sdk\iot\provisioning\tests\cmocka\az_iot_provisioning_test.vcxproj]
```

sdk/core/core/inc/_az_cfg.h
```diff
@@ -91,4 +91,8 @@
 // Get the number of elements in an array
 #define _az_COUNTOF(array) (sizeof(array) / sizeof(array[0]))
 
+#ifndef UNUSED
+#define UNUSED(x) (void)(x)
```
danewalton, line 95: Just as a note: this doesn't work in GCC.
[Link to GCC discussion](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425).

One way we discussed doing this is using `assert()`

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -69,14 +71,14 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
     az_span global_device_endpoint,
     az_span id_scope,
     az_span registration_id,
-    az_iot_provisioning_client_options* options);
+    az_iot_provisioning_client_options const* options);
 
 /**
- * @brief Gets the MQTT client id.
+ * @brief Gets the MQTT user name.
```
danewalton, line 77: nit: here it's two words later it's one. Unify one way or the other?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -247,15 +247,14 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_publish_topic_get(
  * @note The payload of the MQTT publish message should be empty.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] request_id The request id. Must match a received
- *                       #az_iot_provisioning_client_register_response request_id.
+ * @param[in] register_response The received response.
```
danewalton, line 250: nit: I like having the type which is returned because it allows for linking to the struct. In the case:
```c
// The received #az_iot_provisioning_client_register_response* 
```
This way someone reading the docs has easy access to the docs for this parameter.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
 
 #include <_az_cfg.h>
 
+static const az_span provisioning_service_api_version
+    = AZ_SPAN_LITERAL_FROM_STR("/api-version=" AZ_IOT_PROVISIONING_SERVICE_VERSION);
```
danewalton, line 14: does this not need a `?`

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+
+static const az_span str_dps = AZ_SPAN_LITERAL_FROM_STR("$dps");
+static const az_span str_registrations = AZ_SPAN_LITERAL_FROM_STR("/registrations/");
+static const az_span str_put_iotdps_register
+    = AZ_SPAN_LITERAL_FROM_STR("PUT/iotdps-register/?$rid=1");
```
danewalton, line 19: request id number doesn't change?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+  AZ_PRECONDITION_VALID_SPAN(id_scope, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(registration_id, 1, false);
+
+  client->_internal.global_device_endpoint = global_device_endpoint;
+  client->_internal.id_scope = id_scope;
+  client->_internal.registration_id = registration_id;
+
+  if (options != NULL)
+  {
+    client->_internal.options.user_agent = options->user_agent;
```
danewalton, line 47: nit: should you just assign the options to the options? This prevents needing to update it every time an option is added.
```suggestion
    client->_internal.options = options;
```

danewalton: or i might have ran into this myself. does the const disallow this?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+  return AZ_OK;
+}
+
+// $dps/registrations/res/#
+AZ_NODISCARD az_result az_iot_provisioning_client_register_subscribe_topic_filter_get(
+    az_iot_provisioning_client const* client,
+    az_span mqtt_topic_filter,
+    az_span* out_mqtt_topic_filter)
+{
+  UNUSED(client);
```
danewalton, line 109: not sure this is necessarily needed other than as a formality. the `AZ_PRECONDITION_NOT_NULL` will touch it and do away with the "unused param"

danewalton: understood if you're just trying to make it obvious to people reading the code that it is unused.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+
+  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic, str_dps, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_registrations, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_put_iotdps_register, &mqtt_topic));
+
+  *out_mqtt_topic = mqtt_topic;
+  return AZ_OK;
+}
+
+// Topic: $dps/registrations/GET/iotdps-get-operationstatus/?$rid={%s}&operationId=%s
```
danewalton, line 161: nit: should it be
```suggestion
// Topic: $dps/registrations/GET/iotdps-get-operationstatus/?$rid={%s}&operationId={%s}
```

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
danewalton: yea this is why all the tests I have for other parts as `az_span_init()` :/

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -12,9 +12,153 @@
...
+  az_iot_provisioning_client_options options = az_iot_provisioning_client_options_default();
+  options.user_agent = AZ_SPAN_FROM_STR(TEST_USER_AGENT);
+
+  az_result ret = az_iot_provisioning_client_init(
+      &client, test_global_device_endpoint, test_id_scope, test_registration_id, &options);
+  assert_int_equal(AZ_OK, ret);
+
+  uint8_t username_buffer[128];
+  az_span username = AZ_SPAN_LITERAL_FROM_BUFFER(username_buffer);
+  az_span_set(username, 0xCC);
```
danewalton, line 77: these test spans can use the new `az_span_for_test_` stuff
[Link](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/test/inc/az_test_span.h)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <az_iot_provisioning_client.h>
+#include <az_result.h>
+#include <az_span.h>
+
+// Service information
+#define GLOBAL_PROVISIONING_ENDPOINT_DEFAULT "ssl://global.azure-devices-provisioning.net:8883"
```
danewalton, line 14: is this not a value in our SDK?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+static const az_span str_res = AZ_SPAN_LITERAL_FROM_STR("res/");
+
+AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_options_default()
+{
+  return (az_iot_provisioning_client_options){ .user_agent = AZ_SPAN_NULL };
+}
+
+AZ_NODISCARD az_result az_iot_provisioning_client_init(
+    az_iot_provisioning_client* client,
+    az_span global_device_endpoint,
```
hihigupt, line 31: >global_device_endpoint [](start = 12, length = 22)

can this be called endpoint?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+#define ID_SCOPE_ENV "AZ_IOT_ID_SCOPE"
+
+// Device information
+#define REGISTRATION_ID_ENV "AZ_IOT_REGISTRATION_ID"
+// AZ_IOT_DEVICE_X509_CERT_PEM_FILE is the path to a PEM file containing the device certificate and
+// key as well as any intermediate certificates chaining to an uploaded group certificate.
+
+#define DEVICE_X509_CERT_PEM_FILE "AZ_IOT_DEVICE_X509_CERT_PEM_FILE"
+
+static char global_provisioning_endpoint[256] = { 0 };
```
danewalton, line 25: nit nit: here the `{ 0 }` aren't needed

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+      != AZ_OK)
+
+  {
+    printf("Failed to get MQTT PUB register topic, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = az_span_append_uint8(topic_span, '\0', &topic_span)) != AZ_OK)
+  {
+    printf("Failed to get MQTT PUB register topic, return code %d\n", rc);
```
danewalton, line 208: not exactly the error here

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+      != AZ_OK)
+
+  {
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = az_span_append_uint8(client_id_span, '\0', &client_id_span)) != AZ_OK)
+  {
+    printf("Failed to get MQTT username, return code %d\n", rc);
```
danewalton, line 249: not exactly the error here

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+  UNUSED(client);
+
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic_filter, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic_filter);
+
+  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic_filter, str_dps, &mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic_filter, str_registrations, &mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic_filter, str_res, &mqtt_topic_filter));
+  AZ_RETURN_IF_FAILED(az_span_append_uint8(mqtt_topic_filter, '#', &mqtt_topic_filter));
```
hihigupt, line 118: should we build this const or have one?

sdk/iot/provisioning/samples/src/generate_certificate.sh
```diff
@@ -0,0 +1,13 @@
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
+openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
```
danewalton, line 2: will the `"/CN={}"` change?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+{
+  UNUSED(client);
+
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic, str_dps, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_registrations, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_put_iotdps_register, &mqtt_topic));
```
hihigupt, line 155: It looks like to be a constant, can we have one rather then computing it?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+  }
+  else
+  {
+    client->_internal.options = az_iot_provisioning_client_options_default();
+  }
+
+  return AZ_OK;
+}
+
+// <id_scope>/registrations/<registration_id>/api-version=<service_version>
```
hihigupt, line 57: >//  [](start = 0, length = 3)

Is there any guidline to put comments?

I prefer atleast the function comments to be 
/* */

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -10,3 +10,175 @@
...
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic, str_dps, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_registrations, &mqtt_topic));
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, str_put_iotdps_register, &mqtt_topic));
+
+  *out_mqtt_topic = mqtt_topic;
+  return AZ_OK;
```
hihigupt, line 158: nit: please add space before return.
 in all the places

hihigupt:
> #include <stdint.h>

Do you need this?

---
Refers to: sdk/iot/provisioning/src/az_iot_provisioning_client.c:4 in c5dbd85. [](commit_id = c5dbd85b32ba3ba60f785be7e5a5a765ea460be2, deletion_comment = False)

sdk/iot/provisioning/samples/src/generate_certificate.sh
hihigupt, line 2: MS copyright?

sdk/iot/provisioning/samples/src/generate_certificate.sh
```diff
@@ -0,0 +1,13 @@
...
+openssl x509 -text -noout -fingerprint -in device_ec_cert.pem
+
+rm device_cert_store.pem
+cat device_ec_cert.pem >> device_cert_store.pem
+cat device_ec_key.pem >> device_cert_store.pem
```
hihigupt, line 7: cat is concatenate files and print on the standard output.

So you can remove these line and replace with following line:

cat device_ec_cert.pem device_ec_key.pem  > device_cert_store.pem

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+      "Global Device Endpoint",
+      GLOBAL_PROVISIONING_ENDPOINT_ENV,
+      GLOBAL_PROVISIONING_ENDPOINT_DEFAULT,
+      false,
+      endpoint_span,
+      &endpoint_span));
+
+  az_span id_scope_span = AZ_SPAN_LITERAL_FROM_BUFFER(id_scope);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "ID_Scope", ID_SCOPE_ENV, NULL, false, id_scope_span, &id_scope_span));
```
hihigupt, line 77: >ID_Scope [](start = 7, length = 8)

macro, please do same for all

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+      "Registration ID",
+      REGISTRATION_ID_ENV,
+      NULL,
+      false,
+      registration_id_span,
+      &registration_id_span));
+
+  az_span temp = AZ_SPAN_LITERAL_FROM_BUFFER(x509_cert_pem_file);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "X509 Certificate PEM Store File", DEVICE_X509_CERT_PEM_FILE, NULL, false, temp, &temp));
```
hihigupt, line 90: Can we remove stuff that is not getting used?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+static int on_received(void* context, char* topicName, int topicLen, MQTTClient_message* message)
+{
+  UNUSED(context);
+  UNUSED(topicLen);
+
+  int i;
+  char* payloadptr;
+
+  printf("Message arrived\n");
+  printf("     topic: %s\n", topicName);
```
hihigupt, line 107: Can we print this topic with length? Reason of bubbling length is that people think it like buffer.


printf("     topic: %.*s\n", topicLen, topicName);

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+  printf("   message: ");
+
+  payloadptr = message->payload;
+  for (i = 0; i < message->payloadlen; i++)
+  {
+    putchar(*payloadptr++);
+  }
+  putchar('\n');
+  MQTTClient_freeMessage(&message);
+  MQTTClient_free(topicName);
```
hihigupt, line 117: We do not have ownership of these memory, atleast what I read from and implementation I saw

[https://github.com/nivertech/paho.mqtt.c/blob/331190944e874d4bd48712c50e8d3cfb483f6479/src/MQTTClient.c#L473](https://github.com/nivertech/paho.mqtt.c/blob/331190944e874d4bd48712c50e8d3cfb483f6479/src/MQTTClient.c#L473)

hihigupt: Is it possbile to run valgrind on this app?

---
In reply to: [404280052](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404280052) [](ancestors = 404280052)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  mqtt_connect_options.username = (char*)az_span_ptr(username_span);
+  mqtt_connect_options.password = NULL; // Using X509 Client Certificate authentication.
+
+  mqtt_ssl_options.keyStore = (char*)x509_cert_pem_file;
+  mqtt_connect_options.ssl = &mqtt_ssl_options;
+
+  if ((rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options)) != MQTTCLIENT_SUCCESS)
+  {
```
hihigupt, line 151: Can you check what is the scope of memory required by the option parameter.
I am just worried we gave usename that sits in stack.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+
+static int subscribe()
+{
+  int rc;
+
+  char topic_filter[128];
+  az_span topic_filter_span = AZ_SPAN_LITERAL_FROM_BUFFER(topic_filter);
+  if ((rc = az_iot_provisioning_client_register_subscribe_topic_filter_get(
+           &provisioning_client, topic_filter_span, &topic_filter_span))
+      != AZ_OK)
```
hihigupt, line 167: This looks like if we start giving other then const value we might break customer.
Do you still think that exposing const is not right way to go in embedded  sdk?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+  else
+  {
+    printf("(missing) Please set the %s environment variable.\n", env_name);
+    return AZ_ERROR_ARG;
+  }
+
+  return AZ_OK;
+}
+
+static az_result read_configuration()
```
hihigupt, line 64: >read_configuration [](start = 17, length = 18)

Can we name this init client ? Because you are initializing the client based on config.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -0,0 +1,304 @@
...
+  if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
+      != MQTTCLIENT_SUCCESS)
+  {
+    printf("Failed to set MQTT callbacks, return code %d\n", rc);
+    return rc;
+  }
+
+  if ((rc = connect()))
+  {
+    return rc;
```
hihigupt, line 274: Please clean client once it has been created successfully in every exit path.

Now, Valgrind looks to be imp to be run on this and also integrated into pipeline.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Not sure what you mean. This is a const, compiler will take care of it.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I don't recommend it: Device Provisioning has _both_ a service client endpoint (which is particular to each instance) and the global endpoint which is common to _all_ instances in a particular cloud.
In general, gde will be the same for most of our customers unless they are part of one of the non-standard clouds.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: Looks to me like a bug in the macro that needs to be fixed. If this is fine and it'll work, we need to add the correct pragmas to disable the warnings.

sdk/core/core/inc/_az_cfg.h
CIPop: It worked with GCC since that's what's used in the build system and on my box. Should work with clang too since this is what we're using in C-SDK in a lot of places.

Can you give a short explanation/TLDR on what exactly won't work? (I'm, thinking increasing the warning level - didn't try that.)

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: Good point. I think we want username (single word).

sdk/iot/provisioning/samples/src/generate_certificate.sh
CIPop: More work to come in future PRs to have this script print out detailed steps that one needs to take after running it (i.e. link to portal, create DPS registration, use the reg name, etc.)
My plan here: extract into a default variable, override with a parameter.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: No, GDE is not the same in all Azure clouds. This is just the one for the Azure public cloud.
Also, this string format is particular to Paho.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: I like to have it clear that it's all zeros.

danewalton: no prob sounds good 👍

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Actually I wanted this function to read config _only_.
I was _forced_ in doing the client init here to avoid passing a lot of spans over to some other function.

I'll hold on on making this change after we decide on public API.

CIPop: /cc @gilbertw @JeffreyRichter - another point to the char* vs az_span conversation.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
hihigupt: Are we going to expose one more init for Service endpoint?

---
In reply to: [404454304](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404454304) [](ancestors = 404454304)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: If you think of a macro that'll define the spans, then definitely no.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: x509 cert is used.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: This will be removed: no point in printing out every message during provisioning. (I left this in by mistake.)
Paho has good logging that should display the info you need. If that's not true, I'll leave it in and add len.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Interesting: this is from their sample.
I'll run valgrind before resolving this conversation.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: > This looks like if we start giving other then const value we might break customer.

How would that break? Above can give a dynamic topic filter.

> Do you still think that exposing const is not right way to go in embedded sdk?

No, I think it should be OK to expose the constants. I'm still investigating if it would be OK to have _all_ topic filters for SUB const.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: :unamused: Not exactly the code I'd like someone else to write either.

sdk/core/core/inc/_az_cfg.h
danewalton: Shoot sorry I take that back. It works for input parameters but it does NOT work for returned values of functions. So you can't void cast a returned value from one of our functions and silence that warning.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
danewalton: haha i hear ya

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Abnormal termination will immediately stop the process. Only normal termination will clean-up. This is very common, it's also what the official Paho sample does.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: This works today.
Where do you think this needs the '?' - prefix or postfix?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: No: I don't see any value in doing that: RID is fixed. Retry must disconnect.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: Will address in a future PR. I don't want to finalize on tests just yet since we might change the API.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I don't think so: why would it have {}? I think that's just a marker that will be replaced by the string itself :)
I may be wrong. Do we have it in other places? Can you try w/o? (Saving 2 chars over the wire is good.)

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: I only ask because IoT hub seems to be inconsistent where they are needed or not.
On connect, needed: [LINK](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-device)
On telemetry, not needed: [LINK](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#sending-device-to-cloud-messages)
On twin, needed: [LINK](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#retrieving-a-device-twins-properties)

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: I wasn't sure if this was you demonstrating where the strings for those values would go or not. You had it for `$rid={%s}` and wasn't sure if that was actually sent on the wire or if that's your placeholder for what goes there.

CIPop: More to this:
1. OperationID is not containing {} in the json payload.
2. RID has {rid} in the IoT Hub and DPS documentation but works just fine w/o them.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I think you mean new-line. Otherwise I think it's aligned correctly.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: We're investigating that offline. Not part of this PR.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: No, this is mandatory if you disable preconditions. Build fails.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: This was formatted according to the clang rules and it's showing an example of what this function will currently return.
This is _not_ a contract or function documentation - that's done in the declaration site.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Copy paste from the telemetry code :D.
I'll give it a try - if it works, I'll change both.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Same as above - I won't change before we decide on the public API.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: I won't change before we decide on the public API.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: Ahhhh good to know thanks for clarifying

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: How are you going to justify the valgrind output?

---
In reply to: [404466337](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404466337) [](ancestors = 404466337)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: Not sure; but I do not see any ref to temp being held in this function

---
In reply to: [404461063](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404461063) [](ancestors = 404461063)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: Literals, can they come from Macro?

---
In reply to: [404460738](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404460738) [](ancestors = 404460738)

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -12,9 +12,153 @@
...
+      az_span_ptr(expected_topic), az_span_ptr(topic), (size_t)az_span_length(topic));
+  assert_int_equal(0xCC, topic_buffer[az_span_length(topic)]);
+}
+
+static void test_az_iot_provisioning_client_get_operation_status_publish_topic_filter_succeed()
+{
+  az_iot_provisioning_client client;
+  uint8_t topic_buffer[128];
+  az_span topic = AZ_SPAN_LITERAL_FROM_BUFFER(topic_buffer);
+  az_span_set(topic, 0xCC);
```
ericwol-msft, line 130: I don't get what this is testing?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: I'll follow-up with the change in a future PR.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Good point: could you please try the other features w/o the `?`?
I'll leave this one w/o the `?` because I was able to get the server to respond and, apparently, to the point where I got an operationId everything works fine. If we find issues, we can add the ? later.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: > You had it for $rid={%s} and wasn't sure if that was actually sent on the wire or if that's your placeholder for what goes there.

You're right: I've fixed all of them to remove the {} because it's confusing.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: This is indeed a bug but in msvc. I'm switching to az_span_init.

sdk/iot/provisioning/samples/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. [IoT Provisioning using Paho][samples_paho]: Register a device using the IoT Provisioning client and the [Eclipse Paho MQTT C client][Eclipse_Paho].
+
+## Troubleshooting
+
+When interacting with the Azure Device Provisioning Service using this C client library, errors are documented within the [MQTT State Machine][error_codes] requests.
+
+## Next steps
+
+Start using the IoT Provisioning Client in your solutions. Our SDK details could be found at [SDK README][IOT_CLIENT_README].
```
danewalton, line 40: nit: "can be found"

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: On hub it seems to work without the `?`. At least from using MQTT Explorer. As long as it works totally good with me 👍

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: You could have used %.*s to bound the printf 

---
In reply to: [406367096](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r406367096) [](ancestors = 406367096)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: In general, Valgrind detection of overflows in stack and global variables is weak to non-existant. Arguably, Valgrind is the wrong tool for that job. If you are on one of supported platforms, building with -fmudflap and linking with -lmudflap will give you much better results for these kinds of errors.

---
In reply to: [406412124](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r406412124) [](ancestors = 406412124)

hihigupt: Here I was more concerned about stack memory.... I am just worried that if it start failing randomly... it will be harder to debug the issue.

---
In reply to: [406524774](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r406524774) [](ancestors = 406524774,406412124)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt: Clean shutdown is what I prefer.

---
In reply to: [404483628](https://github.com/Azure/azure-sdk-for-c/pull/547#discussion_r404483628) [](ancestors = 404483628,404466337)

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -12,9 +12,160 @@
...
+
+  assert_int_equal(AZ_OK, ret);
+  assert_memory_equal(
+      az_span_ptr(expected_topic), az_span_ptr(topic), (size_t)az_span_length(topic));
+  assert_int_equal(0xCC, topic_buffer[az_span_length(topic)]);
+}
+
+#ifdef _MSC_VER
+// warning C4113: 'void (__cdecl *)()' differs in parameter lists from 'CMUnitTestFunction'
+#pragma warning(disable : 4113)
```
ahsonkhan, line 155: Do we want to re-enable the warning outside this file?

CIPop: I think we need to unify this into something that everybody can use: I don't like having to write and read useless `**state` parameters in all our tests. 

> outside this file

My understanding is that the preprocessor rules apply for this file only and that `#pragma warning` applies only from this line within the file (not above).

PR #551, opened 2020-04-08
Add property_find api

danewalton:


sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -140,3 +142,48 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_append(
...
+  // Keep looking in case the passed name is a substring of a name in props
+  while (az_span_length(remaining) != 0)
+  {
+    index = az_span_find(remaining, name);
+    if (index != -1)
+    {
+      az_span value;
+      az_span found_name;
+      remaining = az_span_slice(remaining, index, -1);
+      found_name = az_span_token(remaining, hub_client_param_equals_span, &value);
```
TiejunMS, line 167: If the property is "key_one=value_one", "value_one" would be found with name "ey_one". Check for length of `found_name` is not restrict.

danewalton: Ahh you’re right. Had that in but removed on a rework. Thanks will update.

danewalton: Added this:
https://github.com/Azure/azure-sdk-for-c/blob/4677584b73519bbe90c7abb4de91dd92df458ab7/sdk/iot/hub/src/az_iot_hub_client.c#L180-L184
And added tests which test if a previous value matches what you are looking for, ie searching for `key` in this:
https://github.com/Azure/azure-sdk-for-c/blob/4677584b73519bbe90c7abb4de91dd92df458ab7/sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c#L49
both in the success case and fail case, and the edge case when the value is the last value (where this addition makes sense as there is no remaining chars to search.)

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
```diff
@@ -554,6 +591,141 @@ static void test_az_iot_hub_client_properties_append_twice_small_buffer_fail(voi
...
+  assert_int_equal(az_iot_hub_client_properties_find(&props, test_key_two, &out_value), AZ_OK);
+  assert_memory_equal(
+      az_span_ptr(out_value), az_span_ptr(test_value_two), (size_t)az_span_length(test_value_two));
+}
+
+static void test_az_iot_hub_client_properties_find_substring_succeed(void** state)
+{
+  (void)state;
+
+  az_span test_span = az_span_from_str(TEST_KEY_VALUE_SUBSTRING);
```
ericwol-msft, line 643: should also test "one_key=value_one&key=value_two"

ericwol-msft: also what about "key_one=value_one&key_one=value_two"?

danewalton: Your first point is valid and I will add a test for that (and fix if it's a problem). As for the second point, is that something that this function should account for? If the user wanted they could use the `az_iot_hub_client_properties_next()` to iterate through the same named keys.

danewalton: updated and reworked using only `az_span_token()`

ericwol-msft: >> If the user wanted they could use the az_iot_hub_client_properties_next() to iterate through the same named keys.

That is what I expect. Test case for it?

danewalton: In this case what would the test case be? That there are two names that are identical and this always returns the first one?

ericwol-msft: humm, I figured that current_property was updated after the find

danewalton: No i think that should only be updated when using next. I think it might be kind of odd to search for a key, find it, underneath the current prop is update, and then another find would return no value. In the case where maybe someone calls it twice expecting the same value.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -140,3 +142,41 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_append(
...
+    {
+      *out_value = az_span_token(remaining, hub_client_param_separator_span, &remaining);
+      return AZ_OK;
+    }
+    else
+    {
+      az_span value;
+      value = az_span_token(remaining, hub_client_param_separator_span, &remaining);
+      (void)value;
+      if (az_span_length(remaining) > 0)
```
TiejunMS, line 170: Is this check duplicated from line 157?
`while (az_span_length(remaining) != 0)`

danewalton: Not exactly. Note on line 168 how remaining is used as the output span. When it is passed to token, it becomes everything after token. Example of `az_span_token`:
```
Token: &
Search string: key=value&my_thing=my_value
Return: [key=value]&<my_thing=my_value>
Where [] is the returned span and <> is the output span. 
On line 168 returned span is "value" and output span is "&remaining"
```

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -13,6 +13,8 @@
 static const uint8_t hub_client_forward_slash = '/';
 static const uint8_t hub_client_param_separator = '&';
 static const uint8_t hub_client_param_equals = '=';
```
ericwol-msft, line 15: why have these defined twice

static const uint8_t hub_client_param_separator = '&';
static const uint8_t hub_client_param_equals = '=';	static const uint8_t hub_client_param_equals = '=';

U can just use az_span_ptr(hub_client_param_separator_span) and only have one definition?

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -140,3 +142,41 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_append(
...
+      return AZ_OK;
+    }
+    else
+    {
+      az_span value;
+      value = az_span_token(remaining, hub_client_param_separator_span, &remaining);
+      (void)value;
+      if (az_span_length(remaining) > 0)
+      {
+        continue;
```
ericwol-msft, line 172: do you really need this "continue" can' you let it fall thru?

sdk/iot/hub/src/az_iot_hub_client.c
danewalton: you are right good call. updated

sdk/iot/hub/src/az_iot_hub_client.c
danewalton: yes this whole if/else is not needed. removed

sdk/iot/hub/src/az_iot_hub_client.c
danewalton: I think Eric might have had the same comment as you below. I removed this if/else logic to simplify. Thanks for the tip Tiejun

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -11,8 +11,8 @@
 #include <_az_cfg.h>
 
 static const uint8_t hub_client_forward_slash = '/';
-static const uint8_t hub_client_param_separator = '&';
-static const uint8_t hub_client_param_equals = '=';
+static const az_span hub_client_param_separator_span = AZ_SPAN_LITERAL_FROM_STR("&");
+static const az_span hub_client_param_equals_span = AZ_SPAN_LITERAL_FROM_STR("=");
```
ahsonkhan, line 15: I am not a big fan of this change which forces us to us `az_span_ptr` in more places to get back the single byte. I would have preferred this remained a `uint8_t `and we created a helper for finding a single byte.

ericwol-msft: ?? obviously we don't want two identical defines in the source code. I don't know what you means by a helper. That sounds worse.

PR #559, opened 2020-04-09
Implement Azure IoT device methods code and tests.

ewertons:


sdk/core/core/inc/az_span.h
```diff
@@ -282,6 +282,7 @@ AZ_NODISCARD az_result az_span_to_uint32(az_span span, uint32_t* out_number);
  */
 AZ_NODISCARD int32_t az_span_find(az_span source, az_span target);
 
+AZ_NODISCARD int32_t az_span_find2(az_span source, int32_t start, az_span target);
```
danewalton, line 285: is this supposed to be here?

JeffreyRichter, line 285: Did you discuss this with anyone?
Do you think we'll make this public, document it, test, it, make samples for it, and support it in perpetuity?
Do we need 2 find functions? Can we remove 1 of them? I'm not sure what your intention is here with this function.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
...
+    uint16_t status,
+    az_span mqtt_topic,
+    az_span* out_mqtt_topic)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(request_id, 1, false);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic);
+
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic, methods_topic_prefix, &mqtt_topic));
```
danewalton, line 90: This first one should be a copy to ignore the input length.
```suggestion
  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic, methods_topic_prefix, &mqtt_topic));
```

sdk/core/core/inc/az_span.h
ewertons: Oh no! please disregard this!

ewertons: Oh no! please disregard this!

ewertons: Just pushed an update, those changes in SDK/core are gone now. Sorry for that.

ewertons: Just pushed an update, those changes in SDK/core are gone now. Sorry for that.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
...
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(received_topic, 1, false);
+  AZ_PRECONDITION_NOT_NULL(out_request);
+
+  int32_t index = az_span_find(received_topic, methods_topic_prefix);
+
+  if (index == -1) 
+  {
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
```
danewalton, line 51: I think we should create a return value for IoT. Spitballing here but something like `AZ_ERROR_IOT_PARSE_NOT_FEATURE`, `AZ_ERROR_IOT_NOT_MINE`....

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
...
+AZ_NODISCARD az_result az_iot_hub_client_methods_subscribe_topic_filter_get(
+    az_iot_hub_client const* client,
+    az_span mqtt_topic_filter,
+    az_span* out_mqtt_topic_filter)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_VALID_SPAN(mqtt_topic_filter, 0, false);
+  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic_filter);
+
+  AZ_RETURN_IF_FAILED(az_span_append(mqtt_topic_filter, methods_topic_prefix, &mqtt_topic_filter));
```
danewalton, line 29: First should be copy to ignore the length
```suggestion
  AZ_RETURN_IF_FAILED(az_span_copy(mqtt_topic_filter, methods_topic_prefix, &mqtt_topic_filter));
```

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
...
+#include <stdint.h>
+
+#include <cmocka.h>
+#include <az_test_precondition.h>
+
+#define TEST_SPAN_BUFFER_SIZE 128
+
+#define TEST_DEVICE_ID_STR "my_device"
+#define TEST_DEVICE_HOSTNAME_STR "myiothub.azure-devices.net"
+#define TEST_MODULE_ID "my_module_id"
```
danewalton, line 23: I don't think this is being used

danewalton: Would some of these api's change if the module id is set? Are tests needed?

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
...
+}
+
+#endif // NO_PRECONDITION_CHECKING
+
+static void test_az_iot_hub_client_methods_subscribe_topic_filter_get_succeed(void** state)
+{
+  (void)state;
+
+  uint8_t mqtt_sub_topic_buf[TEST_SPAN_BUFFER_SIZE];
+  memset(mqtt_sub_topic_buf, 0xFF, _az_COUNTOF(mqtt_sub_topic_buf));
```
danewalton, line 246: these can be updated using the changes John made with `az_span_for_test()`. Simplifies all of these.
[Link here](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/test/inc/az_test_span.h)

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
...
+  az_span mqtt_sub_topic = az_span_init(mqtt_sub_topic_buf, 0, _az_COUNTOF(mqtt_sub_topic_buf));
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_true(az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
+
+  assert_true(az_iot_hub_client_methods_subscribe_topic_filter_get(&client, mqtt_sub_topic, &mqtt_sub_topic) == AZ_OK);
+
+  assert_memory_equal(g_expected_methods_subscribe_topic, az_span_ptr(mqtt_sub_topic), (size_t)az_span_length(mqtt_sub_topic));
```
danewalton, line 255: These would be simplified as well with the api's/link I posted above.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
...
+{
+  (void)state;
+
+  uint8_t mqtt_sub_topic_buf[TEST_SPAN_BUFFER_SIZE];
+  memset(mqtt_sub_topic_buf, 0xFF, _az_COUNTOF(mqtt_sub_topic_buf));
+  az_span mqtt_sub_topic = az_span_init(mqtt_sub_topic_buf, 0, _az_COUNTOF(mqtt_sub_topic_buf));
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_true(az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
```
danewalton, line 251: nit: you can use `assert_int_equal` for returned values

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
...
+  assert_true(az_iot_hub_client_methods_response_publish_topic_get(&client, request_id, status, mqtt_sub_topic, &mqtt_sub_topic) == AZ_ERROR_INSUFFICIENT_SPAN_CAPACITY);
+}
+
+static void test_az_iot_hub_client_methods_received_topic_parse_succeed(void** state)
+{
+  (void)state;
+
+  az_iot_hub_client client;
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_true(az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
```
danewalton, line 360: nit: `hub_client_init` accepts `NULL` as a param for options. It will then call `az_iot_hub_client_options_default()` for you.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,487 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
danewalton, line 1: Can you run the clang formatter on this file? On VSCode you can use `Shift-Alt-F`

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
danewalton, line 1: Can you run the clang formatter on this file?

sdk/iot/hub/src/az_iot_hub_client_methods.c
jspaith: This is expected isn't it, for like case where app has one function handle ALL MQTT incoming traffic and then will "visit" each of the method_parse/twin_parse/etc.?  If so we'd need a new IoT code as and it shouldn't be `ERROR` in it.  `AZ_IOT_NOT_PROCESSED`??

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
...
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+  }
+
+  received_topic = az_span_slice(received_topic, index + az_span_length(methods_topic_prefix), az_span_length(received_topic));
+
+  index = az_span_find(received_topic, methods_topic_filter_suffix);
+  
+  if (index == -1) 
+  {
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
```
jspaith, line 60: This one I think would be an error, though, since we should find this if we've matched 1st part of topic?  Here I tend to agree with Dane that an IoT error would be appropriate.  Off top of head, `AZ_IOT_INVALID_TOPIC`??

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,99 @@
...
+  received_topic = az_span_slice(received_topic, index + az_span_length(methods_topic_filter_suffix), az_span_length(received_topic));
+
+  index = az_span_find(received_topic, methods_response_topic_properties);
+
+  if (index == -1) 
+  {
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+  }
+
+  out_request->name = az_span_slice(received_topic, 0, index);
```
jspaith, line 72: Nit: Recommend that you treat `received_topic` variable as effectively immutable through this function and that when you need a new slice of it, you add a new variable name.  So like above `az_span method_name = az_span_slice(...)`.  Compilers should optimize out extra vars and to my eye at least it's clearer what var is at each state.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Done.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: @jspaith , I think we need to separate the concepts here. `received_topic` could be:
1. A valid direct method request topic;
2. A valid topic name for the other features we support;
3. An invalid topic name.

The function covers 1. because it's the happy path.
But it does not differentiate between 2. and 3. So to say that the return code for these cases should be AZ_IOT_NOT_PROCESSED hides the case where the customer has passed a case 3. input. 
Of course it makes no sense to code separate logics to differentiate between 2. and 3. in the function, it would be duplication of code and expensive.
So I understand that the function should return a *_ERROR_* in case it cannot parse the input. In the upper layer the topic will be submitted to each parsing function and get errors over and over, until one returns OK.
I do agree with Dane though, the return should be an AZ_IOT enum value.
+ @CIPop @hihigupt  to chime in.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Done,

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
ewertons: Indeed itś not being used. Removed.
Well, direct methods do not depend on the device/module identity, so no they would not change unless the hub changes their design.
Adding a test here for that would be P4 I would say, since there is no impact right now. A test here would be just for eventual future changes, so it doesn't make sense to add anything like that now.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
ewertons: Love it! Thank you @jspaith .

ewertons: Resolved.

sdk/iot/hub/src/az_iot_hub_client_methods.c
danewalton: Yea I don't think there is even a way to differentiate between 2 and 3 since each feature is not aware of the others. I think in that case we should have one of two returns:
1. Parsed and it is mine. Success
2. I don't know what this is but it isn't mine. Error

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
ewertons: Thanks. I'll keep it assert_true for now.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
ewertons: That's true. I should simplify that. Thanks.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
ewertons: Done.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Done. For reference, ALT SHIFT F is for VS code on Windows.
On Linux its CTRL SHIFT I

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: I understand the request for more clarity, but I learned something cool recently and I think it's worth sharing here. It's this site that is useful to compare different compilers, called godbolt.org.

I took two functions, 

int function(int a) {
    a = a * 2;
    a = a * 3;
    return a;
}

int function2(int a) {
    int b = a * 2;
    int c = b * 3;
    return c;
}

Using ARM gcc 8.3.1 none (with flag -O3 for optmization) these were the assembly sizes:
function: 4021 bytes
function2:  **4827** bytes.

So there it actually the compiler is doing the optimization but not giving the result we are looking for, which is smaller size.
I know this is not the only compiler we will use, but it makes a case.
What do you think, @jspaith ?
 
Here is the [link to this exercise](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlICsupVs1qhkAUgBMAISnTSAZ0ztkBPHUqZa6AMKpWAVwC2tEJNJb0AGTy1MAOWMAjTMRAA2UgAdUCwuto9QxNBHz81Ohs7RyMXN3MlFQjaBgJmYgIg41ME5UxVANT0giiHZ1cPRTSMrJDOKuLSmLizAEpFVANiZA4AcgB6foBqABUATy9MIbHO4iG0LCGEV0xSIZIh1lRmdCGRIcwAD2YjL3YAOnEABgBBWwIhqgNaAroIe73WofEAdllboaAvbfADMABFgQAqIaScQg/43IHAuEQ5hDaEguEIpHETAELq0PZY643X5gkm9dqsEC9ACsvVIpl6VwZqBpOjkciGClmPW%2BkhBnAZBBpLNa7QA1mZaUIadwGUyWaQ2b0GQoQFdSCLmZTSHBYEg0Kc8OwyBQIEavCaKukjAAOQWkKgmgiudUQJyihlOWzpMY0oWkI1GLQEADytFY/p1pCwRhEwHYXtjeFxBQAbph1TGjvkDK6Awz7spk6w8E5iH69FhC1riHgjIX2jR6Ew2BwePxBMJRChOTIhOX1ZB2qgvMlswBaMMgtV5V6mCCWWqmeqWJrlNz1ML%2BOgr0K%2BXe0DexCr1RL5ZJFGr6bKCC8L68lWxlU9bho34Krj/P6KbrjtDyXQ9ABsp0gqyYqjcABKACyQzAMgyBDHa5wgucnBDBAtB0JgXwQLghAbFIjpDHoxqmvygpfByMhyMKXriqQUqSDK1K9PKjKQTSaoalqjF6jAiAoBgOD4EQZqUC2jAsEmnYCOYPZiP28gPskmjaPu5jri%2BzQVDKO7JFp3iHskJ4tDKamFNUmS3iEuRJNZjS6f%2Blk2cZCg2eZ%2BmAbyfQfE8LzJJI7z0J83x/CSSIfE4oKouiMLEgCQIfMhKJDLFGJJYiQK4vixCEhI8IkmSursfSXExiqtGyDI3J%2BVRkgMTqTEsdw5wAJxdd1PU9Z47GcYqrI8YofHamKgnwBAhqoBRrjkJQlrWm4toOvUzqsK6xDup6MY%2BrQfq1sGoYRlGyZxgmSYxvgaZqJm2ZKrmyD5n0gbFuxSplhWVaibWBD1o2vRCs2dAye2XC8ApQgJn2dEDl9w4QKO44BFOM5zo5GhLppdmrh1FjaN5bi8IZAT7pw%2BOk5ELlviAvBWXQT7k/jDMpF5NMtPT7m41wLPs3%2BtPcL5wEcJIVI0hVQ3KjSMHwYhyGoehmHYbh%2BGERJjVrORVqUSRYtkSpzVipK0pgYN3GqqNmrjZS4u9E1lVKiqRu26QmbbQEdNAA%3D%3D%3D
).

Here are more results:

Compiler: [armv7-clang 9](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlICsupVs1qhkAUgBMAISnTSAZ0ztkBPHUqZa6AMKpWAVwC2tEJNJb0AGTy1MAOWMAjTMRAA2UgAdUCwuto9QxNBHz81Ohs7RyMXN3MlFQjaBgJmYgIg41ME5UxVANT0giiHZ1cPRTSMrJDOKuLSmLizAEpFVANiZA4AcgB6foBqABUATy9MIbHO4iG0LCGEV0xSIZIh1lRmdCGRIcwAD2YjL3YAOnEABgBBWwIhqgNaAroIe73WofEAdllboaAvbfADMABFgQAqIaScQg/43IHAuEQ5hDaEguEIpHETAELq0PZY643X5gkm9dqsEC9ACsvVIpl6VwZqBpOjkciGClmPW%2BkhBnAZBBpLNa7QA1mZaUIadwGUyWaQ2b0GQoQFdSCLmZTSHBYEg0Kc8OwyBQIEavCaKukjAA3H4AWmQwlEjoAnKQqCaCK51RAnKKGU5bOkxjShaQjUYtAQAPK0VjhnWkLBGETAdhB1N4XEFO2YdUpo75Ay%2BiMM%2B7KbOsPBOYhhvRYCta4h4IwV9o0ehMNgcHj8QSusScmRCOvqyDtVBeZJFx1xkFqvKvUwQSy1Uz1SxNcpuephfx0TehXxH2i72IVeqJfLJIo1fTZQS31cPkq2MpX/cNR/BLe/h%2B0R7lw7Q8l0PSgbKdIKtmKq2g6jpoi6GZDO6QwQLghAbFIgprHoxqmvygpfByMhyMKQbiqQUqSDK1K9PKjJwTSaoalqVF6jAiAoBgOD4EQZqUN2jAsFmA4COYw4oKO8ivskmjaCe5g7p%2BzQVDKh7JMp3hnskl4tDK8mFNUmRPiEuRJCZjRqSBRmmTpCimQZGlgbyfQfE8LzJJI7z0J83x/CSSIfE4oKouiMLEgCQIfMg4VDGFGLRYiQK4vixCEhI8IkmSuoMfSzEpiqZGyDI3LucRkiUTq1G0dw5zuk1zUtS1ngMUxiqsqxijsdqYpcfAECGqghGuOQlCWtabgIU6KFup63qsL6xD%2BoGKYhrQYYttGsYJkm2ZphmWYpvgeZqAWRZKiWyBln0kZVgxSq1vWjZ8S2BBth2vRCl2dCiX2XC8JJQgZjJ5Fji9k4QNOs4BPOi7LlZGjrkp5kAapwHfqe4QBCeB56QELk/sZdDvvjigrvezm2djN4OejL401jLScG5EEcJIVI0oVXXKjSs1IfMw5oRhWGCVV%2BGjVaRG4VzQylRRHG1ZK0rQZ1LGqr1mr9ZS3O9NVRVKiqNViu0BarQEIDcEAA%3D)
function: 8285 bytes
function2: **9958** bytes

Compiler: [Arduino UNO (1.8.9)](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlICsupVs1qhkAUgBMAISnTSAZ0ztkBPHUqZa6AMKpWAVwC2tEJNJb0AGTy1MAOWMAjTMRAA2UgAdUCwuto9QxNBHz81Ohs7RyMXN3MlFQjaBgJmYgIg41ME5UxVANT0giiHZ1cPRTSMrJDOKuLSmLizAEpFVANiZA4AcgB6foBqABUATy9MIbHO4iG0LCGEV0xSIZIh1lRmdCGRIcwAD2YjL3YAOnEABgBBWwIhqgNaAroIe73WofEAdllboaAvbfADMABFgQAqIaScQg/43IHAuEQ5hDaEguEIpHETAELq0PZY643X5gkm9dqsEC9ACsvVIpl6VwZqBpOjkciGClmPW%2BkhBnAZBBpLNa7QA1mZaUIadwGUyWaQ2b0GQoQFdSCLmZTSHBYEg0Kc8OwyBQIEavCaKul0AZbJ1aKhOAAOACcpCoJoIrnVECcooZTls6TGNKFpCNRi0BAA8rRWGGdaQsEYRMB2IGU3hcQUAG6YdXJo75Aw%2B8MM%2B7KLOsPBOYihvRYCta4h4IwV9o0ehMNgcHj8QTCUQoTkyIR19WQdqoLzJIsAWljILVeVepggllqpnqlia5Tc9TC/jo29CvhPtH3sQq9US%2BWSRRq%2Bmygnv66fJVsZRvh4az%2BCHd/y/aIDy4doeS6HpwNlOkFSzFUbmIO0HSGABVJ0hggThzhdc43S%2BCBcEIDYpEFNY9GNU1%2BUFL4ORkORhUDcVSClSQZWpXp5UZBCaTVDUtWYvUYEQFAMBwfAiDNShu0YFhMwHARzGHMQx3kd9kk0bQz3MPdv2aCoZWPZIdO8C9kmvFoZQ0wpqkyF8QlyJJbMafSwOsuzTIUOzLMMiDeT6D4nheZJJHeehPm%2BP4SSRD4nFBVF0RhYkASBD5kASoZ4oxFLESBXF8WIQkJHhEkyV1Tj6R45MVXo2QZG5AKaMkJidRYtjuHwt1up63qes8TjuMVVk%2BMUATtTFYT4AgQ1UCo1xyEoS1rTcW17SdZ5nXdT1vV9SgA2TYNaFDFsoxjeNEyzVN00zZN8FzNQCyLJUS2QMs%2BgjKtOKVWt60bcSWwINsO16IUuzoOS%2By4XglKEdNRwY8dfqnCAZznAJF2XVdnI0TdtIcoC9NA39z3CAIzyPcyAl8v8bLoT8KcUNdHx8tySbvTyCbfVniZaTh/KgjhJCpGkquG5UaSQlDMIw1AsJwvCCKw4ipOaii5qtaiyOFoY6sYwS2slaVYKG3jVTGzUJspEXeha6qlRVVqxXaAtiD8DRuCAA)
function: 3408 bytes
function2: **4238** bytes

ewertons: I could try getting the numbers for the actual code in the PR, so we can have a definitive comparison.

danewalton: shout out @ahsonkhan for the website 🙏

sdk/iot/hub/src/az_iot_hub_client_methods.c
danewalton: How about `AZ_IOT_INVALID_FEATURE_TOPIC`. So that we are specific in saying that whatever the topic is, it is not for this feature? `AZ_IOT_INVALID_TOPIC` leads me to believe the topic is not valid for all of IoT.

sdk/iot/hub/src/az_iot_hub_client_methods.c
danewalton: I would also agree with the way that Ewerton currently has this implemented. If the topic was passed as a reference, I would think otherwise. Here we can reuse it here until we either have success or fail (not for this feature). In either case, the `received_topic` will be popped off the stack and the original from the previous caller will be none the wiser.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.h
```diff
@@ -6,3 +6,4 @@ int test_iot_sas_token();
 int test_iot_hub_telemetry();
 int test_iot_hub_client();
 int test_az_iot_hub_client_twin();
+int test_iot_hub_methods();
```
danewalton, line 9: I think the Mac build is going to keep failing without a new line after this.

ewertons: @#$@%@ it! You might be right.

ewertons: Fixed.

sdk/iot/hub/src/az_iot_hub_client_methods.c
jspaith: Agreed with Ewerton here - if there's reason to leave code as is (and there is given size) then let's leave as-is.  Thanks for the tip.  "Weird, wild stuff.  I did not know that."

sdk/iot/hub/src/az_iot_hub_client_methods.c
jspaith: OK - I'm fine with proposal for parsed OK/not mine.  NIT: I'm not convinced though this is an error state -- error implies to me something unexpected that'd be akin to an exception.  This feels more like informational.  I don't have great idea of how to name this though.

RE potentially 3 error codes I'd considered -- you guys are right, not needed.  Even in the case where let's say we get a "/method/POST-thingy/first-partOK" that matches 1st part of topic, and followed by "total-garbage-that-we-can't-parse" I still think we're OK.  We'd return the NOT_MIne, and none of other handlers would process it either and application would still be in state of "I don't know how to process this."  So agree we don't need a special error code for the scenario in this paragraph.

danewalton: So two names possibly here:
`AZ_IOT_INVALID_FEATURE_TOPIC` or
`AZ_ERROR_INVALID_FEATURE_TOPIC`

Just like the core has multiple success cases here:
https://github.com/Azure/azure-sdk-for-c/blob/a86d152733e2dbec9106b230ba283ac6aae282d3/sdk/core/core/inc/az_result.h#L67-L68
We can have our own different IoT successes. Any thoughts?

jspaith: `INVALID` though may be strong?

If I pass a twin topic to this function, it's not topic's fault we passed it in the wrong place whereas invalid makes it sound like a bug in server.  Maybe `no_match` or I was actually even thinking about a `continue` already in this.  (In Windows structured exception handling, if your exception handler doesn't want the exception you mark it with a continue to let the kernel keep unwinding to the next exception handler or else the secondary exception (debugger/crash)).

sdk/iot/hub/src/az_iot_hub_client_methods.c
ahsonkhan: @ewertons, I am a bit confused by the results and exercise you shared.

> Using ARM gcc 8.3.1 none (with flag -O3 for optmization) these were the assembly sizes:
function: 4021 bytes
function2: 4827 bytes.

Where in the godbolt link do you see 4021 and 4827 bytes? That's **a lot** for a 2-4 line function. Where are these large numbers coming from? Also, the assembly looks identical to me in both cases.

Lastly, make sure to pick C and not C++ :)

sdk/iot/hub/src/az_iot_hub_client_methods.c
CIPop: +1 on invalid being strong - it may be valid.
AZ_ERROR_FEATURE_TOPIC_IGNORED + docs on saying this means the topic is not meant for this feature.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,106 @@
...
+
+  if (index == -1)
+  {
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+  }
+
+  received_topic = az_span_slice(
+      received_topic, index + az_span_length(methods_topic_prefix), az_span_length(received_topic));
+
+  index = az_span_find(received_topic, methods_topic_filter_suffix);
```
CIPop, line 57: I recommend just having the whole string ready for this function - since this runs for each receive.

`$iothub/methods/POST/"

That way you do one single scan, no further slicing for static parts.

You can then create the prefix from this longer const using some indexes.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,106 @@
...
+#include <az_result.h>
+#include <az_span.h>
+
+#include <_az_cfg.h>
+
+static const az_span methods_topic_prefix = AZ_SPAN_LITERAL_FROM_STR("$iothub/methods/");
+static const az_span methods_topic_filter_suffix = AZ_SPAN_LITERAL_FROM_STR("POST/");
+static const az_span methods_response_topic_result = AZ_SPAN_LITERAL_FROM_STR("res/");
+static const az_span methods_response_topic_properties = AZ_SPAN_LITERAL_FROM_STR("/?$rid=");
+static const uint8_t hash_tag = '#';
```
CIPop, line 17: We shouldn't define these then include stdint, etc. Makes code harder to read, longer compilation time, etc.
Simply use the "#" within the code - it's part of the contract and it must never change.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,106 @@
...
+#include <stdint.h>
+
+#include "az_iot_hub_client.h"
+#include <az_precondition_internal.h>
+#include <az_result.h>
+#include <az_span.h>
+
+#include <_az_cfg.h>
+
+static const az_span methods_topic_prefix = AZ_SPAN_LITERAL_FROM_STR("$iothub/methods/");
```
CIPop, line 13: See below: some reordering and splicing here will simplify the code.

sdk/iot/hub/src/az_iot_hub_client_methods.c
```diff
@@ -0,0 +1,106 @@
...
+static const az_span methods_response_topic_result = AZ_SPAN_LITERAL_FROM_STR("res/");
+static const az_span methods_response_topic_properties = AZ_SPAN_LITERAL_FROM_STR("/?$rid=");
+static const uint8_t hash_tag = '#';
+
+AZ_NODISCARD az_result az_iot_hub_client_methods_subscribe_topic_filter_get(
+    az_iot_hub_client const* client,
+    az_span mqtt_topic_filter,
+    az_span* out_mqtt_topic_filter)
+{
+  AZ_PRECONDITION_NOT_NULL(client);
```
CIPop, line 24: missing : UNUSED(client). 
This will not compile if you disable preconditions.

Potentially in other places - please try to compile w/o preconditions by adding a 
```C
#define NO_PRECONDITION_CHECKING
```

at the top of this file.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Ok, using AZ_ERROR_IOT_TOPIC_NO_MATCH

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Ok, using AZ_ERROR_IOT_TOPIC_NO_MATCH

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: I will mark that an enhancement question, and will save it for later. I wanna do a comparison of binary sizes before I change this.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: For this one, I'm ok either way. Doing as you suggest.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: We have a gap in tests for this. Need to address that.
Thanks.

sdk/iot/hub/src/az_iot_hub_client_methods.c
ewertons: Do you mean this is related to https://github.com/Azure/azure-sdk-for-c/pull/559#discussion_r406509514 ?
I will save that as an improvement comment, which I want to compare on the compiler before taking.

PR #563, opened 2020-04-09
Add prop next api

danewalton:


sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
```diff
@@ -741,6 +761,107 @@ static void test_az_iot_hub_client_properties_find_value_match_end_fail(void** s
...
+
+  assert_int_equal(az_iot_hub_client_properties_next(&props, &pair_out), AZ_OK);
+  assert_memory_equal(
+      az_span_ptr(pair_out.key), az_span_ptr(test_key_three), (size_t)az_span_length(test_key_three));
+  assert_memory_equal(
+      az_span_ptr(pair_out.value),
+      az_span_ptr(test_value_three),
+      (size_t)az_span_length(test_value_three));
+
+  assert_int_equal(az_iot_hub_client_properties_next(&props, &pair_out), AZ_ERROR_EOF);
```
ewertons, line 798: Perfect.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
```diff
@@ -741,6 +761,107 @@ static void test_az_iot_hub_client_properties_find_value_match_end_fail(void** s
...
+  uint8_t empty_prop_buf[8] = { 0 };
+  az_span test_span = az_span_init(empty_prop_buf, 0, _az_COUNTOF(empty_prop_buf));
+  az_iot_hub_client_properties props;
+  assert_int_equal(az_iot_hub_client_properties_init(&props, test_span), AZ_OK);
+
+  az_pair pair_out;
+
+  assert_int_equal(az_iot_hub_client_properties_next(&props, &pair_out), AZ_ERROR_EOF);
+}
+
```
ewertons, line 864: Should there be a test for when the properties are incorrectly formatted?
Would we want to have that validation in the function? I know it would be expensive.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -182,7 +182,7 @@ typedef struct az_iot_hub_client_properties
   struct
   {
     az_span properties;
-    uint8_t* current_property;
+    uint32_t current_property_index;
```
CIPop, line 185: Indexes should be signed.

CIPop: I think it's safer and faster to use the pointer as it will automatically take the address size of the platform.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -172,3 +172,30 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_find(
...
+    *out = AZ_PAIR_NULL;
+    return AZ_ERROR_EOF;
+  }
+
+  az_span remainder;
+  az_span prop_span = az_span_slice(properties->_internal.properties, index, prop_length);
+
+  out->key = az_span_token(prop_span, hub_client_param_equals_span, &remainder);
+  out->value = az_span_token(remainder, hub_client_param_separator_span, &remainder);
+  properties->_internal.current_property_index = az_span_length(remainder) == 0
```
CIPop, line 196: Please write the full if statement to make it easier to read and debug: your conditional is too complex for the short-hand notation.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
danewalton: I went back and forth on this. But for now, I don't expect hub to be giving us malformed properties. If it's user properties that the user has allocated in the device code (not input from hub), then I believe the burden is on the developer to have them properly formatted. In order to be slim and fast, I think here the developer should accept some responsibility.

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: The problem I ran into is it doesn't gel well with the `az_span_slice` api. There's another way to do this where I initialize a new span with the ptr from ^ that and then calculate length between the span length and pointer. It seems a little more complicated to me though.

sdk/iot/hub/src/az_iot_hub_client.c
danewalton: gotcha updated

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: As to the point about signedness, the compiler didn't like pointer arithmetic resolving to signed integers.

PR #566, opened 2020-04-09
Add az_iot_get_status

danewalton:
While adding the twin parse API, I thought this seems like something we would want. We can set the enum value straight to the number from the response but that could be unsafe. Let me know your thoughts.

### DIFF
--- a/sdk/iot/core/inc/az_iot_core.h
+++ b/sdk/iot/core/inc/az_iot_core.h
@@ -60,9 +60,13 @@ typedef enum
 
 /**
  * @brief Get the #az_iot_status from an int
- * 
+ *
  * @param[in] status_int  The int with the status number.
  * @param[out]  status  The #az_iot_status* with the status enum.
+ * @return  The #az_result with the result of the get operation
+ *  @retval #AZ_OK  If the int is an #az_iot_status enum. `status` will be set to the according
+ * enum.
+ *  @retval #AZ_ERROR_ITEM_NOT_FOUND  If the int is NOT an #az_iot_status enum
  */
 
 AZ_NODISCARD az_result az_iot_get_status(uint32_t status_int, az_iot_status* status);
--- a/sdk/iot/core/inc/az_iot_core.h
+++ b/sdk/iot/core/inc/az_iot_core.h
@@ -69,7 +69,7 @@ typedef enum
  *  @retval #AZ_ERROR_ITEM_NOT_FOUND If the int is NOT an #az_iot_status enum
  */
 
-AZ_NODISCARD az_result az_iot_get_status(uint32_t status_int, az_iot_status* status);
+AZ_NODISCARD az_result az_iot_get_status_from_uint32(uint32_t status_int, az_iot_status* status);
 
 /**
  * @brief Checks if the status indicates a successful operation.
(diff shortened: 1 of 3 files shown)


sdk/iot/core/inc/az_iot_core.h
```diff
@@ -58,6 +58,19 @@ typedef enum
   AZ_IOT_STATUS_TIMEOUT = 504,
 } az_iot_status;
 
+/**
+ * @brief Get the #az_iot_status from an int
+ *
+ * @param[in] status_int  The int with the status number.
+ * @param[out]  status  The #az_iot_status* with the status enum.
+ * @return  The #az_result with the result of the get operation
```
ericwol-msft, line 66: there are extra spaces here

danewalton: I used tab instead of space. Will change.

sdk/iot/core/src/az_iot_core.c
```diff
@@ -10,6 +10,71 @@
 
 #include <_az_cfg.h>
 
+az_result az_iot_get_status(uint32_t status_int, az_iot_status* status)
+{
+  switch(status_int)
```
ericwol-msft, line 15: nit, add a space

danewalton: running the clang formatter on the file

sdk/iot/core/src/az_iot_core.c
```diff
@@ -10,6 +10,71 @@
 
 #include <_az_cfg.h>
 
+AZ_NODISCARD az_result az_iot_get_status(uint32_t status_int, az_iot_status* status)
```
ericwol-msft, line 13: ok, I have been trying to think of a better name for this...but I can't :(

danewalton: it could get `az_iot_get_status_from_int()`?

danewalton: or `..from_uint32()`. kinda like the span API's

danewalton: https://github.com/Azure/azure-sdk-for-c/blob/e789649ee5d56003078e52ac96d45710c6715fac/sdk/core/core/inc/az_span.h#L270

danewalton: updated

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -58,6 +58,19 @@ typedef enum
...
+ *
+ * @param[in] status_int The int with the status number.
+ * @param[out] status The #az_iot_status* with the status enum.
+ * @return The #az_result with the result of the get operation.
+ *  @retval #AZ_OK If the int is an #az_iot_status enum. `status` will be set to the according
+ * enum.
+ *  @retval #AZ_ERROR_ITEM_NOT_FOUND If the int is NOT an #az_iot_status enum.
+ */
+
+AZ_NODISCARD az_result az_iot_get_status_from_uint32(uint32_t status_int, az_iot_status* status);
```
CIPop, line 72: I don't think this should be part of our public API.
The easiest way is to simply cast from an int/uint32_t. 

1. Why would we want the extra code on the device?
2. This is hiding the actual error if it's not in the list - I don't think that's a good idea.

PR #572, opened 2020-04-13
Remove global macro

antkmsft:
We should never declare non-prefiexed entities. So, this macro should've been named as `_az_UNUSED` if used internally, and `AZ_UNUSED` if this is the functionality that we provide as part of the SDK, and are committing to support forever in the future.
I don;t think we need this macro at all, because `(void)something` is the language-native way to say that a variable is unused. The same way that `INCREMENT(x)` does not need to be defined as `++x` (unless the implementation is different between platforms/compilers, or if we want it to do different things depending on the variables defined, such as precondition checks).

### DIFF
--- a/sdk/core/core/inc/_az_cfg.h
+++ b/sdk/core/core/inc/_az_cfg.h
@@ -91,8 +91,4 @@
 // Get the number of elements in an array
 #define _az_COUNTOF(array) (sizeof(array) / sizeof(array[0]))
 
-#ifndef UNUSED
-#define UNUSED(x) (void)(x)
-#endif
-
 #endif // _az_CFG_H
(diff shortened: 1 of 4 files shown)


CIPop:
I've added it originally because there are multiple ways to mark unused arguments. The issue was that they need to be marked in different ways for different compilers but (void)var seems to work on all we tested on.

antkmsft:
I understand. Other way would be to omit parameter name in the function definition if we're not using it, but given that we're compiling with maximum warnings, some compilers would start warning that the argument names mismatch, so you did the right thing, I just don't think we need to use macro, if it is unconditionally implemented and on all compilers the implementation is the same. If we want this in the future, we should call it as `AZ_UNUSED()` or `_az_UNUSED()`.

PR #574, opened 2020-04-13
Clean up MSVC warnings

antkmsft:


sdk/core/core/inc/_az_cfg.h
```diff
@@ -14,8 +14,9 @@
 
 #ifdef _MSC_VER
 
-// warning C4710: '...': function not inlined
```
gilbertw, line 17: Is this removed because we are building /W4?

antkmsft: To my best understanding, they all do - on my local machine with everything enabled, and in the CI. Thanks for the review Rick!

antkmsft: I removed it by commenting out every warning we disable one by one and see it it still builds without warnings. It looks like 2 warnings we had there did no longer apply, because the code has changed over the months.

sdk/core/core/inc/_az_cfg.h
```diff
@@ -14,8 +14,9 @@
...
-#pragma warning(error : 4710)
+// warning C4061: enumerator 'AZ_OK' in switch of enum 'az_result' is not explicitly handled by a
+// case label
+#pragma warning(disable : 4061)
```
ahsonkhan, line 19: Should we fix this in code (maybe by using if/else) instead of disabling the warning? We probably want to make sure that all enum cases are there within a switch/case statement.

sdk/core/core/inc/_az_cfg.h
gilbertw: This one wasn't a disable.  It was making it always an error

sdk/core/core/inc/_az_cfg.h
antkmsft: It does not make sense, and it will be impossible to support.
For instance, if you have an `enum az_error { AZ_OK, AZ_OVERFLOW, AZ_IOT_MQTT, ... 100 more... }`, it would require you to mention _every_ error _every_ switch:
```
switch (error)
{
case AZ_OK: ...
default: // no longer works, you have to handle all 100 errors here
}
```

sdk/core/core/inc/_az_cfg.h
antkmsft: Oops, thank you Rick! Here is the PR to add it back: https://github.com/Azure/azure-sdk-for-c/pull/576/files

sdk/core/core/inc/_az_cfg.h
antkmsft: So basically it will be almost impossible to write any switch/case statement, and for no good reason.

PR #568, opened 2020-04-13
[iot] Refactor SAS token module for IoT Hub Client

ewertons:
Includes movind code to proper location under az_iot_hub_client API and adding support to device modules.

Issue #340, opened 2020-02-07
[QUERY] az_span length and capacity as int32_t


danewalton:
Is there a technical reason why these values are `int32_t` instead of `uint32_t`.
[Link](https://github.com/Azure/azure-sdk-for-c/blob/1ed90fbe699c15cd308cd4c8dcd45c3e40205a31/sdk/core/core/inc/az_span.h#L22-L23)

gilbertw:
The only technical reason I can think of is looping and overflow.  Though I'm not sure that was the actual reason they are signed values.
C++ std:span uses a size_t for length / capacity.  I'm inclined to keep our span similar and switch to unsigned values as well.  @JeffreyRichter thoughts on this design decision?

danewalton:
Yea in my head signed values don't make sense for length or capacity

JeffreyRichter:
We specifically chose signed to help us find bugs. First, we think that a range of 2 billions is good enough, and that we don't need a range of 4 billion - if you think this assumption is wrong, let's hear about it.  Assuming the 2 billion range is good enough, then signed lets us do loops better and catch errors in any math calculations: if the number ever goes negative, we know we have a bug and we can hook up a debugger and have it watch for when the integer goes negative to find the line of code that has the bad math and fix it easily & quickly. 

I understand (and agree) that length and capacity can never be signed but this is why we chose signed; to find bugs quickly (and loops).

andre2007:
I also wondered about the data type and understand now, it is a safety measure to overcome a deficiency of C.

You may have a look at BetterC. While it almost looks like C and you only have to change little coding it comes with major benefits. You still have exact the same build results with exact the same C function signatures.

https://dlang.org/blog/2017/08/23/d-as-a-better-c/

And as example, changing a C application to BetterC:
https://dlang.org/blog/2018/06/11/dasbetterc-converting-make-c-to-d/

tl'dr:

> What may be initially most important to C programmers is memory safety in the form of array  overflow checking, no more stray pointers into expired stack frames, and guaranteed initialization of locals. This is followed by what is expected in a modern language — modules, function overloading, constructors, member functions, Unicode, nested functions, dynamic closures, Compile Time Function Execution, automated documentation generation, highly advanced metaprogramming, and Design by Introspection.

danewalton:
Few thoughts:
1. Every cast that you have to do incurs a possible point for a bug. It would be a good idea to reduce these to the extent we can.
[...]

PR #590, opened 2020-04-18
Azure IoT Provisioning service public API and inbound message processing

CIPop:
Public API changes:
- Changing from `span, out_span` to `char*, size, nullable *out_len` containing NULL-terminated strings.
- Remove custom payload parsing. (@jspaith FYI)
- Public API naming to match Azure SDK guidance.
- Adding SAS token documentation links.

Feature/bug fixes:
- Add inbound Provisioning message parsing. Optimizing string matching.
- Enable the Paho transport in CI.
- Fix User-Agent


sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    "registrationId":"paho-sample-device1",
+    "createdDateTimeUtc":"2020-04-10T03:11:13.0276997Z",
+    "assignedHub":"contoso.azure-devices.net",
+    "deviceId":"paho-sample-device1",
+    "status":"assigned",
+    "substatus":"initialAssignment",
+    "lastUpdatedDateTimeUtc":"2020-04-10T03:11:13.2096201Z",
+    "etag":"IjYxMDA4ZDQ2LTAwMDAtMDEwMC0wMDAwLTVlOGZlM2QxMDAwMCI="}}
+*/
+AZ_INLINE az_result az_iot_provisioning_client_payload_error_code_parse(
```
danewalton, line 190: should these parse methods be this instead to follow the changed "get" methods
```suggestion
AZ_INLINE az_result az_iot_provisioning_client_parse_payload_error_code(
```

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  {
+    *out_mqtt_topic_length = (size_t)required_length;
+  }
 
   return AZ_OK;
 }
 
 // Topic: $dps/registrations/GET/iotdps-get-operationstatus/?$rid=%s&operationId=%s
-AZ_NODISCARD az_result az_iot_provisioning_client_get_operation_status_publish_topic_get(
+AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic(
```
danewalton, line 437:
```suggestion
AZ_NODISCARD az_result az_iot_provisioning_client_get_query_status_publish_topic(
```

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -63,7 +65,8 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * @param[in] id_scope The ID Scope.
  * @param[in] registration_id The Registration ID. This must match the client certificate name (CN
  *                            part of the certificate subject).
- * @param[in] options A reference to an #az_iot_provisioning_client_options structure. Can be NULL.
+ * @param[in] options __[nullable]__ A reference to an 
+ *                                   #az_iot_provisioning_client_options structure. Can be NULL.
```
ericwol-msft, line 69: Can be NULL for default options.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -77,27 +80,37 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
  * @brief Gets the MQTT user name.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT user name.
- * @param[out] out_mqtt_user_name The output #az_span containing the MQTT user name.
+ * @param[out] mqtt_user_name A buffer with sufficient capacity to hold the MQTT user name.
+ *                            If successful, contains a null-terminated string with the user name
+ *                            that needs to be passed to the MQTT client.
+ * @param[in] mqtt_user_name_size The size, in bytes of #mqtt_user_name.
+ * @param[out] out_mqtt_user_name_length __[nullable]__ Contains the string length in bytes, of
```
ericwol-msft, line 87: Contains the string length, in bytes, of

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  az_span mqtt_user_name_span
+      = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
 
   const az_span* const user_agent = &(client->_internal.options.user_agent);
+  az_span str_registrations
+      = az_span_slice(str_dps_registrations_res, idx_registrations_start, idx_registrations_end);
 
   int32_t required_length = az_span_size(client->_internal.id_scope)
       + az_span_size(str_registrations) + az_span_size(client->_internal.registration_id)
       + az_span_size(provisioning_service_api_version);
```
hihigupt, line 76: Please add NULL additional to required_length calculation, that accounts for copy on line 95

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
   AZ_PRECONDITION_NOT_NULL(client);
-  AZ_PRECONDITION_VALID_SPAN(mqtt_client_id, 0, false);
-  AZ_PRECONDITION_NOT_NULL(out_mqtt_client_id);
+  AZ_PRECONDITION_NOT_NULL(mqtt_client_id);
+  AZ_PRECONDITION(mqtt_client_id_size > 0);
+
+  az_span mqtt_client_id_span
+      = az_span_init((uint8_t*)mqtt_client_id, (int32_t)mqtt_client_id_size);
 
   int required_length = az_span_size(client->_internal.registration_id);
```
hihigupt, line 119: same

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_client_id, required_length);
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_client_id_span, required_length);
 
-  az_span_copy(mqtt_client_id, client->_internal.registration_id);
+  az_span remainder = az_span_copy(mqtt_client_id_span, client->_internal.registration_id);
+  remainder = az_span_copy_u8(remainder, '\0');
```
hihigupt, line 124: >'\0' [](start = 41, length = 4)

can we either use '\0' or 0 consistently? Line 95 uses 0 and here we are using '\0'

sdk/iot/provisioning/src/az_iot_provisioning_client.c
hihigupt, line 190: >az_iot_provisioning_client_payload_error_code_parse [](start = 20, length = 51)

how about making it static function rather then inline? You are calling it two places.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    out_state->extended_error_code = (uint32_t)value;
+    out_state->status = (uint32_t)value / 1000;
+
+    return AZ_OK;
+  }
+
+  return AZ_ERROR_ITEM_NOT_FOUND;
+}
+
+AZ_INLINE az_result az_iot_provisioning_client_payload_registration_state_parse(
```
hihigupt, line 207: >az_iot_provisioning_client_payload_registration_state_parse [](start = 20, length = 59)

how about making it static function rather then inline? You are calling it two places.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+  }
+
+  bool found_assigned_hub = false;
+  bool found_device_id = false;
+
+  while ((!(found_device_id && found_assigned_hub))
+         && az_succeeded(az_json_parser_parse_token_member(jp, tm)))
+  {
+    if (az_span_is_content_equal(AZ_SPAN_FROM_STR("assignedHub"), tm->name))
```
hihigupt, line 223: >AZ_SPAN_FROM_STR("assignedHub") [](start = 33, length = 31)

how about making all these const data as static?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
hihigupt, line 190: >az_iot_provisioning_client_payload_error_code_parse [](start = 20, length = 51)

how about making it static function rather then inline? You are calling it two places.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    {
+      AZ_RETURN_IF_FAILED(az_json_parser_skip_children(jp, tm->token));
+    }
+    else
+    {
+      (void)az_iot_provisioning_client_payload_error_code_parse(tm, out_state);
+    }
+  }
+
+  return AZ_OK;
```
hihigupt, line 251: should we return error if (found_assigned_hub && found_device_id) != true?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
   return 0;
 }
 
+static void print_buffer(az_span span)
```
ericwol-msft, line 223: print_span_buffer?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
...
+{
+  int rc;
+  bool registration_complete = false;
+  char* topicName;
+  int topicLen;
+  MQTTClient_message* message;
+
+  while (!registration_complete)
+  {
+    if ((rc = MQTTClient_receive(mqtt_client, &topicName, &topicLen, &message, 60 * 1000))
```
ericwol-msft, line 244: 60 * 1000  -> define this? 1 minute in ms?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  }
+
+  if (!(found_registration_state && found_operation_id))
+  {
+    out_response->operation_id = AZ_SPAN_NULL;
+    out_response->registration_state = AZ_SPAN_FROM_STR("failed");
+
+    if (!found_error)
+    {
+      return AZ_ERROR_ITEM_NOT_FOUND;
```
hihigupt, line 326: item not found or parsing error?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
ericwol-msft: topicLen -> topicNameLen

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
...
+    putchar(*buffer++);
+  }
+
+  putchar('\n');
+}
+
+static int get_operation_status()
+{
+  int rc;
+  bool registration_complete = false;
```
ericwol-msft, line 237: verb?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+
+  AZ_RETURN_IF_FAILED(az_json_parser_init(&jp, received_payload));
+  AZ_RETURN_IF_FAILED(az_json_parser_parse_token(&jp, &tm.token));
+  if (tm.token.kind != AZ_JSON_TOKEN_OBJECT_START)
+  {
+    return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
+  }
+
+  out_response->registration_information = registration_state_default();
+
```
hihigupt, line 273: how about memset?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
...
+      printf("\tError timestamp: ");
+      print_buffer(response.registration_information.error_timestamp);
+      printf("\tError tracking ID: ");
+      print_buffer(response.registration_information.error_tracking_id);
+      if (response.retry_after_seconds > 0)
+      {
+        printf("\tRetry-after: %u seconds.", response.retry_after_seconds);
+      }
+
+      registration_complete = true;
```
ericwol-msft, line 326: is this false? if not, please add comments

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  AZ_PRECONDITION(mqtt_topic_size > 0);
 
-  int32_t required_length = az_span_size(str_dps) + az_span_size(str_registrations)
-      + az_span_size(str_put_iotdps_register);
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  az_span str_dps_registrations = az_span_slice(str_dps_registrations_res, 0, idx_res_start);
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic, required_length);
+  int32_t required_length
+      = az_span_size(str_dps_registrations) + az_span_size(str_put_iotdps_register);
```
hihigupt, line 420: >; [](start = 83, length = 1)

+1 missing for check of line 422

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
```
hihigupt, line 2: General comment: 
please make all the const in the code are some const at the top of file define static like you have done in line 15 for literal "&ClientVersion="

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -0,0 +1,271 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
```
hihigupt, line 2: Is there a line coverage stats that get generated for ctests in sdk-core? Reason I am asking so as to make sure that our parsing logic is properly tested with all branches covered.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
hihigupt, line 244: >60 * 1000 [](start = 79, length = 9)

is this timeout? Can you please declare some macro to associate context with magic number?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: the convention for things like this would be wise to use the default instead of memset. That way any changes to the default behavior gets updated here. Things aren't always guaranteed to be `0`.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
...
+      topicLen = (int)strlen(topicName);
+    }
+
+    printf("Message arrived\n");
+    printf("     topic (%d): %s\n", topicLen, topicName);
+    printf("   message (%d): ", message->payloadlen);
+    print_buffer(az_span_init(message->payload, message->payloadlen));
+
+    az_iot_provisioning_client_register_response response;
+    char query_topic[256];
```
hihigupt, line 264: Can we move all these buffer allocated on stack to data segment?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
...
+                &provisioning_client,
+                az_span_init((uint8_t*)topicName, topicLen),
+                az_span_init((uint8_t*)message->payload, message->payloadlen),
+                &response)))
+    {
+      printf("Message from unknown topic, return code %d\n", rc);
+      return rc;
+    }
+
+    if (az_span_is_content_equal(response.registration_state, AZ_SPAN_FROM_STR("assigning")))
```
hihigupt, line 277: >assigning [](start = 80, length = 9)

How about these schema DPS schema related values come from sdk dps header file: like macros defining these DPS schema key values?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -327,8 +383,10 @@ int main()
     return rc;
   }
 
-  printf("Started registration. [Press ENTER to abort]\n");
-  (void)getchar();
+  if ((rc = get_operation_status()) != 0)
+  {
+    return rc;
+  }
```
hihigupt, line 389: We should try to do clean shutdown. 
E.g: If you do not do clean shutdown when possible, you are hogging server side resources (at least by not sending RST packet),

Note: Every exit of app can be turn into bad shutdown, but that does not mean we should not try when possible.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -200,17 +175,9 @@ static int subscribe()
   int rc;
 
   char topic_filter[128];
```
hihigupt, line 177: can we move these buffer allocation out of stack?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
hihigupt: Well, agree, but we should design things that work in alignment with performance, right?

---
In reply to: [411636942](https://github.com/Azure/azure-sdk-for-c/pull/590#discussion_r411636942) [](ancestors = 411636942)

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
-    az_span mqtt_user_name,
-    az_span* out_mqtt_user_name)
+    char* mqtt_user_name,
+    size_t mqtt_user_name_size,
+    size_t* out_mqtt_user_name_length)
 {
   AZ_PRECONDITION_NOT_NULL(client);
-  AZ_PRECONDITION_VALID_SPAN(mqtt_user_name, 0, false);
-  AZ_PRECONDITION_NOT_NULL(out_mqtt_user_name);
+  AZ_PRECONDITION_NOT_NULL(mqtt_user_name);
```
ericwol-msft, line 64: we should have a not null or empty?

danewalton: since it's a `char*` now we check for not `NULL` here and the size is greater than one on the next line.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ericwol-msft: if it is a string, we should use '\0' to be clear it's a terminating null

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
danewalton: Code coverage for this PR can be found [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=361262&view=codecoverage-tab). 

`az_iot_provisioning_client_c.c` line coverage is at 94.9% and branch coverage is at 63.8%

sdk/iot/provisioning/src/az_iot_provisioning_client.c
danewalton: Yea I don't necessarily disagree. If we are good with that tradeoff we can go for memset here.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: Can you please explain the difference? Also, why does the number of call sites matter when picking between inline and static?

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -5,6 +5,8 @@
  * @file az_iot_hub_client.h
  *
  * @brief definition for the Azure IoT Hub client.
+ * @note The IoT Hub MQTT protocol is described at
```
ahsonkhan, line 8: I wonder how `@note` renders in comparison to `@remarks`.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -141,16 +157,21 @@ AZ_NODISCARD az_result az_iot_provisioning_client_sas_signature_get(
  * @param[in] key_name The Shared Access Key Name (Policy Name). This is optional. For security
  *                     reasons we recommend using one key per device instead of using a global
  *                     policy key.
- * @param[in] mqtt_password An empty #az_span with sufficient capacity to hold the MQTT password.
- * @param[out] out_mqtt_password The output #az_span containing the MQTT password.
+ * @param[out] mqtt_password A buffer with sufficient capacity to hold the MQTT client id.
+ *                           If successful, contains a null-terminated string with the client id
+ *                           that needs to be passed to the MQTT client.
+ * @param[in] mqtt_password_size The size, in bytes of #mqtt_password.
```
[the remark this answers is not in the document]
ahsonkhan: Let's converge on and use `\p` as that is what doxygen says (rather than back ticks): http://www.doxygen.nl/manual/commands.html#cmdp

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
ahsonkhan: Also, add back ticks around language keywords like NULL, true, false.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -165,15 +186,19 @@ AZ_NODISCARD az_result az_iot_hub_client_sas_password_get(
  * @brief Gets the MQTT topic filter to subscribe to register responses.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_topic_filter An empty #az_span with sufficient capacity to hold the MQTT topic
- *                              filter.
- * @param[out] out_mqtt_topic_filter The output #az_span containing the MQTT topic filter.
+ * @param[out] mqtt_topic_filter A buffer with sufficient capacity to hold the MQTT topic filter.
+ *                               If successful, contains a null-terminated string with the topic 
+ *                               filter that needs to be passed to the MQTT client.
+ * @param[in] mqtt_topic_filter_size The size, in bytes of #mqtt_topic_filter.
```
ahsonkhan, line 192: It is a bit odd that `mqtt_topic_filter` is [out] but the `mqtt_topic_filter_size` is [in]. Is the size, the maximum size we can write into the ptr?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -185,10 +210,11 @@ typedef struct az_iot_provisioning_client_registration_state
   az_span assigned_hub_hostname; /**< Assigned Azure IoT Hub hostname. @note This is only
                                     available if error_code is success. */
   az_span device_id; /**< Assigned device ID. */
-  az_span json_payload; /**< Additional JSON payload. */
-  az_iot_status error_code; /**< The register operation status. */
+  az_iot_status status; /**< The register operation status. */
   uint32_t extended_error_code; /**< The extended, 6 digit error code. */
   az_span error_message; /**< Error description. */
+  az_span error_tracking_id; /**< Submit this ID when asking for Azure IoT service-desk help. */
+  az_span error_timestamp; /**< Submit this ID when asking for Azure IoT service-desk help. */
```
ahsonkhan, line 217: Copy/paste error in docs. timestamp != ID.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -13,13 +13,17 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef _WIN32
```
ahsonkhan, line 16: When is this true and is `unistd.h` not available on windows? Maybe add a comment?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -85,7 +98,7 @@ static az_result read_configuration_entry(
 
 static az_result read_configuration_and_init_client()
 {
-  az_span endpoint_span = AZ_SPAN_LITERAL_FROM_BUFFER(global_provisioning_endpoint);
+  az_span endpoint_span = AZ_SPAN_FROM_BUFFER(global_provisioning_endpoint);
```
ahsonkhan, line 101: Are we building this sample now?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -262,6 +220,119 @@ static int register_device()
   return 0;
 }
 
+static void print_buffer(az_span span)
+{
+  char* buffer = (char*)az_span_ptr(span);
+  for (int i = az_span_size(span); i > 0; i--)
```
ahsonkhan, line 226: Why loop backwards?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -11,13 +12,19 @@
 static const az_span provisioning_service_api_version
     = AZ_SPAN_LITERAL_FROM_STR("/api-version=" AZ_IOT_PROVISIONING_SERVICE_VERSION);
 
-static const az_span str_dps = AZ_SPAN_LITERAL_FROM_STR("$dps");
-static const az_span str_registrations = AZ_SPAN_LITERAL_FROM_STR("/registrations/");
+static const az_span user_agent_version_prefix = AZ_SPAN_LITERAL_FROM_STR("&ClientVersion=");
+
+static const az_span str_dps_registrations_res
+    = AZ_SPAN_LITERAL_FROM_STR("$dps/registrations/res/");
+static const int32_t idx_registrations_start = 4;
```
ahsonkhan, line 19: Please add a comment explaining where these numbers came from. Something like right after `$dps` in `str_dps_registrations_res`.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -11,13 +12,19 @@
...
+static const int32_t idx_registrations_end = 19;
+
+static const int32_t idx_res_start = 19;
```
ahsonkhan, line 22: registration end and res start are the same number? Do we want one before the `/` and one after, or is this just to make the calling code more readable.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -77,27 +80,37 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+ *                                                      #mqtt_user_name. Can be NULL.
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_provisioning_client_user_name_get(
+AZ_NODISCARD az_result az_iot_provisioning_client_get_user_name(
     az_iot_provisioning_client const* client,
-    az_span mqtt_user_name,
-    az_span* out_mqtt_user_name);
+    char* mqtt_user_name,
+    size_t mqtt_user_name_size,
```
ahsonkhan, line 94: Why aren't these `int32_t`?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: Nice catch.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan, line 64: Should we add a precondition that `strlen(mqtt_user_name) == mqtt_user_name_size`?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  az_span str_registrations
+      = az_span_slice(str_dps_registrations_res, idx_registrations_start, idx_registrations_end);
 
   int32_t required_length = az_span_size(client->_internal.id_scope)
       + az_span_size(str_registrations) + az_span_size(client->_internal.registration_id)
       + az_span_size(provisioning_service_api_version);
   if (az_span_size(*user_agent) > 0)
   {
-    required_length += az_span_size(*user_agent) + 1;
+    required_length += az_span_size(user_agent_version_prefix) + az_span_size(*user_agent);
```
ahsonkhan, line 79: I wanted to ask this earlier, but is integer overflow possible here or is it very unlikely for anyone to be passing in such large buffers that `required_length` goes beyond int32 max given the domain is IoT devices? Is it OK for us to have undefined behavior in that case?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    remainder = az_span_copy(remainder, user_agent_version_prefix);
+    remainder = az_span_copy(remainder, *user_agent);
   }
 
-  *out_mqtt_user_name = az_span_slice(mqtt_user_name, 0, required_length);
+  remainder = az_span_copy_u8(remainder, 0);
+
+  if (out_mqtt_user_name_length)
+  {
+    *out_mqtt_user_name_length = (size_t)required_length;
```
ahsonkhan, line 99: nit: To be extra cautious, consider adding a precondition here that `_az_span_diff(remainder, mqtt_user_name_span) == required_length`

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    "status":"assigned",
+    "substatus":"initialAssignment",
+    "lastUpdatedDateTimeUtc":"2020-04-10T03:11:13.2096201Z",
+    "etag":"IjYxMDA4ZDQ2LTAwMDAtMDEwMC0wMDAwLTVlOGZlM2QxMDAwMCI="}}
+*/
+AZ_INLINE az_result az_iot_provisioning_client_payload_error_code_parse(
+    az_json_token_member* tm,
+    az_iot_provisioning_client_registration_state* out_state)
+{
+  if (az_span_is_content_equal(AZ_SPAN_FROM_STR("errorCode"), tm->name))
```
ahsonkhan, line 194: This is an incomplete comparison for JSON strings. Doing a byte-by-byte comparison for a known string won't catch escaped strings, where they are also semantically equal.

Either `tm->name` should be the unescaped string value, OR the json APIs should provide a comparison helper for callers.

Ideally: `"errorCode" == "\u0065rrorCode"`

For example, see the remark of the .NET API: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuetextequals?view=netcore-3.1#System_Text_Json_Utf8JsonReader_ValueTextEquals_System_ReadOnlySpan_System_Byte__

This is something we need to fix in the JSON API - we should file an issue. cc @JeffreyRichter, @vhvb1989

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  {
+    double value;
+    AZ_RETURN_IF_FAILED(az_json_token_get_number(&tm->token, &value));
+    out_state->extended_error_code = (uint32_t)value;
```
ahsonkhan, line 198: Is the `value` guaranteed to be a whole number such that this cast is lossless? Maybe add a comment.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    out_state->status = (uint32_t)value / 1000;
```
ahsonkhan, line 199: What is in `az_iot_status` and why is this division by 1000 ok? What does it represent?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    {
+      AZ_RETURN_IF_FAILED(az_json_parser_skip_children(&jp, tm.token));
+    }
+    else if (az_succeeded(az_iot_provisioning_client_payload_error_code_parse(
+                 &tm, &out_response->registration_information)))
+    {
+      found_error = true;
+    }
+
+    // else ignore token.
```
ahsonkhan, line 316: Should this be considered an error condition?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+
+  // Parse the status.
+  az_span remainder = az_span_slice(
+      received_topic, az_span_size(str_dps_registrations_res), az_span_size(received_topic));
+
+  az_span int_slice = az_span_token(remainder, AZ_SPAN_FROM_STR("/"), &remainder);
+  AZ_RETURN_IF_FAILED(az_span_atou32(int_slice, (uint32_t*)(&out_response->status)));
+
+  // Parse the optional retry-after= field.
+  az_span retry_after = AZ_SPAN_FROM_STR("retry-after=");
```
ahsonkhan, line 384: nit: consider making this a global literal span?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+  idx = az_span_find(remainder, retry_after);
+  if (idx != -1)
+  {
+    remainder = az_span_slice(remainder, idx + az_span_size(retry_after), az_span_size(remainder));
```
ahsonkhan, line 388: nit: Use `az_span_slice_to_end`

```suggestion
    remainder = az_span_slice_to_end(remainder, idx + az_span_size(retry_after));
```

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -47,161 +54,420 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+
+  int32_t idx = az_span_find(received_topic, str_dps_registrations_res);
+  if (idx != 0)
+  {
+    return AZ_ERROR_IOT_TOPIC_NO_MATCH;
+  }
+
+  // Parse the status.
+  az_span remainder = az_span_slice(
+      received_topic, az_span_size(str_dps_registrations_res), az_span_size(received_topic));
```
ahsonkhan, line 378: Same here/elsewhere. Use `az_span_slice_to_end`
```suggestion
  az_span remainder = az_span_slice_to_end(received_topic, az_span_size(str_dps_registrations_res));
```

sdk/iot/provisioning/src/az_iot_provisioning_client.c
JeffreyRichter: Yes, it would be nice to fix this (if easy/simple)

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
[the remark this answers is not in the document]
danewalton: The problem I have with `\p` is that it doesn’t work with multi word cases. For that you can either use <tt>...</tt> or backticks. I prefer the backticks so it’s consistent whether it’s single or multi word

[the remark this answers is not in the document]
ahsonkhan: The problem I have with `\p` is that it doesn’t work with multi word cases. 

How can you have a multi word case for parameter names? Aren't they, by definition, a single `snake_case` word?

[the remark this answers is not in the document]
danewalton: I mean any time you want the text to render in the `code` font. There are sometimes when you might want to put multi word things in code font and \p doesn’t work for that case.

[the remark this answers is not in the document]
ahsonkhan: That's fine, we could use back ticks for those (and if doxygen has some sort of `<code>` tag for multi-line, we can use that). For referencing parameters within docs though, which is specifically what I am referring to, we should consider using `\p`.

It also makes sense to use back ticks for language keywords like `true`, `false`, `NULL`, etc.

Btw, regarding multiple words in `code` font (which doxygen calls type writer font), we can use `<tt>`

> To have multiple words in typewriter font use `<tt>multiple words</tt>`.

There must be a reason why doxygen provided separate `\p` support (although reading the comment within the code is a little easier with back ticks).

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -77,27 +80,37 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+ *                                                      of #mqtt_client_id. Can be NULL.
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_provisioning_client_id_get(
+AZ_NODISCARD az_result az_iot_provisioning_client_get_client_id(
     az_iot_provisioning_client const* client,
-    az_span mqtt_client_id,
-    az_span* out_mqtt_client_id);
+    char* mqtt_client_id,
+    size_t mqtt_client_id_size,
```
ewertons, line 112: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -77,27 +80,37 @@ AZ_NODISCARD az_result az_iot_provisioning_client_init(
...
+    size_t* out_mqtt_client_id_length);
```
ewertons, line 113: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -141,16 +157,21 @@ AZ_NODISCARD az_result az_iot_provisioning_client_sas_signature_get(
...
  */
-AZ_NODISCARD az_result az_iot_hub_client_sas_password_get(
+AZ_NODISCARD az_result az_iot_provisioning_client_sas_get_password(
     az_iot_provisioning_client const* client,
     az_span base64_hmac_sha256_signature,
     az_span key_name,
-    az_span mqtt_password,
-    az_span* out_mqtt_password);
+    char* mqtt_password,
+    size_t mqtt_password_size,
```
ewertons, line 173: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -141,16 +157,21 @@ AZ_NODISCARD az_result az_iot_provisioning_client_sas_signature_get(
...
+    size_t* out_mqtt_password_length);
```
ewertons, line 174: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -230,33 +257,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_received_topic_payload_parse(
...
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_provisioning_client_get_operation_status_publish_topic_get(
+AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic(
     az_iot_provisioning_client const* client,
     az_iot_provisioning_client_register_response const* register_response,
-    az_span mqtt_topic,
-    az_span* out_mqtt_topic);
+    char* mqtt_topic,
+    size_t mqtt_topic_size,
```
ewertons, line 297: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -230,33 +257,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_received_topic_payload_parse(
...
+    size_t* out_mqtt_topic_length);
```
ewertons, line 298: I thought we were not going to use size_t (use int32_t instead). MIA?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
ahsonkhan: What does MIA stand for?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: https://github.com/Azure/azure-sdk-for-c/issues/608

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: We're using size_t. The reason is that char* here is a C string (as defined by the C99 standard - either a literal or the way the std libraries define strings). 
The purpose of the API change is to easily interoperate with C99's `sizeof` and `strlen`.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: That would make the code/intent harder to read and understand. 
Why are you proposing that?

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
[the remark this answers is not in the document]
CIPop: I agree with @ahsonkhan : the docs are clear on using \p for parameters. We can use MD-style for anything else.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I forgot to review these since they were internal. Makes sense.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: Yes. I'm now worried it's not clear from the docs. Ideas how to improve?

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: What do you mean? Win32 is Windows, Uni is Posix. This is very common, so I'd rather not add a comment.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Yes: it's IoT Provisioning/Hub protocol.
@ahsonkhan @JeffreyRichter I've also added #612 to track adding this (or alternate solution) to the parser.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: We'd like an option to disable that functionality and use no-copy / no transform operations (this would be one of those cases).
We would be relying on IoT protocol specification and save some CPU cycles for small, battery-operated MCUs.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
hihigupt: inline will duplicate the same code two places ... it will impact size. 

---
In reply to: [412496960](https://github.com/Azure/azure-sdk-for-c/pull/590#discussion_r412496960) [](ancestors = 412496960)

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: In this case I'd like to leave it as-is to clearly show the service supports two separate operations: `register` and `query_status`.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Thanks @hihigupt - I'll add UTs in a future PR.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: is it guaranteed that the service will never return a JSON payload that has escaped strings in them? Let's follow the data end-to-end to make sure. I am surprised you'd want to turn this off (in which case, we might not want to implement it at all). Is that safe to do? We can try to make the overhead as low as possible, in the case where unescaping isn't needed, but I want to make sure we do the correct thing first and capture requirements before discussing perf.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: @hihigupt @ahsonkhan I wrote the functions separately _only_ to make the code easier to read. Inlining functions usually leads to better optimizations (if optimized, it's not a blind copy).

CIPop: Same as above.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Not sure what you mean since this is already const and can only be used within this function.
Assuming you meant defining this at the top of the file as static const I don't agree: currently this is scoped only to this function which makes the code readable and harder to use in the wrong place.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
ahsonkhan: I am not familiar with this header. My question is: Why do we need to do this? What are we doing that is OS/platform specific?

To clarify: I wasn't asking about what Win32 means or suggesting we comment that :p

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: We want the chain-of-responsibility semantic: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/doc/mqtt_state_machine.md#receiving-apis
If the topic was meant for this particular API then we want it to return OK if we were able to parse the payload.

There are several levels of error reporting: API-level, request-level and operation-level. To avoid any confusion, the function returns errors only on API-level (cannot parse, protocol errors, etc).
If the topic/payload are correct, we return OK and require the application to check the request-level or operation level errors through the returned structure.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -141,16 +157,21 @@ AZ_NODISCARD az_result az_iot_provisioning_client_sas_signature_get(
  * @param[in] key_name The Shared Access Key Name (Policy Name). This is optional. For security
  *                     reasons we recommend using one key per device instead of using a global
  *                     policy key.
- * @param[in] mqtt_password An empty #az_span with sufficient capacity to hold the MQTT password.
- * @param[out] out_mqtt_password The output #az_span containing the MQTT password.
+ * @param[out] mqtt_password A buffer with sufficient capacity to hold the MQTT client id.
+ *                           If successful, contains a null-terminated string with the client id
```
ewertons, line 161: c&p typo here: "MQTT client id". Should be "MQTT password" .

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Changed to print_az_span.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Changed topicName -> topic instead.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: See az_result.h - in this case I think the error makes sense as we didn't find the item we were looking for.
I don't think `AZ_ERROR_PARSER_UNEXPECTED_CHAR` is an option here because it actually parsed correctly but the required items were not found.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I don't think memset is appropriate here. The compiler can generate optimized code depending on what the dev selects (size, speed, etc).

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: It's not: the registration operation is complete. It can complete with success or error.

CIPop: I'll change the var name to is_operation_complete since it's confusing.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I actually now think that having the consts at the top makes the code harder to read. I will move most of these within the functions unless they are used in two places.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
CIPop: I plan to add more UTs for limit checks and preconditions later.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: I don't think that would be a good idea for scalability: stack allocation means that the memory isn't always in-use. The reservation is made only within this stack frame.
Having these global means: 
1. We cannot have 2 registrations in parallel.
2. The memory is always allocated even if not used.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: You had this comment in the previous PR. In general it's good to fail fast and avoid corrupting further state. 
The sample does exist gracefully if the registration operation completes in either success or failure.

sdk/iot/hub/inc/az_iot_hub_client.h
CIPop: Found this answer about the differences: https://stackoverflow.com/questions/34372532/choosing-between-attention-note-remark-todo-and-warning-in-doxygen

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Yes, part of this PR.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: I know I had a good reason in some iteration :D.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Right: this is for readability and to avoid to declare global spans that would use more flash. (Maybe I'm over-optimizing here.)

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: To interop with C99's libs and `sizeof` and to avoid casting.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: That wouldn't be true: mqtt_user_name is a buffer that _will_ be filled with the actual name. Initially it can contain anything.
mqtt_user_name_size is the size of that buffer. We cannot verify that is true.

After a successful call we'll set the length (if not null) to strlen(mqtt_user_name) to avoid apps requiring it to call strlen (since we've already calculated the actual len).

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: It could in theory, very unlikely here since the values come from the application.
See https://github.com/Azure/azure-sdk-for-c/issues/340#issuecomment-588405246.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: > is it guaranteed that the service will never return a JSON payload that has escaped strings in them?

I didn't say that. What I said was that all Azure IoT service _protocols_ do not have any escaped characters within them.
This means that the feature will be disabled for our protocol-generated json portions. I.e. our parsers would skip the feature as it's never needed.

That said, users can add json objects (sub-sections) to the larger document. For example the "payload" element - https://github.com/Azure/azure-sdk-for-c/pull/590/files#diff-5392e071b1599b9fab2ffa611e15d224R102.

In these cases, customers may choose to enable the feature to allow escaped chars as well as any UTF-8 encoded multi-byte chars.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: I think global spans will help readability/maintainability more, but your call :)

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Good point. I'll need to think about the best way to achieve that: a function adapting to an enum vs defining the strings, etc.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
ahsonkhan: We could try to doc that it is the maximum available size of the buffer that this method can write into similar to `az_span_to_str`:
https://github.com/Azure/azure-sdk-for-c/blob/31b0dd7d8baea135505bac36d5145f6d018c7f56/sdk/core/core/inc/az_span.h#L217-L219

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I personally like having extra asserts to avoid bugs like the one @hihigupt pointed out above which would be detected by this kind of assert.
What I don't think is appropriate is
1. a "pre"condition which, I thought, were only to be used to verify parameters.
2. having code that cannot be tested in code coverage (if we do things right, this should never happen).

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Won't add a comment - extended_error_code is already public API and has documentation.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: `az_iot_status` and `az_iot_provisioning_client_registration_state.extended_error_code` are public API and have documentation.

```C
// (from enum) @brief Azure IoT service status codes.

  az_iot_status status; /**< The register operation status. */
  uint32_t extended_error_code; /**< The extended, 6 digit error code. */
```

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Definitely not: it would break if it is (why I've added the comment - to make it clear that else, we'll ignore the token).
In parallel, I've opened a task to log all the incoming payloads that we parse. If we will see customer issues we can inspect the logs to figure out the correct repro.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: You were right about idx_res_start - removing that one.
Global spans are larger though right (ptr, size) and this is just the index.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: I understand. I tend to re-use precondition also as something equivalent to .NET's `Debug.Assert` because they are closely related, which clearly isn't its purpose (like you said in 1). Usually code coverage not capturing `Debug.Assert` conditions is considered fine because it in itself is helping catch breaks in what a dev considered an invariant (for consistency in the internal logic) as the code evolves. 

@JeffreyRichter - what do you think about avoiding using the precondition for things outside of contract validation/parameter validation? Should that be avoided and discouraged, or does it not matter and we leave it to dev discretion?

Should we create a concept similar to `Debug.Assert` that extends precondition for more general purpose use?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I can't find other places - let me know if you saw another.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: That's fine. 

That said (just my 2cents), I think a local/in-place comment explaining that a cast from what is a double to  value is fine would help readers in the future. In isolation, I am sure anyone reading this code would wonder the same thing. It is clear that `extended_error_code` is a `uint32_t`, but what's not clear is why we are only taking the integer part of a float value. 

If we are going to fix this with #612 anyway, then the comment won't be necessary.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: Right, but none of this mentioned anything about division by 1000 and that there is some inherent relationship between status and extended_error_code. Are you relying on existing domain/service knowledge of the reader (which I don't have much of)? Then that's fine, but as someone maintaining the code, I will come back to it months from now and ask, why is this 1000 and what happens if I change it to 100?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: Do we expect having more tokens/info within the JSON than precisely what is required? OR is this about ignoring user defined component of the JSON payload?

If so, I think we may want to verify and "skip" past the "payload" JSON property (and its content), but still error on any other arbitrary token.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
JeffreyRichter: In the world of contracts there are preconditions, invariants, and post conditions, see https://en.wikipedia.org/wiki/Design_by_contract. A precondition (be definition) must be true prior to the execution of code. Typically, contracts are implemented via asserts. So, if we want some assert-like thing, that is fine, we can add it. But then, contracts would be implemented on top of it.

I think the PRECONDITION macros tell something to a customer and even our own developers so I like keeping this terminology/macro in our code at the top of function implementations. But, having an assert macro that does something similar is fine with me.

CIPop: In that case I won't add anything here. The check would validate if the code I wrote is correct. If it fails, there's nothing that the developer can do about it. I'll add UTs for this instead.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: We do expect more _valid_ tokens and we want to skip those until we read the ones we care about.

I think it's good to parse only the items we're interested in (for some reason the service is too verbose here). We're not validating the server for items that we don't expose/use or items that won't be useful for applications (at least to our knowledge today).

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
CIPop: Oh, I get it now. I'll add that comment. This is necessary for the sleep function.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: I kept it in-line with the span (buffer, size). I'd like to settle on the right names - changing these is time consuming.
Since the pattern I'm using was already approved by the API review board, please re-open the discussion in the next meeting.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: > If we are going to fix this with #612 anyway, then the comment won't be necessary.

:+1: 
I'll add a TODO pointing at 612. We need that functionality to allow non-FPU enabled MCUs to work without bringing in a floating point lib.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Makes sense. This should actually be a common function used by other services like Twin. 
@danewalton @ewertons - we'll need to unify extended-error to `az_iot_status` transformation. My proposal is to have an inline internal within az_iot_core.

@ahsonkhan I'll follow-up in another PR.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: > Global spans are larger though right (ptr, size) and this is just the index.

Yep, this is just trade offs around code readability/self documenting code vs. perf/size. You know the constraints well so I will defer to your judgement. An index + comment is likely a good enough compromise :)

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: Nope, no other places found.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I think I have a solution to meet both requirements: I'll encapsulate the indexes within two inline functions that will return the span.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
ahsonkhan: Parameter name aside, maybe we can just update the doc comment to be more clear:

```c
 * @param[in] mqtt_topic_filter_size The maximum available space, in bytes, within the buffer referred to by \p mqtt_topic_filter.
```

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -165,15 +187,19 @@ AZ_NODISCARD az_result az_iot_hub_client_sas_password_get(
  * @brief Gets the MQTT topic filter to subscribe to register responses.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_topic_filter An empty #az_span with sufficient capacity to hold the MQTT topic
- *                              filter.
- * @param[out] out_mqtt_topic_filter The output #az_span containing the MQTT topic filter.
+ * @param[out] mqtt_topic_filter A buffer with sufficient capacity to hold the MQTT topic filter.
+ *                               If successful, contains a null-terminated string with the topic
+ *                               filter that needs to be passed to the MQTT client.
+ * @param[in] mqtt_topic_filter_size The size, in bytes of \p mqtt_topic_filter.
```
ahsonkhan, line 193:
```suggestion
 * @param[in] mqtt_topic_filter_size The maximum available space, in bytes, within the buffer referred to by \p mqtt_topic_filter.
```

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -165,15 +187,19 @@ AZ_NODISCARD az_result az_iot_hub_client_sas_password_get(
  * @brief Gets the MQTT topic filter to subscribe to register responses.
  *
  * @param[in] client The #az_iot_provisioning_client to use for this call.
- * @param[in] mqtt_topic_filter An empty #az_span with sufficient capacity to hold the MQTT topic
- *                              filter.
- * @param[out] out_mqtt_topic_filter The output #az_span containing the MQTT topic filter.
+ * @param[out] mqtt_topic_filter A buffer with sufficient capacity to hold the MQTT topic filter.
+ *                               If successful, contains a null-terminated string with the topic
```
ahsonkhan, line 191: Similar docs feedback as https://github.com/Azure/azure-sdk-for-c/pull/568#discussion_r414145132

We should probably come up with some convention on documenting `az_result` returning APIs and apply it consistently.

I think providing the details on what `az_result` values can be returned and what they mean will be really helpful to callers. Whether we keep the similar description in the parameter docs, is up to you.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
ahsonkhan: If you agree, apply the similar wording across the board.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I've added `_az_iot_status_from_extended_status` within az_iot_provisioning_client.c 
I couldn't find any need to promote this to an internal within az_iot_core as Twin is not parsing any errors from payload (yet). 

If we need to have the 6 digit to 3 digit transformation done anywhere else, this should be moved to az_iot_core.h as an internal (or public) API. 

@danewalton @ewertons  @jspaith FYI /cc @hihigupt @ericwol-msft

ahsonkhan:
Q**uestion:** Why did you explicitly rebase/squash all your commits into one before merging the PR? We "squash and merge" when merging the PR anyway, so just curious. Having the commits within the PR (with breakdown on how the particular feedback was addressed), could be useful for some, especially for future reference.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: The current wording is in-line with az_span which has a ptr and size. Changes to doc wording should be made across the Az SDK. If we decide to document `size` as  "the maximum available space" let's do it across all APIs. Please open a tracking bug for that.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: That is non-trivial since we call other APIs and directly return their `az_result`.
@JeffreyRichter proposed that only results that should be handled by the application should be documented: e.g. if the incoming message parser attempted to parse something that was not meant for itself we can document the `AZ_ERROR_IOT_TOPIC_NO_MATCH`. 

Documenting common ones returned by all AZ_CORE won't scale and isn't really that useful since applications should treat them as unhandled exceptions (crash, report, restart) since usually there's nothing much that can be done to recover.

CIPop:
> Question: Why did you explicitly rebase/squash all your commits into one before merging the PR? We "squash and merge" when merging the PR anyway, so just curious. Having the commits within the PR (with breakdown on how the particular feedback was addressed), could be useful for some, especially for future reference.

I think that's up for the dev submitting the PR.
In this case the PR was large do to the large dependency changes and the merge-block in place for the last few weeks. Having the PR large meant that the review took too long and was blocking me from making progress. I rebased & squashed to simplify my work-flow when building on top of the PR branch.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
ahsonkhan: https://github.com/Azure/azure-sdk-for-c/issues/638

PR #591, opened 2020-04-20
Add wildcard to desired props

danewalton:
I was testing functionality and couldn't get the desired properties to work without using the wildcard. I suspect because of the properties sent on the topic. Verified with current SDK.


sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -20,7 +20,7 @@ static const az_span az_iot_hub_twin_get_pub_topic = AZ_SPAN_LITERAL_FROM_STR("$
 static const az_span az_iot_hub_twin_patch_pub_topic
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/reported/");
 static const az_span az_iot_hub_twin_patch_sub_topic
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/");
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/#");
```
ewertons, line 23: This seems like extra space we don't need to have. How about concatenating the hashtag from az_iot_hub_client_twin_hashtag?

danewalton: fixed

PR #592, opened 2020-04-20
Add twin parse API

danewalton:


sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -103,16 +119,118 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_patch_publish_topic_get(
...
+
+      // Get request id prop value
+      az_iot_hub_client_properties props;
+      az_span prop_span = az_span_slice(remainder, 1, az_span_size(remainder));
+      AZ_RETURN_IF_FAILED(
+          az_iot_hub_client_properties_init(&props, prop_span, az_span_size(prop_span)));
+      AZ_RETURN_IF_FAILED(az_iot_hub_client_properties_find(
+          &props, az_iot_hub_client_request_id_span, &out_twin_response->request_id));
+
+      if (out_twin_response->status == AZ_IOT_STATUS_NO_CONTENT)
```
danewalton, line 187: Caveat here for those reviewing: We currently can't map an error return value on the `res/#` to it's source. Meaning if we get a `$iothub/twin/res/400/` here, we can't natively tell if it was from a reported prop request or a get twin request because we don't have a queue of messages to resolve the request ID with. Proposal here could be to have a fourth `out_twin_response->response_type` here to designate it an error. Something like `AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_ERROR`.

sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -12,15 +12,17 @@
 #include <_az_cfg.h>
 
 static const uint8_t az_iot_hub_client_twin_hashtag = '#';
-static const az_span az_iot_hub_client_twin_request_id_suffix = AZ_SPAN_LITERAL_FROM_STR("?$rid=");
-
-static const az_span az_iot_hub_twin_response_sub_topic
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/");
-static const az_span az_iot_hub_twin_get_pub_topic = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/GET/");
+static const uint8_t az_iot_hub_client_twin_question = '?';
```
ewertons, line 15: nit: Also leaving comment that we later need to consolidate all these in a separate file, to avoid repeated vars throughout the SDK.

sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -12,15 +12,17 @@
...
+static const az_span az_iot_hub_twin_get_pub_topic = AZ_SPAN_LITERAL_FROM_STR("GET/");
+static const az_span az_iot_hub_twin_version_prop = AZ_SPAN_LITERAL_FROM_STR("$version");
 static const az_span az_iot_hub_twin_patch_pub_topic
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/reported/");
+    = AZ_SPAN_LITERAL_FROM_STR("PATCH/properties/reported/");
 static const az_span az_iot_hub_twin_patch_sub_topic
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/");
+    = AZ_SPAN_LITERAL_FROM_STR("PATCH/properties/desired/");
 
 AZ_NODISCARD az_result az_iot_hub_client_twin_response_subscribe_topic_filter_get(
```
ewertons, line 27: According to the last email from @CIPop summarizing the C99 meeting discussion, we need to change the functions to be something like: az_iot_hub_client_twin_response_get_subscribe_topic, or similar (get is not at the end). Is he doing that on a separate PR?

sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -103,16 +119,119 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_patch_publish_topic_get(
...
+  remainder = az_span_copy(remainder, az_iot_hub_client_request_id_span);
+  remainder = az_span_copy_u8(remainder, az_iot_hub_client_twin_equals);
+  remainder = az_span_copy(remainder, request_id);
 
   *out_mqtt_topic = az_span_slice(mqtt_topic, 0, required_length);
 
   return AZ_OK;
 }
+
+AZ_NODISCARD az_result az_iot_hub_client_twin_received_topic_parse(
```
ewertons, line 142: Same comment here for the name of the function, it should change according to C99 discussions, right? `parse` should come before `received`

sdk/iot/hub/src/az_iot_hub_client_twin.c
ewertons: On the other hand, we could leave that part of the parsing to the implementing layer, documenting it well. IT would be the most efficient way, although slightly more complicated for the customer to implement.

sdk/iot/hub/src/az_iot_hub_client_twin.c
danewalton: Added #605

sdk/iot/hub/src/az_iot_hub_client_twin.c
danewalton: I think those should be changed as we edit the files. AKA I can do that here

sdk/iot/hub/src/az_iot_hub_client_twin.c
danewalton: yes agreed. will change

sdk/iot/hub/src/az_iot_hub_client_twin.c
danewalton: Scratch that I will do it for the parse here and the other names in another PR. Actually I can do that with the `char*` updates.

ewertons:
LGTM

PR #593, opened 2020-04-20
Change C2D get api to use char*

danewalton:


sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -12,28 +12,38 @@
...
+  az_span mqtt_topic_filter_span
+      = az_span_init((uint8_t*)mqtt_topic_filter, (int32_t)mqtt_topic_filter_size);
   int32_t required_length = az_span_size(c2d_topic_prefix)
-      + az_span_size(client->_internal.device_id) + az_span_size(c2d_topic_suffix) + 1;
+      + az_span_size(client->_internal.device_id) + az_span_size(c2d_topic_suffix)
+      + (int32_t)sizeof(hash_tag);
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_filter, required_length);
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(
+      mqtt_topic_filter_span, required_length + (int32_t)sizeof(null_terminator));
```
danewalton, line 35: CALLOUT: Here the `sizeof` the `null_terminator` will be accounted for instead of in `required_length`. That way we don't have to add one above and then have to subtract it away before for `*out_mqtt_topic_filter_length`

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -12,28 +12,38 @@
...
-  AZ_PRECONDITION_NOT_NULL(out_mqtt_topic_filter);
+  AZ_PRECONDITION_NOT_NULL(mqtt_topic_filter);
+  AZ_PRECONDITION(mqtt_topic_filter_size > 0);
 
+  az_span mqtt_topic_filter_span
+      = az_span_init((uint8_t*)mqtt_topic_filter, (int32_t)mqtt_topic_filter_size);
   int32_t required_length = az_span_size(c2d_topic_prefix)
-      + az_span_size(client->_internal.device_id) + az_span_size(c2d_topic_suffix) + 1;
+      + az_span_size(client->_internal.device_id) + az_span_size(c2d_topic_suffix)
+      + (int32_t)sizeof(hash_tag);
```
danewalton, line 32: CALLOUT: Any individual `uint8_t` will have a call to `sizeof()` in the required size so that we can easily see which characters are accounted for and which are not. If we have a situation where 3+ characters are copied to the output and we just put a number here, it can be easy to lose track.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -300,15 +300,18 @@ AZ_NODISCARD az_result az_iot_hub_client_telemetry_publish_topic_get(
  * @note C2D MQTT Publish messages will have QoS At Least Once (1).
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
- * @param[in] mqtt_topic_filter An empty #az_span with sufficient capacity to hold the MQTT topic
+ * @param[out] mqtt_topic_filter A char buffer with sufficient capacity to hold the MQTT topic
```
ewertons, line 303: What about explicitly mentioning that the output string will be null terminated? It is not understandable from the description that it would be so.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -12,28 +12,38 @@
 
 static const az_span c2d_topic_prefix = AZ_SPAN_LITERAL_FROM_STR("devices/");
 static const az_span c2d_topic_suffix = AZ_SPAN_LITERAL_FROM_STR("/messages/devicebound/");
+static const uint8_t null_terminator = '\0';
```
ewertons, line 15: Should be marked for post-public-preview that we should consolidate all these constants in one file.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
ewertons: could just move the "+ sizeof(null_terminator) to the end of the sum stored in `required_length`

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: updated

ewertons:
Re-reviewed. Approved.

PR #611, opened 2020-04-22
added defaut user agent for client sdk version

ericwol-msft:


sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
```diff
@@ -24,17 +25,19 @@
...
 #define TEST_MODULE_ID "my_module_id"
 #define TEST_USER_AGENT "os=azrtos"
+#define PLATFORM_USER_AGENT "DeviceClientType=c/" AZ_SDK_VERSION_STRING
 
 static const az_span test_device_id = AZ_SPAN_LITERAL_FROM_STR(TEST_DEVICE_ID_STR);
 static const az_span test_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_HUB_HOSTNAME_STR);
 
 static const char test_correct_user_name[]
-    = "myiothub.azure-devices.net/my_device/?api-version=2018-06-30";
+    = "myiothub.azure-devices.net/my_device/?api-version=2018-06-30&" PLATFORM_USER_AGENT;
```
danewalton, line 34: Nit: I used to compose the "correct" test strings with a bunch of macros which I also used as the test params. After a little coaxing from @CIPop, I changed my mind to have test params and "correct" output to be unconnected. It provides an opportunity to sort of "double check" to make sure that its right instead of using the exact same value in param and output. So here, it would have whatever the `PLATFORM_USER_AGENT` spelled out in the string literal.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -18,7 +19,8 @@ static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-v
 
 AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default()
 {
-  return (az_iot_hub_client_options){ .module_id = AZ_SPAN_NULL, .user_agent = AZ_SPAN_NULL };
+  return (az_iot_hub_client_options){ .module_id = AZ_SPAN_NULL,
+                                      .user_agent = AZ_SPAN_LITERAL_FROM_STR("DeviceClientType=c/" AZ_SDK_VERSION_STRING) };
```
danewalton, line 23: can we have this be defined at the top as a static?

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
ericwol-msft: yea, but this changes per build because of PLATFORM_USER_AGENT contains the sdk version

PR #617, opened 2020-04-23
Change IoT API's to use char* and int instead of span

danewalton:
List of updates in this PR:
- Updated output of all get functions to use `char*` and length instead of `az_span` for the cases in which the returned "thing" will be input to some outside module/lib/function, which usually expect either a `NULL` terminated string or `char*` with length.
- Length and size for these functions are now `int32_t` instead of `size_t` to mirror sdk wide usage.
- Changed the names of functions to new guidelines where the verb comes after the module instead of at the end of the function. 
EX: `az_iot_hub_client_telemetry_publish_topic_get` -> `az_iot_hub_client_telemetry_get_publish_topic`
- Removed the unit tests which "get twice" since we don't append anything anymore (don't seem useful).
- Updated precondition on `az_iot_hub_client_methods_get_response_publish_topic` which requires the status to be one of three values (200, 404, 504). Link to docs for that here: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-direct-methods#response

Things needing more attention (since a lot of it is sort of cosmetic):
- The name for `az_iot_hub_client_twin_get_get_publish_topic`
- The reworked code for pnp telemetry topic to work (`az_iot_pnp_client_telemetry_get_publish_topic` and the static function `_az_add_telemetry_property`). @jspaith maybe for this one

### DIFF
--- a/sdk/iot/README.md
+++ b/sdk/iot/README.md
@@ -16,7 +16,7 @@ TBD
 
 To use IoT Hub connectivity, the first action by a developer should be to initialize the
 client with the `az_iot_hub_client_init()` API. Once that is initialized, you may use the
-`az_iot_hub_client_user_name_get()` and `az_iot_hub_client_client_id_get()` to get the
+`az_iot_hub_client_get_user_name()` and `az_iot_hub_client_client_id_get()` to get the
 user name and client id to establish a connection with IoT Hub.
 
 An example use case is below.
@@ -45,7 +45,7 @@ int main()
   az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, &options);
 
   //Get the MQTT user name to connect
-  az_iot_hub_client_user_name_get(&my_client, my_mqtt_user_name, 
+  az_iot_hub_client_get_user_name(&my_client, my_mqtt_user_name, 
                 sizeof(my_mqtt_user_name), &my_mqtt_user_name_length);
 
   //Get the MQTT client id to connect
--- a/sdk/iot/README.md
+++ b/sdk/iot/README.md
@@ -16,7 +16,7 @@ TBD
 
 To use IoT Hub connectivity, the first action by a developer should be to initialize the
 client with the `az_iot_hub_client_init()` API. Once that is initialized, you may use the
-`az_iot_hub_client_get_user_name()` and `az_iot_hub_client_client_id_get()` to get the
+`az_iot_hub_client_get_user_name()` and `az_iot_hub_client_get_client_id()` to get the
 user name and client id to establish a connection with IoT Hub.
 
 An example use case is below.
@@ -49,7 +49,7 @@ int main()
                 sizeof(my_mqtt_user_name), &my_mqtt_user_name_length);
 
   //Get the MQTT client id to connect
-  az_iot_hub_client_client_id_get(&my_client, my_mqtt_client_id, 
+  az_iot_hub_client_get_client_id(&my_client, my_mqtt_client_id, 
                 sizeof(my_mqtt_client_id), &my_mqtt_client_id_length);
 
   //At this point you are free to use my_mqtt_client_id and my_mqtt_user_name to connect using
--- a/sdk/iot/README.md
+++ b/sdk/iot/README.md
@@ -29,12 +29,12 @@ static az_span my_iothub_hostname = AZ_SPAN_LITERAL_FROM_STR("constoso.azure-dev
 static az_span my_device_id = AZ_SPAN_LITERAL_FROM_STR("contoso_device");
 
 //Make sure to size the buffer to fit the user name (100 is an example)
-static uint8_t my_mqtt_user_name_buffer[100];
-static az_span my_mqtt_user_name = AZ_SPAN_LITERAL_FROM_BUFFER(my_mqtt_user_name_buffer);
+static char my_mqtt_user_name[100];
+static size_t my_mqtt_user_name_length;
 
 //Make sure to size the buffer to fit the client id (16 is an example)
-static uint8_t my_mqtt_client_id_buffer[16];
-static az_span my_mqtt_client_id = AZ_SPAN_LITERAL_FROM_BUFFER(my_mqtt_client_id_buffer);
+static char my_mqtt_client_id_buffer[16];
+static size_t my_mqtt_client_id_length;
 
 int main()
 {
@@ -45,10 +45,12 @@ int main()
   az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, &options);
 
   //Get the MQTT user name to connect
-  az_iot_hub_client_user_name_get(&my_client, my_mqtt_user_name, &my_mqtt_user_name)
+  az_iot_hub_client_user_name_get(&my_client, my_mqtt_user_name, 
+                sizeof(my_mqtt_user_name), &my_mqtt_user_name_length);
 
   //Get the MQTT client id to connect
-  az_iot_hub_client_client_id_get(&my_client, my_mqtt_client_id, &my_mqtt_client_id);
+  az_iot_hub_client_client_id_get(&my_client, my_mqtt_client_id, 
+                sizeof(my_mqtt_client_id), &my_mqtt_client_id_length);
 
   //At this point you are free to use my_mqtt_client_id and my_mqtt_user_name to connect using
   //your MQTT client.
@@ -108,15 +110,13 @@ void my_telemetry_func()
   //Initialize the client to then pass to the telemetry API
   az_iot_hub_client_init(&my_client, my_iothub_hostname, my_device_id, NULL);
 
-  //Allocate a span with capacity large enough to put the telemetry topic.
-  //Optionally, the size can include space for a null terminator.
-  //Here, the buffer is zero initialized to null terminate the topic.
-  uint8_t telemetry_topic_buffer[64] = { 0 };
-  az_span topic_span = az_span_init(telemetry_topic_buffer,
-                sizeof(telemetry_topic_buffer) / sizeof(telemetry_topic_buffer[0]));
+  //Allocate a char buffer with capacity large enough to put the telemetry topic.
+  char telemetry_topic[64];
+  size_t telemetry_topic_length;
 
-  //Get the NULL terminated topic and put in topic_span to send the telemetry
-  az_iot_hub_client_telemetry_get_publish_topic(&my_client, NULL, topic_span, &topic_span);
+  //Get the NULL terminated topic and put in telemetry_topic to send the telemetry
+  az_iot_hub_client_telemetry_get_publish_topic(&my_client, NULL, telemetry_topic, 
+                                    sizeof(telemetry_topic), &telemetry_topic_length);
 }
 ```
 
--- a/sdk/iot/README.md
+++ b/sdk/iot/README.md
@@ -94,7 +94,7 @@ void my_property_func()
 
 ### Telemetry
 
-Telemetry functionality can be achieved by sending a user payload to a specific topic. In order to get the appropriate topic to which to send, use the `az_iot_hub_client_telemetry_publish_topic_get()` API. An example use case is below.
+Telemetry functionality can be achieved by sending a user payload to a specific topic. In order to get the appropriate topic to which to send, use the `az_iot_hub_client_telemetry_get_publish_topic()` API. An example use case is below.
 
 ```C
 //FOR SIMPLICITY THIS DOES NOT HAVE ERROR CHECKING. IN PRODUCTION ENSURE PROPER ERROR CHECKING.
@@ -116,7 +116,7 @@ void my_telemetry_func()
                 sizeof(telemetry_topic_buffer) / sizeof(telemetry_topic_buffer[0]));
 
   //Get the NULL terminated topic and put in topic_span to send the telemetry
-  az_iot_hub_client_telemetry_publish_topic_get(&my_client, NULL, topic_span, &topic_span);
+  az_iot_hub_client_telemetry_get_publish_topic(&my_client, NULL, topic_span, &topic_span);
 }
 ```
 
--- a/sdk/iot/doc/mqtt_state_machine.md
+++ b/sdk/iot/doc/mqtt_state_machine.md
@@ -132,7 +132,7 @@ _Example:_
         switch (twin_response.response_type)
         {
             case AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_GET:
-                // This is a response to a az_iot_hub_client_twin_get_get_publish_topic.
+                // This is a response to a az_iot_hub_client_twin_document_get_publish_topic.
                 break;
             case AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_DESIRED_PROPERTIES:
                 // This is received as the Twin desired properties were changed using the service client.
--- a/sdk/iot/doc/mqtt_state_machine.md
+++ b/sdk/iot/doc/mqtt_state_machine.md
@@ -124,7 +124,7 @@ _Example:_
     {
         // This is a Method request:
         //  method_request.name contains the method
-        //  method_request.request_id contains the request ID that must be used to submit the response using az_iot_hub_client_methods_get_response_publish_topic()
+        //  method_request.request_id contains the request ID that must be used to submit the response using az_iot_hub_client_methods_response_get_publish_topic()
     }
     else if (az_succeeded(ret = az_iot_hub_client_twin_parse_received_topic(client, received_topic, &twin_response)))
     {
@@ -138,7 +138,7 @@ _Example:_
                 // This is received as the Twin desired properties were changed using the service client.
                 break;
             case AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES:
-                // This is a response received after patching the reported properties using az_iot_hub_client_twin_get_patch_publish_topic().
+                // This is a response received after patching the reported properties using az_iot_hub_client_twin_patch_get_publish_topic().
                 break;
             default:
                 // error.
--- a/sdk/iot/doc/mqtt_state_machine.md
+++ b/sdk/iot/doc/mqtt_state_machine.md
@@ -93,7 +93,7 @@ The application is responsible for filling in the MQTT payload with the format e
 _Example:_
 
 ```C
-if(az_failed(az_iot_hub_client_telemetry_publish_topic_get(client, NULL, mqtt_topic, &mqtt_topic)))
+if(az_failed(az_iot_hub_client_telemetry_get_publish_topic(client, NULL, mqtt_topic, &mqtt_topic)))
 {
     // error.
 }
@@ -114,17 +114,17 @@ _Example:_
 
     //az_span received_topic is filled by the application.
 
-    if (az_succeeded(az_iot_hub_client_c2d_received_topic_parse(client, received_topic, &c2d_request)))
+    if (az_succeeded(az_iot_hub_client_c2d_parse_received_topic(client, received_topic, &c2d_request)))
     {
         // This is a C2D message: 
         //  c2d_request.properties contain the properties of the message.
         //  the MQTT message payload contains the data.
     }
-    else if (az_succeeded(ret = az_iot_hub_client_methods_received_topic_parse(client, received_topic, &method_request)))
+    else if (az_succeeded(ret = az_iot_hub_client_methods_parse_received_topic(client, received_topic, &method_request)))
     {
         // This is a Method request:
         //  method_request.name contains the method
-        //  method_request.request_id contains the request ID that must be used to submit the response using az_iot_hub_client_methods_response_publish_topic_get()
+        //  method_request.request_id contains the request ID that must be used to submit the response using az_iot_hub_client_methods_get_response_publish_topic()
     }
     else if (az_succeeded(ret = az_iot_hub_client_twin_parse_received_topic(client, received_topic, &twin_response)))
     {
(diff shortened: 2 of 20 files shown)


sdk/iot/doc/resources/iot_hub_flow.puml
```diff
@@ -26,32 +26,32 @@ az_iot_hub_client_sas_password_get --> application_mqtt_connect : password is a
...
-application_mqtt_connect --> az_iot_hub_client_twin_patch_subscribe_topic_filter_get
-az_iot_hub_client_twin_patch_subscribe_topic_filter_get --> application_mqtt_subscribe
+application_mqtt_connect --> az_iot_hub_client_twin_get_patch_subscribe_topic_filter
+az_iot_hub_client_twin_get_patch_subscribe_topic_filter --> application_mqtt_subscribe
 
 
-application_mqtt_subscribe --> az_iot_hub_client_twin_get_publish_topic_get
-az_iot_hub_client_twin_get_publish_topic_get --> application_mqtt_publish
+application_mqtt_subscribe --> az_iot_hub_client_twin_get_get_publish_topic
+az_iot_hub_client_twin_get_get_publish_topic --> application_mqtt_publish
```
CIPop, line 51: This won't work: we can change it to `twin_request`, `twin_document`.
/cc @ewertons @ericwol-msft @jspaith @hihigupt for ideas.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -89,14 +89,17 @@ AZ_NODISCARD az_result az_iot_hub_client_init(
  * {iothubhostname}/{device_id}/{module_id}/?api-version=2018-06-30&{user_agent}
  *
  * @param[in] client The #az_iot_hub_client to use for this call.
- * @param[in] mqtt_user_name An empty #az_span with sufficient capacity to hold the MQTT user name.
- * @param[out] out_mqtt_user_name The output #az_span containing the MQTT user name.
+ * @param[out] mqtt_user_name A char buffer with sufficient capacity to hold the MQTT user name.
```
CIPop, line 92: Please keep exactly the same format as in Provisioning for documentation. That way, if (when) we decide to change, it'll be simpler to just use a reg-ex.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -89,14 +89,17 @@ AZ_NODISCARD az_result az_iot_hub_client_init(
...
+ * @param[out] out_mqtt_user_name_length The optional output length of the mqtt user name. Can
+ * be `NULL`.
  * @return #az_result.
  */
 AZ_NODISCARD az_result az_iot_hub_client_user_name_get(
     az_iot_hub_client const* client,
-    az_span mqtt_user_name,
-    az_span* out_mqtt_user_name);
+    char* mqtt_user_name,
+    int32_t mqtt_user_name_size,
```
CIPop, line 101: These must be `size_t`.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -409,30 +426,36 @@ AZ_NODISCARD az_result az_iot_hub_client_methods_response_publish_topic_get(
...
- * @param[out] out_mqtt_topic_filter The output #az_span containing the MQTT topic filter.
+ * @param[out] mqtt_topic_filter A char buffer with sufficient capacity to hold the MQTT topic
+ *                              filter. On success, will be `NULL` terminated.
+ * @param[in] mqtt_topic_filter_size The size of the passed buffer. Must be greater than 0.
+ * @param[out] out_mqtt_topic_filter_length The optional output length of the mqtt topic filter. Can
+ * be `NULL`.
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_hub_client_twin_response_subscribe_topic_filter_get(
+AZ_NODISCARD az_result az_iot_hub_client_twin_get_response_subscribe_topic_filter(
```
CIPop, line 436: I think twin has several functionalities: twin_get, twin_patch, twin_response.
I think the API name should consider the functionality such that all API names end in `get_subscribe_topic_filter`, etc.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -482,15 +505,19 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
...
- * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @param[out] mqtt_topic A char buffer with sufficient capacity to hold the MQTT topic
+ *                              filter. On success, will be `NULL` terminated.
+ * @param[in] mqtt_topic_size The size of the passed buffer. Must be greater than 0.
+ * @param[out] out_mqtt_topic_length The optional output length of the mqtt topic filter. Can
+ * be `NULL`.
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
+AZ_NODISCARD az_result az_iot_hub_client_twin_get_get_publish_topic(
```
CIPop, line 515: We need a better name for `twin_get`: `twin_download_document` maybe?

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -499,15 +526,19 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_get_publish_topic_get(
...
- * @param[out] out_mqtt_topic The output #az_span containing the MQTT topic.
+ * @param[out] mqtt_topic A char buffer with sufficient capacity to hold the MQTT topic
+ *                              filter. On success, will be `NULL` terminated.
+ * @param[in] mqtt_topic_size The size of the passed buffer. Must be greater than 0.
+ * @param[out] out_mqtt_topic_length The optional output length of the mqtt topic filter. Can
+ * be `NULL`.
  * @return #az_result
  */
-AZ_NODISCARD az_result az_iot_hub_client_twin_patch_publish_topic_get(
+AZ_NODISCARD az_result az_iot_hub_client_twin_get_patch_publish_topic(
```
CIPop, line 536: Same as above: twin_patch_get_publish_topic (so that all of these end in `get_publish_topic`).

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: Yea I think I ran into some issue using `size_t` when doing the pnp changes when casting. I can try it again

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: I will wait till DPS is merged since some of the wording still seems to be in flux. But will be updated.

sdk/iot/README.md
```diff
@@ -116,7 +116,7 @@ void my_telemetry_func()
                 sizeof(telemetry_topic_buffer) / sizeof(telemetry_topic_buffer[0]));
 
   //Get the NULL terminated topic and put in topic_span to send the telemetry
-  az_iot_hub_client_telemetry_publish_topic_get(&my_client, NULL, topic_span, &topic_span);
+  az_iot_hub_client_telemetry_get_publish_topic(&my_client, NULL, topic_span, &topic_span);
```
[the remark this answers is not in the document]
danewalton: updated regardless :)

sdk/iot/doc/mqtt_state_machine.md
```diff
@@ -95,7 +95,7 @@ The application is responsible for filling in the MQTT payload with the format e
 _Example:_
 
 ```C
-if(az_failed(az_iot_hub_client_telemetry_publish_topic_get(client, NULL, mqtt_topic, &mqtt_topic)))
+if(az_failed(az_iot_hub_client_telemetry_get_publish_topic(client, NULL, mqtt_topic, &mqtt_topic)))
```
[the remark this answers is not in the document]
danewalton: updated

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: updated

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -53,32 +54,42 @@ AZ_NODISCARD az_result az_iot_pnp_client_init(
...
 
-  int32_t written = az_span_size(user_name);
-  remainder = az_span_slice_to_end(remainder, written);
+  // First get hub user name since it is unknown how long it will be
+  AZ_RETURN_IF_FAILED(az_iot_hub_client_user_name_get(
+      &client->_internal.iot_hub_client, mqtt_user_name, mqtt_user_name_size, &written));
 
-  int32_t required_length
-      = az_span_size(pnp_model_id) + az_span_size(client->_internal.root_interface_name) + 2;
+  az_span mqtt_user_name_span
```
jspaith, line 71: Nit: I think the code is easier to read if you do the length checks before creating the span, which may get not used if the `AZ_RETURN_IF_NOT_ENOUGH_SIZE` isn't satisfied.

Something like
```c
int required_length = ...;
AZ_WHATEVER_ASSERT_SHOULD_BE(required_length > mqtt_user_name_size);
az_span mqtt_user_name_span = ... // we'll know it'll be OK
```
```

File: sdk/iot/pnp/inc/az_iot_pnp_client.h [hunk 10]
```diff
@@ -182,17 +192,21 @@ AZ_NODISCARD AZ_INLINE az_result az_iot_pnp_client_get_sas_password(
...
  */
 AZ_NODISCARD az_result az_iot_pnp_client_telemetry_get_publish_topic(
     az_iot_pnp_client const* client,
     az_span component_name,
-    az_span mqtt_topic,
+    char* mqtt_topic,
+    int32_t mqtt_topic_size,
     void* reserved,
-    az_span* out_mqtt_topic);
+    int32_t* out_mqtt_topic_length);
```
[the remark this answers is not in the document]
jspaith: Actually.... I don't think we need a full on [Yalta Conference](https://en.wikipedia.org/wiki/Yalta_Conference) (as one of my buddies used to call overly drawn out philosophy meetings) on this :).  If you think this change is right, as I do, then just make it and see if we get feedback on next PR round.  It's not hard to change around if needed.

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -53,32 +54,42 @@ AZ_NODISCARD az_result az_iot_pnp_client_init(
...
-  AZ_PRECONDITION_VALID_SPAN(mqtt_user_name, 0, false);
-  AZ_PRECONDITION_NOT_NULL(out_mqtt_user_name);
+  AZ_PRECONDITION_NOT_NULL(mqtt_user_name);
+  AZ_PRECONDITION(mqtt_user_name_size > 0);
 
-  az_span remainder = mqtt_user_name;
-  az_span user_name;
-  AZ_RETURN_IF_FAILED(
-      az_iot_hub_client_user_name_get(&client->_internal.iot_hub_client, remainder, &user_name));
+  size_t written;
```
jspaith, line 65: Nit: consider a `written_size` or `written_length` or `bytes_written` (I have another comment on size v length).  I sometimes use `written` as a pointer argument that traverses through.

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -53,32 +54,42 @@ AZ_NODISCARD az_result az_iot_pnp_client_init(
...
+  AZ_RETURN_IF_FAILED(az_iot_hub_client_user_name_get(
+      &client->_internal.iot_hub_client, mqtt_user_name, mqtt_user_name_size, &written));
 
-  int32_t required_length
-      = az_span_size(pnp_model_id) + az_span_size(client->_internal.root_interface_name) + 2;
+  az_span mqtt_user_name_span
+      = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(remainder, required_length);
+  int32_t required_length = (int32_t)written + (int32_t)sizeof(pnp_client_param_separator)
```
jspaith, line 74: Nit: I think this is `required_size` because we're checking it against the size input param?

jspaith, line 74: We should consider whether this `required_size|length` should include the sizeof(null_terminalor) here.  I get we're trying to return what a strlen() would have returned on the output string buffer, but coming from Windows world at least they tend to include # of bytes and not length of string.  (See [this blast from the past for registry query](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexa)].

If we want to just have the strlen() value of the string, OK I guess, but we need to document it on each and every place that we use this pattern.  I had to look this up to be sure almost every time I used a Windows (or preferabbly Windows CE :)) API.

Needs word-smithing but a pattern could be something like:
```
 * @param[in] mqtt_client_id_size The size, in bytes of \p mqtt_client_id.  Must include enough room for trailing \0.
 * @param[out] out_mqtt_client_id_length __[nullable]__ Contains the string length, in bytes, of
 *                                                      of \p mqtt_client_id, not counting trailing \0.  Can be `NULL`.
```

sdk/iot/pnp/src/az_iot_pnp_client.c
jspaith: Actually - not a nit, please don't call this `written` in 1st place.  This is just the `original_mqtt_user_name_length` (needs a word smith).  I'd name it that, as written tends to be something incremented over course of function.

sdk/iot/pnp/src/az_iot_pnp_client.c
```diff
@@ -53,32 +54,42 @@ AZ_NODISCARD az_result az_iot_pnp_client_init(
...
 
-  az_span remainder = mqtt_user_name;
-  az_span user_name;
-  AZ_RETURN_IF_FAILED(
-      az_iot_hub_client_user_name_get(&client->_internal.iot_hub_client, remainder, &user_name));
+  size_t written;
 
-  int32_t written = az_span_size(user_name);
-  remainder = az_span_slice_to_end(remainder, written);
+  // First get hub user name since it is unknown how long it will be
```
jspaith, line 67: "since it is unknown how long it will be" - I think we should remove this.  We need to make this call in any case because we're writing this user name into the `mqtt_user_name` to build up protocol.  That segment of wording looks like we're just doing a length check but throwing away the results.

jspaith: Similar comment in the PnP telemetry file with same verbiage.  I'm promoting this beyond a nit:), because it kind of threw my code read at 1st.

danewalton: okay understood. The original intent of the comment was to say that we must get this part before going to calculate the required size for pnp.

sdk/iot/pnp/src/az_iot_pnp_client.c
danewalton: I would agree but currently `AZ_RETURN_IF_NOT_ENOUGH_SIZE` requires the first param to be a span. I can alternatively just have something like:

```c
if (required_size > mqtt_user_name_size)
{
    return AZ_ERROR_INSUFFICIENT_SIZE;
}
```
and do away with the macro.

danewalton: I think I will resolve this one and say that if we decide we want to do this (and I can ask the team), then we can do it across all the API's instead of just these.

sdk/iot/pnp/src/az_iot_pnp_client.c
danewalton: This currently matches the rest of the SDK and I think it's okay if it stays since it eventually gets assigned to `out_mqtt_topic_length` at the end. Seems to match.

sdk/iot/pnp/src/az_iot_pnp_client.c
danewalton: okay I have it changed to `hub_user_name_written_length`. let me know if you have an issue with that.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
danewalton: Yea I think it should be change around. I'll make it before the `char*` and length stuff. So the last three will be the `char*`, size, and out_length

sdk/iot/pnp/src/az_iot_pnp_client.c
jspaith: @danewalton - nit could it just be `hub_user_name_length`?  The written feels unnecessary to my eye at least.

sdk/iot/pnp/src/az_iot_pnp_client.c
jspaith: I do agree we should be consistent and I wouldn't block checkin trying to resolve this style issue.  

But I do think your `if` block up higher in comments is way easier to read even if more verbose, because we're checking before we're assigning.  Maybe (eventually) a macro that just deals with lengths and bringing in consistency?

sdk/iot/pnp/src/az_iot_pnp_client.c
danewalton: yup no problem will do. for user_name and topic

danewalton: updated @jspaith

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: updated

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: updated

sdk/iot/doc/resources/iot_hub_flow.puml
danewalton: updated

sdk/iot/hub/inc/az_iot_hub_client.h
danewalton: all have been updated

sdk/iot/pnp/src/az_iot_pnp_client.c
danewalton: Noted. For now we are not including the null terminator as `strlen()` does.

sdk/iot/hub/src/az_iot_hub_client_twin.c
```diff
@@ -11,6 +11,7 @@
 
 #include <_az_cfg.h>
 
+static const uint8_t null_terminator = '\0';
```
CIPop, line 14: Nit: I prefer using the actual chars throughout the code instead of defining them here. This makes the code harder to read as people now need to go to the top of the file to understand what these are.
I think it's useful to declare constants when they might change in the future or their meaning is unclear. The null terminator in C is always '\0'. Same goes for the characters below.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
```diff
@@ -93,68 +93,76 @@ static void test_az_iot_hub_client_init_NULL_hub_hostname_id_fails(void** state)
   assert_precondition_checked(az_iot_hub_client_init(&client, test_device_id, AZ_SPAN_NULL, NULL));
 }
 
-static void test_az_iot_hub_client_user_name_get_NULL_client_fails(void** state)
```
CIPop, line 96: Nit: we may want to simplify the cmocka state/(void)state some time in the future. 2x less lines to scroll through for each test.

sdk/iot/pnp/inc/az_iot_pnp_client.h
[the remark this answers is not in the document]
CIPop: @danewalton this is marked as resolved but I can't see the change. Unresolving to make this visible in the main discussion.

(I do agree we don't need to meet since the only requirements I got were around the 3 params. These need to be grouped for obvious reasons but the reserved can be in any position.)

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
danewalton: Yea I agree. As it stands the compiler flags prevent us from doing it. One of the builds fails

sdk/iot/hub/src/az_iot_hub_client_twin.c
danewalton: I'm willing to go one way or another. Is it okay if we have it open for a team discussion and go whichever way we collectively decide? Democracy and all

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client.c
CIPop: See provisioning tests: it requires a pragma where tests are added.

sdk/iot/pnp/src/az_iot_pnp_client.c
ahsonkhan: A length based macro seems like a reasonable addition. Same with just doing the explicit length check and not having a macro. The macro was more useful when span had 3 fields, because the required size check was a bit more complicated. Although, it is useful to have a centralized place that returns the `AZ_ERROR_INSUFFICIENT_SIZE` so if we were to change that, it only needs to be updated in one place.

danewalton: It would need to be changed at the macro but the arguments to it would need to be update too so slightly more intensive. But yes I agree. We can think about changing the macro since the check based on lengths/size numbers seems like it would be more appropriate

PR #624, opened 2020-04-24
IoT Provisioning length fix, adding tests

CIPop:
- Fixing length bug.
- Adding UTs for 100% lcov coverage

### DIFF
--- a/sdk/iot/provisioning/src/az_iot_provisioning_client.c
+++ b/sdk/iot/provisioning/src/az_iot_provisioning_client.c
@@ -76,15 +76,15 @@ AZ_NODISCARD az_result az_iot_provisioning_client_get_user_name(
   const az_span* const user_agent = &(client->_internal.options.user_agent);
   az_span str_registrations = get_str_registrations();
       
-  int32_t required_length = 1 + az_span_size(client->_internal.id_scope)
+  int32_t required_size = 1 + az_span_size(client->_internal.id_scope)
       + az_span_size(str_registrations) + az_span_size(client->_internal.registration_id)
       + az_span_size(provisioning_service_api_version);
   if (az_span_size(*user_agent) > 0)
   {
-    required_length += az_span_size(user_agent_version_prefix) + az_span_size(*user_agent);
+    required_size += az_span_size(user_agent_version_prefix) + az_span_size(*user_agent);
   }
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_user_name_span, required_length);
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_user_name_span, required_size);
 
   az_span remainder = az_span_copy(mqtt_user_name_span, client->_internal.id_scope);
   remainder = az_span_copy(remainder, str_registrations);
@@ -101,7 +101,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_get_user_name(
 
   if (out_mqtt_user_name_length)
   {
-    *out_mqtt_user_name_length = (size_t)required_length;
+    *out_mqtt_user_name_length = (size_t)required_size - 1;
   }
 
   return AZ_OK;
@@ -121,16 +121,16 @@ AZ_NODISCARD az_result az_iot_provisioning_client_get_client_id(
   az_span mqtt_client_id_span
       = az_span_init((uint8_t*)mqtt_client_id, (int32_t)mqtt_client_id_size);
 
-  int32_t required_length = 1 + az_span_size(client->_internal.registration_id);
+  int32_t required_size = 1 + az_span_size(client->_internal.registration_id);
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_client_id_span, required_length);
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_client_id_span, required_size);
 
   az_span remainder = az_span_copy(mqtt_client_id_span, client->_internal.registration_id);
   remainder = az_span_copy_u8(remainder, '\0');
 
   if (out_mqtt_client_id_length)
   {
-    *out_mqtt_client_id_length = (size_t)required_length;
+    *out_mqtt_client_id_length = (size_t)required_size - 1;
   }
 
   return AZ_OK;
@@ -151,9 +151,9 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
 
   az_span mqtt_topic_filter_span
       = az_span_init((uint8_t*)mqtt_topic_filter, (int32_t)mqtt_topic_filter_size);
-  int32_t required_length = 1 + az_span_size(str_dps_registrations_res);
+  int32_t required_size = 1 + 1 + az_span_size(str_dps_registrations_res);
 
-  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_filter_span, required_length);
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_filter_span, required_size);
 
   az_span remainder = az_span_copy(mqtt_topic_filter_span, str_dps_registrations_res);
   remainder = az_span_copy_u8(remainder, '#');
@@ -161,7 +161,80 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
 
   if (out_mqtt_topic_filter_length)
   {
-    *out_mqtt_topic_filter_length = (size_t)required_length;
+    *out_mqtt_topic_filter_length = (size_t)required_size - 1;
+  }
+
+  return AZ_OK;
+}
+
+
+// $dps/registrations/PUT/iotdps-register/?$rid=%s
+AZ_NODISCARD az_result az_iot_provisioning_client_register_get_publish_topic(
+    az_iot_provisioning_client const* client,
+    char* mqtt_topic,
+    size_t mqtt_topic_size,
+    size_t* out_mqtt_topic_length)
+{
+  (void)client;
+
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_NOT_NULL(mqtt_topic);
+  AZ_PRECONDITION(mqtt_topic_size > 0);
+
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  az_span str_dps_registrations = get_str_dps_registrations();
+
+  int32_t required_size
+      = 1 + az_span_size(str_dps_registrations) + az_span_size(str_put_iotdps_register);
+
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, required_size);
+
+  az_span remainder = az_span_copy(mqtt_topic_span, str_dps_registrations);
+  remainder = az_span_copy(remainder, str_put_iotdps_register);
+  remainder = az_span_copy_u8(remainder, '\0');
+
+  if (out_mqtt_topic_length)
+  {
+    *out_mqtt_topic_length = (size_t)required_size - 1;
+  }
+
+  return AZ_OK;
+}
+
+// Topic: $dps/registrations/GET/iotdps-get-operationstatus/?$rid=%s&operationId=%s
+AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic(
+    az_iot_provisioning_client const* client,
+    az_iot_provisioning_client_register_response const* register_response,
+    char* mqtt_topic,
+    size_t mqtt_topic_size,
+    size_t* out_mqtt_topic_length)
+{
+  (void)client;
+
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_NOT_NULL(mqtt_topic);
+  AZ_PRECONDITION(mqtt_topic_size > 0);
+
+  AZ_PRECONDITION_NOT_NULL(register_response);
+  AZ_PRECONDITION_VALID_SPAN(register_response->operation_id, 1, false);
+
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  az_span str_dps_registrations = get_str_dps_registrations();
+
+  int32_t required_size = 1 + az_span_size(str_dps_registrations)
+      + az_span_size(str_get_iotdps_get_operationstatus)
+      + az_span_size(register_response->operation_id);
+
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, required_size);
+
+  az_span remainder = az_span_copy(mqtt_topic_span, str_dps_registrations);
+  remainder = az_span_copy(remainder, str_get_iotdps_get_operationstatus);
+  remainder = az_span_copy(remainder, register_response->operation_id);
+  remainder = az_span_copy_u8(remainder, '\0');
+
+  if (out_mqtt_topic_length)
+  {
+    *out_mqtt_topic_length = (size_t)required_size - 1;
   }
 
   return AZ_OK;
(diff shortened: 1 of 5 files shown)


sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -151,23 +151,96 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
...
+
+// Topic: $dps/registrations/GET/iotdps-get-operationstatus/?$rid=%s&operationId=%s
+AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic(
+    az_iot_provisioning_client const* client,
+    az_iot_provisioning_client_register_response const* register_response,
+    char* mqtt_topic,
+    size_t mqtt_topic_size,
+    size_t* out_mqtt_topic_length)
+{
+  (void)client;
```
ahsonkhan, line 211: Do we intend to use the `client` in the future?

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -75,60 +91,162 @@ static void test_az_iot_provisioning_client_custom_options_get_username_succeed(
...
+      test_global_device_endpoint,
+      AZ_SPAN_FROM_STR(TEST_ID_SCOPE),
+      AZ_SPAN_FROM_STR(TEST_REGISTRATION_ID),
+      &options);
+  assert_int_equal(AZ_OK, ret);
 
-  char* expected_topic = "$dps/registrations/res/#";
+  char client_id[sizeof(TEST_REGISTRATION_ID) - 1];
+  memset(client_id, 0xCC, sizeof(client_id));
+  size_t client_id_len = 0xBAADC0DE;
```
ahsonkhan, line 125: Nice :p

3131949278 > int32.max (0x7FFFFFFF)

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -75,60 +91,162 @@ static void test_az_iot_provisioning_client_custom_options_get_username_succeed(
...
+  ret = az_iot_provisioning_client_get_user_name(
+      &client, user_name, sizeof(user_name), &user_name_len);
   assert_int_equal(AZ_OK, ret);
   assert_memory_equal(expected_username, user_name, strlen(user_name));
   assert_int_equal((uint8_t)0xCC, (uint8_t)user_name[strlen(user_name) + 1]);
+  assert_int_equal(strlen(expected_username), user_name_len);
 }
 
-static void test_az_iot_provisioning_client_get_subscribe_topic_filter_succeed()
+static void test_az_iot_provisioning_client_get_connect_info_insufficient_space_fails()
```
ahsonkhan, line 109: What is connect info?

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -75,60 +91,162 @@ static void test_az_iot_provisioning_client_custom_options_get_username_succeed(
...
+  ret = az_iot_provisioning_client_register_get_subscribe_topic_filter(
+      &client, topic, sizeof(topic), &topic_len);
+  assert_int_equal(AZ_OK, ret);
+  assert_memory_equal(expected_topic, topic, strlen(topic));
+  assert_int_equal((uint8_t)0xCC, (uint8_t)topic[strlen(topic) + 1]);
+  assert_int_equal(strlen(expected_topic), topic_len);
 }
 
-static void test_az_iot_provisioning_client_get_register_publish_topic_filter_succeed()
+static void test_az_iot_provisioning_client_get_subscribe_topic_filter_insufficient_space_fails()
```
ahsonkhan, line 177: Lol, such long test names really makes you miss objects/instances and nesting from other languages :)

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -130,13 +156,14 @@ static void test_az_iot_provisioning_client_received_topic_payload_parse_assigne
...
-test_az_iot_provisioning_client_received_topic_payload_parse_invalid_certificate_error_succeed()
+test_az_iot_provisioning_client_parse_received_topic_and_payload_invalid_certificate_error_succeed()
 {
   az_iot_provisioning_client client;
   az_span received_topic = AZ_SPAN_FROM_STR("$dps/registrations/res/401/?$rid=1");
-  az_span received_payload = AZ_SPAN_FROM_STR(
-      "{\"errorCode\":401002,\"trackingId\":\"" TEST_ERROR_TRACKING_ID "\","
-      "\"message\":\"" TEST_ERROR_MESSAGE_INVALID_CERT "\",\"timestampUtc\":\"" TEST_ERROR_TIMESTAMP "\"}");
+  az_span received_payload
+      = AZ_SPAN_FROM_STR("{\"errorCode\":401002,\"trackingId\":\"" TEST_ERROR_TRACKING_ID "\","
```
ahsonkhan, line 164: Add test with extra properties in the JSON to validate we are skipping them (if one already doesn't exist).

ahsonkhan: Also, do we care about the behavior of having repeated properties in the JSON? It looks like for us, first one wins and we ignore subsequent ones.

For example: "{errorCode: 1, errorCode: 2, ... }"

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -245,6 +272,94 @@ test_az_iot_provisioning_client_received_topic_payload_parse_allocation_error_st
...
+}
+
+static void
+test_az_iot_provisioning_client_received_topic_and_payload_parse_registration_state_not_found_fails()
+{
+  az_iot_provisioning_client client;
+  az_span received_topic = AZ_SPAN_FROM_STR("$dps/registrations/res/200/?$rid=1");
+  az_span received_payload
+      = AZ_SPAN_FROM_STR("{\"operationId\":\"" TEST_OPERATION_ID
+                         /*"\",\"status\":\"" TEST_STATUS_FAILED */ "\",\"registrationState\":{"
```
ahsonkhan, line 317: ~Why do we have a commented out property within the JSON?~

I was initially confused by this (especially since I saw status again below, but as part of the `registrationState` object), but I get that is the point of the test and the comment highlights what's missing.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -151,23 +151,96 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
 
   az_span mqtt_topic_filter_span
       = az_span_init((uint8_t*)mqtt_topic_filter, (int32_t)mqtt_topic_filter_size);
-  int32_t required_length = 1 + az_span_size(str_dps_registrations_res);
+  int32_t required_size = 1 + 1 + az_span_size(str_dps_registrations_res);
```
danewalton, line 154: can we make an explicit call to `sizeof()`?
```suggestion
  int32_t required_size = (int32_t)sizeof('#') + (int32_t)sizeof('\0') + az_span_size(str_dps_registrations_res);
```

danewalton: But in this case, along with my below suggestion, the sizeof null would be moved to the `AZ_RETURN_IF_NOT_ENOUGH_SIZE`

CIPop: That would be incorrect as '#' is an `int`. So in C, sizeof('#') would be 4. Which means your code would always require 6 bytes extra.
It's also extremely confusing as it will work as expected in C++ where sizeof('#') is 1.

Please check other code if you believe we have that mistake.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: Not that I know of.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -151,23 +151,96 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
...
+  AZ_PRECONDITION_VALID_SPAN(register_response->operation_id, 1, false);
+
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  az_span str_dps_registrations = _az_iot_provisioning_get_str_dps_registrations();
+
+  int32_t required_size = 1 + az_span_size(str_dps_registrations)
+      + az_span_size(str_get_iotdps_get_operationstatus)
+      + az_span_size(register_response->operation_id);
+
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, required_size);
```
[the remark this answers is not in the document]
CIPop: I agree. In that case I'll change back to `required_length` (as size in our terminology includes the null).
I also like the idea of defining the `null_terminator` to add clarity on what that 1 is.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: MQTT CONNECT is one of the MQTT message types. The information that our client is computing in connect is `user_name`, `client_id` and `password`.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: The idea here is to verify it's not changed if we error out.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
CIPop: Yeah and I can't easily read the unit, action and expected results either. Maybe we change to some other naming convention?

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client.c
```diff
@@ -139,6 +257,42 @@ static void test_az_iot_provisioning_client_get_operation_status_publish_topic_f
...
+  az_span operation_id = AZ_SPAN_LITERAL_FROM_STR(TEST_OPERATION_ID);
+  az_iot_provisioning_client_register_response response = { 0 };
+  response.operation_id = operation_id;
+
+  size_t topic_len = 0xBAADC0DE;
+  az_result ret = az_iot_provisioning_client_query_status_get_publish_topic(
+      &client, &response, topic, sizeof(topic), &topic_len);
+
+  assert_int_equal(AZ_ERROR_INSUFFICIENT_SPAN_SIZE, ret);
+  for (size_t i = 0; i < sizeof(topic); i++)
```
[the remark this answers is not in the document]
CIPop: Thought about creating another buffer and comparing that way... that would be slower, more code.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
CIPop: Won't add to this PR. The goal here was to get 100% coverage based on current AzSDK tools.
See my other PR enabling branch coverage: based on that we should definitely add the cases you mention and others (future PR though).

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
CIPop: Yeah: it's a very interesting choice of protocol schema :). Don't get me started...

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
CIPop: Also, I'd like to dig into the protocol itself more and uncover actual possible JSON messages returned from the service today (higher pri than the other tests).

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -17,13 +17,13 @@ static const az_span str_dps_registrations_res
...
 // /registrations/
-AZ_INLINE az_span get_str_registrations()
+AZ_INLINE az_span _az_iot_provisioning_get_str_registrations()
 {
   return az_span_slice(str_dps_registrations_res, 4, 19);
 }
 
 // $dps/registrations/
-AZ_INLINE az_span get_str_dps_registrations()
+AZ_INLINE az_span _az_iot_provisioning_get_str_dps_registrations()
```
hihigupt, line 26: macro for online function?

CIPop: I think inline is better than a macro. Let's postpone optimizations for later.
One thing that I will do is properly prefix all of these with `_`. If we decide to make changes, we can find them faster.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -167,7 +170,80 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
...
+  AZ_PRECONDITION_VALID_SPAN(register_response->operation_id, 1, false);
+
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  az_span str_dps_registrations = _az_iot_provisioning_get_str_dps_registrations();
+
+  int32_t required_length = az_span_size(str_dps_registrations)
+      + az_span_size(str_get_iotdps_get_operationstatus)
+      + az_span_size(register_response->operation_id);
+
+  AZ_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, required_length + (int32_t)sizeof((uint8_t)'\0'));
```
danewalton, line 230: just as a note here...no action items for you. But I have a bug task open to go through and consolidate all the commonly used `const`'s like the null terminator, question mark, etc into one of the common files. So this will in the future be changed to something like 
```c
.. + sizeof(null_terminator)
```
or something

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -167,7 +170,80 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_subscribe_topic_f
...
+  (void)client;
+
+  AZ_PRECONDITION_NOT_NULL(client);
+  AZ_PRECONDITION_NOT_NULL(mqtt_topic);
+  AZ_PRECONDITION(mqtt_topic_size > 0);
+
+  AZ_PRECONDITION_NOT_NULL(register_response);
+  AZ_PRECONDITION_VALID_SPAN(register_response->operation_id, 1, false);
+
+  az_span mqtt_topic_span = az_span_init((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
```
danewalton, line 223: Convo for another PR.... but John had an interesting thought for making this simpler. Below on line 230 we return if the size is not enough, but we don't actually need to go through to initialize the span in order to check that.
We would not init the span until after the check and line 230 would become something like
```c
if (required_length > mqtt_topic_size)
{
    return AZ_ERROR_SPAN_INSUFFICIENT_SIZE;
}
```
Thoughts?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I don't like hiding single characters behind named constants. '#' should not be `hash_tag`. It just makes it harder to read - now someone new to the code needs to go to some obscure top-level header file to figure out what otherwise is a single character.

The only benefit to define constants is to avoid spreading common values _that may change_ through the code. The null terminator will never change. Same thing goes for `'#'`.

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: My thinking on that is premature optimization for a case that's never supposed to happen at runtime.
What I propose instead is to run the samples against the real service under a profile. Optimize hot-paths.

PR #632, opened 2020-04-27
escape the useragent string

ericwol-msft:


sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -17,7 +17,7 @@ static const az_span hub_client_param_equals_span = AZ_SPAN_LITERAL_FROM_STR("="
 
 static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2018-06-30");
 static const az_span client_sdk_version
-    = AZ_SPAN_LITERAL_FROM_STR("DeviceClientType=c/" AZ_SDK_VERSION_STRING);
+    = AZ_SPAN_LITERAL_FROM_STR("DeviceClientType=c%2F" AZ_SDK_VERSION_STRING);
```
CIPop, line 20: Can we instead rethink the string to avoid the `/` which is incompatible with the MQTT protocol?

PR #633, opened 2020-04-27
Make _az_url_encode internal

antkmsft:
Ewerton has a request to make URL encoding function accessible.


sdk/core/core/internal/az_url_internal.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * @brief Copies character from the \p source #az_span to the \p destination #az_span by
+ * URL-encoding the \p source span characters.
+ *
+ * @param[in] destination The #az_span whose bytes will receive the URL-encoded source.
+ * @param[in] source The #az_span containing the non-URL-encoded bytes.
+ * @param[out] out_length A pointer to a #uint32_t that is going to be assigned the length
```
ahsonkhan, line 21: It isn't a uint32_t.
```suggestion
 * @param[out] out_length A pointer to an int32_t that is going to be assigned the length
```

sdk/core/core/internal/az_url_internal.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <stdint.h>
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * @brief Copies character from the \p source #az_span to the \p destination #az_span by
+ * URL-encoding the \p source span characters.
+ *
+ * @param[in] destination The #az_span whose bytes will receive the URL-encoded source.
```
ahsonkhan, line 19: nit:
```suggestion
 * @param[in] destination The #az_span whose bytes will receive the URL-encoded \p source.
```

sdk/core/core/src/az_aad.c
```diff
@@ -113,29 +59,36 @@ AZ_NODISCARD az_result _az_aad_build_body(
...
+
     remainder = az_span_copy(remainder, client_secret_key);
 
-    AZ_RETURN_IF_FAILED(_az_url_encode(client_secret, &remainder));
+    AZ_RETURN_IF_FAILED(_az_url_encode(remainder, client_secret, &url_length));
+    remainder = az_span_slice_to_end(remainder, url_length);
   }
 
-  *out_body = az_span_slice(body, 0, _az_span_diff(remainder, body));
+  *out_body = az_span_slice(body, 0, url_length);
```
ahsonkhan, line 91: Are you sure this `url_length` is correct and the `out_body` contains everything we need? Do we have tests to verify because it doesn't look like things like `"&client_secret="` are accounted for here. If we don't, please add some and all we are doing is returning a small slice of the body from the last `url_length` value.

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,66 @@
+#include "az_hex_private.h"
```
ahsonkhan, line 1: Needs license header.

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,66 @@
...
+    case '~':
+      return false;
+    default:
+      return !(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'));
+  }
+}
+
+AZ_NODISCARD az_result _az_url_encode(az_span destination, az_span source, int32_t* out_length)
+{
+  AZ_PRECONDITION_NOT_NULL(out_length);
```
ahsonkhan, line 26: precondition that destination size must be at least >= source?

antkmsft: It should be covered by `if (az_span_size(destination) < result_size)`

sdk/core/core/src/az_aad.c
antkmsft: Thank you I actually just debugged and fixed this, fix is coming soon.

sdk/core/core/src/az_url.c
antkmsft: Fixed.

sdk/core/core/src/az_aad.c
antkmsft: Done.

sdk/core/core/src/az_url.c
ahsonkhan: Right, I get that, but that is for the "dynamic" portion, where destination size being 2x source size might not be enough if everything needs encoding.

The precondition that destination must be at least as large as source seems like a reasonable thing the caller should make sure of. There is no case where smaller destination will succeed (since it has to be at least as big as the source).

sdk/core/core/src/az_aad.c
ahsonkhan: Add more unit tests to capture bugs like these in the future.

sdk/core/core/src/az_url.c
antkmsft: I think you're right. I'll add it.

sdk/core/core/src/az_aad.c
antkmsft: Thanks for the suggestion, I'm working on it, if everything plays out well, it it will be a next commit to this PR. Otherwise if it takes too much time, I'll put that to backlog.

sdk/core/core/src/az_url.c
antkmsft: After adding it and thinking more 2 hours later, maybe it is not the best behavior, after all.
for the cases when you convert statically defined things (AZ_SPAN_FROMSTRs), it is fine. It is also fine when you have two buffers - with input and output close to each other - then you can make sure one buffer is greater than or equal to each other.
But imagine you get the string to encode from a HTTP response, as part of JSON. And you pass one of the string values from JSON to encoder. Now, you may have some idea what the reasonable size for the input data, or maybe you only have as much buffer for the encoded result, and you want to make your best effort to encode. But you don't actually control the size of the input data. Let's say it can get huge on rare occasions, and the you have no options, but to fail, and that is fine logic for your device - for example, if the response gives too long of a string, you light up a red LED and ignore the data. Now, to be fully compliant with the contract, you'd have to have an external check for both output span being >= of input, AND the return code for AZ_ERROR_INSUFFICIENT_SPAN / not being AZ_OK.

antkmsft: Also, we probably shouldn't be checking for destination >= source, but `>`. `destination >= source + 2 characters`, to be exact. Because, if the size is the same, why are you calling this function at all?

antkmsft: But maybe we don't need to check. I am not sure. It is a borderline decision. After all, if you have one of these case where you don't control the source size relative to destionation, so be it - write that extra size check beore invocation. In other cases, you don't have to.

JeffreyRichter: It is possible that nothing in the source requires encoding and so the destination size CAN be equal to the source size.

sdk/core/core/src/az_aad.c
```diff
@@ -30,73 +29,20 @@ AZ_NODISCARD az_result _az_token_set(_az_token* self, _az_token const* new_token
...
-  *ref_destination = az_span_slice_to_end(*ref_destination, result_size);
-
-  return AZ_OK;
-}
-
 // https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-access-token
 AZ_NODISCARD az_result _az_aad_build_url(az_span url, az_span tenant_id, az_span* out_url)
 {
-  az_span root_url = AZ_SPAN_FROM_STR("https://login.microsoftonline.com/");
+  az_span const root_url = AZ_SPAN_FROM_STR("https://login.microsoftonline.com/");
```
gilbertw, line 35: This needs to be a configuration setting.  It can change based on which cloud the application is running.  By default we would use the above.
https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,69 @@
...
+  {
+    result_size += _az_url_should_encode(az_span_ptr(source)[i]) ? 3 : 1;
+  }
+
+  if (az_span_size(destination) < result_size)
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
+  }
+
+  uint8_t* p_s = az_span_ptr(source);
```
gilbertw, line 44: NitL p_s, p_d, s are not descriptive.  Its very easy to mix them up

ahsonkhan: Given this is primarily an implementation move, I would prefer any such changes be made in a separate PR so it can be reviewed independantly.

sdk/core/core/src/az_url.c
ahsonkhan: > Now, to be fully compliant with the contract, you'd have to have an external check for both output span being >= of input, AND the return code for AZ_ERROR_INSUFFICIENT_SPAN / not being AZ_OK.

I understand the concern here. But I think it is worth it to avoid unnecessary O(n^2) processing in the worst case and the caller should be bailing early if it is  clear that the dest is too small. To me, the best way to approach this is to write sample usage code that highlights how callers would approach this API in the case you are concerned with. 

My intuition is, keep the precondition of (dest >= src) as it is conveys intent clearly since there is no value in calling this function otherwise,and allows bailing out early.

ahsonkhan: FWIW, I think we should think about all this and fix it in a separate PR to let upstack consumes make progress.

For this PR, I recommend adding the precondition (or leave it as is if you prefer not to add it), add the extra tests, and check it in. :)

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,71 @@
...
+      return false;
+    default:
+      return !(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'));
+  }
+}
+
+AZ_NODISCARD az_result _az_url_encode(az_span destination, az_span source, int32_t* out_length)
+{
+  AZ_PRECONDITION_NOT_NULL(out_length);
+  AZ_PRECONDITION_VALID_SPAN(source, 0, true);
```
ahsonkhan, line 30:
```suggestion
  AZ_PRECONDITION_VALID_SPAN(source, 0, false);
```

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,71 @@
...
+
+  int32_t const input_size = az_span_size(source);
+  AZ_PRECONDITION_VALID_SPAN(destination, input_size, true);
```
ahsonkhan, line 33:
```suggestion
  AZ_PRECONDITION_VALID_SPAN(destination, input_size, false);
```

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,176 @@
...
+#include <cmocka.h>
+
+#include <_az_cfg.h>
+
+static void test_credential_client_secret(void** state)
+{
+  (void)state;
+
+  az_credential_client_secret credential = { 0 };
+  // init credential_credentials struc
```
ahsonkhan, line 25: remove or fix comment?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,176 @@
...
+      (void)ignore;
+
+      if (az_span_is_content_equal(AZ_SPAN_FROM_STR("authorization"), header.key))
+      {
+        assert_true(az_span_is_content_equal(AZ_SPAN_FROM_STR("Bearer AccessToken"), header.value));
+        has_auth_header = true;
+      }
+    }
+
+#ifdef MOCK_ENABLED
```
ahsonkhan, line 144: Why do we need this #ifdef if `send_request` is only ever called if `MOCK_ENABLED` is defined?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,176 @@
...
+    {
+      response->_internal.http_response = AZ_SPAN_FROM_STR("SubsequentResponse");
+    }
+  }
+
+  return AZ_OK;
+}
+
+#ifdef MOCK_ENABLED
+az_result __wrap_az_http_client_send_request(_az_http_request* request, az_http_response* response)
```
ahsonkhan, line 164: Who calls this?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include "az_credentials.h"
```
ahsonkhan, line 4: What changed in this file? I am asking to know what to review. If it was just a file rename, why didn't git just detect it as a rename from `test_az_aad.c` to `test_az_credential_client_secret.c`.

sdk/core/core/src/az_aad.c
antkmsft: Thank you Rick. Yes, I think it must not only be a config setting, but the client secret credential should be configurable for that, i.e. you should be able to create credentilas that use different auth URLs.

antkmsft: I updated the issue with my comment above.

sdk/core/core/src/az_url.c
antkmsft: @JeffreyRichter , if "It is possible that nothing in the source requires encoding and so the destination size CAN be equal to the source size" - then why to call url_encode() function at all?
If you tell me the scenario I described above - i.e. when at a compile time you can't reasonably guarantee that destination will always be greater than buffer, then why do we have "destination_size >= source_size" at all.

@ahsonkhan I think I am rather FOR the "destination_size >= source_size" after all. Exceptions, such as the one I described above, can be avoided by checking the size before invoking. But I think we need it to be "destination_size >= source_size + 2".

But I agree that "destination_size >= source_size + 2" is better to go as a separate PR.
And also I think that since we're pass things like "span_size + n" to AZ_PRECONDITION_SPAN_IS_VALID(), then SPAN_SIZE_IS_VALID() should check min_size for being >= 0, because it is possible to overflow an int, or mess up your math operations for min_sie calculations. You never want to say that "span size should be -5 or greater".

sdk/core/core/src/az_url.c
antkmsft: pointer to source, pointer to destination... I find it ok, but it is subjective. I can change it, I think I am fine with more descriptive names. Should I go with src_ptr, and dest_ptr?

sdk/core/core/src/az_url.c
antkmsft: No, I mean that null is fine, if size is 0.
Because why do I treat `az_span empty = { 0 }` and `az_span empty = AZ_SPAN_FROM_STR("")` differently?
but actually, I am working on the proposal PR to remove is_null_valid parameter all together.

sdk/core/core/src/az_url.c
antkmsft: Same as above.
`encode({ 0 }, { 0 }, &result)` is ok.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: CMocka does substitute the next call to az_http_client_send_request() with this. az_http_client_send_request() is called by client secret credential, and by the HTTP pipeline.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Missed it after copying-pasting. I'll remove it. And I will update in the place where I copied it from.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Thanks! I'll remove this ifdef.

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake-modules")
 
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
 option(BUILD_CURL_TRANSPORT "Build internal http transport implementation with CURL for HTTP Pipeline" OFF)
-option(UNIT_TESTING "Build unit test projects" OFF)
+option(UNIT_TESTING "Build unit test projects" ON)
```
antkmsft, line 10:
```suggestion
option(UNIT_TESTING "Build unit test projects" OFF)
```

CMakeLists.txt
```diff
@@ -90,4 +90,3 @@ if (UNIT_TESTING)
     add_subdirectory(sdk/samples/keyvault/keyvault/test/cmocka)
   endif()
 endif()
```
antkmsft, line 92:
```suggestion
endif()

```

sdk/core/core/src/az_url.c
JeffreyRichter: Because the caller doesn't know whether there is any characters that require encoding. The caller has some string, they don't know what's in it. But, some of the characters MAY require encoding so this call this function to do the work. Otherwise, if the caller has to examine the string characters first, see that at least 1 needs encoding and then call the encode function, then the string could be traversed multiple times which is inefficient AND an inconvenient programming model from the caller's perspective.

The encoding will definitely always fail if the destination can't hold the source so the precondition is useful to catch this bug.

CMakeLists.txt
antkmsft, line 92:
```suggestion
endif()

```

antkmsft, line 92:
```suggestion
endif()


```

sdk/core/core/src/az_aad.c
antkmsft: https://github.com/Azure/azure-sdk-for-c/issues/636

sdk/core/core/src/az_url.c
ahsonkhan: > But I agree that "destination_size >= source_size + 2" is better to go as a separate PR.

I don't think we should do the `+ 2`.

The caller shouldn't need to know the specific of this operation (such as which specific characters are being encoded, and that encoding turns 1 char into 3 - those are impl details). 

All they care about is that sanitizing their string by calling this function will make it url safe.

sdk/core/core/src/az_url.c
ahsonkhan: I like `src_ptr`, `dest_ptr`.

sdk/core/core/src/az_url.c
ahsonkhan: You don't want accidental null pointer access (or dereference).

This method is grabbing the az_span_ptr and I don't think we should allow null ptrs to avoid accidental bugs.

sdk/core/core/src/az_url.c
antkmsft: How all 3 things can be true at once?

```
void fn(az_span span_from_http_response_json)
az_span src1 = AZ_SPAN_FROM_STR("100percent") // 10 characters
az_span src2 = AZ_SPAN_FROM_STR("1hundred%") // 9 characters
az_span src3 = span_from_http_response_json; // we can't know what the size is.

uint8_t buf[10] = { 0 } // just enough to fit "100percent";
az_span_dst = AZ_SPAN_FROM_BUF(buf);

_az_url_encode(dst, src1, out); // can only succeed when a no-op, why to call encode at all?
_az_url_encode(dst, src2, out); // doomed to always fail, but preconditions will not fire
_az_url_encode(dst, src3, out); // we have no way to know, preconditions may fire depending on the input
```

sdk/core/core/src/az_url.c
ahsonkhan: I don't know if it should be. Why do you want to support this case?

sdk/core/core/src/az_url.c
ahsonkhan: Consider a broader look which includes more real usage with appropriate buffer size and az_result checks before/after. I don't see how all three are true at once as you just showcased potential inputs and it won't be the code that folks write anyway. I think the `why to call encode at all?` question has already been answered.

Maybe we should take this offline and get to a consensus/shared understanding.

antkmsft: In short, I don't understand how any of the arguments provided above is an argument against `dest_size >= source size + 2` precondition. We either don't need the size precondition, or it is +2.

sdk/core/core/src/az_url.c
antkmsft: I don't want my code to care whether the spaniter pointer is NULL or not when span size is 0.
I find it convenient to use { 0 } spans.
And it is true, { 0 } is an empty span. A result of URL enconding an empty span is an empty span, no problem.
https://github.com/Azure/azure-sdk-for-c/pull/642

sdk/core/core/src/az_url.c
antkmsft: Dereferencing any pointer without checkin the size first is an error already, regardless of whether it is pointer to null, orpointer to garbage. The size is 0, you should not be accessing the data that is out of bounds.
This code does not dereference the ponter, it checks the size, and only accesses the indices that are in range.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: It is a new test. And I completely removed the old one. I would not consider it a rename, as I wrote this test from scratch. Previous code did not check the client secret credential end to end, just the add part. It did not check whether the credential applies auth header to the request, it did not check whether the credential handles retriable errors, it did not check whether the credential will reuse the token when it did not expire.

ahsonkhan: Gotcha. Thanks.

Will the test now fail if we change the following line to catch incorrectly created body?

https://github.com/Azure/azure-sdk-for-c/blob/04ca0526cf798513d9640a345eedea664429aec9/sdk/core/core/src/az_aad.c#L138
 ```diff
-*out_body = az_span_slice(body, 0, _az_span_diff(remainder, body));
+*out_body = az_span_slice(body, 0, url_length);
```

File: sdk/core/core/src/az_url.c (hunk 7 above)
Comment (antkmsft) [reply sdk/core/core/src/az_url.c]: Done.

File: sdk/core/core/test/cmocka/test_az_credential_client_secret.c (hunk 10 above)
Comment (antkmsft) [reply sdk/core/core/test/cmocka/test_az_credential_client_secret.c:25]: Done.

File: sdk/core/core/test/cmocka/test_az_credential_client_secret.c (hunk 11 above)
Comment (antkmsft) [reply sdk/core/core/test/cmocka/test_az_credential_client_secret.c]: Done.

**Comment by ahsonkhan** [comment]
@antkmsft - let me know when the PR is ready for review :)

File: sdk/core/core/test/cmocka/test_az_credential_client_secret.c (hunk 13 above)
Comment (antkmsft) [reply sdk/core/core/test/cmocka/test_az_credential_client_secret.c]: Yes, I just confirmed that, it catches the bug.

File: sdk/core/core/src/az_url.c (hunk 9 above)
Comment (antkmsft) [reply sdk/core/core/src/az_url.c]: Per our earlier online conversation, I'll update another PR to allow null input spans, but disallow null output spans in preconditions.

File: sdk/core/core/src/az_url.c (hunk 8 above)
Comment (antkmsft) [reply sdk/core/core/src/az_url.c]: Per our online conversation, I'll update preconditions to allow NULL-ptr input spans (where applicable), and disallow output spans to have NULLs, even if input size is 0.

Comment (antkmsft) [reply sdk/core/core/src/az_url.c]: You can see example in this function, I'll apply this principle in more places in our code.

File: sdk/core/core/src/az_url.c (hunk 5 above)
Comment (antkmsft) [reply sdk/core/core/src/az_url.c]: I lost some energy to find the right answer, whether the `source`, or `source+2` is correct, and a few days after I am a bit disoriented now. Maybe you are right. There are two of you at least, so I am not going to add `+2` to this precondition.

File: sdk/core/core/src/az_aad.c (hunk 3 above)
Comment (antkmsft) [reply sdk/core/core/src/az_aad.c]: Done :) New test covers token retrieval, token caching, token refresh, and then token caching once again. It covers client secret credential end -to-end - from initialization to getting a HTTP response via the request with the authentication header set to the expected bearer token value.

**Comment by antkmsft** [comment]
> @antkmsft - let me know when the PR is ready for review :)
@ahsonkhan, it is ready now. There is one more change to undo an unneccessary new line in the CMakeLists.txt, but other than that, I expect the unit test to pass, and I would check it in.

File: sdk/core/core/test/cmocka/test_az_credential_client_secret.c [hunk 16]
```diff
@@ -0,0 +1,227 @@
...
+#include <cmocka.h>
+
+#include <_az_cfg.h>
+
+static void test_credential_client_secret(void** state)
+{
+  (void)state;
+
+  az_credential_client_secret credential = { 0 };
+  // init credential_credentials struc
```
ahsonkhan, line 25: Apply the fix for this comment again?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,227 @@
...
+        az_span_size(request_url),
+        AZ_SPAN_FROM_BUFFER(header_buf),
+        AZ_SPAN_FROM_BUFFER(body_buf));
+
+    az_http_response response = { 0 };
+    uint8_t response_buf[500] = { 0 };
+    ignore = az_http_response_init(&response, AZ_SPAN_FROM_BUFFER(response_buf));
+
+#ifdef _az_MOCK_ENABLED
+    will_return(__wrap_az_platform_clock_msec, ((i / 2) + 1) * 100000000);
```
ahsonkhan, line 73: Add a comment explaining the meaning behind this calculation. why are we doing `i / 2 + 1` and what does the multiplication by 100000000 mean?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,227 @@
...
+#endif // _az_MOCK_ENABLED
+    (void)ignore;
+  }
+}
+
+az_result send_request(_az_http_request* request, az_http_response* response);
+
+az_result send_request(_az_http_request* request, az_http_response* response)
+{
+  static bool redo_auth = false;
```
ahsonkhan, line 88: Does this have to be static?

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,227 @@
...
+
+int64_t __wrap_az_platform_clock_msec() { return (int64_t)mock(); }
+#endif // _az_MOCK_ENABLED
+
+int test_az_credential_client_secret()
+{
+  const struct CMUnitTest tests[] = {
+    cmocka_unit_test(test_credential_client_secret),
+  };
+  return cmocka_run_group_tests_name("az_core_credential_client_secret", tests, NULL, NULL);
```
ahsonkhan, line 226: I am having difficulty understanding what this test is trying to validate. It seems more like an end-to-end test rather than a unit test. Is it possible to also add unit tests that test the behavior of individual components or methods as well?

ahsonkhan: For example add unit test that only test the behavior of `_az_aad_build_url`, `_az_aad_build_body`, `_az_aad_request_token`, etc.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Yes, it has to be. The way CMocka works is that it allows you to inject a function in compile time, but only one. I use this static variable to remember the state between invocations - whether we do simulate the situation when token was reset already, or not.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: this is what az_platform_clock_msec() will return. I will comment.

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: I actually like this. This is much more thorough test than the one we had in the past, which did not caught a bug. This goes through all the flow realistically, making sure that we do retries, parse json, handle expiration, handle caching. And it does not invoke private functions like the previous one did - so you can change the internals, and as long as the implementation passes the test that is only using public/internal API, you know that you are good to go.

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,69 @@
...
+
+  int32_t result_size = 0;
+  for (int32_t i = 0; i < input_size; ++i)
+  {
+    result_size += _az_url_should_encode(az_span_ptr(source)[i]) ? 3 : 1;
+  }
+
+  if (az_span_size(destination) < result_size)
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
```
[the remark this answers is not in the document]
antkmsft: I guess we can. The code should handle the errors, and we're also on an embedded device, so it is better to do just the necessary minimum, but I guess we can afford this luxury :)

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Ha, good catch, thank you. I wonder how it got here, given that I don't like to comment self-documenting code. I think I got it via some set of either undos or merges, but I think it proves that my eyes are rather trained to ignore comments in code :)

sdk/core/core/src/az_url.c
```diff
@@ -0,0 +1,69 @@
...
+      return !(('0' <= c && c <= '9') || ('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'));
+  }
+}
+
+AZ_NODISCARD az_result _az_url_encode(az_span destination, az_span source, int32_t* out_length)
+{
+  _az_PRECONDITION_NOT_NULL(out_length);
+  _az_PRECONDITION_VALID_SPAN(source, 0, true);
+
+  int32_t const input_size = az_span_size(source);
```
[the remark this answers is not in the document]
antkmsft: Fixed

sdk/core/core/src/az_url.c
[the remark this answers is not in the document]
antkmsft: Fixed

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Fixed

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
antkmsft: Fixed

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,262 @@
...
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse3"),
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse4"),
+  };
+
+  // Cmocka works in a way that you have to pre-load it with a value for every time it will be
+  // invoked (tt does not resure previously set value).
+  int const clock_requests[]{
+    2, // wait to retry, set token expiration
+    1, // check if token has expired
+    3, // check if token has expired, wait to retry, settoken expiration
```
antkmsft, line 56:
```suggestion
    3, // check if token has expired, wait to retry, set token expiration
```

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,262 @@
...
+        az_span_size(request_url),
+        AZ_SPAN_FROM_BUFFER(header_buf),
+        AZ_SPAN_FROM_BUFFER(body_buf));
+
+    az_http_response response = { 0 };
+    uint8_t response_buf[500] = { 0 };
+    ignore = az_http_response_init(&response, AZ_SPAN_FROM_BUFFER(response_buf));
+
+#ifdef _az_MOCK_ENABLED
+    for (var j = 0; j < clock_requests[i]; ++j)
```
antkmsft, line 91:
```suggestion
    for (int j = 0; j < clock_requests[i]; ++j)
```

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
ahsonkhan: I agree that e2e tests are valuable and we should certainly keep them. I am not suggesting replacing one with the other. However, if the test were to fail, the developer will have very little idea on what went wrong or where the bug is, requiring them to debug. That is why having granular unit tests help. Generally, a unit test should at least help guide/narrow down where the bug could potentially be.

https://www.guru99.com/unit-test-vs-integration-test.html
https://stackoverflow.com/questions/5357601/whats-the-difference-between-unit-tests-and-integration-tests

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,258 @@
...
+
+  az_span const expected_responses[] = {
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse1"),
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse2"),
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse3"),
+    AZ_SPAN_FROM_STR("HTTP/1.1 200 OK\r\n\r\nResponse4"),
+  };
+
+  // Cmocka works in a way that you have to pre-load it with a value for every time it will be
+  // invoked (tt does not resure previously set value).
```
ahsonkhan, line 52: What is `tt`?
```suggestion
  // invoked (tt does not reset previously set value).
```

sdk/core/core/test/cmocka/test_az_credential_client_secret.c
```diff
@@ -0,0 +1,258 @@
...
+  int const clock_nrequests[] = {
+    2, // wait to retry, set token expiration
+    1, // check if token has expired
+    3, // check if token has expired, wait to retry, set token expiration
+    1, // check if token has expired
+  };
+
+  // Some value that is big enough so that when you add 3600000 milliseconds to it (1 hour),
+  // and while in debuger, the vallue you see is seen as "103600000", which is easy to debug.
```
ahsonkhan, line 61:
```suggestion
  // and while in debugger, the value you see is seen as "103600000", which is easy to debug.
```

PR #642, opened 2020-04-29
Proposal: Remove second parameter from az_span_is_valid()

antkmsft:
Why do we want to allow a function to accept `az_span x = AZ_SPAN_FROM_STR("")`, but not accept `az_span x = { 0 }`? Why do you want your code to even care whether the ptr is null, if the size is 0? Is it worth to carry that extra boolean parameter around?

I think the following:
```
(az_span) { .ptr = nonnull, .size = 0 } - valid span, given that you are fine with its size
(az_span) { .ptr = NULL, .size = 0 } - also valid span, given that you are fine with its size
(az_span) { .ptr = nonnull, .size = 5 } - valid span, given that you are fine with its size
(az_span) { .ptr = any, .size = -5 } - never a valid span
(az_span) { .ptr = NULL, .size = 5 } - not a valid span for our SDK*
```

If you agree with the above, you can see that the validity of span is an input of only one parameter: `size`. Depending on the size, `is_null_allowed = (size != 0)`.

`*` - I know that on embedded devices, 0x00000000 may be valid, for firmware updates, for example. 3 questions then:
1. Do we see it being realistic that it is our code that writes to address 0x00000000, and not some customer's code memcpy()?
2. If it is realistic, then how can we ever assume that 0x00000000 is not a valid value for a span anywhere in our code? The az_span_copy() proably should allow { .ptr = NULL, .size = 5 } then.
But, how can you be sure that the span you created can make to that final az_span_copy() call, given that az_span_copy()  is probably is going to be called from az_curl_http_response_handler(), and in order for the span to make it to az_curl_http_response_handler(), it needs to make it through the pipeline code and az_http_response_init(), which probably do not think that NULLptr span is fine.
I think that in preconditions, we should either allow (az_span) { NULL, 5 }, or disallow it then (null check could be an ifdef). But I don't think we need it now, so we should disallow this types of spans. 
3. If you say, yes, we only want the az_span_copy() to be able to accept spans like { NULL, 5 }. Then can we write a custom precondition check for just that one place, and not require us to pass `false` in 99% else of our code?

### DIFF
--- a/sdk/core/core/internal/az_precondition_internal.h
+++ b/sdk/core/core/internal/az_precondition_internal.h
@@ -61,33 +61,16 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
 #define AZ_PRECONDITION_NOT_NULL(arg) AZ_PRECONDITION((arg != NULL))
 #define AZ_PRECONDITION_IS_NULL(arg) AZ_PRECONDITION((arg == NULL))
 
-AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, bool null_is_valid)
+AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size)
 {
-  uint8_t* ptr = az_span_ptr(span);
+  uint8_t* const ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 
-  bool result = false;
-
-  /* Span is valid if:
-     The size is greater than or equal to a user defined minimum value AND one of the
-     following:
-        - If null_is_valid is true and the pointer in the span is null, the size must also be 0.
-        - In the case of the pointer not being NULL, the size is greater than or equal to zero.
-  */
-  if (null_is_valid)
-  {
-    result = ptr == NULL ? span_size == 0 : span_size >= 0;
-  }
-  else
-  {
-    result = ptr != NULL && span_size >= 0;
-  }
-
-  return result && min_size <= span_size;
+  return span_size >= 0 && (span_size == 0 || ptr != NULL) && span_size >= min_size;
 }
 
-#define AZ_PRECONDITION_VALID_SPAN(span, min_size, null_is_valid) \
-  AZ_PRECONDITION(az_span_is_valid(span, min_size, null_is_valid))
+#define AZ_PRECONDITION_VALID_SPAN(span, min_size) \
+  AZ_PRECONDITION(az_span_is_valid(span, min_size))
 
 #include <_az_cfg_suffix.h>
 
--- a/sdk/core/core/src/az_span.c
+++ b/sdk/core/core/src/az_span.c
@@ -52,7 +52,7 @@ AZ_NODISCARD az_span az_span_from_str(char* str)
 
 AZ_NODISCARD az_span az_span_slice(az_span span, int32_t start_index, int32_t end_index)
 {
-  AZ_PRECONDITION_VALID_SPAN(span, 0, true);
+  AZ_PRECONDITION_VALID_SPAN(span, 0);
 
   // The following set of preconditions validate that:
   //    0 <= end_index <= span.size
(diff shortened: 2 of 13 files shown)


danewalton:
Haven't delved too much into the PR itself but right off the top I can offer up a scenario for the:

> Why do you want your code to even care whether the ptr is null, if the size is 0?

Currently, with `az_span_token()` the remainder can have this exact scenario. Take the following:
Source: `abcdefgh`
Delimeter: `de`

The return span here would be `abc` and the `out_remainder` span param would be `fgh`.

Now take the following:
Source: `abcdefgh`
Delimeter: `defgh`

The return span here would be `abc` but the `out_remainder` span would be a valid pointer address but with size 0. 

This isn't to say this can't change, but that is a current behavior.

antkmsft:
@danewalton , I understand. I think, if you want to check this very specific case, it is possible to use 
AZ_PRECONDITION_NOT_NULL() for the span ptr, or even write your own precondition to check that the pointer is exactly somewhere inside the original range, which could be even better, and pass it to AZ_PRECONDITION(). Or use AZ_PRECONDITION_RANGE().

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -61,33 +61,16 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
...
-  {
-    result = ptr == NULL ? span_size == 0 : span_size >= 0;
-  }
-  else
-  {
-    result = ptr != NULL && span_size >= 0;
-  }
-
-  return result && min_size <= span_size;
+  return span_size >= 0 && (span_size == 0 || ptr != NULL) && span_size >= min_size;
```
ahsonkhan, line 69: Can you keep the comment explaining this expression OR re-write it if you prefer.

sdk/core/core/src/az_span.c
```diff
@@ -279,7 +279,7 @@ void az_span_to_str(char* destination, int32_t destination_max_size, az_span sou
   // Implementations of memmove generally do the right thing when number of bytes to move is 0, even
   // if the ptr is null, but given the behavior is documented to be undefined, we disallow it as a
   // precondition.
-  AZ_PRECONDITION_VALID_SPAN(source, 0, false);
+  AZ_PRECONDITION_VALID_SPAN(source, 0);
```
ahsonkhan, line 282: Most APIs like these that rejected a "null span" (but allowed an empty span), need to retain this behavior. We can't let users pass in `AZ_SPAN_NULL` here.

This is especially true for methods that call underlying STL APIs like memmove which are documented to have undefined behavior for null ptr. With the precondition, the caller knows at dev time not to pass in a "null span".

The problem with this change is that it forces us to turn a precondition check to a "real" check to guard against null ptr.

If you want to separate the "null_is_valid" check into a separate precondition so that all callers don't have to pass in true/false, that is fine, but I don't think we should be changing the behavior of APIs.

That said, since this is preconditions, saving a bool is not worth it, since that cost shouldn't affect the user in production anyway.

Alternatively, we should change the precondition to always reject a span with `null` ptr.

antkmsft: Do you want it to look like this? - https://github.com/Azure/azure-sdk-for-c/pull/644/files

antkmsft: I don't follow why to exclude spans with { .ptr = NULL, .size = 0 } in all the places across our SDK (with single exception for az_span_slice), just because memmove() from standard library is not formally documented for the case when size=0 ptr=NULL, and that case is enclosed in az_span_to_str() function, while the rest of our SDK would work correctly handling spans of size 0 reagrdless of the pointer. Even if you get a ptr, and pass it to a function, and it imposes the limitation that ptr can't be NULL even when size is 0, then we should handle that in the code that invokes that function, and I am not even sure it should be a precondition then, and not a simple if(). I don't think we should impose the limitations on the all the spans across all our code, just because of one external function.
We only use memmove in az_span_to_str() and in az_span_copy(). The rest of the code is using az_span_copy() when it needs to memmove. And BTW, az_span_copy hadles it with `if(size == 0) return destination`. I think az_span_to_str() should have the same sort of candling for size == 0 case.

PR #643, opened 2020-04-29
Add min_size check to az_span_is_valid()

antkmsft:
Regardless of what you think of the url_encode()'s precondition requiring `dest_size >= src_size + 2`, I think we should check min_size parameter in span validation. It can be an integer overflow when adding several span lenghts, it can be an error in your math expression, it can be bad input data.

Passing -5 as min_size is an unsatisfiable condition, so we should return false. You never want to meaningfully say "a span should be of size -5 or greater". It is an error, and we want to warn you. If we ignore it, we could be masking errors - you think your preconditions work, butin reality they are effectively no-ops.

And if you are concerned about perf/code size impact, az_span_check_is_valid() is an inline function. If you pass an integer literal `5` there, the compiler should be able to figure out that it can throw that `if()` code away. Plus, az_span_check_is_valid() only impacts the code compiled with preconditions enabled, which is already not as fast as possible code as it can be, and we should rather being consistent in preferring correctness over performance.

### DIFF
--- a/sdk/core/core/internal/az_precondition_internal.h
+++ b/sdk/core/core/internal/az_precondition_internal.h
@@ -63,6 +63,11 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
 
 AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, bool null_is_valid)
 {
+  if (min_size < 0)
+  {
+    return false;
+  }
+
   uint8_t* ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 


sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -63,6 +63,11 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
 
 AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, bool null_is_valid)
 {
+  if (min_size < 0)
```
ahsonkhan, line 66: This is a good change. However, it doesn't guarantee correctness in all cases (and might give us a false sense that we have fixed the overflow issue).

The caller is still responsible for making sure their expression/calculation is safe from overflow and I don't think the precondition can help in that case.

```C
//gcc 5.4.0

#include <stdio.h>
#include <stdbool.h>
#include <stdint.h>

bool az_span_is_valid(int32_t span_size, int32_t min_size)
{
    if (min_size < 0) return false;
    
    return span_size >= min_size;
}

int main(void)
{
    printf("%d\n", az_span_is_valid(10, 1)); // success (1), as expected
    printf("%d\n", az_span_is_valid(10, 20)); // fail (0), as expected
    
    // should fail, but doesn't, let's add a min_size < 0 check.
    printf("%d\n", az_span_is_valid(10, 2147483647 + 1)); // 0 after adding the check
    
    // should fail, but doesn't, because it is now positive, but this is still wrong!
    printf("%d\n", az_span_is_valid(10, 2147483647 + 2147483647 + 2)); // 1
    return 0;
}

```

antkmsft: I miss the assembly language. You add two numbers, and it is one instruction to check the carry flag and jump overflow occurred :) In C, we would probably need a function to do arithmetic operations, so it would work on all compilers and platforms, and on some of them it would use that JC (jump if carry flag is set) logic, and on others it would use complier-specific language features (probably MSVC has something to handle this).

ahsonkhan: Here's a potential solution, but I don't know if it is worth the trade offs (isn't totally foolproof).

```C
//gcc 5.4.0

#include <stdio.h>
#include <stdbool.h>
#include <stdint.h>

bool az_span_is_valid(int32_t span_size, uint64_t min_size)
{
    return span_size >= min_size;
}

int main(void)
{
    printf("%d\n", az_span_is_valid(10, 1)); // success (1), as expected
    printf("%d\n", az_span_is_valid(10, 20)); // fail (0), as expected
    
    printf("%d\n", az_span_is_valid(10, 2147483647ul + 1)); // 0 
    
    printf("%d\n", az_span_is_valid(10, 2147483647 + 2147483647ul + 2)); // 0
    return 0;
}
```

PR #644, opened 2020-04-29
Spans with NULL ptr are never valid, with one exception

antkmsft:


sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -66,7 +66,7 @@ AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size)
   uint8_t* const ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 
-  return span_size >= 0 && (span_size == 0 || ptr != NULL) && span_size >= min_size;
+  return span_size >= 0 && ptr != NULL && span_size >= min_size;
```
JeffreyRichter, line 69: You don't need the first size test since you have the other size test.

sdk/core/core/src/az_span.c
```diff
@@ -52,13 +52,14 @@ AZ_NODISCARD az_span az_span_from_str(char* str)
 
 AZ_NODISCARD az_span az_span_slice(az_span span, int32_t start_index, int32_t end_index)
 {
-  AZ_PRECONDITION_VALID_SPAN(span, 0);
+  int32_t const span_size = az_span_size(span);
+  AZ_PRECONDITION(span_size >= 0);
```
JeffreyRichter, line 56: This should just be a check that the span is valid; not just its size.
It IS OK to slice a NULL span as long as the start and end are both 0.

You should have the precondition test for start <= end first. AND, both must be >=0.
The start must also be between 0 and size+1 (I think) inclusive.

PR #647, opened 2020-04-29
Update c2d parse for NO_MATCH case and allow props to init with AZ_SPAN_NULL

danewalton:
This also updates C2D unit tests to use `assert_int_equal` instead of `assert_return_code` since we get false failures on some `az_result` return values.

EDIT: Updated the PR to allow for properties to be initialized with an `AZ_SPAN_NULL` or an empty buffer. The reason being in cases when a response/message comes in from the service which don't have properties, we should be able to communicate to the developer that the topic was successfully parsed but that it came with no properties. 

Therefore, if `az_iot_hub_client_properties_init` is called with an `az_span` with size 0 and pointer equal to `NULL` or otherwise, we should allow it. All the other property functionality will work with this as they check to make sure the required size is enough in the case that it appends. And in the case for finding a property or getting the `next` property, there are no properties to find so either `AZ_ERROR_ITEM_NOT_FOUND` is returned in `find` or `AZ_ERROR_EOF` in the `next` case.

### DIFF
--- a/sdk/iot/hub/src/az_iot_hub_client_c2d.c
+++ b/sdk/iot/hub/src/az_iot_hub_client_c2d.c
@@ -60,9 +60,16 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
 
   az_span token;
   token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  AZ_RETURN_IF_FAILED(
-      az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
+  if (az_span_ptr(received_topic) != NULL)
+  {
+    token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
+    AZ_RETURN_IF_FAILED(
+        az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
+  }
+  else
+  {
+    return AZ_ERROR_IOT_TOPIC_NO_MATCH;
+  }
 
   return AZ_OK;
 }
--- a/sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
+++ b/sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
@@ -33,6 +33,8 @@ static const az_span test_URL_ENCODED_topic
     = AZ_SPAN_LITERAL_FROM_STR("devices/useragent_c/messages/devicebound/"
                                "%24.to=%2Fdevices%2Fuseragent_c%2Fmessages%2FdeviceBound&abc=123&"
                                "ghi=%2Fsome%2Fthing&jkl=%2Fsome%2Fthing%2F%3Fbla%3Dbla");
+static const az_span test_parse_method_topic_fail
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/methods/POST/foo/?$rid=one");
 
 #ifndef AZ_NO_PRECONDITION_CHECKING
 enable_precondition_check_tests()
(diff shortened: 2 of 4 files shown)


sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -60,9 +60,16 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
...
-      az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
+  if (az_span_ptr(received_topic) != NULL)
+  {
+    token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
+    AZ_RETURN_IF_FAILED(
+        az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
+  }
+  else
+  {
+    return AZ_ERROR_IOT_TOPIC_NO_MATCH;
```
CIPop, line 71: We should accept C2D w/o properties. If I'm reading this right, this is the case where the topic matches the filter but there are no properties?

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -204,29 +189,66 @@ static void test_az_iot_hub_client_c2d_parse_received_topic_URL_ENCODED_succeed(
...
   assert_true(az_span_is_content_equal(pair.value, AZ_SPAN_FROM_STR("%2Fsome%2Fthing")));
 
-  assert_return_code(az_iot_hub_client_properties_next(&out_request.properties, &pair), AZ_OK);
+  assert_int_equal(az_iot_hub_client_properties_next(&out_request.properties, &pair), AZ_OK);
   assert_true(az_span_is_content_equal(pair.key, AZ_SPAN_FROM_STR("jkl")));
   assert_true(az_span_is_content_equal(pair.value,
   AZ_SPAN_FROM_STR("%2Fsome%2Fthing%2F%3Fbla%3Dbla")));
 }
 
+static void test_az_iot_hub_client_c2d_parse_received_topic_fail(void** state)
```
CIPop, line 215: Nit: Naming should be updated to `unit_action_result`. E.g. c2d_parse_received_topic_unrelated_topic_reject. (This is part of normal operation so not a failure.)

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -204,29 +189,66 @@ static void test_az_iot_hub_client_c2d_parse_received_topic_URL_ENCODED_succeed(
...
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+  assert_true(
+      az_iot_hub_client_init(&client, test_device_hostname, test_device_id, &options) == AZ_OK);
+
+  az_span received_topic = test_parse_method_topic_fail;
+
+  az_iot_hub_client_c2d_request out_request;
+
+  assert_int_equal(
+      az_iot_hub_client_c2d_parse_received_topic(
```
CIPop, line 229: Nit: assert(expected, actual) is easier to read since the first arg is usually short.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -204,29 +189,66 @@ static void test_az_iot_hub_client_c2d_parse_received_topic_URL_ENCODED_succeed(
...
+          &client, received_topic, &out_request),
+      AZ_ERROR_IOT_TOPIC_NO_MATCH);
+}
+
+static void test_az_iot_hub_client_c2d_parse_received_topic_MALFORMED_fail(void** state)
```
CIPop, line 234: Please keep names lowercase.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -60,9 +60,16 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
 
   az_span token;
   token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  token = az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  AZ_RETURN_IF_FAILED(
-      az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
+  if (az_span_ptr(received_topic) != NULL)
```
TiejunMS, line 63: When there is no properties, az_span_ptr(received_topic) pointers to "".

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
danewalton: fixed

sdk/iot/hub/src/az_iot_hub_client_c2d.c
danewalton: updated

sdk/iot/hub/src/az_iot_hub_client_c2d.c
danewalton: updated

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
danewalton: Understood. That could come in a later PR since I think all of hub is done this way.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
danewalton: good point. updated

PR #648, opened 2020-04-29
Add IoT Hub Samples for Twin, Telemetry, C2D, and Methods

danewalton:


sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+  if (topicLen == 0)
+  {
+    // The length of the topic if there are one more NULL characters embedded in topicName,
+    // otherwise topicLen is 0.
+    topicLen = (int)strlen(topicName);
+  }
+
+  az_result result;
+
+  az_span topic_span = az_span_init((uint8_t*)topicName, topicLen);
```
CIPop, line 113: Nit: I'd chain az_span_init within the call to c2d_parse instead of adding another variable name. (In this case topic_span is not used for anything else and it would make the code shorter thus faster to read and understand.)

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+
+  az_iot_hub_client_c2d_request c2d_request;
+  if (az_iot_hub_client_c2d_parse_received_topic(&client, topic_span, &c2d_request) == AZ_OK)
+  {
+    printf("C2D Message arrived\n");
+    printf("Payload:\n%.*s\n", message->payloadlen, (char*)message->payload);
```
CIPop, line 119: NIT: Had no idea printf knows this. Still, I'd avoid it to allow optimized `printf` implementations when used in IAR, etc.

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+    // otherwise topicLen is 0.
+    topicLen = (int)strlen(topicName);
+  }
+
+  az_result result;
+
+  az_span topic_span = az_span_init((uint8_t*)topicName, topicLen);
+
+  az_iot_hub_client_c2d_request c2d_request;
+  if (az_iot_hub_client_c2d_parse_received_topic(&client, topic_span, &c2d_request) == AZ_OK)
```
CIPop, line 116: `if(az_succeeded`. There's no guarantee that AZ_OK is the only success return code.

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+  {
+    printf("C2D Message arrived\n");
+    printf("Payload:\n%.*s\n", message->payloadlen, (char*)message->payload);
+  }
+
+  (void)result;
```
CIPop, line 122: Maybe remove `result`? Doesn't appear to be used anywhere.

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+
+  putchar('\n');
```
CIPop, line 124: Not needed - printf above took care of it.
Also: see `print_az_span`.

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Getting started
+
+Getting started explained in detail [here][SDK_README_GETTING_STARTED].
+
+## Examples
+
+The following section documents various examples. All of them use the [Eclipse Paho MQTT C client][Eclipse_Paho].
+
+1. [IoT Hub C2D][paho_c2d_sample]: Receive and view incoming C2D messages using the IoT Hub client.
```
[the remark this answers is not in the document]
danewalton: I went with alphabetical/the way the directory structure is set up. It's a toss up either way. Might let other clues in (I know it's a nit)?

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+//              Note: this is required to work-around MQTTClient.h as well as az_span init issues.
+#include <_az_cfg.h>
+
+// Service information
+#define DEVICE_ID "<YOUR DEVICE ID HERE>"
+#define HUB_FQDN "<YOUR IOT HUB FQDN HERE>"
+#define HUB_URL "ssl://" HUB_FQDN ":8883"
+
+// Device information
+#define REGISTRATION_ID_ENV "AZ_IOT_REGISTRATION_ID"
```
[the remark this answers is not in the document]
danewalton: fixed

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: done

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: makes sense. updated

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: yup removed

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: needed now that I changed to `putchar` on the payload instead of the size limited `printf`

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: Sure but in this case wouldn't it be possible that the values of the returned params be different such that someone might want to take a different set of actions on them? In either case you would probably want this updated. In the case that a different success value is returned, we would want to reevaluate this condition just as we would if this is `az_succeeded`.

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,249 @@
...
+#include <az_iot_hub_client.h>
+#include <az_result.h>
+#include <az_span.h>
+
+// TODO: #564 - Remove the use of the _az_cfh.h header in samples.
+//              Note: this is required to work-around MQTTClient.h as well as az_span init issues.
+#include <_az_cfg.h>
+
+// Service information
+#define DEVICE_ID "<YOUR DEVICE ID HERE>"
```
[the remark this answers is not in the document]
danewalton: updated

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+// AZ_IOT_DEVICE_X509_TRUST_PEM_FILE is the path to a PEM file containing the server trusted CA
+// This is usually not needed on Linux or Mac but needs to be set on Windows.
+#define DEVICE_X509_TRUST_PEM_FILE "AZ_IOT_DEVICE_X509_TRUST_PEM_FILE"
+
+static char device_id[64];
+static char iot_hub_fqdn[128];
+static char x509_cert_pem_file[512];
+static char x509_trust_pem_file[256];
+
+static char paho_endpoint[128];
```
ericwol-msft, line 44: mqtt endpoint?

ericwol-msft: mqtt hub endpoint?

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  else
+  {
+    printf("(missing) Please set the %s environment variable.\n", env_name);
+    return AZ_ERROR_ARG;
+  }
+
+  return AZ_OK;
+}
+
+static az_result create_paho_endpoint(char* destination, int32_t size, az_span iot_hub)
```
ericwol-msft, line 87: I don't think we should use paho, and use mqtt instead?

sdk/iot/hub/samples/telemetry/src/paho_iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+static char x509_cert_pem_file[512];
+static char x509_trust_pem_file[256];
+char telemetry_topic[128];
+
+static char paho_endpoint[128];
+static az_span paho_prefix = AZ_SPAN_LITERAL_FROM_STR("ssl://");
+static az_span paho_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+static const char* telemetry_message_payloads[NUMBER_OF_MESSAGES] = {
+  "Message One", "Message Two", "Message Three", "Message Four", "Message Five",
```
ericwol-msft, line 62: should we make these JSON payloads?

JeffreyRichter: Here's a thought: don't put any JSON-speciifc things in your client library code if you think you might offer other protocols (like CBOR) in the future. The "fear" is that you'll carry data/code in your library that will be useless for a CBOR customer.

danewalton: telemetry doesn't need to be JSON so it isn't here. Things like methods responses do need to be in JSON so they are.

JeffreyRichter: Will your statement remain true if CBOR becomes a customer choice?

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: updated

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: removed all paho labels that can be removed

sdk/iot/hub/samples/telemetry/src/paho_iot_hub_telemetry_example.c
danewalton: Not necessarily. But in that case we can change these samples to included CBOR capabilities. Something like having a `generate_payload()` and having it do JSON or CBOR. For now, only the features that fundamentally require JSON like DPS have it internally. Otherwise, it is up to the user to create the correct payloads, in the necessary encoding. Example:
https://github.com/Azure/azure-sdk-for-c/blob/1848fd042adceedc2c3c0202c3ceb228435c33d4/sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c#L306-L318
In ⬆️ case we can just as easily do this encoding in CBOR using [tinyCBOR](https://github.com/intel/tinycbor)

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,418 @@
...
+  {
+    printf("Direct Method arrived\n");
+    if(az_span_is_content_equal(double_method_span, method_request.name))
+    {
+      //Get payload value
+      az_span payload_span = az_span_init((uint8_t*)message->payload, message->payloadlen);
+      int64_t value_to_double;
+      if(parse_value_to_double(payload_span, &value_to_double) != AZ_OK)
+      {
+        //Error
```
danewalton, line 267: Update these

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
+---
+page_type: sample
+languages:
+  - c
+products:
+  - azure
+  - azure-iot
+  - azure-iot-hub
+urlFragment: iot-hub-samples
+---
```
ahsonkhan, line 10: Remove this preamble? Is this meant to act as hidden metadata? It affects readability of the markdown on GitHub:
image

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Getting started
+
+Getting started explained in detail [here][SDK_README_GETTING_STARTED].
+
+## Examples
+
+The following section documents various examples. All of them use the [Eclipse Paho MQTT C client][Eclipse_Paho].
+
+### [IoT Hub C2D][c2d_sample]
+Receive and view incoming C2D messages using the IoT Hub client.
```
ahsonkhan, line 33: nit: consistency
```suggestion
Receive and view incoming C2D messages using the IoT Hub Client.
```

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### [IoT Hub Methods][methods_sample]
```
ahsonkhan, line 35: This link is 404.

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+Invoke methods from the cloud. The sample supports a method named "double"
+which will return back to you the value sent as the payload. The payload must be of the form:
+```json
+{
+  "value": 10
+}
+```
+where 10 can be substituted for any number between `UINT64_T_MIN / 2` and `UINT64_T_MAX / 2`.
```
ahsonkhan, line 43: Isn't UINT64_T_MIN just 0? Was this meant to be INT64_T? Also, consider using the names from `limits.h`
https://www.tutorialspoint.com/c_standard_library/limits_h.htm

I don't get much search hit for `UINT64_T_MAX`.

```suggestion
where 10 can be substituted for any number between `0 and `ULONG_MAX / 2`.
```

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+### [IoT Hub Telemetry][telemetry_sample]
+Send 5 telemetry messages using the IoT Hub client.
```
ahsonkhan, line 45: Same consistency nit here. Should the `c` in `client` be capital everywhere or lower case?
```suggestion
Send 5 telemetry messages using the IoT Hub Client.
```

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+### [IoT Hub Twin][twin_sample]
+Use twin features such as updating reported properties, receiving the twin document, and receiving desired properties using the IoT Hub client.
+
+## Troubleshooting
+
+When interacting with the Azure IoT Hub using this C client library, errors are documented within the [MQTT State Machine][error_codes] requests.
+
+## Next steps
+
+Start using the IoT Hub Client in your solutions. Our SDK details can be found at [SDK README][IOT_CLIENT_README].
```
ahsonkhan, line 55:
```suggestion
Start using the IoT Hub Client in your solutions. More details can be found in the [IOT SDK README][IOT_CLIENT_README].
```

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Additional Documentation
+
+For extensive documentation on Azure IoT Hub, see the [API reference documentation][iot_hub_mqtt].
+
+## Contributing
+
+This project welcomes contributions and suggestions. Find [more contributing][SDK_README_CONTRIBUTING] details here.
```
ahsonkhan, line 63: Is this section needed here?

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  return AZ_OK;
+}
+
+static int on_received(void* context, char* topicName, int topicLen, MQTTClient_message* message)
+{
+  (void)context;
+
+  if (topicLen == 0)
+  {
+    // The length of the topic if there are one more NULL characters embedded in topicName,
```
ahsonkhan, line 140:
```suggestion
    // The length of the topic if there are one or more NULL characters embedded in topicName,
```

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+<!-- LINKS -->
+[IOT_CLIENT_README]: ../../README.md
+[SDK_README_CONTRIBUTING]:../../README.md#contributing
+[SDK_README_GETTING_STARTED]: ../../README.md#getting-started
+[SDK_README_KEY_CONCEPTS]: ../../README.md#key-concepts
+[c2d_sample]: c2d/src/iot_hub_c2d_example.c
+[methods_sample]: methods/src/iot_methods_example.c
```
ahsonkhan, line 71: I think this needs to be updated to match file name: `iot_hub_methods_example.c`

sdk/iot/hub/samples/README.md
danewalton: yes it should be `INT32` thank you

sdk/iot/hub/samples/README.md
danewalton: I'd say it couldn't hurt 🤷‍♂️

sdk/iot/hub/samples/README.md
danewalton: yup fixed thank you

sdk/iot/hub/samples/README.md
danewalton: odd. I think it is this way across all of our stuff. I will likely leave it here for now but make a note to fix across the board if necessary.

sdk/iot/hub/samples/README.md
```diff
@@ -0,0 +1,76 @@
...
+products:
+  - azure
+  - azure-iot
+  - azure-iot-hub
+urlFragment: iot-hub-samples
+---
+
+# Azure IoT Hub Sample
+
+This document explains samples for the Azure Embedded C SDK IoT Hub Client and how to use them.
```
gilbertw, line 14: ... Azure SDK for Embedded C ...

danewalton: updated

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  az_span remainder = az_span_copy(destination_span, mqtt_url_prefix);
+  remainder = az_span_copy(remainder, az_span_slice(iot_hub, 0, iot_hub_length));
+  az_span_copy(remainder, mqtt_url_suffix);
+
+  return AZ_OK;
+}
+
+static az_result read_configuration_and_init_client()
+{
+
```
hihigupt, line 131: nit remove blank line

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  }
+  az_span destination_span = az_span_init((uint8_t*)destination, size);
+  az_span remainder = az_span_copy(destination_span, mqtt_url_prefix);
+  remainder = az_span_copy(remainder, az_span_slice(iot_hub, 0, iot_hub_length));
+  az_span_copy(remainder, mqtt_url_suffix);
+
+  return AZ_OK;
+}
+
+static az_result read_configuration_and_init_client()
```
hihigupt, line 129: >read_configuration_and_init_client [](start = 17, length = 34)

init_clien_using_configuration?

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: updated

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "X509 Certificate PEM Store File", DEVICE_X509_CERT_PEM_FILE, NULL, false, cert, &cert));
+
+  az_span trusted = AZ_SPAN_FROM_BUFFER(x509_trust_pem_file);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "X509 Trusted PEM Store File", DEVICE_X509_TRUST_PEM_FILE, "", false, trusted, &trusted));
+
+  az_span device_id_span = AZ_SPAN_FROM_BUFFER(device_id);
+  AZ_RETURN_IF_FAILED(
+      read_configuration_entry("Device ID", DEVICE_ID, "", false, device_id_span, &trusted));
```
hihigupt, line 142: >trusted [](start = 83, length = 7)

should this be &device_id_span?

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+
+  az_span iot_hub_fqdn_span = AZ_SPAN_FROM_BUFFER(iot_hub_fqdn);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "IoT Hub FQDN", IOT_HUB_FQDN, "", false, iot_hub_fqdn_span, &trusted));
```
hihigupt, line 146: >trusted [](start = 67, length = 7)

same

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: shoot yes updating

danewalton: fixing the other samples as well

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+
+  char username[128];
+  if ((rc = az_iot_hub_client_get_user_name(&client, username, sizeof(username), NULL)) != AZ_OK)
+
+  {
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    return rc;
+  }
+
+  mqtt_connect_options.username = username;
```
hihigupt, line 175: > mqtt_connect_options.username = username; [](start = 1, length = 42)

do you know if username is copied inside paho struct?

If not username will be lost in unwinding of stack.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+#include <_az_cfg.h>
+
+// Device ID
+#define DEVICE_ID "AZ_IOT_DEVICE_ID"
+
+// IoT Hub FQDN
+#define IOT_HUB_FQDN "AZ_IOT_HUB_FQDN"
+
+// Device information
+#define REGISTRATION_ID_ENV "AZ_IOT_REGISTRATION_ID"
```
hihigupt, line 39: how about checking if user has passed these macro using compiler option use that,  otherwise use default in c file?

ifndef DEVICE_ID
#define DEVICE_ID ""
#endif

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+             telemetry_message_payloads[i],
+             0,
+             0,
+             NULL))
+        != MQTTCLIENT_SUCCESS)
+    {
+      printf("Failed to publish telemetry topic, return code %d\n", rc);
+      return rc;
+    }
+    sleep_seconds(1);
```
hihigupt, line 214: >1 [](start = 18, length = 1)

Please use macro for magic numbers

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+{
+  int rc;
+
+  if ((rc = read_configuration_and_init_client()) != AZ_OK)
+  {
+    printf("Failed to read configuration from environment variables, return code %d\n", rc);
+    return rc;
+  }
+
+  char client_id[128];
```
hihigupt, line 229: do not allocate buffers in stack, move them to data segment.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  {
+    return rc;
+  }
+
+  if ((rc = az_iot_hub_client_telemetry_get_publish_topic(
+           &client, NULL, telemetry_topic, sizeof(telemetry_topic), NULL))
+      != AZ_OK)
+  {
+    return rc;
+  }
```
hihigupt, line 257: can we move this inside the send telemetry function, because looks like that uses this

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  // Loop and send 5 messages
+  if ((rc = send_telemetry_messages()) != 0)
+  {
+    return rc;
+  }
+
+  printf("Messages Sent [Press ENTER to shut down]\n");
+  (void)getchar();
+
+  if ((rc = MQTTClient_disconnect(mqtt_client, 10000)) != MQTTCLIENT_SUCCESS)
```
hihigupt, line 268: what happens when send telemetry fails? Do we avoid using clean shutdhown?

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  {
+    printf("Failed to disconnect MQTT client, return code %d\n", rc);
+    return rc;
```
hihigupt, line 271: Is there a reason we are not following clean shut down approach in failure case as we did in sdkv2
[https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/samples/iothub_ll_telemetry_sample/iothub_ll_telemetry_sample.c](https://github.com/Azure/azure-iot-sdk-c/blob/master/iothub_client/samples/iothub_ll_telemetry_sample/iothub_ll_telemetry_sample.c)

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+#define DEVICE_X509_CERT_PEM_FILE "AZ_IOT_DEVICE_X509_CERT_PEM_FILE"
+
+// AZ_IOT_DEVICE_X509_TRUST_PEM_FILE is the path to a PEM file containing the server trusted CA
+// This is usually not needed on Linux or Mac but needs to be set on Windows.
+#define DEVICE_X509_TRUST_PEM_FILE "AZ_IOT_DEVICE_X509_TRUST_PEM_FILE"
+
+#define NUMBER_OF_MESSAGES 5
+
+static char device_id[64];
+static char iot_hub_fqdn[128];
```
hihigupt, line 52: >128 [](start = 25, length = 3)

put all the allocation number as macros

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+static char x509_cert_pem_file[512];
+static char x509_trust_pem_file[256];
```
hihigupt, line 54: >256 [](start = 32, length = 3)

Interesting buffer size for trusted cert ? Did you test cert path?

hihigupt: hope I am incorrect.

---
In reply to: [418658585](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418658585) [](ancestors = 418658585)

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: https://github.com/eclipse/paho.mqtt.c/blob/3148fe2d5f4b87e16266dfe559c0764e16ca0546/src/MQTTClient.c#L1589

Malloc'd and copied

hihigupt: Hmm malloc mess.... Can we atleast move this buffer to data segment from stack memory?

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: I disagree: having seldom used values on the stack is more efficient. That way, RAM is free for other things instead of being always used by the uninitialized data segment. 
Especially since these variables are used just during connect (which should ideally be only a very small percentage of time compared with normal operation).

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  return 0;
+}
+
+static int subscribe()
+{
+  int rc;
+
+  char c2d_topic[128];
```
danewalton, line 350: change name

sdk/iot/hub/samples/telemetry/src/paho_iot_hub_telemetry_example.c
CIPop: I don't think 99% of devices will report variable sized data (e.g. sensors are normally 4-8 bytes). They don't need to be descriptive because IOT Hub already knows which device is sending data and so it can be as compressed as possible (i.e. send only the actual binary values, perform adaptation based on device capabilities on the service side where battery and CPU power are not a concern).

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: It depends on whether it is in an RTOS thread or not. The stack can be blown through with these topic names. It's hard to pin down which is "better" since I think it is highly customer, use case dependent. EG how much RAM/Flash do they have, are they using threads, etc. Should we maybe put it to a vote? Unless someone has a very concrete, critical reason for otherwise

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
CIPop: That would be a breaking change in the contract described in our MQTT state machine. 
That's exactly why we need to avoid checking _exact_ error codes: in the future we may need to differentiate several "SUCCESS" cases for other purposes. That differentiation will need to be backward compatible (apps are expected to use az_success if they can't handle the new differences).

sdk/iot/hub/samples/README.md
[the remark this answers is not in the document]
CIPop: Basic SKUs don't support the other features; Telemetry is the most important feature (the rest may or may not be required by devices). Reading the doc, users should find the most important features at top.
I would argue that C2D be moved last: most people should use Methods and Twin instead of C2D.

sdk/iot/hub/samples/c2d/src/paho_iot_hub_c2d_example.c
danewalton: > (apps are expected to use az_success if they can't handle the new differences).

But I think this argument goes both ways. If someone uses `az_succeed()`, doesn't that bake into their code that going into that `if()` block can handle the newly introduced succeed case?

sdk/iot/hub/samples/README.md
CIPop: Please keep readme.md files as short as possible. I think we already have a contributing guide in root.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: Made the comment elsewhere - we shouldn't keep unused variables in the data segment as they would occupy memory continuously even if they are not needed most of the time.

sdk/iot/hub/samples/README.md
danewalton: removed

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
hihigupt: If this sample is for non- embedded device, I am ok because windows and linux has Mb of stack. But If this is targeted for embedded. stack is not that much it is generally 1K or 2K, making these big chunk of allocation even temporary is not wise. 

---
In reply to: [418690382](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418690382) [](ancestors = 418690382)

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: I left comments in the provisioning sample: apps should crash immediately unless it can handle the error code.
If the error is out of memory or an AV continuing would destroy the state and make the bug impossible to debug.

It's more or less like catching Exception in java/C# which is wrong: https://softwareengineering.stackexchange.com/questions/270406/is-it-a-good-practice-to-catch-all-exception

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: It's a cert path, not a cert.

hihigupt: aha .. my bad

---
In reply to: [418697312](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418697312) [](ancestors = 418697312)

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  {
+    az_result result;
+
+    printf("Direct Method arrived\n");
+    if (az_span_is_content_equal(double_method_span, method_request.name))
+    {
+      // Get payload value
+      az_span payload_span = az_span_init((uint8_t*)message->payload, message->payloadlen);
+      int64_t value_to_double;
+      if ((result = parse_value_to_double(payload_span, &value_to_double)) != AZ_OK)
```
CIPop, line 271: I think this is too complicated for a sample. We should not need/use the json parser - we can simply print out the payload.
We can also change to a simpler scenario such as `reboot_device` or `turn_on_led`.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+      {
+        printf("Unable to parse input value, status %d", result);
+        return result;
+      }
+
+      // Invoke Method
+      int64_t return_value;
```
CIPop, line 278: We should avoid using int64_t - these are usually small devices and I really doubt most would need 64bit integers.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  int rc;
+
+  MQTTClient_SSLOptions mqtt_ssl_options = MQTTClient_SSLOptions_initializer;
+  MQTTClient_connectOptions mqtt_connect_options = MQTTClient_connectOptions_initializer;
+
+  char username[128];
+  size_t username_length;
+  if ((rc = az_iot_hub_client_get_user_name(&client, username, sizeof(username), &username_length))
+      != AZ_OK)
+
```
CIPop, line 320: Nit: extra NL

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
CIPop: +1 - several places.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  if ((rc = subscribe()) != 0)
+  {
+    return rc;
+  }
+
+  printf("Subscribed to topics.\n");
+
+  printf("Waiting for activity. [Press ENTER to abort]\n");
```
CIPop, line 417: See other comment: this will likely timeout at MQTT-protocol level. We should instead convert to sync MQTT_recv and use timeouts instead of getchar.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: Having security or privacy related information within C files is not a good idea. We shouldn't promote that practice.

For larger embedded devices the configuration should be either read from a file, environment or, ideally from a Hardware Security Module.
For smaller devices a separate flash segment, eeprom or HSM should be used.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: What would be the benefit though: it will add another indirection and line of code.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: moved

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: You were right though: likely my mistake - all paths should have the same value - here's one case where the comment above does make sense - we should have a MAX_PATH define.
This needs to be done to all samples incl. provisioning. 

@danewalton we can do that in a different PR - let me know what you decided.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: its the same elsewhere so for now I'm inclined to keep as is.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+static void sleep_seconds(uint32_t seconds)
+{
+#ifdef _WIN32
+  Sleep((DWORD)seconds * 1000);
+#else
+  sleep(seconds);
+#endif
+}
+
+static az_result read_configuration_entry(
```
CIPop, line 77: @JeffreyRichter @gilbertw @ahsonkhan This code is copy/pasted now throughout several of our sample files and I see (buggy) code throughout other code.
Any recommendation where to place it (common environment variable reader).

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: I think this is how provisioning has it named. If we change, please update all samples.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: These macro's aren't the values but the environment variables names to look for. I believe it is settled that anything user specific for samples will be done via environment variables

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+  char username[128];
+  if ((rc = az_iot_hub_client_get_user_name(&client, username, sizeof(username), NULL)) != AZ_OK)
+
+  {
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    return rc;
+  }
+
+  mqtt_connect_options.username = username;
+  mqtt_connect_options.password = NULL; // Using X509 Client Certificate authentication.
```
CIPop, line 176: Nit: the comment is missing from some of the other samples.

Future thinking: we'll need to update _all_ samples with another environment var to allow SAS tokens or SharedAccessKeys so maybe don't fix it now since we're going to shortly touch the code anyway.

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
```diff
@@ -0,0 +1,278 @@
...
+    return rc;
+  }
+
+  return 0;
+}
+
+static int send_telemetry_messages()
+{
+  int rc;
+  for (int i = 0; i < NUMBER_OF_MESSAGES; ++i)
```
CIPop, line 198: We should replace this with a timeout (send messages for x seconds).

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
CIPop: I don't understand: in RTOS or bare metal is exactly where you wouldn't want to keep unused data in a small RAM especially since it would be available (through computation) from the larger Flash.
Assembling and keeping it temporarily (as needed) in RAM is ideal (which the stack is really good at).

This information is used _only_ during CONNECT/CONNACK. keeping it in RAM all the time would be wasteful IMO.

Embedded Linux / Windows won't have the problem because they keep everything in RAM anyway.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  int32_t required_size
+      = az_span_size(mqtt_url_prefix) + iot_hub_length + az_span_size(mqtt_url_suffix);
+  if (required_size > size)
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
+  }
+  az_span destination_span = az_span_init((uint8_t*)destination, size);
+  az_span remainder = az_span_copy(destination_span, mqtt_url_prefix);
+  remainder = az_span_copy(remainder, az_span_slice(iot_hub, 0, iot_hub_length));
+  az_span_copy(remainder, mqtt_url_suffix);
```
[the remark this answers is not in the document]
danewalton: very true I should have null terminated it. Updating

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: Alright I will change it to do a "ping/pong". Method will be named "ping" with no body. Response will always be
```json
{"response":"pong"}
```

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: The JSON parses out to a `double` so was attempting to unify. But since I'm changing the method sample to simplify this won't apply.

sdk/iot/hub/samples/README.md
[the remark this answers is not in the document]
danewalton: reordered

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+static char iot_hub_fqdn[128];
+static char x509_cert_pem_file[512];
+static char x509_trust_pem_file[256];
+
+static char mqtt_endpoint[128];
+static az_span mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("ssl://");
+static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+static char get_twin_topic[128];
+static az_span get_twin_topic_request_id = AZ_SPAN_LITERAL_FROM_STR("get_twin");
```
CIPop, line 53: To set a good example, I think this should be as short as possible (starting with one byte). The only requirement from hub is to have this unique.

danewalton: The idea I had here was to easily be able to associate a returned response to the request which originated. This way if someone gets an error return, they can tell if it was from a `get_twin` or `reported_properties`

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  return 1;
+}
+
+static int connect_device()
+{
+  int rc;
+
+  MQTTClient_SSLOptions mqtt_ssl_options = MQTTClient_SSLOptions_initializer;
+  MQTTClient_connectOptions mqtt_connect_options = MQTTClient_connectOptions_initializer;
+
```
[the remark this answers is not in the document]
danewalton: updated all to have these

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+
+  az_iot_hub_client_twin_response twin_response;
+  if (az_iot_hub_client_twin_parse_received_topic(&client, topic_span, &twin_response) == AZ_OK)
+  {
+    printf("Twin Message Arrived\n");
+    print_twin_response_type(twin_response.response_type, message);
+    printf("Response status was %i\n", twin_response.status);
+  }
+
+  (void)result;
```
CIPop, line 192: Remove?

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+{
+  int rc;
+
+  char twin_desired_topic[128];
+  if ((rc = az_iot_hub_client_twin_patch_get_subscribe_topic_filter(
+           &client, twin_desired_topic, sizeof(twin_desired_topic), NULL))
+      != AZ_OK)
+
+  {
+    printf("Failed to get C2D MQTT SUB topic filter, return code %d\n", rc);
```
CIPop, line 248: copy-paste issue: C2D here and a few other places (should be twin)

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+  return 0;
+}
+
+static int send_get_twin()
+{
+  int rc;
+  printf("Requesting twin document\n");
+
+  //Get the topic if we haven't already
+  if (get_twin_topic[0] == '\0')
```
CIPop, line 282: This is too complicated for a sample - we should have the sample assume that the topic was not generated.

CIPop: Also, it's likely not OK to submit two requests with the same request_id.

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+  return result;
+}
+
+static int send_reported_property()
+{
+  int rc;
+  printf("Sending reported property\n");
+
+  //Get the topic if we haven't already
+  if (reported_property_topic[0] == '\0')
```
CIPop, line 326: Same as above: we should remove this check. Also, the topic should always be constructed in a way that ensures that the request_id is different.

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+    return rc;
+  }
+
+  printf("Subscribed to topics.\n");
+
+  printf("\nWaiting for activity:\nPress 'g' to get the twin document\nPress 'r' to send a reported "
+         "property\n[Press 'q' to quit]\n");
+
+  int input;
+  while (1)
```
CIPop, line 414: We shouldn't have IoT samples interact with the user. It's unusual to have IoT devices requiring user input.

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: ah true. was trying to optimize. will change

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
hihigupt: Every failure can't be categorized as unhandled failure and state of application should be closed properly if we are able to handle failure. By abruptly closing the application you are not playing well with our service.
e.g by not sending explicit RST packet because you were not able to full file some application logic is not playing well with service and is causing resource hogging on service side. As this example come from Azure, we should be mindful of this.  

---
In reply to: [418697016](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418697016) [](ancestors = 418697016)

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  return AZ_OK;
+}
+
+static az_result create_mqtt_endpoint(char* destination, int32_t size, az_span iot_hub)
+{
+  int32_t iot_hub_length = (int32_t)strlen(iot_hub_fqdn);
+  int32_t required_size = az_span_size(mqtt_url_prefix) + iot_hub_length + az_span_size(mqtt_url_suffix);
+  if (required_size > size)
```
hihigupt, line 91: Can we please add a horizontal space between logical blocks?

hihigupt: I am not sure what Sdk has the style guidance for spacing or it is missing. But here is one example of horizontal spacing guidance
[https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace](https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace) 

---
In reply to: [418755362](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418755362) [](ancestors = 418755362)

danewalton: yup will try to go through and look for this elsewhere too

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: fixed here and elsewhere

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
hihigupt: Often, the consequences of a stack overflow manifest themselves far removed from the cause of the overflow itself, making the cause that much more difficult to identify and fix.
Having a temporary buffer in global space and solve your problem.


---
In reply to: [418720161](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418720161) [](ancestors = 418720161)

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+  az_result result;
+
+  az_span topic_span = az_span_init((uint8_t*)topicName, topicLen);
+
+  az_iot_hub_client_twin_response twin_response;
+  if (az_iot_hub_client_twin_parse_received_topic(&client, topic_span, &twin_response) == AZ_OK)
+  {
+    printf("Twin Message Arrived\n");
+    print_twin_response_type(twin_response.response_type, message);
+    printf("Response status was %i\n", twin_response.status);
```
hihigupt, line 189: >i [](start = 33, length = 1)

lets use %d, that is more popular

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+  if ((rc = az_iot_hub_client_twin_patch_get_subscribe_topic_filter(
+           &client, twin_desired_topic, sizeof(twin_desired_topic), NULL))
+      != AZ_OK)
+
+  {
+    printf("Failed to get C2D MQTT SUB topic filter, return code %d\n", rc);
+    return rc;
+  }
+
+  char twin_response_topic[128];
```
hihigupt, line 252: you have already allocated 256B on stack.

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  {
+    return rc;
+  }
+
+  printf("Subscribed to topics.\n");
+
+  printf("Waiting for activity. [Press ENTER to abort]\n");
+  (void)getchar();
+
+  if ((rc = MQTTClient_disconnect(mqtt_client, 10000)) != MQTTCLIENT_SUCCESS)
```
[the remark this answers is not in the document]
danewalton: updated

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
hihigupt: We should have guidance:
1. Should we allocate at start of function or closer to use (If we go with later we will cause the issue of creating many buffer without realizing the consequences)
2. Where should we create large buffer/even temporary,

---
In reply to: [418762214](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418762214) [](ancestors = 418762214)

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+
+// TODO: #564 - Remove the use of the _az_cfh.h header in samples.
+//              Note: this is required to work-around MQTTClient.h as well as az_span init issues.
+#include <_az_cfg.h>
+
+//Device ID
+#define DEVICE_ID "AZ_IOT_DEVICE_ID"
+
+//IoT Hub FQDN
+#define IOT_HUB_FQDN "AZ_IOT_HUB_FQDN"
```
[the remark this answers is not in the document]
danewalton: I'll change it to `AZ_IOT_HUB_HOSTNAME`

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
hihigupt: @jeffrichter@live.com @ericwol@microsoft.com 

---
In reply to: [418763205](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418763205) [](ancestors = 418763205,418762214)

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+    printf("(missing) Please set the %s environment variable.\n", env_name);
+    return AZ_ERROR_ARG;
+  }
+
+  return AZ_OK;
+}
+
+static az_result create_mqtt_endpoint(char* destination, int32_t size, az_span iot_hub)
+{
+  int32_t iot_hub_length = (int32_t)strlen(iot_hub_fqdn);
```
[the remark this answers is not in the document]
danewalton: No since the span is the full length of the buffer (which is unfilled). We only want the size of the actual string not the whole buffer.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+        AZ_RETURN_IF_FAILED(az_json_token_get_number(&json_token_member.token, &double_value));
+        *value = (int64_t)double_value;
+        break;
+      }
+    }
+  } while (result != AZ_OK);
+  return result;
+}
+
+static const az_span double_method_span = AZ_SPAN_LITERAL_FROM_STR("double");
```
hihigupt, line 166: start of file?

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
hihigupt, line 271: >parse_value_to_double [](start = 20, length = 21)

this name of function sounds like you are converting string to double type

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+      if ((result = send_method_response(&method_request, 200, return_value)) != AZ_OK)
+      {
+        printf("Unable to send %d response, status %d", 200, result);
+        return result;
+      }
+    }
+    else
+    {
+      // Unsupported Method
+      if ((result = send_method_response(&method_request, 404, 0xFFFF)) != AZ_OK)
```
hihigupt, line 295: >0xFFFF [](start = 63, length = 6)

suggestion: -1?

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+      {
+        printf("Unable to send %d response, status %d", 404, result);
+        return result;
```
hihigupt, line 298: >return result; [](start = 8, length = 14)

on_received return bool and indicate to lower layer that you have consumer the message. Now here you are returning non-zero value and not freeing memory of message.

Are you not freeing memory in failure case?
Can you run valgrind on these sample happy path and also while mimicking some failure case (atleast application level.)

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: N/A since the sample has been simplified

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: N/A with simplified sample

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: updated here and elsewhere

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: All buffers will be statically allocated at the top of the file.

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: Noted. Will update in a later PR.

sdk/iot/hub/samples/src/iot_hub_c2d_sample.c
```diff
@@ -0,0 +1,293 @@
...
+static char mqtt_endpoint[128];
+static az_span mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("ssl://");
+static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+static char c2d_topic[128];
+
+static az_iot_hub_client client;
+static MQTTClient mqtt_client;
+
+static az_result read_configuration_entry(
```
hihigupt, line 58: General question, which I was even asked by RTOS team, why in our code we do not have comments both in src and these samples.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: yes I need to change this to fail a little more gracefully. thanks

sdk/iot/hub/samples/src/iot_hub_c2d_sample.c
danewalton: What kind of comments are you looking for? What it will do from a high level? Function details?

hihigupt: Comments types :
1. What function does why it will return error.
2. If there are multiple way to do things why we are using this specific approach.
3. Who owns the memory.
4. if function produce some output in string, some possible example of that 

this help us to maintain code.

---
In reply to: [418773973](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418773973) [](ancestors = 418773973)

hihigupt: This comment might not be related to this but overall our iot folder

---
In reply to: [418775327](https://github.com/Azure/azure-sdk-for-c/pull/648#discussion_r418775327) [](ancestors = 418775327,418773973)

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
JeffreyRichter: OK, but if you ever decide to allocate buffers on the stack, then I recommend you allocate them close to use. First, this makes the code more readable/maintainable. Second, the compiler can be smart about it. For example:
if (...) {
   char buf[100];
   // ...
} else {
   char buf[200];
   // ...
}

in this case, the compiler should just allocate 200 bytes on the stack; not 300.

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: updated for single return point

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
hihigupt: In threadX general guidance is to allocate stack for worst case. Reason for this is that you allocate stack for worst case, so that your code does not overflow in random failure case. One of the tough problem in embedded is to measure the right size of stack that application will use. By hiding these allocation under If condition this exercise really become tough.

sdk/iot/hub/samples/src/iot_hub_c2d_sample.c
danewalton: Yup there are definitely improvements to documentation and comments that can be done. And I will be sure to add as many as I can in the coming days. But ultimately this PR is going to sit forever if we wait for things like that. And it'll keep having to be rebased, updated, etc.

I promise I'll continue to document the heck out of it.

sdk/iot/hub/samples/src/iot_hub_methods_sample.c
```diff
@@ -0,0 +1,374 @@
...
+  remainder = az_span_copy(remainder, mqtt_url_suffix);
+  az_span_copy_u8(remainder, null_terminator);
+
+  return AZ_OK;
+}
+
+static az_result read_configuration_and_init_client()
+{
+
+  az_span cert = AZ_SPAN_FROM_BUFFER(x509_cert_pem_file);
```
hihigupt, line 122: nit: remove this space

sdk/iot/hub/samples/src/iot_hub_telemetry_sample.c
```diff
@@ -0,0 +1,286 @@
...
+  remainder = az_span_copy(remainder, mqtt_url_suffix);
+  az_span_copy_u8(remainder, null_terminator);
+
+  return AZ_OK;
+}
+
+static az_result read_configuration_and_init_client()
+{
+
+  az_span cert = AZ_SPAN_FROM_BUFFER(x509_cert_pem_file);
```
hihigupt, line 138: same in all file

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
danewalton: From the service docs:

> Request ID can be any valid value for a message property value

So while not entirely useful in this case to have it as the same thing over and over again, it is valid. And since we aren't differentiating between sent/received messages, I think it's okay for now.

sdk/iot/hub/samples/src/iot_hub_telemetry_sample.c
danewalton: removed

sdk/iot/hub/samples/src/iot_hub_methods_sample.c
danewalton: removed

sdk/iot/hub/samples/src/iot_hub_c2d_sample.c
danewalton: added to backlog ✔️

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+static int connect_device()
+{
+  int rc;
+
+  MQTTClient_SSLOptions mqtt_ssl_options = MQTTClient_SSLOptions_initializer;
+  MQTTClient_connectOptions mqtt_connect_options = MQTTClient_connectOptions_initializer;
+
+  char username[128];
+  size_t username_length;
+  if ((rc = az_iot_hub_client_get_user_name(&client, username, sizeof(username), &username_length))
```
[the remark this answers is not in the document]
danewalton: added to backlog ✔️

sdk/iot/hub/samples/c2d/src/iot_hub_c2d_example.c
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  if ((rc = subscribe()) != 0)
+  {
+    return rc;
+  }
+
+  printf("Subscribed to topics.\n");
+
+  printf("Waiting for activity. [Press ENTER to abort]\n");
```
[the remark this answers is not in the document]
danewalton: backlog ✔️

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
```diff
@@ -0,0 +1,447 @@
...
+
+static char mqtt_endpoint[128];
+static az_span mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("ssl://");
+static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+static char get_twin_topic[128];
+static az_span get_twin_topic_request_id = AZ_SPAN_LITERAL_FROM_STR("get_twin");
+static char reported_property_topic[128];
+static az_span reported_property_topic_request_id = AZ_SPAN_LITERAL_FROM_STR("reported_prop");
+static az_span reported_property_name = AZ_SPAN_LITERAL_FROM_STR("foo");
```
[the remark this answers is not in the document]
danewalton: backlog ✔️

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: backlog ✔️

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: backlog ✔️

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
danewalton: backlog ✔️

sdk/iot/hub/samples/methods/src/iot_hub_methods_example.c
```diff
@@ -0,0 +1,430 @@
...
+  else
+  {
+    printf("(missing) Please set the %s environment variable.\n", env_name);
+    return AZ_ERROR_ARG;
+  }
+
+  return AZ_OK;
+}
+
+static az_result create_mqtt_endpoint(char* destination, int32_t size, az_span iot_hub)
```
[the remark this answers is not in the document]
danewalton: updated the second param to be `destination_size`

sdk/iot/hub/samples/telemetry/src/iot_hub_telemetry_example.c
danewalton: not anymore 😃

danewalton: will be up for discussion

sdk/iot/hub/samples/twin/src/iot_hub_twin_example.c
hihigupt: I tried to compile following code both with -Os and -O1

void test_fn(int cond)
{
    char buf[100];

    if (cond)
    {
        char buff_nested[100] = "nested_condition";

        printf("nested : %s, parent scope : %s\r\n", buff_nested, buf);
    }
    else
    {
        printf("parent scope %s\r\n", buf);
    }
}

int main(int argc, char* argv[])
{
    test_fn(0);
    printf("test_fn(0) ===> done");

    test_fn(1);
    printf("test_fn(1) ===> done");
}

https://godbolt.org/z/5AncQ3

Irrespective of which condition get executed stack pointer (SP) is always increased by 200, which make sense also because compiler produces code at compile time and has to generate code for all the cases runtime will take (Use this link to see the disassembly https://godbolt.org/z/5AncQ3). So, it will always allocate (increase the SP) all the memory irrespective where you declared in the function. 

I agree with your point that we should declare variable where we use them. But advantage of making sure that we declare at the start of function is that we can remove any doubts about how much memory our function is using (like confusion we had here). It really helps both coder/reviewer to easily reason about memory requirement of function and avoid any potential stack overflow bugs.

JeffreyRichter: I was thinking more like this: https://godbolt.org/z/-xAjoa
#include <string.h>

int  test_fn(int cond)
{
    if (cond)
    {
        char buf[300] = {0}; // change this to anything under 980 and "sub rsp, 1000" always says 1000
        return strlen(buf);
    }
    else
    {
        char buf[980] = {0};
        return strlen(buf);
    }
}

int main(int argc, char* argv[])
{
    test_fn(0);
    test_fn(1);
}

hihigupt: Yes, compiler will try to do max of all branches. But do you think if we follow same pattern as AZRTOS, we will be careful with our allocation and avoid any possible stack overflow bugs?

JeffreyRichter: I think you may be conflating 2 separate issues: the first issue is whether we should declare variables near their use or near the top of the function. I think near use is preferred because of code readability & maintainability. In addition, the example I show here means that compiler will allocate less stack space than if you declared both buffers (with different variable names) at the top of the function. 

The 2nd issue is about stack usage. Sure, we always want to minimize stack usage regardless of where an engineer declares the variables inside a function.

The fact of the matter is that we and the customer have no idea how much stack space is required to do ANYTHING. And, as we change function implementations over times, this changes dynamically. In other words, a function may need 1KB of stack today and 1.5KB of stack tomorrow. We don't document the amount a stack a function needs and therefore, this is not a defined CONTACT that we must adhere to as time goes on. 

Another thing to be aware of is that Azure services were not really designed well for embedded devices. For example, XML and JSON are not tiny payloads to create/parse. And, there is no way for a client to pass a payload size limit to a service that it has to adhere to and the service also doesn't make a contract over this so it can change over time.

For our Embedded C SDK HTTP processing, we do use the stack for somethings (like logging) but we do allow the customer to specify the maximum size of a log entry. We fill this stack buffer as best we can and then hand the buffer to the customer to log it how they wish. They can always turn off logging or reduce the buffer size (to reduce stack usage) but this reduces the usefulness of the logs.  We expose the knobs as I'm unaware of any better solutions.

PR #653, opened 2020-04-29
Update CMakeLists for use as submodule

danewalton:
[`CMAKE_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_SOURCE_DIR.html) refers to the top level of the source tree which could be a user's project using us as a submodule.

Instead, we can use [`<PROJECT-NAME>_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html) to find the top level of this project.

Additionally, some previous target includes were being made `PRIVATE` which need to be public for include files to resolve for someone using them. Core already does this:
https://github.com/Azure/azure-sdk-for-c/blob/3011f16b299e5aaf5a752d36918c0bfc48444269/sdk/core/core/CMakeLists.txt#L67-L72
I have updated those.

### DIFF
--- a/sdk/iot/core/CMakeLists.txt
+++ b/sdk/iot/core/CMakeLists.txt
@@ -25,6 +25,6 @@ add_library (az::iot::core ALIAS ${TARGET_NAME})
 
 # set coverage excluding for az_core. Don't show coverage outside iot
 set(COV_EXCLUDE
-    ${CMAKE_SOURCE_DIR}/sdk/core/core/inc/*
-    ${CMAKE_SOURCE_DIR}/sdk/core/core/internal/*)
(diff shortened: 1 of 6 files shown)


barcharcraz:
This is pretty much correct, however in some cases it may be shorter to use `PROJECT_SOURCE_DIR` and friends (that is, the source dir for the most recent project() call) that can happen in another PR, however.

danewalton:
> This is pretty much correct, however in some cases it may be shorter to use `PROJECT_SOURCE_DIR` and friends (that is, the source dir for the most recent project() call) that can happen in another PR, however.

Yup in some cases it can but we have several calls to `project()` which can muddy that. I can double check in another like you said 👍🏽

PR #654, opened 2020-04-29
Prefix build macros, internal prefix for precondition checks

antkmsft:
NO_PRECONDITION_CHECKING and BUILD_MOCK are not AZ_ prefixed. Therefore, we can less rely on the fact that they are not taken by something else in customer's build environment. We should add prefixes.

NO_PRECONDITION_CHECKING is part of public interface, it is documented. Therefore, it should be named AZ_NO_PRECONDITION_CHECKING.

BUILD_MOCK as a macro is not documented. Therefore, it should be at least _az_BUILD_MOCK.

Precondition checks are internal macros, therefore prefix is changed from `AZ_` to `_az_`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,12 +14,12 @@ option(BUILD_PRECONDITIONS "Build SDK with preconditions enabled" ON)
 
 #enable mock functions with link option -ld
 if (NOT BUILD_PRECONDITIONS)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNO_PRECONDITION_CHECKING")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DAZ_NO_PRECONDITION_CHECKING")
 endif()
 
 #enable mock functions with link option -ld
 if(UNIT_TESTING_MOCK_ENABLED)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMOCK_ENABLED")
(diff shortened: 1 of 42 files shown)


antkmsft:
Updated: added prefixes for precondition check macros

CMakeLists.txt
```diff
@@ -14,12 +14,12 @@ option(BUILD_PRECONDITIONS "Build SDK with preconditions enabled" ON)
 
 #enable mock functions with link option -ld
 if (NOT BUILD_PRECONDITIONS)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNO_PRECONDITION_CHECKING")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DAZ_NO_PRECONDITION_CHECKING")
```
antkmsft, line 17:
```suggestion
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DAZ_NO_PRECONDITION_CHECKING") 
```

CMakeLists.txt
```diff
@@ -14,12 +14,12 @@ option(BUILD_PRECONDITIONS "Build SDK with preconditions enabled" ON)
...
```
antkmsft, line 17:
```suggestion
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DAZ_NO_PRECONDITION_CHECKING")
```

PR #658, opened 2020-04-30
IoT Core, Provisioning state APIs implementations

CIPop:
- Implementations for az_iot_core state and retry public APIs + UT.
- Adding operation_state parsing APIs + UT + updated sample.
- Clarifying naming for Provisioning APIs (breaking API change).
- Removing some az_iot_core functions from the public API surface.
- Fixing Provisioning invalid JSON parsing bug + UT.

Doc:
- Updating PUML (switching to svg rendering).
- Changing from @note to @remark for non-important items (leaving @note for paragraphs that _must_ be understood, @remarks for clarifications).

### DIFF
--- a/sdk/iot/core/inc/az_iot_core.h
+++ b/sdk/iot/core/inc/az_iot_core.h
@@ -21,12 +21,16 @@
 /**
  * @brief Azure IoT service MQTT connection properties.
  *
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
  */
+
 enum
 {
-  AZ_CLIENT_DEFAULT_MQTT_CONNECT_PORT = 8883,
-  AZ_CLIENT_DEFAULT_MQTT_CONNECT_CLEAN_SESSION = 0x20,
-  AZ_CLIENT_DEFAULT_MQTT_CONNECT_KEEPALIVE_SECONDS = 240
+  AZ_IOT_DEFAULT_MQTT_CONNECT_PORT = 8883,
+  AZ_IOT_DEFAULT_MQTT_CONNECT_KEEPALIVE_SECONDS = 240
 };
 
 /**
@@ -59,34 +63,27 @@ typedef enum
 } az_iot_status;
 
 /**
- * @brief Get the #az_iot_status from an int.
- *
- * @param[in] status_int The int with the status number.
- * @param[out] status The #az_iot_status* with the status enum.
- * @return The #az_result with the result of the get operation.
- *  @retval #AZ_OK If the int is an #az_iot_status enum. `status` will be set to the according
- * enum.
- *  @retval #AZ_ERROR_ITEM_NOT_FOUND If the int is NOT an #az_iot_status enum.
- */
-
-AZ_NODISCARD az_result az_iot_get_status_from_uint32(uint32_t status_int, az_iot_status* status);
-
-/**
  * @brief Checks if the status indicates a successful operation.
  *
  * @param[in] status The #az_iot_status to verify.
- * @return True if the status indicates success. False otherwise.
+ * @return `true` if the status indicates success. `false` otherwise.
  */
-AZ_NODISCARD bool az_iot_is_success_status(az_iot_status status);
+AZ_INLINE bool az_iot_is_success_status(az_iot_status status)
+{
+  return status < AZ_IOT_STATUS_BAD_REQUEST;
+}
 
 /**
  * @brief Checks if the status indicates a retriable error occurred during the
  *        operation.
  *
  * @param[in] status The #az_iot_status to verify.
- * @return True if the operation should be retried. False otherwise.
+ * @return `true` if the operation should be retried. `false` otherwise.
  */
-AZ_NODISCARD bool az_iot_is_retriable_status(az_iot_status status);
+AZ_INLINE bool az_iot_is_retriable_status(az_iot_status status)
+{
+  return ((status == AZ_IOT_STATUS_THROTTLED) || (status == AZ_IOT_STATUS_SERVER_ERROR));
+}
 
 /**
  * @brief Calculates the recommended delay before retrying an operation that failed.
@@ -94,7 +91,7 @@ AZ_NODISCARD bool az_iot_is_retriable_status(az_iot_status status);
  * @param[in] operation_msec The time it took, in milliseconds, to perform the operation that
  *                           failed.
  * @param[in] attempt The number of failed retry attempts.
- * @param[in] retry_delay_msec The minimum time, in milliseconds, to wait before a retry.
+ * @param[in] min_retry_delay_msec The minimum time, in milliseconds, to wait before a retry.
  * @param[in] max_retry_delay_msec The maximum time, in milliseconds, to wait before a retry.
  * @param[in] random_msec A random value between 0 and the maximum allowed jitter, in milliseconds.
  * @return The recommended delay in milliseconds.
@@ -102,12 +99,12 @@ AZ_NODISCARD bool az_iot_is_retriable_status(az_iot_status status);
 AZ_NODISCARD int32_t az_iot_retry_calc_delay(
     int32_t operation_msec,
     int16_t attempt,
-    int32_t retry_delay_msec,
+    int32_t min_retry_delay_msec,
     int32_t max_retry_delay_msec,
     int32_t random_msec);
 
 /**
- * @brief az_span_token is a string tokenizer for az_span.
+ * @brief _az_span_token is a string tokenizer for az_span.
  *
  * @param[in] source The #az_span with the content to be searched on. It must be a non-empty
  * #az_span.
@@ -120,7 +117,7 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
  * occurrence of (but not including the) `delimiter`, or the end of `source` if `delimiter` is not
  * found. If `source` is empty, AZ_SPAN_NULL is returned instead.
  */
-AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
+AZ_NODISCARD az_span _az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
 
 /**
  * @brief _az_iot_u32toa_size gives the length, in bytes, of the string that would represent the given number.
--- a/sdk/iot/core/inc/az_iot_core.h
+++ b/sdk/iot/core/inc/az_iot_core.h
@@ -5,6 +5,11 @@
  * @file az_iot_core.h
  *
  * @brief Azure IoT common definitions.
+ *
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
  */
 
 #ifndef _az_IOT_CORE_H
@@ -17,15 +22,6 @@
 #include <stdint.h>
 
 #include <_az_cfg_prefix.h>
-
-/**
- * @brief Azure IoT service MQTT connection properties.
- *
- * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
- * prefixed with an underscore ('_') directly in your application code. These symbols
- * are part of Azure SDK's internal implementation; we do not document these symbols
- * and they are subject to change in future versions of the SDK which would break your code.
- */
 
 enum
 {
@@ -68,7 +64,7 @@ typedef enum
  * @param[in] status The #az_iot_status to verify.
  * @return `true` if the status indicates success. `false` otherwise.
  */
-AZ_INLINE bool az_iot_is_success_status(az_iot_status status)
+AZ_NODISCARD AZ_INLINE bool az_iot_is_success_status(az_iot_status status)
 {
   return status < AZ_IOT_STATUS_BAD_REQUEST;
 }
@@ -80,7 +76,7 @@ AZ_INLINE bool az_iot_is_success_status(az_iot_status status)
  * @param[in] status The #az_iot_status to verify.
  * @return `true` if the operation should be retried. `false` otherwise.
  */
-AZ_INLINE bool az_iot_is_retriable_status(az_iot_status status)
+AZ_NODISCARD AZ_INLINE bool az_iot_is_retriable_status(az_iot_status status)
 {
   return ((status == AZ_IOT_STATUS_THROTTLED) || (status == AZ_IOT_STATUS_SERVER_ERROR));
 }
@@ -104,7 +100,7 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
     int32_t random_msec);
 
 /**
- * @brief _az_span_token is a string tokenizer for az_span.
+ * @brief String tokenizer for #az_span.
  *
  * @param[in] source The #az_span with the content to be searched on. It must be a non-empty
  * #az_span.
@@ -120,7 +116,7 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
 AZ_NODISCARD az_span _az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
 
 /**
- * @brief _az_iot_u32toa_size gives the length, in bytes, of the string that would represent the given number.
+ * @brief Gives the length, in bytes, of the string that would represent the given number.
  *
  * @param[in] number The number whose length, as a string, is to be evaluated.
  * @return The length (not considering null terminator) of the string that would represent the given
--- a/sdk/iot/core/src/az_iot_core.c
+++ b/sdk/iot/core/src/az_iot_core.c
@@ -19,11 +19,11 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
     int32_t max_retry_delay_msec,
     int32_t random_msec)
 {
-  AZ_PRECONDITION_RANGE(0, operation_msec, INT32_MAX - 1);
-  AZ_PRECONDITION_RANGE(0, attempt, INT16_MAX - 1);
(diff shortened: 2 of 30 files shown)


CIPop:
@gilbertw @JeffreyRichter @ericwol-msft PTAL at public API changes (.h files)

CIPop:
Master got updated and broke this.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -228,16 +233,19 @@ typedef struct az_iot_provisioning_client_registration_state
 typedef struct az_iot_provisioning_client_register_response
 {
   az_iot_status status; /**< The current request status.
-                         * @note The authoritative response for the device registration operation
+                         * @remark The authoritative response for the device registration operation
                          * (which may require several requests) is available only through
-                         * #registration_state.  */
+                         * #operation_status.  */
```
danewalton, line 238: I think this should be backticks. `operation_status` isn't a type and won't resolve.

sdk/iot/provisioning/samples/src/paho_iot_provisioning_example.c
```diff
@@ -300,42 +312,42 @@ static int get_operation_status()
...
       }
-
-      is_operation_complete = true;
     }
 
     MQTTClient_freeMessage(&message);
     MQTTClient_free(topic);
   }
-
+
```
danewalton, line 350: unnecessary space?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -335,6 +335,11 @@ AZ_INLINE az_result _az_iot_provisioning_client_payload_registration_state_parse
     }
   }
 
+  if (found_assigned_hub != found_device_id)
```
danewalton, line 338:
```suggestion
  if (!(found_assigned_hub && found_device_id))
```
having not equals makes me believe `false == false` would pass here. I have to look back up at the implementation to see if that is caught or not.

danewalton: or even have `if (found_assigned && found_device_id)` and then after default to `AZ_ERROR_ITEM_NOT_FOUND`

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -58,56 +62,49 @@ typedef enum
...
+ * @param[in] min_retry_delay_msec The minimum time, in milliseconds, to wait before a retry.
  * @param[in] max_retry_delay_msec The maximum time, in milliseconds, to wait before a retry.
  * @param[in] random_msec A random value between 0 and the maximum allowed jitter, in milliseconds.
  * @return The recommended delay in milliseconds.
  */
 AZ_NODISCARD int32_t az_iot_retry_calc_delay(
     int32_t operation_msec,
     int16_t attempt,
-    int32_t retry_delay_msec,
+    int32_t min_retry_delay_msec,
```
gilbertw, line 102: Why split min and max?  Guarantees to never be late are extremely hard to enforce due to schedulers.

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -120,7 +117,7 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
  * occurrence of (but not including the) `delimiter`, or the end of `source` if `delimiter` is not
  * found. If `source` is empty, AZ_SPAN_NULL is returned instead.
  */
-AZ_NODISCARD az_span az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
+AZ_NODISCARD az_span _az_span_token(az_span source, az_span delimiter, az_span* out_remainder);
```
gilbertw, line 120: Based on the method name this should be in az_span.  This also sounds like it might be a json parser api.  Is there a reason for not adding to az_span or using the json parser?

CIPop: That's right and the reason I moved it to internal - it should definitely not be a public API within the IoT headers.
There was some discussion a while back (@JeffreyRichter @ewertons have more context) and it was decided to remain outside of the span API - not sure why.

I do think the token API is useful: 1. we used it and liked it, 2. C's StdLib provides `strtok`

CIPop: I will defer this to offline discussions and a later PR.

sdk/iot/core/inc/az_iot_core.h
CIPop: The API is using the internal function provided by AZ SDK which has retry_delay_msec defined as the minimum retry time.
If found that confusing so I renamed the arg in our public API.

I don't understand the timing enforcement guarantee problem. This API is implementing an exponential back-off with jitter. It's very common with distributed (network) protocols to implement a randomized back-off. The exponential aspect is in-place to avoid servers being DDoSed by devices.

More details on how this should be used here: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/doc/mqtt_state_machine.md#retry-timing

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -21,12 +21,16 @@
...
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
  */
+
 enum
 {
-  AZ_CLIENT_DEFAULT_MQTT_CONNECT_PORT = 8883,
-  AZ_CLIENT_DEFAULT_MQTT_CONNECT_CLEAN_SESSION = 0x20,
```
ericwol-msft, line 28: why was this removed, I was using it?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: > having not equals makes me believe false == false would pass here

And you would be correct :). C doesn't have XOR. false, false and true,true are the only acceptable values: IoT Hub either sends the hub and device or _none at all_ and instead sends the error code.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: Right: I'm going to fix all of them (searching for # in generated html).

sdk/iot/core/inc/az_iot_core.h
JeffreyRichter: It should not be public and I think it was moved into core (maybe its still an outstanding PR?).
Maybe @ahsonkhan or @vhvb1989 know?

sdk/iot/doc/mqtt_state_machine.md
```diff
@@ -61,14 +61,23 @@ if(az_failed(az_iot_hub_client_sas_get_signature(client, unix_time + 3600, signa
     // error.
 }
 
-// Base64Encode the HMAC256 of the az_span_ptr(signature) with the Shared Access Key.
+// Base64Encode the HMAC256 of the az_span_ptr(signature)/az_span_len(signature) with the Shared Access Key.
```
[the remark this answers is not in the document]
CIPop: Oh, I meant using those two, not divide. I see how that can be misinterpreted.
Fixed both.

sdk/iot/core/inc/az_iot_core.h
CIPop: It's not portable to a lot of the MQTT libraries. Most of them control clean session with a boolean flag and not the actual bit within the `CONNACK` message.
Also, it's not clear what the default meant. The guidance is to always have CleanSession=false. (see https://github.com/Azure/azure-sdk-for-c/pull/658/files#diff-19fef307f19572073d4a230f548b6b41R72 within the MQTT state machine MD file updated in this PR.)

sdk/iot/core/inc/az_iot_core.h
ahsonkhan: It hasn't moved to core yet. Final decision on semantics of the API was pending, which is why we decided to leave it in iot/common for now (btw, we probably want to rename iot/core to iot/common to avoid confusion).

We can (and should) move it to core in the near future (post preview?).

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -21,12 +21,16 @@
 /**
  * @brief Azure IoT service MQTT connection properties.
  *
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
```
ahsonkhan, line 24: Consider moving this `@note` to the `@file` comment block above, like other places. I don't think it belongs to this particular enum.

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -21,12 +21,16 @@
...
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
  */
+
```
ahsonkhan, line 29: Remove new line?

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -58,56 +62,49 @@ typedef enum
...
-
 /**
  * @brief Checks if the status indicates a successful operation.
  *
  * @param[in] status The #az_iot_status to verify.
- * @return True if the status indicates success. False otherwise.
+ * @return `true` if the status indicates success. `false` otherwise.
  */
-AZ_NODISCARD bool az_iot_is_success_status(az_iot_status status);
+AZ_INLINE bool az_iot_is_success_status(az_iot_status status)
```
ahsonkhan, line 71: Keep `AZ_NODISCARD`?

sdk/iot/core/inc/az_iot_core.h
```diff
@@ -58,56 +62,49 @@ typedef enum
...
     int32_t operation_msec,
     int16_t attempt,
-    int32_t retry_delay_msec,
+    int32_t min_retry_delay_msec,
     int32_t max_retry_delay_msec,
     int32_t random_msec);
 
 /**
- * @brief az_span_token is a string tokenizer for az_span.
+ * @brief _az_span_token is a string tokenizer for az_span.
```
ahsonkhan, line 107: Remove method name from doc to avoid redundant info and having to make changes in multiple places in the future.

```suggestion
 * @brief String tokenizer for #az_span.
```

sdk/iot/core/src/az_iot_core.c
```diff
@@ -8,74 +8,41 @@
...
+
 #include <_az_cfg.h>
 
-AZ_NODISCARD az_result az_iot_get_status_from_uint32(uint32_t status_int, az_iot_status* status)
+AZ_NODISCARD int32_t az_iot_retry_calc_delay(
+    int32_t operation_msec,
+    int16_t attempt,
+    int32_t min_retry_delay_msec,
+    int32_t max_retry_delay_msec,
+    int32_t random_msec)
```
ahsonkhan, line 20: This is a relatively complex API signature, especially for something that is public. Any way to design it more simply?

Does attempt need to be an `int16_t`, or can we just keep it `int32_t` like the rest?

sdk/iot/core/src/az_iot_core.c
```diff
@@ -8,74 +8,41 @@
...
 {
-  switch (status_int)
+  _az_PRECONDITION_RANGE(0, operation_msec, INT32_MAX - 1);
```
ahsonkhan, line 22: Why is `INT32_MAX` not allowed?

sdk/iot/core/tests/cmocka/test_az_iot_core.c
```diff
@@ -157,13 +92,65 @@ static void _az_iot_u32toa_size_success(void** state)
...
+
+  assert_int_equal(
+      INT32_MAX - 1,
+      az_iot_retry_calc_delay(0, INT16_MAX - 1, INT32_MAX - 1, INT32_MAX - 1, INT32_MAX - 1));
+}
+
+#ifdef _MSC_VER
+// warning C4113: 'void (__cdecl *)()' differs in parameter lists from 'CMUnitTestFunction'
+#pragma warning(disable : 4113)
+#endif
```
ahsonkhan, line 143: Can you please explain exactly what about `CMUnitTestFunction` is requiring this warning disable?

sdk/iot/doc/mqtt_state_machine.md
```diff
@@ -146,6 +155,8 @@ _Example:_
     }
 ```
 
+__Important:__ C2D messages are not enqueued until the device establishes the first MQTT session (connects for the first time to IoT Hub). The C2D message queue is preserved (according to the per-message time-to-live) as long as the device connects with Clean Session `false`.
```
ahsonkhan, line 158: What is `Clean Session`? Is it a term that folks familiar with the IoT Hub domain would be aware off, or is it part of our API surface? Consider adding a link to a reference.

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -361,7 +366,7 @@ typedef struct az_iot_hub_client_c2d_request
  * @param[out] out_request If the message is a C2D request, this will contain the
  *                         #az_iot_hub_client_c2d_request
  * @return #az_result
- *         - #AZ_ERROR_IOT_TOPIC_NO_MATCH if the topic is not matching the expected format.
+ *         - `AZ_ERROR_IOT_TOPIC_NO_MATCH` if the topic is not matching the expected format.
```
ahsonkhan, line 369: Why was this change necessary? It should link to the correct `az_result` value with `#`

sdk/iot/hub/inc/az_iot_hub_client.h
```diff
@@ -411,7 +416,7 @@ typedef struct az_iot_hub_client_method_request
  * @param[out] out_request If the message is a method request, this will contain the
  *                         #az_iot_hub_client_method_request.
  * @return #az_result
- *         - #AZ_ERROR_IOT_TOPIC_NO_MATCH if the topic is not matching the expected format.
+ *         - `AZ_ERROR_IOT_TOPIC_NO_MATCH` if the topic is not matching the expected format.
```
ahsonkhan, line 419: Same here/elsewhere related to `az_result`. Keep `#`.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
```diff
@@ -251,18 +259,63 @@ typedef struct az_iot_provisioning_client_register_response
...
+/**
+ * @brief Returns the #az_iot_provisioning_client_operation_status of a
+ * #az_iot_provisioning_client_register_response object.
+ *
+ * @param[in] response The #az_iot_provisioning_client_register_response obtained after a successful
+ *                     call to #az_iot_provisioning_client_parse_received_topic_and_payload.
+ * @param[out] out_operation_status The registration operation status.
+ * @return #az_result
+ *         - #AZ_ERROR_PARSER_UNEXPECTED_CHAR if the string contains an unexpected value.
+ *         @remark See #az_iot_provisioning_client_register_response for acceptable values.
```
ahsonkhan, line 295: I don't believe this will render as intended. Move the `@remark` after the @brief in its own standalone line, if it isn't important that it be connected to and display next to the `AZ_ERROR_PARSER_UNEXPECTED_CHAR` result.

JeffreyRichter:
Post preview is fine.


sdk/iot/core/inc/az_iot_core.h
CIPop: Moved the entire comment in the file section.

sdk/iot/core/inc/az_iot_core.h
CIPop: Changed this and the one below.

sdk/iot/core/src/az_iot_core.c
CIPop: > This is a relatively complex API signature, especially for something that is public. Any way to design it more simply?

I had a different approach that would require state. We all agreed the overheads are too big.

> Does attempt need to be an int16_t, or can we just keep it int32_t like the rest?

It's the same as the internal that Az SDKs proposed and so I can't use a larger int unless the internal also changes.

sdk/iot/core/src/az_iot_core.c
CIPop: It's allowed but it'll generate a compile-time overflow warning and break the build.

sdk/iot/core/tests/cmocka/test_az_iot_core.c
CIPop: It's the cmocka `void **state` parameter. I think it's not adding any value (we don't use the state feature) and it adds 3 lines of extra code.
The warning is for not using the same signature (I'm using va-args `test()` and it's expecting `test(void **)`). Using va_args (any empty () definition) is a valid C99 language feature.

sdk/iot/doc/mqtt_state_machine.md
CIPop: I won't add it here though: it has its own paragraph under Connecting.

sdk/iot/hub/inc/az_iot_hub_client.h
CIPop: Unfortunately it doesn't and leaves `#AZ_ERROR_IOT_TOPIC_NO_MATCH` in the doc.

sdk/iot/hub/inc/az_iot_hub_client.h
CIPop: See above.

sdk/iot/provisioning/inc/az_iot_provisioning_client.h
CIPop: You're right: it renders at top-level, not within the return paragraph.

CIPop: On second thought, I'm removing this completely since the value comes from a struct that we've parsed and it's unclear.

sdk/iot/hub/inc/az_iot_hub_client.h
ahsonkhan: I think that is because that error code isn't documented. I use `#AZ_OK` and others in `az_span.h` etc. and it works fine, because it has a `///<` comment.

PR #665, opened 2020-05-01
Illustrate how we can make az_span_is_valid non-inlineable (up to -~2KB lib size on x86-64)

antkmsft:
…


ahsonkhan:
Does this result in 2KB higher code size or lower? What was this PR trying to showcase and what's the outcome? Do we want it inlineable or not?

PR #667, opened 2020-05-01
_az_url_encode => _az_span_url_encode

antkmsft:


sdk/core/core/internal/az_span_internal.h
```diff
@@ -5,8 +5,11 @@
 #define _az_SPAN_INTERNAL_H
 
 #include <az_precondition_internal.h>
+#include <az_result.h>
 #include <az_span.h>
 
+#include <stdint.h>
```
ahsonkhan, line 11: We shouldn't need this `#include`. What's requiring this?

sdk/core/core/src/az_span.c
```diff
@@ -618,3 +620,62 @@ _az_span_scan_until(az_span self, _az_predicate predicate, int32_t* out_index)
...
+  uint8_t* const src_ptr = az_span_ptr(source);
+  uint8_t* dest_ptr = az_span_ptr(destination);
+
+  for (int32_t src_idx = 0; src_idx < source_size; ++src_idx)
+  {
+    uint8_t c = src_ptr[src_idx];
+    if (!_az_span_url_should_encode(c))
+    {
+      *dest_ptr = c;
+      dest_ptr += 1;
```
ahsonkhan, line 667: Here's the potential overflow, btw, that I was referring to.

sdk/core/core/internal/az_span_internal.h
antkmsft: we use int32_t below

sdk/core/core/src/az_span.c
antkmsft: Check this out: https://github.com/Azure/azure-sdk-for-c/pull/671

sdk/core/core/internal/az_span_internal.h
ahsonkhan: I get that, we also use `int32_t` in `_az_span_diff`, so we didn't need it before. I think it is already being transitively included with `az_span.h` or some other header.

So, did you just add it as an abundance of  precaution, or was the build somehow failing now without it, when it should have been working fine before?

sdk/core/core/src/az_span.c
ahsonkhan: I don't think that fixes it, especially since that only address preconditions and we want that safety at release/runtime as well.

sdk/core/core/internal/az_span_internal.h
antkmsft: Generally, we shouldn't rely on the headers we include to include the headers we expect. This is also the reason why we include our headers first, and then the standard library headers - so that we know that our header would compile even if customer did not include some stdlib header before including us.
If a function is declared in a header, and I am including that header, and implementing that function, then (within our SDK) I can expect that it did include the header to not only compile, but to use all the types that are input and output parameters for the function.
So, we use NULL in headers, we include stddef, bool - stdbool, `*int*_t` - stdint.

sdk/core/core/src/az_span.c
antkmsft: Then what is the overflow? We checked the destination range above.

antkmsft: This is not a precondition: 
```
if (az_span_size(destination) < result_size)
  {
    *out_length = 0;
    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
  }
```

antkmsft: See the update, to that PR I stenghtened the size check for this function.

ahsonkhan: It's the unguarded addition by 3 for every byte in source (so 3x expansion) and then followed by direct pointer access on the destination (instead of using az_span APIs).

See this example:
```C
static void test_url_encode_fail(void** state)
{
  (void)state;

  uint8_t* p_array = (uint8_t*)malloc(715827883);
  memset(p_array, '@', 715827883); // Everything needs to be escaped.

  az_span source = az_span_init(p_array, 715827883);

  uint8_t dest_buffer[10];
  // destination is certainly too small
  az_span destination = AZ_SPAN_FROM_BUFFER(dest_buffer);

  int32_t length;

  // Exit code 0xc0000409 - indicates a stack buffer overflow, i.e. potential AV
  // result_size overflows to -2147483647
  az_result res = _az_span_url_encode(destination, source, &length);

  printf("Result: %d\n", res); // We never get here because of write access violation.
  assert_true(res == AZ_ERROR_INSUFFICIENT_SPAN_SIZE);

  free(p_array);
}
```

image

A change like this could help prevent invalid memory access, and only result in data loss (but likely too slow, so I would fix this differently, but want to illustrate the point):
```C
  for (int32_t src_idx = 0; src_idx < source_size; ++src_idx)
  {
    uint8_t c = src_ptr[src_idx];
    if (!_az_span_url_should_encode(c))
    {
      destination = az_span_copy_u8(destination, c);
    }
    else
    {
      destination = az_span_copy_u8(destination, '%');
      destination = az_span_copy_u8(destination, _az_number_to_upper_hex(c >> 4));
      destination = az_span_copy_u8(destination, _az_number_to_upper_hex(c & 0x0F));
    }
  }
```

Also, we end up returning a negative length, even if we somehow got lucky and didn't AV:
https://github.com/Azure/azure-sdk-for-c/blob/ba67151753118b484347b16ccc18d1ba1d5342b9/sdk/core/core/src/az_span.c#L678

PR #670, opened 2020-05-02
Adding Azure IoT Logging

CIPop:
- Adding logging for MQTT received topic and payload, retry event and SAS token generation.
- Removing (void **state) from all IoT cmocka tests.
- Formatting changes

### DIFF
--- a/sdk/core/core/inc/az_log.h
+++ b/sdk/core/core/inc/az_log.h
@@ -44,6 +44,17 @@ typedef enum
   AZ_LOG_HTTP_RETRY = _az_LOG_MAKE_CLASSIFICATION(
       _az_FACILITY_HTTP,
       3), ///< First HTTP request did not succeed and will be retried.
+
+  AZ_LOG_MQTT_RECEIVED_TOPIC
+  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_MQTT, 1), ///< Accepted MQTT topic received.
+
+  AZ_LOG_MQTT_RECEIVED_PAYLOAD
+  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_MQTT, 2), ///< Accepted MQTT payload received.
+
+  AZ_LOG_IOT_RETRY = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 1), ///< IoT Client retry.
+
+  AZ_LOG_IOT_SAS_TOKEN
+  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 2), ///< IoT Client generated new SAS token.
 } az_log_classification;
 
 /**
--- a/sdk/iot/core/src/az_iot_core.c
+++ b/sdk/iot/core/src/az_iot_core.c
@@ -8,6 +8,7 @@
 #include <az_result.h>
 #include <az_span.h>
 
+#include <az_log_internal.h>
 #include <az_retry_internal.h>
 
 #include <_az_cfg.h>
@@ -25,9 +26,11 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
   _az_PRECONDITION_RANGE(0, max_retry_delay_msec, INT32_MAX - 1);
   _az_PRECONDITION_RANGE(0, random_msec, INT32_MAX - 1);
 
+  az_log_write(AZ_LOG_IOT_RETRY, AZ_SPAN_NULL);
+
   int32_t delay = _az_retry_calc_delay(attempt, min_retry_delay_msec, max_retry_delay_msec);
-  
-  if (delay < 0) 
+
+  if (delay < 0)
   {
     delay = max_retry_delay_msec;
   }
@@ -86,6 +89,6 @@ AZ_NODISCARD int32_t _az_iot_u32toa_size(uint32_t number)
       digit_count--;
     }
 
-    return digit_count; 
+    return digit_count;
   }
 }
--- a/sdk/iot/core/tests/cmocka/test_az_iot_core.c
+++ b/sdk/iot/core/tests/cmocka/test_az_iot_core.c
@@ -3,9 +3,10 @@
 
 #include "test_az_iot_core.h"
 #include <az_iot_core.h>
+#include <az_log.h>
+#include <az_precondition.h>
 #include <az_span.h>
 
-#include <az_precondition.h>
 #include <az_precondition_internal.h>
 
 #include <setjmp.h>
@@ -137,6 +138,37 @@ static void test_az_iot_retry_calc_delay_overflow_time_success()
       az_iot_retry_calc_delay(0, INT16_MAX - 1, INT32_MAX - 1, INT32_MAX - 1, INT32_MAX - 1));
 }
 
+static int _log_retry = 0;
+static void _log_listener(az_log_classification classification, az_span message)
+{
+  switch (classification)
+  {
+    case AZ_LOG_IOT_RETRY:
+      _log_retry++;
+      assert_int_equal(az_span_ptr(message), 0);
+      assert_int_equal(az_span_size(message), 0);
+      break;
+    default:
+      assert_true(false);
+  }
+}
+
+static void test_az_iot_provisioning_client_logging_succeed()
+{
+  az_log_classification const classifications[]
+      = { AZ_LOG_IOT_RETRY, AZ_LOG_END_OF_LIST };
+  az_log_set_callback(_log_listener);
+  az_log_set_classifications(classifications);
+
+  assert_int_equal(0, _log_retry);
+  _log_retry = 0;
+  assert_int_equal(2229, az_iot_retry_calc_delay(5, 1, 500, 100000, 1234));
+  assert_int_equal(1, _log_retry);
+
+  az_log_set_classifications(NULL);
+  az_log_set_callback(NULL);
+}
+
 #ifdef _MSC_VER
 // warning C4113: 'void (__cdecl *)()' differs in parameter lists from 'CMUnitTestFunction'
 #pragma warning(disable : 4113)
@@ -151,6 +183,7 @@ int test_az_iot_core()
     cmocka_unit_test(test_az_iot_is_retriable_status_translate_success),
     cmocka_unit_test(test_az_iot_retry_calc_delay_common_timings_success),
     cmocka_unit_test(test_az_iot_retry_calc_delay_overflow_time_success),
+    cmocka_unit_test(test_az_iot_provisioning_client_logging_succeed),
   };
   return cmocka_run_group_tests_name("az_iot_core", tests, NULL, NULL);
 }
--- a/sdk/iot/core/tests/cmocka/test_az_iot_core.c
+++ b/sdk/iot/core/tests/cmocka/test_az_iot_core.c
@@ -145,7 +145,7 @@ static void _log_listener(az_log_classification classification, az_span message)
   {
     case AZ_LOG_IOT_RETRY:
       _log_retry++;
-      assert_int_equal(az_span_ptr(message), 0);
+      assert_ptr_equal(az_span_ptr(message), (void *)0);
       assert_int_equal(az_span_size(message), 0);
       break;
     default:
--- a/sdk/iot/core/tests/cmocka/test_az_iot_core.c
+++ b/sdk/iot/core/tests/cmocka/test_az_iot_core.c
@@ -157,16 +157,16 @@ static void test_az_iot_provisioning_client_logging_succeed()
 {
   az_log_classification const classifications[]
       = { AZ_LOG_IOT_RETRY, AZ_LOG_END_OF_LIST };
-  az_log_set_callback(_log_listener);
   az_log_set_classifications(classifications);
+  az_log_set_callback(_log_listener);
 
   assert_int_equal(0, _log_retry);
   _log_retry = 0;
   assert_int_equal(2229, az_iot_retry_calc_delay(5, 1, 500, 100000, 1234));
   assert_int_equal(1, _log_retry);
 
-  az_log_set_classifications(NULL);
   az_log_set_callback(NULL);
+  az_log_set_classifications(NULL);
 }
 
 #ifdef _MSC_VER
--- a/sdk/iot/hub/src/az_iot_hub_client_c2d.c
+++ b/sdk/iot/hub/src/az_iot_hub_client_c2d.c
@@ -60,9 +60,8 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
   _az_PRECONDITION_NOT_NULL(out_request);
   (void)client;
 
-  az_span token;
   az_span reminder;
-  token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
+  az_span token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
   if (az_span_ptr(reminder) == NULL)
   {
     return AZ_ERROR_IOT_TOPIC_NO_MATCH;
--- a/sdk/iot/hub/src/az_iot_hub_client_c2d.c
+++ b/sdk/iot/hub/src/az_iot_hub_client_c2d.c
@@ -4,9 +4,11 @@
 #include <stdint.h>
 
 #include "az_iot_hub_client.h"
-#include <az_precondition_internal.h>
 #include <az_result.h>
 #include <az_span.h>
+
+#include <az_log_internal.h>
+#include <az_precondition_internal.h>
 
 #include <_az_cfg.h>
 
@@ -59,17 +61,18 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
   (void)client;
 
   az_span token;
-  token = _az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  if (az_span_ptr(received_topic) != NULL)
-  {
-    token = _az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-    AZ_RETURN_IF_FAILED(
-        az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
-  }
-  else
+  az_span reminder;
+  token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
+  if (az_span_ptr(reminder) == NULL)
   {
     return AZ_ERROR_IOT_TOPIC_NO_MATCH;
   }
+
+  az_log_write(AZ_LOG_MQTT_RECEIVED_TOPIC, received_topic);
+
+  token = _az_span_token(reminder, c2d_topic_suffix, &reminder);
+  AZ_RETURN_IF_FAILED(
+      az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
 
   return AZ_OK;
 }
--- a/sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
+++ b/sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
@@ -3,10 +3,10 @@
 
 #include "test_az_iot_hub_client.h"
 #include <az_iot_hub_client.h>
-#include <az_span.h>
-#include <az_test_span.h>
-
+#include <az_log.h>
 #include <az_precondition.h>
+#include <az_span.h>
+
 #include <az_precondition_internal.h>
 
 #include <setjmp.h>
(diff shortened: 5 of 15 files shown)


sdk/iot/core/src/az_iot_core.c
```diff
@@ -25,9 +26,11 @@ AZ_NODISCARD int32_t az_iot_retry_calc_delay(
   _az_PRECONDITION_RANGE(0, max_retry_delay_msec, INT32_MAX - 1);
   _az_PRECONDITION_RANGE(0, random_msec, INT32_MAX - 1);
 
+  az_log_write(AZ_LOG_IOT_RETRY, AZ_SPAN_NULL);
```
ahsonkhan, line 29: Do you intentionally want to log null (i.e. empty span)?

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -59,17 +61,18 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
...
-  token = _az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-  if (az_span_ptr(received_topic) != NULL)
-  {
-    token = _az_span_token(received_topic, c2d_topic_suffix, &received_topic);
-    AZ_RETURN_IF_FAILED(
-        az_iot_hub_client_properties_init(&out_request->properties, token, az_span_size(token)));
-  }
-  else
+  az_span reminder;
+  token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
```
ahsonkhan, line 65: nit: Any reason the declaration isn't just inline with the call?
```suggestion
  az_span token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
```

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
```diff
@@ -208,14 +195,11 @@ static void test_az_iot_hub_client_c2d_parse_received_topic_no_props_succeed(voi
   az_iot_hub_client_c2d_request out_request;
 
   assert_int_equal(
-      az_iot_hub_client_c2d_parse_received_topic(&client, test_url_no_props, &out_request),
-      AZ_OK);
+      az_iot_hub_client_c2d_parse_received_topic(&client, test_url_no_props, &out_request), AZ_OK);
 }
 
-static void test_az_iot_hub_client_c2d_parse_received_topic_reject(void** state)
+static void test_az_iot_hub_client_c2d_parse_received_topic_reject()
```
ahsonkhan, line 201: I might be mistaken, but isn't the `void**` parameter required for testing without preconditions or filtering test runs? @vhvb1989 - do you remember, or is it fine to remove these parameters?

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -466,6 +469,55 @@ static void test_az_iot_provisioning_client_operation_complete_translate_succeed
...
+static const az_span _log_received_payload = AZ_SPAN_LITERAL_FROM_STR("LOG_PAYLOAD");
+
+static int _log_invoked_topic = 0;
+static int _log_invoked_payload = 0;
+static void _log_listener(az_log_classification classification, az_span message)
+{
+  switch (classification)
+  {
+    case AZ_LOG_MQTT_RECEIVED_TOPIC:
+      assert_memory_equal(
```
ahsonkhan, line 482: Is it feasible to use assert_string_equal here so that if the assert fails we can see the expected vs actual?

If `assert_memory_equal` also show that comparison, then ignore this.

sdk/iot/core/src/az_iot_core.c
CIPop: I would like to log an int similar to ETW but the logging callback doesn't allow that.
I don't want to stack-alloc a span that contains the int values so there's nothing else I can do here.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
CIPop: I don't think so. I'll give that a try.

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_c2d.c
CIPop: I doubt it - provisioning has the state removed since I first checked it in.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
CIPop: Unfortunately az_span is not a string and so doesn't guarantee the null terminator.
UTs are used as programming patterns so I don't want to give the wrong impression (in this particular case, but not others - see the sas token logger- that would not be true).

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -491,6 +426,53 @@ static void test_az_iot_hub_client_methods_parse_received_topic_response_topic_f
...
+      _log_invoked_topic++;
+      break;
+    default:
+      assert_true(false);
+  }
+}
+
+static void test_az_iot_hub_client_methods_logging_succeed()
+{
+  az_log_classification const classifications[]
```
JeffreyRichter, line 448: order

sdk/iot/hub/tests/cmocka/test_az_iot_hub_client_methods.c
```diff
@@ -491,6 +426,53 @@ static void test_az_iot_hub_client_methods_parse_received_topic_response_topic_f
...
+  assert_true(az_iot_hub_client_init(&client, test_device_hostname, test_device_id, NULL) == AZ_OK);
+
+  az_iot_hub_client_method_request out_request;
+  assert_true(
+      az_iot_hub_client_methods_parse_received_topic(&client, _log_expected_topic, &out_request)
+      == AZ_OK);
+
+  assert_int_equal(1, _log_invoked_topic);
+
+  az_log_set_classifications(NULL);
```
JeffreyRichter, line 467: order

sdk/iot/provisioning/src/az_iot_provisioning_client.c
```diff
@@ -465,6 +467,9 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
     return AZ_ERROR_IOT_TOPIC_NO_MATCH;
   }
 
+  az_log_write(AZ_LOG_MQTT_RECEIVED_TOPIC, received_topic);
```
JeffreyRichter, line 470: NOTE: 2 separate calls to az_log_write are NOT guaranteed to be atomic as another thread could also be calling az_log_write at the same time. If you want the 2 log entries to be considered as an atomic unit, YOU must combine the messages into 1 and call az_log_write ONCE.

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -466,6 +469,55 @@ static void test_az_iot_provisioning_client_operation_complete_translate_succeed
...
+
+  az_iot_provisioning_client client;
+  az_iot_provisioning_client_register_response response;
+  assert_true(az_failed(az_iot_provisioning_client_parse_received_topic_and_payload(
+      &client, _log_received_topic, _log_received_payload, &response)));
+
+  assert_int_equal(1, _log_invoked_topic);
+  assert_int_equal(1, _log_invoked_payload);
+
+  az_log_set_classifications(NULL);
```
JeffreyRichter, line 517: order

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_parser.c
```diff
@@ -466,6 +469,55 @@ static void test_az_iot_provisioning_client_operation_complete_translate_succeed
...
+    default:
+      assert_true(false);
+  }
+}
+
+static void test_az_iot_provisioning_client_logging_succeed()
+{
+  az_log_classification const classifications[]
+      = { AZ_LOG_MQTT_RECEIVED_TOPIC, AZ_LOG_MQTT_RECEIVED_PAYLOAD, AZ_LOG_END_OF_LIST };
+  az_log_set_callback(_log_listener);
```
JeffreyRichter, line 500: order

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_sas.c
```diff
@@ -216,6 +267,54 @@ static void az_iot_provisioning_client_sas_get_signature_device_signature_overfl
...
+    default:
+      assert_true(false);
+  }
+}
+
+static void test_az_iot_provisioning_client_sas_logging_succeed()
+{
+  az_log_classification const classifications[] = { AZ_LOG_IOT_SAS_TOKEN, AZ_LOG_END_OF_LIST };
+  az_log_set_callback(_log_listener);
+  az_log_set_classifications(classifications);
```
JeffreyRichter, line 289: order

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_sas.c
```diff
@@ -216,6 +267,54 @@ static void az_iot_provisioning_client_sas_get_signature_device_signature_overfl
...
+  uint8_t signature_buffer[TEST_SPAN_BUFFER_SIZE];
+  az_span signature = az_span_init(signature_buffer, _az_COUNTOF(signature_buffer));
+  az_span out_signature;
+
+  assert_true(az_succeeded(az_iot_provisioning_client_sas_get_signature(
+      &client, test_sas_expiry_time_secs, signature, &out_signature)));
+
+  assert_int_equal(1, _log_invoked_sas);
+
+  az_log_set_classifications(NULL);
```
JeffreyRichter, line 309: order

sdk/iot/hub/src/az_iot_hub_client_c2d.c
```diff
@@ -58,18 +60,18 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
...
-  }
-  else
+  az_span reminder;
+  az_span token = _az_span_token(received_topic, c2d_topic_suffix, &reminder);
+  if (az_span_ptr(reminder) == NULL)
   {
     return AZ_ERROR_IOT_TOPIC_NO_MATCH;
   }
 
+  az_log_write(AZ_LOG_MQTT_RECEIVED_TOPIC, received_topic);
```
hihigupt, line 70: >az_log_write [](start = 2, length = 12)

is there a way to compile out logging, if user do not want them?

JeffreyRichter: The calls to the log function are always compiled. But, if the user doesn't register for a classification or register a callback, the log message didn't have to be generated. In fact,  code that creates messages should call https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/internal/az_log_internal.h#L15 first. This returns false if the user doesn't care about this classification and then the code can avoid producing the message and calling az_log_write.

hihigupt: I agree that runtime may be able to not generate the message (this will save cpu IO), but the string part of log message will always have to be part of binary as compiler has no clue about runtime decision. To reduce the size of binary and remove these unnecessary literal from binary, we do should support for compile time removal of logs.

JeffreyRichter: In this case, the received topic comes from the MQTT stack, right? There is no string in flash.

hihigupt: Agree in this case we are not using any literal and only thing we are wasting is stack space to call this function.

In general if we support compile time logging strip-off, we can avoid following things:
1. Remove unnecessary logging function from the user binary fully.
2. If we are using some literal in logging, we can avoid these literals to part of binary if logging is compile time opted out.
3. Any stack usage via the logging function can also be removed, if user remove logging at compile time.

What do you think?

sdk/core/core/inc/az_log.h
```diff
@@ -44,6 +44,17 @@ typedef enum
   AZ_LOG_HTTP_RETRY = _az_LOG_MAKE_CLASSIFICATION(
       _az_FACILITY_HTTP,
       3), ///< First HTTP request did not succeed and will be retried.
+
```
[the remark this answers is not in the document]
CIPop: @JeffreyRichter I think that makes sense for the IoT ones but would the MQTT ones be common?
Otherwise maybe the HTTP ones need to be moved as well.

sdk/iot/core/tests/cmocka/test_az_iot_core.c
```diff
@@ -137,6 +138,37 @@ static void test_az_iot_retry_calc_delay_overflow_time_success()
...
+      assert_true(false);
+  }
+}
+
+static void test_az_iot_provisioning_client_logging_succeed()
+{
+  az_log_classification const classifications[]
+      = { AZ_LOG_IOT_RETRY, AZ_LOG_END_OF_LIST };
+  az_log_set_callback(_log_listener);
+  az_log_set_classifications(classifications);
```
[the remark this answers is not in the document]
CIPop: Sounds to me like a design bug on the `az_log` APIs. I'll open a bug - maybe one solution is to bundle the two _set_ functions in one to avoid any possible race condition?

sdk/iot/provisioning/src/az_iot_provisioning_client.c
CIPop: I think it's interesting to keep topics and payload logging separate. The main purpose of the logs is to enable our CSS team to debug - with some clever filters I think we should be fine (I have some sample filters for the .NET SDK here as example: https://github.com/Azure/azure-iot-sdk-csharp/tree/master/tools/CaptureLogs/query).

sdk/iot/hub/src/az_iot_hub_client_c2d.c
CIPop: > but the string part of log message will always have to be part of binary

@JeffreyRichter I kind of agree with @hihigupt here in one aspect: the logger only accept `az_span` and that means we can only log numbers or other items if we convert to span. This means we would never be able to log those because all memory comes from the application (unless we want to alloc on the stack which we've decided against according to AzRTOS rules mandating short and deterministic stack lengths).

@hihigupt I was careful not to add more variables in logs. All added cases will be compiled out - there should be zero impact if the compiler optimizes it out. 
That said, we _need_ logs to do any kind of CSS investigations and so we should recommend our customers to keep these enabled and accessible if they need our help investigating any device-side issues.

sdk/core/core/inc/az_log.h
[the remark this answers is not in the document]
CIPop: For now, to solve this faster, I'll move all to az_iot_core.h. We can then promote them back to az_log.h later, as needed.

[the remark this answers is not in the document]
CIPop: Actually there are issues with this: if I move them out the `enum` name cannot be the same and that causes all call-sites to fail:
```
sdk/iot/core/src/az_iot_core.c:29:16: error: implicit conversion from enumeration type 'az_log_iot_classification' to different enumeration type 'az_log_classification'
      [-Werror,-Wenum-conversion]
  az_log_write(AZ_LOG_IOT_RETRY, AZ_SPAN_NULL);
```

If I give the same name the enum conflicts with the one in az_log.h. 

What's the recommendation?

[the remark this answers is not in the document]
CIPop: The only az_result value for IoT is defined in az_result.h today. 
The entire SDK has all values defined within az_result.h and az_log.h - I can't find any example of logs defined outside these two files.

Leaving this as not resolved - we need guidance on what the design pattern is. I will address in a future PR if needed.

/cc @gilbertw @ahsonkhan @danewalton

sdk/iot/core/tests/cmocka/test_az_iot_core.c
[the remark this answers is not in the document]
CIPop: I've opened https://github.com/Azure/azure-sdk-for-c/issues/675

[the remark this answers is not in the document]
CIPop: Also, for now, I've fixed the code on our side.

sdk/iot/hub/src/az_iot_hub_client_c2d.c
CIPop: I consider this resolved for the purpose of the first issue: the code should be compiled out as long as nobody sets a callback. We don't make any new allocations.

sdk/core/core/inc/az_log.h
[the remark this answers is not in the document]
JeffreyRichter: I understand that IOT has no result codes of their own today. When/if this ever changes, the pattern is for IOT to define their own enum like this:
typedef enum {
   AZ_ERROR_IOT_FOO = AZ_MAKE_ERROR(AZ_IOT_FACILITY, 1),
   AZ_ERROR_IOT_BAR = AZ_MAKE_ERROR(AZ_IOT_FACILITY, 2)
} az_iot_result;

Similarly, we'd like IOT to use the same pattern to define its own log classifications in your own .h file(s).

@CIPop : This pattern was discussed in an "az_result range allocation for Azure IoT SDKs" email dated January 13, 2020. @barcharcraz or @gilbertw should add these patterns to our official C guidelines.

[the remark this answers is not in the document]
JeffreyRichter: @barcharcraz : @CIPop has a good point about the enum type not being the same and cause compiler failures. Can you suggest a better pattern?

sdk/iot/hub/src/az_iot_hub_client_c2d.c
JeffreyRichter: I don't have a solution for how to log numbers as strings without converting the number to a string somewhere. It either has to go on the stack or in some global buffer. If we use a global buffer, then thread synchronization is required which we also don't want. If you have a better solution, I'm open to ideas. Or, don't log numbers at all.

hihigupt: I am more worried about not allowing user the flexibility to fully opt out of sdk logging and allowing them to reduce there binary size/stack requirements.

JeffreyRichter: I don't think stack size issue is a concern because we allow the customer the ability to restrict the log message size with this: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/inc/az_config.h#L27

We also allow client library developers the ability to not waste CPU on creating log messages by calling https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/internal/az_log_internal.h#L15.  

So, the only thing to possible consider is code size. The actual logging infrastructure code in tiny so I don't think it is a big deal at all. The code to create messages with possible literal strings is the only issue I think worth considering. I think we should ignore this for now and, IF customers really complain, we can surround the code with #ifdefs to address this. But, I'd rather wait until we knows it's a problem and we know that this will add significant benefit.

sdk/core/core/inc/az_log.h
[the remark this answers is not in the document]
CIPop: @JeffreyRichter we have one code already : https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/inc/az_result.h#L132

sdk/iot/hub/src/az_iot_hub_client_c2d.c
hihigupt: One thing to add to this is function call also incurs stack use, which is what I am more focusing on. 
Default value for LOG message (1024) is scary that will be allocated on stack.

Adding people from AZRTOS if they can any data points for this problem, if exists.
@TiejunMS
@yuxin-azrtos

TiejunMS: I agree with @hihigupt , allocating 1KB buffer on stack by default is terrible. Unless the SDK core is designed for desktop applications only, it will be very hard to debug the issue of stack overflow.

If the log function can not be fully opt off, I personally would be hesitate to use az_log. Instead of complain, I would write my own version of log function, since it is simple to write one. In order to let customers to use it, I would suggest to provide the feature to be able to opt off for the first release version.

JeffreyRichter: Great feedback. Let me ask you two questions that I'd like very specific answers to:
1. What should we change about the logging mechanism we have today to better address what you'd like to see. Please be specific.
2. What amount of stack space is OK to use and why? 10 bytes? 100 bytes? 300 bytes?

PR #671, opened 2020-05-02
Check address space wrap for spans

antkmsft:


sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -68,9 +68,15 @@ AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, boo
     return false;
   }
 
-  uint8_t* ptr = az_span_ptr(span);
+  uint8_t* const ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 
+  // Can't wrap over the end of the address space
+  if (ptr + span_size < ptr)
```
ahsonkhan, line 75: Can you please verify that this doesn't get optimized out and change the behavior? It's possible this may not achieve our intended goal.

Also, can you explain, maybe with a code snippet/test case what edge case this protects us against?

I read that some compilers in C consider `(X + 1 < X)` as a given by definition and hence optimize this out, leading to undefined behavior.
See http://blog.llvm.org/2011/05/what-every-c-programmer-should-know.html
https://blog.regehr.org/archives/213

antkmsft: Ok, now I rewrote it, it won't wrap an int.

antkmsft: It protects us from the case (assuming that you have 16-bit address space)
```
(az_span) {
    .ptr = (uint8_t*)0xFFF0,
    .size = 100,
}
```

sdk/core/core/src/az_span.c
```diff
@@ -646,10 +646,24 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
   int32_t result_size = 0;
   for (int32_t i = 0; i < source_size; ++i)
   {
-    result_size += _az_span_url_should_encode(az_span_ptr(source)[i]) ? 3 : 1;
+    if (_az_span_url_should_encode(az_span_ptr(source)[i]))
+    {
+      if (result_size <= INT32_MAX - 3)
```
ahsonkhan, line 651: I have a feeling we can re-write this with less overhead.

Consider also adding a precondition that source cannot be `> INT_MAX/3`. I don't think we need to support anything larger.

Maybe you can figure out ways to reduce the number of checks happening within a loop. If I have some suggestion, I will let you know.

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -68,9 +68,15 @@ AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, boo
     return false;
   }
 
-  uint8_t* ptr = az_span_ptr(span);
+  uint8_t* const ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 
+  // Can't wrap over the end of the address space
+  if (span_size > (uint8_t*)~0 - ptr)
```
ahsonkhan, line 75: Can you add a comment explaining what exactly you are doing here and why it works? For example, why are we negating 0, etc.

sdk/core/core/src/az_span.c
```diff
@@ -646,10 +646,24 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
...
+    {
+      result_size += 1;
+      continue;
+    }
+    
+    result_size = -1;
   }
 
-  if (az_span_size(destination) < result_size)
+  if (result_size == -1 || az_span_size(destination) < result_size)
```
ahsonkhan, line 666: Add unit tests that capture the overflow conditions so we have branch coverage and confidence this fixes the underlying issue.

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Maybe it is end of the week, but I struggle to find the words that won't end up repeating the code.
Any more concrete suggestions? :)

sdk/core/core/src/az_span.c
antkmsft: It turns out, we don't have a test for az_span_is_valid at all, I have to wand will write it.

sdk/core/core/src/az_span.c
antkmsft: Again, if you have more concrete suggestions, please send them to me. I had a couple of ideas, but many of them ending up relying on adding up two integers, which we especially want to avoid in this case, or this entire check can't be trusted.
I would try not to disallow INT32_MAX/3 if possible, that's about 715MB, large but still relatively reasonable number.
I also do not keep the goal to optimize "fail fast" scenario, as this might never happen.
Another thought, if you run this multithreaded, and change contents of the source buffer (not the size), then the size of the encoded result we calculated and checked won't be sufficient, we still may write outside of the destination span. To guard against that, we have to always require destination being 3x larger than the source.
I'm gonna think about that over the weekend.

antkmsft: This all reminds me... At a time, we used to have az_assumptions.c - a file that we compile and that produces no output, but it was checking all our assumptions at a compile time, so everywhere in the code we would know whether it is safe rely on some behavior, or not. And if that file won't compile, our SDK won't work correctly on that platform, and needs to be updated: https://github.com/Azure/azure-sdk-for-c/blob/669fe8ca88b3defb1d8ad4a9c8b69ced8b6f3267/sdk/core/core/src/az_assumptions.c

We could've write static assert of (INT32_MAX + 1 == INT32_MIN in that file, or maybe right above the function, so that if that breaks at compile time, you know that you need to update the block that checks destination span size, or remove it entirely at your own risk) - with that assert we could write much simper check of size being negative, and it would probably work on 100% of platforms our SDK ever gets compiled on.

ahsonkhan: > with that assert we could write much simper check of size being negative, and it would probably work on 100% of platforms our SDK ever gets compiled on.

That wouldn't completely work either. Since we are multiplying by 3, it could overflow back to a positive number and still result in data loss.

sdk/core/core/src/az_span.c
ahsonkhan: I meant specifically about the encode url method. For example, encoding a buffer that is larger than INT_MAX/3 in size.

sdk/core/core/src/az_span.c
antkmsft: Check out the new commit, I think it is less overhead, I like it.

sdk/core/core/src/az_span.c
antkmsft: I understand, but still, I don't want to give up. Check out the new version, it does not require all that :)

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Commented

sdk/core/core/src/az_span.c
antkmsft: I wrote yet another version, with size calculation extracted to another (inline) function, which has the loop code duplicated a bit, but it should fail fast / work bit faster in cases when there's we know upfront there's enough space.

antkmsft: If we remove the guarantee which we never explicitly gave, that we won't modify destination if result can't fit, we can make everything faster with just one pass when reading the source. Which makes a lot of sense to do IF/when this code needs to process tons of data. Plus, it will behave a bit better if the source buffer gets modified by another thread. The nature of algorithm is that to get that final size is to scan all the input once. It seems like a reasonable tradeoff, I don't see much code that would be unavoidably inconvenienced by this "relaxed" contract.

antkmsft: What's good is that for the cases when you provide destination buffer which is 3x in size, you pay almost no correctness check penalty. This technically can be achieved today, even with the upfront check size check, but in the case when destination is less than 3x of source, we scan the source twice.

antkmsft: Yes, I went all in with that idea, it is yet another version now. If needed, I can roll back a few commits, but I think it is reasonable to not scan twice on IOT devices for the small tradeoff which can be avoided.

ahsonkhan: > I would try not to disallow INT32_MAX/3 if possible, that's about 715MB, large but still relatively reasonable number.

As an aside, I really don't think 715MB is too small a limitation for a URL. In this context, I think we can totally limit us to a much smaller number, if we really needed to.

sdk/core/core/internal/az_span_internal.h
```diff
@@ -30,6 +30,8 @@ AZ_INLINE AZ_NODISCARD int32_t _az_span_diff(az_span sliced_span, az_span origin
  * URL-encoding the \p source span characters.
  *
  * @param[in] destination The #az_span whose bytes will receive the URL-encoded \p source.
+ * If \p destination can't fit the \p source, some data may still be written to it, but the \p
+ * out_length will be set to 0, and the function will return #AZ_ERROR_INSUFFICIENT_SPAN_SIZE.
```
ahsonkhan, line 34: Move this to a `@remark` section.

sdk/core/core/test/cmocka/test_az_span.c
```diff
@@ -759,6 +760,43 @@ static void az_span_copy_empty(void** state)
...
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), -1, true));
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), -1, false));
+
+  assert_true(_az_span_is_valid(az_span_init((uint8_t*)~0, 0), 0, false));
+  assert_false(_az_span_is_valid(az_span_init((uint8_t*)~0 - 1, 1), 0, false));
+  assert_false(_az_span_is_valid(az_span_init((uint8_t*)~0 - 1, 2), 0, false));
+  assert_true(_az_span_is_valid(az_span_init((uint8_t*)((~0) / 2), ((~0) / 2) + 1), 0, false));
+  assert_false(_az_span_is_valid(az_span_init((uint8_t*)((~0) / 2), ((~0) / 2) + 2), 0, false));
+
+  assert_false(_az_span_is_valid(az_span_init((uint8_t*)((~0) / 2), -1, 0, true));
```
ahsonkhan, line 797: I don't think the brackets match up correctly here. Where is the size parameter for `az_span_init`?

sdk/core/core/test/cmocka/test_az_span.c
```diff
@@ -759,6 +760,43 @@ static void az_span_copy_empty(void** state)
...
+  assert_true(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 1, true));
+  assert_true(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 1, false));
+  assert_true(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 5, true));
+  assert_true(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 5, false));
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 6, true));
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), 6, false));
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), -1, true));
+  assert_false(_az_span_is_valid(AZ_SPAN_FROM_STR("Hello"), -1, false));
+
+  assert_true(_az_span_is_valid(az_span_init((uint8_t*)~0, 0), 0, false));
```
ahsonkhan, line 791: Add the cases where we pass `true` for `null_is_valid` for this block similar to the ones we did above.

sdk/core/core/test/cmocka/test_az_span.c
```diff
@@ -759,6 +760,43 @@ static void az_span_copy_empty(void** state)
   assert_true(az_span_is_content_equal(az_span_copy(dst, AZ_SPAN_NULL), dst));
 }
 
+static void test_az_span_is_valid(void** state)
+{
+  (void)state;
+  assert_true(_az_span_is_valid({ 0 }, 0, true));
```
ahsonkhan, line 766: nit: Consider adding similar test cases for `AZ_SPAN_NULL`.

sdk/core/core/test/cmocka/test_az_span.c
```diff
@@ -759,6 +760,43 @@ static void az_span_copy_empty(void** state)
   assert_true(az_span_is_content_equal(az_span_copy(dst, AZ_SPAN_NULL), dst));
 }
 
+static void test_az_span_is_valid(void** state)
```
ahsonkhan, line 763: Great test!

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -51,16 +54,112 @@ static void test_url_encode(void** state)
...
+
+    // Insufficient size
+    url_length = 0xFF;
+    assert_true(
+        _az_span_url_encode(az_span_slice(0, 2), AZ_SPAN_FROM_STR("/"), &url_length)
+        == AZ_ERROR_INSUFFICIENT_SPAN_SIZE);
+
+    assert_int_equal(url_length, 0);
+    assert_true(az_span_is_content_equal(
+        az_span_slice(buffer5, 0, sizeof(buf5)), AZ_SPAN_FROM_STR("%2Flo")));
```
ahsonkhan, line 87: This test makes it seem like we wrote past the destination slice. If the destination is only 2 bytes, this should be "%2llo", no? If we are changing the destination to "%2Flo", that is a bug.

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -51,16 +54,112 @@ static void test_url_encode(void** state)
...
+    assert_true(az_succeeded(_az_span_url_encode(buffer7, AZ_SPAN_FROM_STR("he/lo"), &url_length)));
+
+    assert_true(az_span_is_content_equal(
+        az_span_slice(buffer7, 0, url_length), AZ_SPAN_FROM_STR("he%2Flo")));
+
+    assert_int_equal(url_length, sizeof("he%2Flo"));
+  }
+#ifdef AZ_NO_PRECONDITION_CHECKING
+  {
+    uint8_t buf5[5] = { '*', '*', '*', '*', '*' };
```
ahsonkhan, line 104: I would consider moving the precondition based test into a separate, isolated, test function rather than inlining it.

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -51,16 +54,112 @@ static void test_url_encode(void** state)
...
+    az_span const buffer5 = AZ_SPAN_FROM_BUFFER(buf5);
+
+    url_length = 0xFF;
+    assert_true(
+        _az_span_url_encode(buffer5, AZ_SPAN_FROM_STR("hello_world"), &url_length)
+        == AZ_ERROR_INSUFFICIENT_SPAN_SIZE);
+
+    assert_int_equal(url_length, 0);
+
+    url_length = 0xFF;
```
ahsonkhan, line 114: nit: Remove this line

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -51,16 +54,112 @@ static void test_url_encode(void** state)
...
+    assert_true(az_span_is_content_equal(
+        az_span_slice(buffer5, 0, sizeof(buf5)), AZ_SPAN_FROM_STR("*****")));
```
ahsonkhan, line 116: The slice isn't necessary here.
```suggestion
        buffer5, AZ_SPAN_FROM_STR("*****")));
```

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -61,16 +61,28 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
...
+  uint8_t* const ptr = az_span_ptr(span);
   int32_t const span_size = az_span_size(span);
 
+  // Can't wrap over the end of the address space.
+  // The biggest theoretical pointer value is "(void*)~0" (0xFFFF...), which is the end of address
+  // space. We don't attempt to read/write beyond the end of the address space - it is unlikely a
+  // desired behavior, and it is not defined. So, if the span size is greater than the addresses
+  // left until the theoretical end of the address space, it is not a valid span.
+  // Example: (az_span) { .ptr = (uint8_t*)(~0 - 5), .size = 10 } is not a valid span, because most
+  // likely you end up pointing to 0x0000 at .ptr[6], &.ptr[7] is 0x...0001, etc.
```
ahsonkhan, line 80: Awesome. Thanks :)

sdk/core/core/src/az_span.c
```diff
@@ -643,39 +643,77 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
   int32_t const source_size = az_span_size(source);
   _az_PRECONDITION_VALID_SPAN(destination, source_size, false);
 
-  int32_t result_size = 0;
-  for (int32_t i = 0; i < source_size; ++i)
+  if (source_size == 0)
```
ahsonkhan, line 646: As far as I can tell, this check isn't necessary and removing it wouldn't change the behavior when `source_size == 0`. It should just fall down to the end and return `AZ_OK` with `*out_length = 0`.

sdk/core/core/src/az_span.c
```diff
@@ -643,39 +643,77 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
...
   {
-    result_size += _az_span_url_should_encode(az_span_ptr(source)[i]) ? 3 : 1;
+    *out_length = 0;
+    return AZ_OK;
   }
 
-  if (az_span_size(destination) < result_size)
+  int32_t const destination_size = az_span_size(destination);
+  if (destination_size < source_size)
```
ahsonkhan, line 653: Similarly here, I think we can remove this check (especially given it is already checked as a precondition), since we won't overflow anything even if we let this pass through.

sdk/core/core/src/az_span.c
ahsonkhan: > If we remove the guarantee which we never explicitly gave, that we won't modify destination if result can't fit, we can make everything faster with just one pass when reading the source.

I think it is fine for us to remove that guarantee.

sdk/core/core/src/az_span.c
```diff
@@ -643,39 +643,77 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
...
+      uint8_t c = src_ptr[src_idx];
+      if (!_az_span_url_should_encode(c))
+      {
+        *dest_ptr = c;
+        dest_ptr += 1;
+      }
+      else
+      {
+        ++extra_space_used;
+        if (extra_space_used > extra_space_have)
```
ahsonkhan, line 703: This code duplication and extra-check-inside-loop is unfortunate.

Why don't we have something like:
1) result size = src.size
2) only if dest < src.size * 3, calculate result size and bail early if dest too small.
3) then, just loop through from i to result size without worrying about extra det length checks or logic duplication

sdk/core/core/src/az_span.c
antkmsft: Completed.

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Completed.

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Fixed, clarified.

sdk/core/core/src/az_span.c
antkmsft: No problem now.

sdk/core/core/internal/az_span_internal.h
antkmsft: Done.

sdk/core/core/test/cmocka/test_az_span.c
antkmsft: Fixed.

sdk/core/core/test/cmocka/test_az_span.c
antkmsft: Done.

sdk/core/core/test/cmocka/test_az_span.c
antkmsft: Done, in terms of comparing the contents with memcmp.

sdk/core/core/test/cmocka/test_az_span.c
antkmsft: Thank you :)

sdk/core/core/test/cmocka/test_az_url_encode.c
antkmsft: It should've been be either "%2Fllo", or "hello" (unmodified). We either fit a single 3-character sequence or not. I updated unit tests to include more cases like this, it should be clearer now.

sdk/core/core/test/cmocka/test_az_url_encode.c
antkmsft: Done.

sdk/core/core/test/cmocka/test_az_url_encode.c
antkmsft: Done.

sdk/core/core/test/cmocka/test_az_url_encode.c
antkmsft: Done.

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Thank you :)

sdk/core/core/src/az_span.c
antkmsft: I'd prefer to keep it this way: I know that the code below would work correctly for source_size == 0, but it is a convenience to write it below knowing that I don't have to proofread every line for source being 0. And also if there's nothing to do, we exit rather fast, not doing the extra job, eating battery and so on. The whole check should be pretty cheap, since checking for zero are sometimes even cheaper than checking with any other value, something along the lines of "test EAX, EAX; jz $exit0".

sdk/core/core/src/az_span.c
antkmsft: I'd keep it with the same reasoning as above.

ahsonkhan:
FWIW @antkmsft - you could just "resolve" feedback that you have addressed/close conversations that are no longer actionable, instead of adding a "done" comment, since comments add notifications to the reviewers/watchers. Up to you :)

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -44,29 +47,168 @@ static uint8_t url_decoded_buf[] = {
...
+  {
+    // May be sufficient to perform encoding
+    uint8_t buf7[7] = { '*', '*', '*', '*', '*', '*', '*' };
+    az_span const buffer7 = AZ_SPAN_FROM_BUFFER(buf7);
+    az_span const buffer5 = az_span_slice(buffer7, 0, 5);
+    az_span const buffer6 = az_span_slice(buffer7, 0, 6);
+    {
+      url_length = 0xFF;
+      assert_true(
+          _az_span_url_encode(buffer5, AZ_SPAN_FROM_STR("he/lo"), &url_length)
```
ahsonkhan, line 106: nit: store these in named locals and reuse (place around line 99-102 in the containing scope)

```C
az_span input = AZ_SPAN_FROM_STR("he/lo");
az_span expected = AZ_SPAN_FROM_STR("*****");
```

sdk/core/core/test/cmocka/test_az_span.c
```diff
@@ -760,6 +761,54 @@ static void az_span_copy_empty(void** state)
...
+  assert_true(_az_span_is_valid(az_span_init(max_ptr, 0), 0, true));
+
+  assert_false(_az_span_is_valid(az_span_init(max_ptr, 1), 0, false));
+  assert_false(_az_span_is_valid(az_span_init(max_ptr, 1), 0, true));
+
+  assert_true(_az_span_is_valid(az_span_init(max_ptr - 1 , 1), 0, false));
+  assert_true(_az_span_is_valid(az_span_init(max_ptr - 1, 1), 0, true));
+
+  assert_false(_az_span_is_valid(az_span_init(max_ptr - 1, 2), 0, false));
+  assert_false(_az_span_is_valid(az_span_init(max_ptr - 1, 2), 0, true));
```
ahsonkhan, line 809: How about also for `max_size` (i.e. INT_MAX).
```C
assert_false(_az_span_is_valid(az_span_init(max_ptr - 1, max_size), 0, false));
assert_false(_az_span_is_valid(az_span_init(max_ptr - 1, max_size), 0, true));

assert_true(_az_span_is_valid(az_span_init(max_ptr - max_size, max_size), 0, false));
assert_true(_az_span_is_valid(az_span_init(max_ptr - max_size, max_size), 0, true));
```

sdk/core/core/test/cmocka/test_az_url_encode.c
antkmsft: I reworked these buffers and I think I added more clarity. Instead of reusing, I declare everything inside a very small scope, that is sufficient just for one scenario, I think it is better now.

sdk/core/core/test/cmocka/test_az_span.c
antkmsft: Done.

sdk/core/core/src/az_span.c
antkmsft: I wrote the loops with do-while, so that we don't do the very first index comparison now.
Regarding the rest that you described, we currently do have 2 branches, for `dest >= src * 3` and `dest >= src && dest < src * 3`. In the second case, we encode while there is space available, and check while there's still space available, and only if there isn't enough space to encode, we fail. Basically, by avoidng and checking, we avoid looping through the source twice, plus we avoid overrunning the destination in case the source gets changed by another thread after we calculated the size neccessary (although corrupting the memory is not _too_ bad option in this case either).

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -79,20 +79,50 @@ AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, boo
...
 #define _az_PRECONDITION_VALID_SPAN(span, min_size, null_is_valid) \
-  _az_PRECONDITION(az_span_is_valid(span, min_size, null_is_valid))
+  _az_PRECONDITION(_az_span_is_valid(span, min_size, null_is_valid))
+
+AZ_NODISCARD AZ_INLINE bool _az_span_overlap(az_span a, az_span b)
+{
+  uint8_t* const a_ptr = az_span_ptr(a);
+  uint8_t* const b_ptr = az_span_ptr(b);
+
+  return a_ptr <= b_ptr ? (a_ptr + az_span_size(a) > b_ptr) : (b_ptr + az_span_size(b) > a_ptr);
```
ahsonkhan, line 122: Does an empty span overlap with any span (empty or non empty)? I say no.

Add checks:
```C
if (az_span_size(a) == 0 || az_span_size(b) == 0)
  return false;
```

Also, does it make sense to consider `a_ptr + az_span_size(a) == b_ptr` to be overlapping?

ahsonkhan: Also, is there a way to support overlapping spans when doing url encode?

One concern with this check is that now the following scenario isn't allowed, even though it is "safe":
```C
buffer[100];
buffer[99] = 'A';
dest = buffer;
src = buffer.Slice(99, 100);

// fails because the spans overlap, but the dest has plenty of space to hold the src.
url_encode(dest, src);
```

Maybe that's OK, just asking to see how feasible it would be.

sdk/core/core/src/az_span.c
```diff
@@ -714,40 +714,86 @@ AZ_NODISCARD az_result _az_span_url_encode(az_span destination, az_span source,
...
+      else
+      {
+        dest_ptr[0] = '%';
+        dest_ptr[1] = _az_number_to_upper_hex(c >> 4);
+        dest_ptr[2] = _az_number_to_upper_hex(c & 0x0F);
+        dest_ptr += 3;
+      }
+
+      ++src_idx;
+    } while (src_idx < source_size);
```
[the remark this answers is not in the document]
antkmsft: Because we don't do the first comparison to figure out whether we need to go into the loop. Here, now that we checked source for being empty, we know that we need to go through at least one character.

For two iterations: `for (A; B; C) { D; }` does `A, B, D, C, B, D, C, B`.
`A; do { D; C; } while(B)` does `A, D, C, B, D, C, B`.

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -15,58 +16,417 @@
...
+      assert_int_equal(url_length, 0);
+    }
+    { // Overlapping buffers, same pointer.
+      uint8_t buf[13] = { 'a', 'B', 'c', '*', '*', '*', '*', '*', '*', '*', '*', '*', '*' };
+      az_span const in_buffer = az_span_slice(AZ_SPAN_FROM_BUFFER(buf), 0, 3);
+      az_span const out_buffer = az_span_slice(AZ_SPAN_FROM_BUFFER(buf), 0, 3);
+
+      int32_t url_length = 0xFF;
+      assert_true(az_succeeded(_az_span_url_encode(out_buffer, in_buffer, &url_length)));
+      assert_int_equal(url_length, sizeof("aBc") - 1);
```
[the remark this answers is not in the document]
antkmsft: Because it also includes +1 for the 0-terminating character of the string literal.

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -15,58 +16,417 @@
...
+      uint8_t buf[13] = { 'a', 'B', 'c', '/', '/', '/', '*', '*', '*', '*', '*', '*', '*' };
+
+      az_span const in_buffer = az_span_slice(AZ_SPAN_FROM_BUFFER(buf), 0, 6);
+      az_span const out_buffer = az_span_slice(AZ_SPAN_FROM_BUFFER(buf), 1, 13);
+
+      int32_t url_length = 0xFF;
+      assert_true(az_succeeded(_az_span_url_encode(out_buffer, in_buffer, &url_length)));
+      assert_int_equal(url_length, sizeof("aaaaaa") - 1);
+      assert_true(
+          az_span_is_content_equal(AZ_SPAN_FROM_BUFFER(buf), AZ_SPAN_FROM_STR("aaaaaaa******")));
```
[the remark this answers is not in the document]
antkmsft: Yes, that is what it is will end up doing if you pass overlapping spans. The main goal here is rather to verify that the function does not go nuts completely. Result is likely a garbage, yes (unless you pass source = destination for the span that does not need encoding at all)

sdk/core/core/test/cmocka/test_az_url_encode.c
```diff
@@ -15,58 +16,417 @@
...
+
+  az_span const buffer = AZ_SPAN_FROM_BUFFER(buf);
+
+  int32_t url_length = 0xFF;
   assert_true(
-      az_succeeded(_az_span_url_encode(buffer, AZ_SPAN_FROM_BUFFER(url_decoded_buf), &url_length)));
+      az_succeeded(_az_span_url_encode(buffer, AZ_SPAN_FROM_BUFFER(values256), &url_length)));
 
-  assert_true(az_span_is_content_equal(az_span_slice(buffer, 0, url_length), url_encoded));
+  int const nunreserved = sizeof("-_.~"
```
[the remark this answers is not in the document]
antkmsft: nunreserved = "number of unreserved" :)

sdk/core/core/internal/az_precondition_internal.h
antkmsft: There might be some cases when source and destination may overlap, and everything is going to be just fine. It is harder and IMO not with the cost to check for all these cases.

When it is safe to encode in the same buffer, it should be possible to write a code that clearly slices source and destination into two non-overlapping buffers, and calls the function. This way, you are always guaranteed to not overrun the buffer and get hardly predictable results, when the nature of the input changes.

Other case would be to pass two spans pointing to the same buffer with the contents of "abc". Today it is fine, tomorrow you change "abc" to "a/c", and may end up with unexpected behavior. Preconditions will help you to catch it earlier and either reword your intent more explicitly, or avoid the unexpected behavior in all cases.

`a_ptr + az_span_size(a) == b_ptr` are not overlapping because buffer `b` starts just as the buffer `a` ends, they are right next to each other, but don't overlap.

As of the empty buffer pointing inside another buffer being considered overlapping, that was a little dilemma, but I decided to keep it this way for the same reason we don't allow NULL destination spans even when the source is empty. Should you change the size of the inner span from 0 to 1, they will start to overlap. It is a slight difference, and we can remove that restriction once we have a data to back up that we really don't want the buffer to be considered overlapping in this case.
Interestingly, two empty buffers with the same ptr, or an empty span just at the end of the non-empty span are not considered overlapping, but being next to each other.

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -79,20 +79,46 @@ AZ_NODISCARD AZ_INLINE bool az_span_is_valid(az_span span, int32_t min_size, boo
         - If null_is_valid is true and the pointer in the span is null, the size must also be 0.
         - In the case of the pointer not being NULL, the size is greater than or equal to zero.
   */
+
+  // On some platforms, in some cpmpilation configurations (Debug), NULL is not 0x0...0. But if you
```
ahsonkhan, line 83:
```suggestion
  // On some platforms, in some compilation configurations (Debug), NULL is not 0x0...0. But if you
```

sdk/core/core/internal/az_span_internal.h
```diff
@@ -37,6 +37,10 @@ AZ_INLINE AZ_NODISCARD int32_t _az_span_diff(az_span sliced_span, az_span origin
  *         - #AZ_OK if successful
  *         - #AZ_ERROR_INSUFFICIENT_SPAN_SIZE if the \p destination is not big enough to contain
  * the encoded bytes
+ *
+ * @remark If \p destination can't fit the \p source, some data may still be written to it, but the
+ * \p out_length will be set to 0, and the function will return #AZ_ERROR_INSUFFICIENT_SPAN_SIZE. \p
+ * destination and \p source must not overlap.
```
ahsonkhan, line 43: nit:
```suggestion
 * \p out_length will be set to 0, and the function will return #AZ_ERROR_INSUFFICIENT_SPAN_SIZE. The \p
 * destination and \p source must not overlap.
```

sdk/core/core/internal/az_span_internal.h
```diff
@@ -37,6 +37,10 @@ AZ_INLINE AZ_NODISCARD int32_t _az_span_diff(az_span sliced_span, az_span origin
  *         - #AZ_OK if successful
  *         - #AZ_ERROR_INSUFFICIENT_SPAN_SIZE if the \p destination is not big enough to contain
  * the encoded bytes
+ *
+ * @remark If \p destination can't fit the \p source, some data may still be written to it, but the
+ * \p out_length will be set to 0, and the function will return #AZ_ERROR_INSUFFICIENT_SPAN_SIZE.
+ * The \p destination and \p source must not overlap.
```
ahsonkhan, line 43: Btw, how does this render in docs? Do we need to split this into its own `@remarks` line?

antkmsft: "Multiple adjacent \remark commands will be joined into a single paragraph. Each remark will start on a new line." http://www.doxygen.nl/manual/commands.html#cmdremark

PR #673, opened 2020-05-02
AZ_FALLTHROUGH => _az_

antkmsft:
We don't explicitly ship these as part of the SDK, correct? We don't commit to support them between versions, we don't want customers to use these macros, and they don't give much value to customer anyway, but they would pop up in intellisense.
I propose to name them as internal - with `_az_` prefix.

Update: now AZ_FALLTHROUGH is the only macro we make an `_az_`

### DIFF
--- a/sdk/core/core/src/az_json_parser.c
+++ b/sdk/core/core/src/az_json_parser.c
@@ -30,7 +30,7 @@ typedef enum
  * from tokens
  *
  */
-AZ_NODISCARD AZ_INLINE bool az_json_is_white_space(uint8_t c)
+_az_NODISCARD _az_INLINE bool az_json_is_white_space(uint8_t c)
 {
   switch (c)
   {
@@ -48,7 +48,7 @@ AZ_NODISCARD AZ_INLINE bool az_json_is_white_space(uint8_t c)
  * numbers like 10e10
  *
  */
-AZ_NODISCARD AZ_INLINE bool az_json_is_e(uint8_t c)
+_az_NODISCARD _az_INLINE bool az_json_is_e(uint8_t c)
 {
   switch (c)
   {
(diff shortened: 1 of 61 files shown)


ahsonkhan:
I am not sure about this change. These are part of our public surface area (as part of API signatures), are visible to callers, and do affect API behavior and how customers write calling code. 

For instance, `AZ_NODISCARD` implies the user must check the result and can't ignore it. Given that, it is a bit odd to see `_az_` anywhere in public API. I would be curious to hear what others think though.

sdk/core/core/src/az_json_parser.c
```diff
@@ -283,7 +283,7 @@ AZ_NODISCARD static az_result az_span_reader_get_json_number_digit_rest(
     {
       case '-':
         e_sign = -1;
-        AZ_FALLTHROUGH;
+        _az_FALLTHROUGH;
```
ahsonkhan, line 286: Looks like we have only one case of `fallthrough`, maybe we should just re-write this code and remove it.

JeffreyRichter: Yes, I think we should leave AZ_NODISCARD and AZ_INLINE as they were and get rid of AZ_FALLTHROUGH if it's only used in 1 place. 

FYI: If we kept AZ_FALLTHROUGH, then it should probably be _az_ prefixed since it's not in the public API surface; it's an implementation convenience for our code.

ahsonkhan: > FYI: If we kept AZ_FALLTHROUGH, then it should probably be _az_ prefixed since it's not in the public API surface; it's an implementation convenience for our code.

To clarify, since GitHub treats the underscores to mean italics.

> FYI: If we kept AZ_FALLTHROUGH, then it should probably be `_az_` prefixed since it's not in the public API surface; it's an implementation convenience for our code.

antkmsft: I would keep `_az_FALLTHROUGH` usage - it is the most efficient way to write that code where it is used. Although it is tempting from an aesthetical perspective to remove that usage, I do not think it is the right priority. I reverted `INLINE` and `FALLTHROUGH`.

ahsonkhan: I don't think it is worth it to expose it from the header just for a single use within the SDK, especially one that might end up getting re-written anyway.

But given it is internal now, it is fine :)

antkmsft: Agree; But I'll add that it is possible that we may use fallthrough in more places in the future. If that's the most efficient way to write code in some circumstances, and the language allows it, we should use it. Fallthrough macro is the price we have to pay to dodge warnings and say explicitly that we really mean to write the code that we wrote, but other than that, we should not be avoiding fallthrough statements, in my opinion. Even more, with the _`az_FALLTHROUGH` macro we are rather getting the best of both worlds - warnings AND the ability to write fallthrough cases. Aregarding the location, maybe it is not the best, but probably should be fine - since it is internal, and at the same time as a language extension it is readily available everywhere in our code, where `_az_cfg.h`. is included - i.e. practically everywhere. Maybe when we get more macros like these, it would be worth to make something like `_az_cfg_internal.h`, but even then I am not sure if it is worth doing, given that the `_az_cfh.h` header itself is pretty much an internal utility, as the name suggests, plus fallthrough now follows an internal macro naming rules.

PR #689, opened 2020-05-06
Do not log even the smallest part of auth token header value; volatile; _az_.

antkmsft:
+ Apply Jeffrey's suggestions and asks.

### DIFF
--- a/sdk/core/core/src/az_http_policy_logging.c
+++ b/sdk/core/core/src/az_http_policy_logging.c
@@ -205,7 +205,7 @@ void _az_http_policy_logging_log_http_request(_az_http_request const* request)
 
   (void)_az_http_policy_logging_append_http_request_msg(request, &log_msg);
 
-  az_log_write(AZ_LOG_HTTP_REQUEST, log_msg);
+  _az_log_write(AZ_LOG_HTTP_REQUEST, log_msg);
 }
 
 void _az_http_policy_logging_log_http_response(
@@ -221,7 +221,7 @@ void _az_http_policy_logging_log_http_response(
   (void)_az_http_policy_logging_append_http_response_msg(
       &response_copy, duration_msec, request, &log_msg);
 
-  az_log_write(AZ_LOG_HTTP_RESPONSE, log_msg);
+  _az_log_write(AZ_LOG_HTTP_RESPONSE, log_msg);
 }
 
 AZ_NODISCARD az_result az_http_pipeline_policy_logging(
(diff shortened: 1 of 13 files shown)


sdk/core/core/src/az_log.c
```diff
@@ -14,8 +14,8 @@
 
 #include <_az_cfg.h>
 
-static az_log_classification const* _az_log_classifications = NULL;
-static az_log_message_fn _az_log_message_callback = NULL;
+static az_log_classification const* volatile _az_log_classifications = NULL;
+static az_log_message_fn volatile _az_log_message_callback = NULL;
```
ahsonkhan, line 18: Do we have a test that highlights why these need to be volatile?

sdk/core/core/src/az_http_policy_logging.c
```diff
@@ -58,6 +58,8 @@ static az_result _az_http_policy_logging_append_http_request_msg(
     _az_http_request const* request,
     az_span* ref_log_msg)
 {
+  extern az_span const _az_auth_header_name;
```
ahsonkhan, line 61: This link between `az_http_policy_logging.c` and `az_credential_client_secret.c` is unfortunate. Is there any way to decouple them?

Also, do we know if "authorization" is the only header we need to avoid logging?

sdk/core/core/src/az_log.c
JeffreyRichter: No, you can't create a test for timing

ahsonkhan: OK. Can you please explain why these need to be volatile?

JeffreyRichter: One thread sets the value while another thread reads the value. The use of volatile ensures that this works correctly in that the compiler won't cache the value in a register or temporary variable.

JeffreyRichter: Your question made me think about the code and I modified it and sent revised code to @antkmsft to test and check in.

sdk/core/core/src/az_http_policy_logging.c
JeffreyRichter: Yes, repeat this line in this source code file:
az_span const _az_auth_header_name = AZ_SPAN_LITERAL_FROM_STR("authorization");
Or, just hard-code it in line 113.

antkmsft: What is not to like? The values are kept in sync. Compilation error if _az_auth_header_name gets removed.

PR #690, opened 2020-05-06
Standardize on Embedded C name

gilbertw:
Fixes readme files to use a consistent name throughout.  

For this review please only focus on the "Embedded C" name change.  Yes I know there are other issues in the readme's.  You can call out other issues if you see them, but do not block the review on changes unrelated to the name change.  

Issues called out that are unrelated to the name change will be addressed in subsequent PRs.

### DIFF
--- a/sdk/core/core/samples/README.md
+++ b/sdk/core/core/samples/README.md
@@ -7,7 +7,7 @@ products:
 urlFragment: core-samples
 ---
 
-# Azure Core Samples client library for C
+# Azure Core Samples client Library for Embedded C
 This document explains samples and how to use them.
 
 ## Key concepts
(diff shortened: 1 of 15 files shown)


sdk/core/core/samples/README.md
```diff
@@ -7,7 +7,7 @@ products:
 urlFragment: core-samples
 ---
 
-# Azure Core Samples client library for C
+# Azure Core Samples client Library for Embedded C
```
ahsonkhan, line 10: Why capitalize  the L in library, and if we are doing it also capitalize the C in client?

gilbertw: The name doesn't make a ton of sense.  Samples are never packaged as libraries.  I'll clean up these readme's in one of my subsequent readme cleanup PRs

PR #692, opened 2020-05-06
Fix comment format to what doxygen expects by adding an asterisk

ahsonkhan:
Multi-line comments need to start with `/**` instead of `/*` for the public API doc comments to show up in the doxygen generated docs.


ahsonkhan:
How did  this get checked-in to master and past the previous CI runs?

I am guessing that because we weren't using correctly formatted docs, the comment/parameter name validation wasn't running.

I am glad we caught this :)

```text
/Users/runner/runners/2.166.4/work/1/s/sdk/core/core/inc/az_json.h:341:11: error: parameter 'out_token' not found in the function declaration [-Werror,-Wdocumentation]
 * @param out_token A pointer to an az_json_token containing the next parsed JSON token.
          ^~~~~~~~~
/Users/runner/runners/2.166.4/work/1/s/sdk/core/core/inc/az_json.h:341:11: note: did you mean 'token'?
 * @param out_token A pointer to an az_json_token containing the next parsed JSON token.
          ^~~~~~~~~
          token
```

PR #697, opened 2020-05-07
Log all classifications via single-item array

antkmsft:
Jeffrey, doing this the way you proposed, with the one exception that we can't create array inside the scope of `if()` assign it's pointer, and use that pointer in an outer scope.
If you think it is more clear, I can apply this. I don't think it is more readable, and it is not the most direct way to emulate with the single-item array, although I agree there is some elegance in that solution. Honestly I can understand it all, no matter how it is written.
I'd either write a separate logic for the if statement, that conditionally invokes the callback and returns true (https://github.com/Azure/azure-sdk-for-c/pull/698), or I'd use the JMP method (https://github.com/Azure/azure-sdk-for-c/pull/696), which generates less code. Up to you, you are the author, I'll do as you suggest. I personally prefer JMP :)

### DIFF
--- a/sdk/core/core/src/az_log.c
+++ b/sdk/core/core/src/az_log.c
@@ -40,7 +40,7 @@ static bool _az_log_write_engine(bool log_it, az_log_classification classificati
 {
   // Copy the volatile fields to local variables so that they don't change within this function
   az_log_message_fn const callback = _az_log_message_callback;
-  az_log_classification const* const classifications = _az_log_classifications;
+  az_log_classification const* classifications = _az_log_classifications;
 
   if (callback == NULL)
   {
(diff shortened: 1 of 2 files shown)


sdk/core/core/src/az_log.c
```diff
@@ -40,35 +40,33 @@ static bool _az_log_write_engine(bool log_it, az_log_classification classificati
...
   if (callback == NULL)
   {
     // If no one is listening, don't attempt to log.
     return false;
   }
 
-  // If the user hasn't registered any classifications, then we log everything.
-  bool const log_everything = (classifications == NULL);
-  switch (log_everything)
+  az_log_classification current_classification[2] = { classification, AZ_LOG_END_OF_LIST };
```
JeffreyRichter, line 51: I think this is the best (most sustainable) pattern that removes the repeating of code.
Although, as i look at it, it can be simplified more:
if (classifications == null) {
   // If the user hasn't registered any classifications, then we log everything (this classification).
   classifications = &classification;
}

antkmsft: OK, I'll use this one.
However, the new code `classifications = &classification;`, where would AZ_LOG_END_OF_LIST come from in this case?

JeffreyRichter: We don't need AZ_LOG_END_OF_LIST because there will always be a match on the first entry and then the function returns and so the loop never loops around.

antkmsft: See, this time it was me who needed an explanation to understand how the code would work :)
I like this method (second after the switch/case jump), and I'll update the original PR to use it, and will add the comment.

JeffreyRichter: Thanks. I agree it's not obvious which is why the comments are useful. It's just very odd to jump into a loop. In fact, many programming languages simply don't allow it and won't compile the code.

antkmsft: Ha, other languages. C is better, it ALLOWS you to do what others can't :) I updated the original PR with `&classification` and a comment.

PR #700, opened 2020-05-07
Change to az_failed and az_succeeded in samples

danewalton:


danewalton:
> 
> 
> Thanks @danewalton . Could you apply this to Provisioning Sample as well please? (Or create a card/issue to track if you don't have time.)

Updated provisioning. Take another look if you would like 👍

PR #703, opened 2020-05-07
Sample readme updates

danewalton:
Add further clarifications to the IoT Sample readme.

### DIFF
--- a/sdk/iot/hub/samples/README.md
+++ b/sdk/iot/hub/samples/README.md
@@ -39,7 +39,8 @@ named `fingerprint.txt` for your convenience).
 1. Create a device in IoT Hub with `X.509 Self-Signed` authentication.
 1. Paste in the certificate fingerprint printed previously.
 
-After this is generated, make sure to set `AZ_IOT_DEVICE_X509_CERT_PEM_FILE` as an environment variable.
+After this is generated, make sure to set `AZ_IOT_DEVICE_X509_CERT_PEM_FILE` as an environment variable. It should be the
+full path to the generated `device_cert_store.pem` file.
 
 For more details on X509 authentication, refer to [this online documentation](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-x509ca-overview#how-to-register-the-x509-ca-certificate-to-iot-hub)
 
--- a/sdk/iot/hub/samples/README.md
+++ b/sdk/iot/hub/samples/README.md
@@ -17,9 +17,12 @@ This document explains samples for the Azure SDK for Embedded C IoT Hub Client a
 
 Key concepts are explained in detail [here][SDK_README_KEY_CONCEPTS].
 
-## Samples for Azure IoT Hub Client APIs
+## Prerequisites
 
-This document describes how to use samples and what is done in each sample.
+- To generate the device certificate, the provided script uses [OpenSSL](https://www.openssl.org/). Please
+install OpenSSL prior to using the script.
+- To use the samples, we use [Eclipse Paho MQTT C client][Eclipse_Paho]. You can use the directions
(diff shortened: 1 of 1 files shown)


sdk/iot/hub/samples/README.md
```diff
@@ -17,9 +17,12 @@ This document explains samples for the Azure SDK for Embedded C IoT Hub Client a
 
 Key concepts are explained in detail [here][SDK_README_KEY_CONCEPTS].
 
-## Samples for Azure IoT Hub Client APIs
+## Prerequisites
 
-This document describes how to use samples and what is done in each sample.
+- To generate the device certificate, the provided script uses [OpenSSL](https://www.openssl.org/). Please
```
ericwol-msft, line 22: I think this also comes from VCPKG @CIPop

danewalton: [I think that might be curl](https://github.com/Azure/azure-sdk-for-c/blob/master/CONTRIBUTING.md#vcpkg)

CIPop: Yes: all prereqs come from vcpkg.

sdk/iot/hub/samples/README.md
```diff
@@ -46,16 +49,27 @@ For more details on X509 authentication, refer to [this online documentation](ht
 
 ## Samples
 
-The following section documents various samples. All of them use the [Eclipse Paho MQTT C client][Eclipse_Paho].
-For all of these samples, the following environment variables will need to be set:
+For all of these samples, the following environment variables will need to be set.
 
+**NOTE**: These need to be set before running the cmake commands.
+
+- `VCPKG_DEFAULT_TRIPLET`: The triplet created using the instructions [here][VCPKG_DIRECTIONS].
```
CIPop, line 56: Duplicating these instructions will likely lead to issues in the future when contributing.md changes (I don't think we'll maintain both places).
We should just point to VCPKG_Directions saying those need to be done initially.

sdk/iot/hub/samples/README.md
```diff
@@ -46,16 +49,27 @@ For more details on X509 authentication, refer to [this online documentation](ht
...
 - `AZ_IOT_HUB_HOSTNAME`: The hostname of your IoT Hub.
 - `AZ_IOT_DEVICE_X509_CERT_PEM_FILE`: The full path to your device cert (in `.pem` format) concatenated
  with its private key.
 - `AZ_IOT_DEVICE_X509_TRUST_PEM_FILE`: The full path to the trusted server cert (in `.pem` format). This is usually
 not needed on Mac or Linux but might be needed for Windows.
 
+Once these are set, you MUST compile the SDK with the `BUILD_PAHO_TRANSPORT` option turned on. For example, on the command
+line, it might look like the following:
+```bash
+cmake -DBUILD_PAHO_TRANSPORT=ON ..
```
CIPop, line 68: Not sure if this is the right place to add cmake descriptions: they are likely going to run out of sync.

sdk/iot/hub/samples/README.md
danewalton: I'm not duplicating the instructions here, only stating that the env var needs to be set. I understand your point but I want to be explicit everything that needs to be set in order for these to work. If I don't, someone will try to compile, paho linking will fail, and they'll have to hunt around everywhere for how to resolve that. I tried to be as minimal as possible about the details and reroute people elsewhere

sdk/iot/hub/samples/README.md
danewalton: I understand it could get out of sync but in this case, if the cmake option gets changed, it should be done with a find all and replace all and this section would come up. I think it complete the flow here of getting off the ground.

PR #705, opened 2020-05-08
Remove _az_cfg.h from samples

### LINKED ISSUE #564: Remove use of _azcfg.h from tests and samples
This is tracking fixing the MSVC compiler warnings related to span initialization with auto-variables.
Using `AZ_SPAN_FROM_BUFFER` causes MSVC to issue a warning (treated as error) in our build system.

~~The work should be completed after `az_span` is changed to take only buffer and size (no capacity).~~

danewalton:
Closes #564


danewalton:
While it worked on my machine it looks like the issue still happens on Windows. Didn't check my bad.

PR #707, opened 2020-05-08
Logging: option to remove all logging code

antkmsft:


antkmsft:
Forgot the logging policy! One moment.

antkmsft:
Next steps: verify that CI/CD succeeds now, then disable logging and verify again.

antkmsft:
> Please wait till Mon/Tue to commit this. i.e. after the release is cut

Thank you, this is the plan!

antkmsft:
@JeffreyRichter , I made a few minor changes:
* I renamed `AZ_NO_LOGGING` to `_az_NO_LOGGING`.
* I updated unit tests, and made a macro along with a header (`az_test_log.h`) that helps in testing both in cases when built with and without `_az_NO_LOGGING`. IoT had about 5 verifications, we had one in Core.
* I updated include guards in two other `az_test_*.h` files.
* I updated the CI/CD to include a single configuration to build without logging, `Linux_x64_with_unit_tests_and_samples_no_logging_no_preconditions`.
* I verified that CI/CD is going to fail if client code forgets to #ifdef az_policy_logging. Here is how the failure looks like: https://dev.azure.com/azure-sdk/public/_build/results?buildId=386497&view=logs&j=7005464e-0e2d-5609-c3eb-49069ce2a341&t=ae28e88a-aad8-50ed-b970-f7d4c51e5bf5

JeffreyRichter:
This all looks good but the AZ_NO_LOGGING should be public (like AZ_NO_PRECONDITION_CHECKING). Customers have to be able to specify this symbol when building to turn off logging.

antkmsft:
> [...] AZ_NO_LOGGING should be public (like AZ_NO_PRECONDITION_CHECKING). Customers have to be able to specify this symbol when building to turn off logging.

Ok, I changed it back to `AZ_`. But for the users, the straightforward way it is controlled is via `BUILD_LOGGING` CMake option, they don't _have_ _to_ use `AZ_NO_LOGGING` directly (we do the same with Cmake option `UNIT_TESTING_MOCK_ENABLED` which defines `_az_MOCK_ENABLED` macro).

By the way, we probably need a few sentences said about this in docs as well.

antkmsft:
@JeffreyRichter , I updated readme:
```
If the SDK is built with `AZ_NO_LOGGING` macro defined (or adding option -DBUILD_LOGGING=OFF with cmake), it should reduce the binary size and slightly improve performance.
The logging should have negligible performance impact if no listener is registered, or if classifications include only the small portion of messages, but when the SDK is built with `AZ_NO_LOGGING`, the parts of logging messages text won't be included in the output binary, and it should not be possible to make logging function even in smallest capacity - recompilation without `AZ_NO_LOGGING` would be required.
When the code is built with `AZ_NO_LOGGING`, all the logging-related code could still be present and get compiled, but it is expected to get completely removed during compilation.
```

Now the code technically is ready to go, from my perspective. I will not commit this PR until after the Release is cut, and confirming with you and @gilbertw.

If there is more feedback, either from you, or from the rest of the team, feel free to comment, but as of now I cannot think of anything more to add to this PR, so I am marking it as "Ready for Review".

JeffreyRichter:
@antkmsft : Yes, you did everything the IoT would like. Thanks for doing this so quickly and hold off on merging. I'll mail the IoT team (and CC you) next week.

JeffreyRichter:
FYI: We don't force users to use CMake; they may make their own projects using their own IDE and just build from there. So, we shouldn't do anything that 'requires' Cmake.

sdk/core/core/internal/az_log_internal.h
```diff
@@ -12,9 +12,23 @@
...
+  (void)classification;
+  return false;
+}
+
+AZ_INLINE void _az_log_write(az_log_classification classification, az_span message)
+{
+  (void)classification;
+  (void)message;
+}
+#endif // AZ_NO_LOGGING
```
hihigupt, line 31: any reason not to use same pattern as used by AZ_NO_PRECONDITION_CHECKING?

#ifdef AZ_NO_PRECONDITION_CHECKING
#define _az_PRECONDITION(condition)
#else
#define _az_PRECONDITION(condition) \
  do \
  { \
    if (!(condition)) \
    { \
      az_precondition_failed_get_callback()(); \
    } \
  } while (0)



Also, current implementation can you confirm that there is no code in binary for  _az_log_write, when using no logging,

hihigupt: Tagging @TiejunMS for code review

TiejunMS: I'm OK with the inline function. The compiler is smart enough to strip out the log functions when `AZ_NO_LOGGING` is defined. Here are the results of experiments I did.
1. Keep source code of convenience layer (CL) unchanged and disable logging in CL. RO code is 8000 and RO data is 303. [Link](https://expresslogic.visualstudio.com/X-Ware/_build/results?buildId=14313&view=logs&j=84d3beb9-ba6d-5951-beee-b2a2468262b8&t=9c5b81df-f200-5a06-d6b4-ebad7c262ef0&l=46).
1. Use `_az_log_write` in CL with az_log **enabled**. RO code is 10052 and RO data is 3903. [Link](https://expresslogic.visualstudio.com/X-Ware/_build/results?buildId=14314&view=logs&j=84d3beb9-ba6d-5951-beee-b2a2468262b8&t=9c5b81df-f200-5a06-d6b4-ebad7c262ef0&l=46).
1.  Use `_az_log_write` in CL with az_log **disabled**. RO code is 7912 and RO data is 303. [Link](https://expresslogic.visualstudio.com/X-Ware/_build/results?buildId=14317&view=logs&j=84d3beb9-ba6d-5951-beee-b2a2468262b8&t=9c5b81df-f200-5a06-d6b4-ebad7c262ef0&l=45).

JeffreyRichter:
Actually, why don't we change the preconditions to be inline functions instead of macros?
Macros cause lots of problems with debugging and are less error prone.
Any objections to us converting precondition macros to functions?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -169,6 +174,21 @@ jobs:
         vcpkg.deps: 'cmocka'
         VCPKG_DEFAULT_TRIPLET: 'x64-osx'
         build.args: ' -DUNIT_TESTING=ON -DBUILD_PRECONDITIONS=OFF'
+
+      # Build with no logging
+      # WARNINGS_AS_ERRORS          ON
```
CIPop, line 179: Nit: Is there a reason to keep comments that may run out of sync with the code below?
E.g. Platform doesn’t seem to be the same as your comment.

sdk/core/core/README.md
```diff
@@ -120,6 +120,9 @@ Log classifications allow your application to select which specific log messages
    }
    ```
 
+If the SDK is built with `AZ_NO_LOGGING` macro defined (or adding option -DBUILD_LOGGING=OFF with cmake), it should reduce the binary size and slightly improve performance.
+Logging has a negligible performance impact if no listener is registered or if you specify few classifications. However, if you'd like to exclude all of the logging code to make your final executable smaller, define the `AZ_NO_LOGGING` symbol when building the SDK.
+
```
CIPop, line 125: We should probably add that without logging, Azure support may not be able to understand or resolve runtime device-caused issues.

sdk/core/core/internal/az_log_internal.h
hihigupt: But if customer wants to debug a debug build and the binary is not fitting (because of logs and no optimization). In that case customer can try AZ_NO_LOGGING to reduce the binary size and see if binary fits in the device (because literal will be removed), but with current change it looks like this is not possible [Link](https://expresslogic.visualstudio.com/X-Ware/_build/results?buildId=14313&view=logs&j=84d3beb9-ba6d-5951-beee-b2a2468262b8&t=0e610488-0db1-5249-3946-6cea4ec1597a), as literal are still pinned in the binary.

TiejunMS: You raise a good point! I didn't think about it. In this case, it is worth converting az_log_write to macros.

antkmsft: @TiejunMS, could you please try to see what numbers you can get from this branch with the optimization flags (https://www.rapidtables.com/code/linux/gcc/gcc-o.html - I played on https://godbolt.org/, it looks like any `-O` would be sufficient - either just `-O`, or `-O3`) and also what numbers you can get with the branch `antk/preconditions-inline-fns` (https://github.com/Azure/azure-sdk-for-c/pull/714).

TiejunMS: @antkmsft , we don't have project for ARM GCC. So I didn't check the code size generated by it. For `antk/preconditions-inline-fns` branch, here is [link](https://expresslogic.visualstudio.com/X-Ware/_build/results?buildId=14597&view=logs&j=84d3beb9-ba6d-5951-beee-b2a2468262b8&t=9c5b81df-f200-5a06-d6b4-ebad7c262ef0&l=46) to map file.

ahsonkhan: @TiejunMS what steps do you take to generate those, what prerequisites do you need, and how long does it take? It would be great if you could share those so we could run it as well, especially based on local changes.

TiejunMS: @ahsonkhan , it is the build pipeline of convenience layer development. The agent for IAR build is run on a physical Windows machine with IAR installed. This [pipeline](https://expresslogic.visualstudio.com/X-Ware/_build?definitionId=139&_a=summary) is triggered by PR and CI. And we can queue the build manually also. It takes 5 minutes from checkout till test is done. 

SDK repository is a submodule of [convenience layer repository](https://expresslogic.visualstudio.com/X-Ware/_git/azure-services-port). If you need to do any tests by this pipeline, you need to update the submodule to your local branch and push to DevOps. Then manually queue a build in this pipeline.

antkmsft: @JeffreyRichter , in case your compiler is not very good, and/or if you don't use optimization flags, function calls and string literals are not being removed. Should we go with the macros? (https://github.com/Azure/azure-sdk-for-c/pull/710). They don't rely on the compiler optimizing, and ensure that the code is not included under any circumstances.

I propose to merge https://github.com/Azure/azure-sdk-for-c/pull/710 here, and close https://github.com/Azure/azure-sdk-for-c/pull/714.

JeffreyRichter:
Yes, go with the macros


sdk/core/core/README.md
antkmsft: Thank you Cristian! Logs do not get sent to us either way, so even with the logging enabled, we will not be able to see more that with the logging disabled. Unless customer implements logging callback function that saves logs in some way, and sends it to us some way "by hand", and on our end there is someone whom they will be in contact who agrees to read these logs and help to debug. All this involves some steps - such as us guiding the customer how to better save the logs and send them to us, and if that is the case, we should mention to customer that `NO_LOGGING` should be disabled.
You are probably thinking of something like "unique request ID" policy which we (are supposed to) include with every HTTP request, and it could help us to diagnose problems with the request on our side, but this macro does not control that - it is logging, that is being handled on the device (whether it saves it to a file or sends to some logging server), and the purpose is more to log what's happening on the device, than what is happening on our side.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
antkmsft: was just keeping in sync with what we already have. @danieljurek , do we want to keep these as comments? Do other languages/pipelines have this?
I am fine wither way, but what you saying is also true, comments may get out of sync with the actual config, and maybe there isn't much sense to repeat in comments what you can just see several lines below.
I don't see where the Platform mismatches the comment, could you please show me? :)

sdk/core/core/internal/az_log_internal.h
antkmsft: Jeffrey agreed to use macros, now I need someone's approval for the https://github.com/Azure/azure-sdk-for-c/pull/710 please. It won't go to `master`, it will go to this branch (the rules are too strict).

antkmsft: @hihigupt , we use macros now.

antkmsft:
@hihigupt, @gilbertw, the release is cut, you can now unblock this PR.

antkmsft:
Holding off until tomorrow, while partners still doing smoke testing on the release.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
danieljurek: Let's keep them here for now and synchronized. I'm going to be refactoring the pipelines to make better use of agent resources (https://github.com/Azure/azure-sdk-for-c/issues/669) and I want to make sure I'm bucketing these appropriately. 

Other languages do not currently keep these kinds of comments because they're testing only different OS platforms instead of OS platforms AND build configurations.

PR #714, opened 2020-05-11
PoC, WIP Preconditions as inline functions

antkmsft:
A possible thing that Jeff has proposed in https://github.com/Azure/azure-sdk-for-c/pull/707

### DIFF
--- a/sdk/core/core/internal/az_precondition_internal.h
+++ b/sdk/core/core/internal/az_precondition_internal.h
@@ -55,7 +55,10 @@ AZ_INLINE void _az_precondition(bool condition)
 #endif // AZ_NO_PRECONDITION_CHECKING
 }
 
-#define _az_PRECONDITION_RANGE(low, arg, max) _az_precondition((low <= arg && arg <= max))
+AZ_INLINE void _az_precondition_range(int32_t low, int32_t arg, int32_t max)
+{
+  _az_precondition((low <= arg && arg <= max));
+}
 
 AZ_INLINE void _az_precondition_not_null(void const* arg) { _az_precondition(arg != NULL); }
 AZ_INLINE void _az_precondition_null(void const* arg) { _az_precondition(arg == NULL); }
--- a/sdk/core/core/internal/az_precondition_internal.h
+++ b/sdk/core/core/internal/az_precondition_internal.h
@@ -43,23 +43,22 @@
 
 az_precondition_failed_fn az_precondition_failed_get_callback();
 
+AZ_INLINE void _az_precondition(bool condition)
+{
 #ifdef AZ_NO_PRECONDITION_CHECKING
-#define _az_PRECONDITION(condition)
+  (void)condition;
 #else
-#define _az_PRECONDITION(condition) \
-  do \
-  { \
-    if (!(condition)) \
-    { \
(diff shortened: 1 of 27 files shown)


sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -43,23 +43,22 @@
...
-  } while (0)
+  if (!condition)
+  {
+    az_precondition_failed_get_callback()();
+  }
 #endif // AZ_NO_PRECONDITION_CHECKING
+}
 
-#define _az_PRECONDITION_RANGE(low, arg, max) _az_PRECONDITION((low <= arg && arg <= max))
+#define _az_PRECONDITION_RANGE(low, arg, max) _az_precondition((low <= arg && arg <= max))
```
JeffreyRichter, line 58: Can't this be an inline function too?

antkmsft: I'll need to think about it, this is why I left it as a macro :) Maybe it needs to work for signed and unsigned ints of different sizes, and maybe even pointers. I'll take a look at usage. Maybe we'll remove this one at all (convert to conditions you pass to `_az_precondition()`)

sdk/core/core/internal/az_precondition_internal.h
```diff
@@ -43,23 +43,22 @@
 
 az_precondition_failed_fn az_precondition_failed_get_callback();
 
+AZ_INLINE void _az_precondition(bool condition)
```
ahsonkhan, line 46: I haven't been following the other threads too closely. Can you summarize why we are turning macros into inline functions (maybe in the PR description)? How do we verify this is doing what it wants us to do.

sdk/core/core/internal/az_precondition_internal.h
antkmsft: Jeff, I made `_az_precondition_range()` to take int32_t (other option would be to replace all `_az_PRECONDITION_RANGE()` usages with `_az_precondition(low <= arg && arg <= max))`

sdk/core/core/internal/az_precondition_internal.h
antkmsft: This is a draft and an experiment. I am linking to the PR where Jeff had this idea in comments.
The idea is that with inline functions, the compiler is smart enough to remove the code that has no effect, so we turn macros into inline functions, and they are more readable and better debuggable.

sdk/core/core/internal/az_precondition_internal.h
JeffreyRichter: Were all examples you found using int32_t?
No uses with other integer types?

antkmsft: All but one. One is using int16_t.

JeffreyRichter: Great, then this int32_t version should work good enough then.

antkmsft: Yes, I think it might be good enough for most cases. If there's any problem (using it with int64, for example), there should be warnings treated as errors. And for all the non-typical cases (pointers, unsigned, int64s) - they are not common, and can be written as `_az_precondition(min <= x && x <= max)`.

PR #718, opened 2020-05-11
Add brief keyword to make sure the description shows up in docs and remove repeated type names

ahsonkhan:
My goal with this PR was to improve the `Data Structures` docs page.
1) Remove redundant/unnecessary links and info
2) Keep the `@brief` short (1-2 short sentences at most). Move the rest to `@details`.
3) Make sure that each typedef has a `@brief`.

**Before:**
image

**After:**
image

cc @danieljurek

### DIFF
--- a/sdk/core/core/inc/az_http.h
+++ b/sdk/core/core/inc/az_http.h
@@ -128,7 +128,7 @@ typedef enum
 } _az_http_response_kind;
 
 /**
- * @brief Represents an HTTP response.
+ * @brief Allows you to parse an HTTP response's status line, headers, and body.
  *
  * @details Users create an instance of this and pass it in to an Azure service client's operation
  * function. The function initializes the az_http_response and application code can query the
--- a/sdk/core/core/inc/az_http.h
+++ b/sdk/core/core/inc/az_http.h
@@ -108,8 +108,9 @@ typedef enum
 /**
  * @brief Allows you to customize the retry policy used by SDK clients whenever they performs an I/O
  * operation.
- * @details Applications should acquire an initialized instance of this struct and then modify any
- * fields you desire before passing a pointer to this struct when initializing the specific client.
+ * @details Client libraries should acquire an initialized instance of this struct and then modify
+ * any fields necessary before passing a pointer to this struct when initializing the specific
+ * client.
  */
 typedef struct
 {
--- a/sdk/core/core/inc/az_http.h
+++ b/sdk/core/core/inc/az_http.h
@@ -106,11 +106,10 @@ typedef enum
 } az_http_status_code;
 
 /**
- * @brief An az_http_policy_retry_options instance allows you to customize the retry policy
- * used by an XxxClient type whenever it performs an I/O operation. Applications should
- * acquire an initialized instance of this struct by calling az_http_policy_retry_options_default
- * first and then modify any fields you desire before passing a pointer to this struct
- * when initializing the XxxClient.
+ * @brief Allows you to customize the retry policy used by SDK clients whenever they performs an I/O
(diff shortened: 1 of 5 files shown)


sdk/core/core/inc/az_http.h
```diff
@@ -106,11 +106,10 @@ typedef enum
 } az_http_status_code;
 
 /**
- * @brief An az_http_policy_retry_options instance allows you to customize the retry policy
- * used by an XxxClient type whenever it performs an I/O operation. Applications should
- * acquire an initialized instance of this struct by calling az_http_policy_retry_options_default
```
ahsonkhan, line 111: Btw, `_az_http_policy_retry_options_default` is an internal API, so how is the caller supposed to create and modify the retry_options? We can't refer to an internal method within the docs.

https://github.com/Azure/azure-sdk-for-c/blob/a941c50ffa218d3a70db99fa0347cacca031e9cb/sdk/core/core/src/az_http_policy_retry.c#L30-L39

cc @JeffreyRichter, @antkmsft

sdk/core/core/inc/az_http.h
```diff
@@ -129,12 +128,12 @@ typedef enum
 } _az_http_response_kind;
 
 /**
- * @brief az_http_response represents an HTTP response.
+ * @brief Represents an HTTP response.
```
ahsonkhan, line 131: Maybe we can improve this a bit to be more meaningful. I am open to suggestions here.

JeffreyRichter: Allows you to parse an HTTP response's status line, headers, and body.

sdk/core/core/inc/az_http.h
JeffreyRichter: Customers do not call this function directly. Instead, when using a specific client library (like Storage), the customer call the client's *_options_default function and it internally calls this function. See https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/storage/blobs/src/az_storage_blobs_blob_client.c#L32

The client library's function needs to document its function. But, the retry structure is publicly exposed/documented from the Core library: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/inc/az_http.h#L109.

antkmsft: In addition to that, customers aren't supposed to create pipelines, and it is the pipeline where you pass these retry options.

sdk/core/core/inc/az_span.h
```diff
@@ -26,7 +26,7 @@
 #include <_az_cfg_prefix.h>
 
 /**
- * A span is a "view" over a byte buffer that represents a contiguous region of memory. It contains
+ * @brief A span is a "view" over a byte buffer that represents a contiguous region of memory. It contains
```
gilbertw, line 29: Nit: In some places the pattern is "A <type> is a ..." and in other cases the brief is "Represents ...."

sdk/core/core/inc/az_http.h
ahsonkhan: So, are the doc changes fine then? Or do you want me to reword it?

sdk/core/core/inc/az_span.h
ahsonkhan: I don't know if consistency here is that important given different people end up writing it differently. But you are right, and I will fix this particular one since I am already touching it in this PR.

sdk/core/core/inc/az_http.h
JeffreyRichter: Docs for this function? You can change it to say that "Client libraries should acquire..."

PR #727, opened 2020-05-12
Rename the /core/core/ directory to /core/az_core/

ahsonkhan:
Trying to be consistent with what other languages do:
https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/core/Azure.Core
https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/core/azure-core
https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/core/azure-core
https://github.com/Azure/azure-sdk-for-ios/tree/master/sdk/core/AzureCore

JavaScript, Go, Android, etc. have some differences:
https://github.com/Azure/azure-sdk-for-go/tree/master/sdk/azcore (no /core/ directory)
https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-auth (separate core-* sub-directories)
https://github.com/Azure/azure-sdk-for-android/tree/master/azurecore (no /sdk/ directory - likely will change, it is fairly new)


danewalton:
This becomes the only directory of a sub-component with the `az_` prefix?

ahsonkhan:
> This becomes the only directory of a sub-component with the `az_` prefix?

Yes, that's true, at least for now. Is that a concern? Do you prefer spelling it out, i.e. `azure_core`?

danewalton:
> 
> 
> > This becomes the only directory of a sub-component with the `az_` prefix?
> 
> Yes, that's true, at least for now. Is that a concern? Do you prefer spelling it out, i.e. `azure_core`?

I only say that because if this gets done and you look at, for example, core vs iot, it would like almost like an incomplete name change across the repo. You have some things following `az_<component>` and others as just `<component>`. Was there an original worry/impetus behind this change?

ahsonkhan:
> Was there an original worry/impetus behind this change?

Consistency with what other languages are doing and some offline discussions where having `/core/core` was causing friction (it begs the question why do we have two?).

danewalton:
> Consistency with what other languages are doing and some offline discussions where having `/core/core` was causing friction (it begs the question why do we have two?).

Could `core` be the exception to the `/<service>/<component>` and it just be `/core` while others might be `/iot/hub`?

danewalton:
Actually after looking at the other language repos, I might be an advocate for putting `az_` in front of everything else.
So something like 
```
sdk
├── core
│   └── az_core
├── iot
│   ├── az_iot_common
│   ├── az_iot_hub
│   ├── az_iot_provisioning
├── platform
│   ├── az_http_client
│   ├── az_noplatform
│   ├── az_posix
│   └── az_win32
├── samples
│   └── az_keyvault
└── storage
    ├── az_blobs
```

And at that point we might be able to change `az_iot_common` to `az_iot_core`?

ahsonkhan:
> And at that point we might be able to change `az_iot_common` to `az_iot_core`?

I think those are separate things and I would keep `az_iot_common` as is. We have used the term `core` broadly in different languages and SDKs to mean Azure Core, so there is little value to overload it. `common` or `shared` make more sense within the context anyway.

JeffreyRichter:
I also prefer az_iot_common. I'd like to keep the Core term reserved for azure core components similar to what other languages do.

ahsonkhan:
> I don't think we should make the change at this point

I will hold off on merging this and will follow up post-preview.

antkmsft:
Oooh, I am looking at all these merge conflicts now and I have bad feeling for all my PRs currently in review :) But do it! - some tools should ultimately be able handle it, we shouldn't optimize for bad tools. (or maybe we're lucky and git just handles renames well)

ahsonkhan:
This branch is in a weird state now so it is easier for me to just restart on top of master and close this PR.
Please review https://github.com/Azure/azure-sdk-for-c/pull/778

PR #728, opened 2020-05-12
Fix up doxyfile based on the recent iot/core to iot/common directory rename.

ahsonkhan:
Follow up to https://github.com/Azure/azure-sdk-for-c/pull/674

Fyi @ewertons 

@danieljurek do you know how the fact that the input wasn't set to the right directory name would impact running `doxygen Doxyfile` at the root of the repo?


ahsonkhan:
> do you know how the fact that the input wasn't set to the right directory name would impact running `doxygen Doxyfile` at the root of the repo?

Before, things like `az_iot_is_success_status` weren't showing up in the locally generated docs.

PR #730, opened 2020-05-12
Gracefully exit on hub samples

danewalton:
Update samples to gracefully exit (disconnect from service and cleanup resources)

### DIFF
--- a/sdk/iot/hub/samples/src/paho_iot_hub_c2d_example.c
+++ b/sdk/iot/hub/samples/src/paho_iot_hub_c2d_example.c
@@ -264,40 +264,53 @@ int main()
     printf("Failed to create MQTT client, return code %d\n", rc);
     return rc;
   }
-
-  // Set the callback for incoming MQTT messages
-  if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
-      != MQTTCLIENT_SUCCESS)
+  else
   {
-    printf("Failed to set MQTT callbacks, return code %d\n", rc);
-    return rc;
+    // Set the callback for incoming MQTT messages
+    if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
+        != MQTTCLIENT_SUCCESS)
+    {
+      //Failed to set the callback - move to destroy mqtt client and return
+      printf("Failed to set MQTT callbacks, return code %d\n", rc);
+    }
+    else
+    {
+      // Connect to IoT Hub
+      if ((rc = connect_device()) != 0)
+      {
+        // Connect failed - move to clean up resources
+      }
+      else
+      {
+        // Subscribe to the C2D topic to receive cloud-to-device messages
+        if ((rc = subscribe()) != 0)
+        {
+          // Failed to subscribe - move to disconnect and clean up resources
+        }
+        else
+        {
+          printf("Subscribed to topics.\n");
+
+          // Connected and subscribed: Wait for any incoming C2D messages
+          printf("Waiting for activity. [Press ENTER to abort]\n");
+          (void)getchar();
+        }
+
+        // In success case or on error subscribing, gracefully disconnect from service
+        if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS))
+            != MQTTCLIENT_SUCCESS)
+        {
+          printf("Failed to disconnect MQTT client, return code %d\n", rc);
+        }
+        else
+        {
+          printf("Disconnected.\n");
+        }
+      }
+    }
   }
 
-  // Connect to IoT Hub
-  if ((rc = connect_device()) != 0)
-  {
-    return rc;
-  }
-
-  // Subscribe to the C2D topic to receive cloud-to-device messages
-  if ((rc = subscribe()) != 0)
-  {
-    return rc;
-  }
-
-  printf("Subscribed to topics.\n");
-
-  // Wait for any incoming C2D messages
-  printf("Waiting for activity. [Press ENTER to abort]\n");
-  (void)getchar();
-
-  if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
-  {
-    printf("Failed to disconnect MQTT client, return code %d\n", rc);
-    return rc;
-  }
-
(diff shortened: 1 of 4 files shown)


sdk/iot/hub/samples/src/paho_iot_hub_c2d_example.c
```diff
@@ -264,40 +264,53 @@ int main()
...
+        // Connect failed - move to clean up resources
+      }
+      else
+      {
+        // Subscribe to the C2D topic to receive cloud-to-device messages
+        if ((rc = subscribe()) != 0)
+        {
+          // Failed to subscribe - move to disconnect and clean up resources
+        }
+        else
```
ahsonkhan, line 290: nit: Can we restructure it to avoid the deep if/else nesting, but rather have it as a sequence of `if(fail condition) print and gracefully fail` stataements?

sdk/iot/hub/samples/src/paho_iot_hub_twin_example.c
```diff
@@ -392,65 +392,76 @@ int main()
...
+          printf("\nWaiting for activity:\nPress 'g' to get the twin document\nPress 'r' to send a "
+                 "reported "
+                 "property\n[Press 'q' to quit]\n");
+          int input;
+          while (1)
+          {
+            input = getchar();
+            if (input != '\n')
+            {
+              switch (input)
```
ahsonkhan, line 430: This particular case highlights the deep nesting issue.

sdk/iot/hub/samples/src/paho_iot_hub_c2d_example.c
danewalton: So something like:

```c
  // Connect to IoT Hub
  if ((rc = connect_device()) != 0)
  {
    return rc;
  }

  // Subscribe to the necessary twin topics to receive twin updates and responses
  if ((rc = subscribe()) != 0)
  {
    return rc;
  }
```

becomes

```c
  // Connect to IoT Hub
  if ((rc = connect_device()) != 0)
  {
    MQTTClient_destroy(&mqtt_client);
    return rc;
  }

  // Subscribe to the necessary twin topics to receive twin updates and responses
  if ((rc = subscribe()) != 0)
  {
    MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS);
    MQTTClient_destroy(&mqtt_client);
    return rc;
  }
```

It's a tradeoff...Cascading if/else or repeated destruction code :/

ahsonkhan: Yep. You are right. Either is fine :)
Would using a macros help or is each condition and its block somewhat unique?

danewalton: or....dare I say
.
.
.
.
.
.
.
`goto`

danewalton: But in seriousness they're each a little different. One assumes that the client is initialized and the other initialized and connected

ahsonkhan: I like gotos, so I would be all for using them :)

ewertons: Hmmm. I would consider it safer not to have gotos... 
Some can use it well. Some can cause havoc for using them. Better be safe.

ericwol-msft: or

`    // Set the callback for incoming MQTT messages
    if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
        != MQTTCLIENT_SUCCESS)
    {
      //Failed to set the callback - move to destroy mqtt client and return
      printf("Failed to set MQTT callbacks, return code %d\n", rc);
    }

    if (rc == 0)
    {
      // Connect to IoT Hub
      if ((rc = connect_device()) != 0)
      {
        // Connect failed - move to clean up resources
      }
    }

    if (rc == 0)
    {
        // Subscribe to the C2D topic to receive cloud-to-device messages
        if ((rc = subscribe()) != 0)
        {
          // Failed to subscribe - move to disconnect and clean up resources
        }
    }

    if (rc == 0)
    {
      printf("Subscribed to topics.\n");

      // Connected and subscribed: Wait for any incoming C2D messages
      printf("Waiting for activity. [Press ENTER to abort]\n");
      (void)getchar();
    }`

danewalton: The only problem here I see is that we still need to differentiate between the case in which we need to disconnect and destroy and the case that we just need to destroy.

sdk/iot/hub/samples/src/paho_iot_hub_c2d_example.c
```diff
@@ -264,40 +264,53 @@ int main()
...
+    if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
+        != MQTTCLIENT_SUCCESS)
+    {
+      //Failed to set the callback - move to destroy mqtt client and return
+      printf("Failed to set MQTT callbacks, return code %d\n", rc);
+    }
+    else
+    {
+      // Connect to IoT Hub
+      if ((rc = connect_device()) != 0)
```
CIPop, line 279: Here, always attempting to do cleanup is bad practice. If the error is indicative of memory corruption, out of memory or access violation.
Example:  MQTTClient_Connect could throw `PAHO_MEMORY_ERROR`. Handling this will:
1. likely lead to even more corruption within the cleanup phase.
2. fail due to out of memory within printf and other library functions.

In general, teaching our customers to _always attempt to recover from all errors_ may lead in almost impossible to understand issues in production settings.

CIPop: Another example: `MQTTCLIENT_BAD_STRUCTURE` is (outside of dev-time errors) indicative of memory corruption as the internal struct is missing the expected signature.

danewalton:
After discussion, because we don't know the full depth of the errors which Paho could return, we shouldn't be trying to gracefully shutdown. However, added comments about where/why we shut down and what a graceful shutdown is should be added. I will do that in another PR.

A possible worthwhile addition in the future is if `MQTTCLIENT_DISCONNECTED` is returned from trying to send an MQTT message, we can demonstrate some retry logic to connect again. For now that doesn't have to do with this PR.

PR #734, opened 2020-05-12
Add documentation to the storage header

gilbertw:


sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -16,6 +27,9 @@
 
 #include <_az_cfg_prefix.h>
 
+/**
+ * @brief Client is fixed to a specific version of the Azure Blob Storage service
```
ahsonkhan, line 31: We use `Azure Blob Storage` service here, but above we say  `Azure Storage Blob` client

Is the client ASB and service ABS?

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
   } _internal;
 } az_storage_blobs_blob_client;
 
+
```
ahsonkhan, line 59: nit: Remove extra line
```suggestion
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+/**
+ * @brief Init a client with default options
```
ahsonkhan, line 61:
```suggestion
 * @brief Initialize a client with default options.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+ *
+ * @param client blob_client to initialize
```
ahsonkhan, line 63: Capitalize and add a period.
```suggestion
 * @param client The blob client instance to initialize.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+ * @param url a url to a blob storage account
```
ahsonkhan, line 64:
```suggestion
 * @param url A url to a blob storage account.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+ * @param credential credential object for authentication
+ *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS
+ * @param options client options
```
ahsonkhan, line 67: Taking inspiration from other places where we init an object passing in `options` pointer.

Note, you'll likely have to format this file with Ctrl+K/Ctrl+D after you accept this suggestion to get line wrapping since it will go over 100 characters. 
```suggestion
 * @param options  A reference to an #az_storage_blobs_blob_client_options structure which defines custom behavior of the client.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
   } _internal;
 } az_storage_blobs_blob_client;
 
+
+/**
+ * @brief Init a client with default options
+ *
+ * @param client blob_client to initialize
+ * @param url a url to a blob storage account
+ * @param credential credential object for authentication
```
ahsonkhan, line 65:
```suggestion
 * @param credential The object used for authentication.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+ *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS
```
ahsonkhan, line 66:
```suggestion
 *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
   az_span option;
 } az_storage_blobs_blob_upload_options;
 
+/**
+ * @brief Init a client with default options
+ * This is convenient method to create a client with basic settings
+ * Specific options can be modified after initializing the options
```
ahsonkhan, line 86: This should not be part of the `@brief` but rather in the `@details` section.



```suggestion
* @details Call this to obtain an initialized #az_storage_blobs_blob_client_options structure that can be modified and passed to #az_storage_blobs_blob_client_init().
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
   az_span option;
 } az_storage_blobs_blob_upload_options;
 
+/**
+ * @brief Init a client with default options
```
ahsonkhan, line 84: This method doesn't initialize a client, but rather just creates the default options that can be passed into the init client method.

```suggestion
 * @brief Gets the default blob storage options.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
+ * This is convenient method to create a client with basic settings
+ * Specific options can be modified after initializing the options
+ *
+ * @remark Use this, for instance, when only caring about setting one option by calling this method
+ * and then overriding that specific option
```
ahsonkhan, line 89: I am not sure what you mean here.

```suggestion
 * @remark Use this, for instance, when only caring about setting one option by calling this method
 * and then overriding that specific option.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
+ */
 AZ_NODISCARD az_storage_blobs_blob_client_options az_storage_blobs_blob_client_options_default();
 
+/**
+ * @brief Init a blob upload with default options
+ * This is convenient method to create a blob upload options with basic settings
+ * Specific options can be modified after initializing the options
```
ahsonkhan, line 96: Same here.
Move to `@details` and add periods.

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
+ * This is convenient method to create a client with basic settings
+ * Specific options can be modified after initializing the options
+ *
+ * @remark Use this, for instance, when only caring about setting one option by calling this method
+ * and then overriding that specific option
+ */
 AZ_NODISCARD az_storage_blobs_blob_client_options az_storage_blobs_blob_client_options_default();
 
+/**
+ * @brief Init a blob upload with default options
```
ahsonkhan, line 94: Similar feedback here/elsewhere.

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
 }
 
-typedef struct
-{
-  az_span option;
-} az_storage_blobs_blob_download_options;
-
 /**
- * @brief Creates a new blob
+ * @brief Uploads to a blob
```
ahsonkhan, line 106:
```suggestion
 * @brief Uploads the contents to blob storage.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
  *
  * @param client a storage blobs client structure
```
ahsonkhan, line 108:
```suggestion
 * @param client A storage blobs client structure.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
+ * @param context supports cancelling long running operations
```
ahsonkhan, line 109:
```suggestion
 * @param context Supports cancelling long running operations.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
  * @param content blob content
```
ahsonkhan, line 110:
```suggestion
 * @param content The blob content to upload.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
- * @param options create options for blob. It can be NULL so nothing is added to http request
- * headers
+ * @param options upload options for blob.
```
ahsonkhan, line 111:
```suggestion
 * @param options The upload options for blob. 
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
...
- * @brief Creates a new blob
+ * @brief Uploads to a blob
  *
  * @param client a storage blobs client structure
+ * @param context supports cancelling long running operations
  * @param content blob content
- * @param options create options for blob. It can be NULL so nothing is added to http request
- * headers
+ * @param options upload options for blob. 
  * @param response a pre allocated buffer where to write http response
```
ahsonkhan, line 112:
```suggestion
 * @param response A pre-allocated buffer where to write HTTP response into.
```

sdk/storage/blobs/inc/az_storage_blobs.h
ahsonkhan: Hmm... interesting. We don't use the options at all, currently.

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -53,28 +80,39 @@ typedef struct
   az_span option;
```
ahsonkhan, line 80: Should this field be internal? What does it mean?

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -42,6 +56,19 @@ typedef struct
...
+ * @param credential credential object for authentication
+ *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS
+ * @param options client options
+ *
+ * @return An #az_result value indicating the result of the operation:
+ *         - #AZ_OK if successful
+ */
 AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
     az_storage_blobs_blob_client* client,
     az_span uri,
```
ahsonkhan, line 74: uri vs url - parameter vs comment mismatch, this is why CI is failing.

@JeffreyRichter what should the parameter name be? We use url in most places (az_http_transport, az_aad, az_credential_client_secret, az_http_request), but the keyvault and storage clients use uri.

JeffreyRichter: Actually, we should be using 'endpoint' everywhere.

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: Should be Azure Storage Blobs client, good catch

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: The options exists for future proofing.  We don't need it now,  We can't change the signature in the future when we may need them.

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: Is endpoint the azure sdk standard for all Uri & Url parameters?

JeffreyRichter: At one point the arch board agreed to use endpoint, see https://azure.github.io/azure-sdk/java_design.html and https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L63

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: Moved to internal

ahsonkhan: nit: also rename it to unused?

sdk/storage/blobs/inc/az_storage_blobs.h
ahsonkhan: Sorry, let me clarify.

Isn't the service called Azure Blob Storage (swap the words storage and blog)?
https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction
https://azure.microsoft.com/en-us/services/storage/blobs/

Looks like the client/packages are called Azure Storage Blobs in other languages, but that ordering reads like a weird service name.
https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/storage/Azure.Storage.Blobs

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -1,6 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+/**
+ * @file az_storage_blobs.h
+ *
+ * @brief Definition for the Azure Storage Blob client.
```
ahsonkhan, line 7:
```suggestion
 * @brief Definition for the Azure Storage Blobs client.
```

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -16,6 +27,9 @@
 
 #include <_az_cfg_prefix.h>
 
+/**
+ * @brief Client is fixed to a specific version of the Azure Storage Blobs service
```
ahsonkhan, line 31:
```suggestion
 * @brief Client is fixed to a specific version of the Azure Blob Storage service.
```

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: I followed what other languages did.  We can revisit for the next release

sdk/storage/blobs/inc/az_storage_blobs.h
gilbertw: For now I'm going to stay with what's in the code for consistency.  We can revisit for the next release

sdk/storage/blobs/inc/az_storage_blobs.h
```diff
@@ -33,48 +47,76 @@ typedef struct
...
+    az_span endpoint,
     void* credential,
     az_storage_blobs_blob_client_options* options);
 
 typedef struct
 {
-  az_span option;
+  struct
+  {
+    az_span option;
```
ahsonkhan, line 81: nit: In case you missed it, rename to `unused`. And maybe make it a char/bool to keep the struct small?
```suggestion
    az_span unused;
```

sdk/storage/blobs/inc/az_storage_blobs.h
ahsonkhan: Makes sense.

ahsonkhan:
Fixed https://github.com/Azure/azure-sdk-for-c/issues/722

PR #736, opened 2020-05-12
Doc typos

danewalton:
Reading the new example doxygen docs found a couple typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ With this in mind, there are many tenants or principles that we follow in order 
 To get help with the SDK:
 
 - File a [Github Issue](https://github.com/Azure/azure-sdk-for-c/issues/new/choose).
-- Ask new qustions or see others' questions on [Stack Overflow](https://stackoverflow.com/questions/tagged/azure+c) using the `azure` and `c` tags.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,7 +82,7 @@ When we make an official release, we will create a unique git tag containing the
     cmake --build .
     ```
 
-   This results in building each library as a static library file, placed in the output directory you created (for example `build\sdk\core\core\Debug`). At a minimum, you must have an `Azure Core` library, a `Platform` library, and an `HTTP` library. Then, you can build any additional Azure service client library you intend to use from within your application (for example `build\sdk\storage\blobs\Debug`). To use our client libraries in your application, just `#include` our public header files and then link your application's object files with our libray files.
+   This results in building each library as a static library file, placed in the output directory you created (for example `build\sdk\core\core\Debug`). At a minimum, you must have an `Azure Core` library, a `Platform` library, and an `HTTP` library. Then, you can build any additional Azure service client library you intend to use from within your application (for example `build\sdk\storage\blobs\Debug`). To use our client libraries in your application, just `#include` our public header files and then link your application's object files with our library files.
```
danewalton, line 85: Don't know why the diff doesn't show it but it's `library` second to last word.

ahsonkhan: I love going to the libray :p

PR #741, opened 2020-05-12
Update readme names

danewalton:
Tending to @ahsonkhan comment here:
https://github.com/Azure/azure-sdk-for-c/pull/735#issuecomment-627107333

### DIFF
--- a/sdk/iot/README.md
+++ b/sdk/iot/README.md
@@ -1,4 +1,4 @@
-# Embedded C for Azure IoT Clients
+# Azure IoT Clients
 
 Azure SDK for Embedded C official IoT Hub client library.
 
--- a/sdk/iot/provisioning/readme.md
+++ b/sdk/iot/provisioning/readme.md
@@ -1,4 +1,4 @@
-# Azure IoT Clients for Embedded C: Azure IoT Device Provisioning Client
+# Azure IoT Device Provisioning Client
 
 Official Embedded C client for Azure IoT Provisioning Service.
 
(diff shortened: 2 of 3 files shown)


sdk/iot/README.md
```diff
@@ -1,4 +1,4 @@
-# Embedded C for Azure IoT Clients
+# Azure IoT Clients
 
 Azure SDK for Embedded C official IoT Hub client library.
```
ahsonkhan, line 3: Is this accurate?

danewalton: on line 3? I believe so. "Azure SDK for Embedded C" is the top level name so what follows seems reasonable

ahsonkhan: Yes, I meant the other part `official IoT Hub client library.`

The title is `Azure IoT Clients`, but the tagline mentions `Hub`. Is that intentional?

This seems similar to the one in sdk/iot/hub/readme.md
`"Official Embedded C client for Azure IoT Hub.`

CIPop: Good catch @ahsonkhan .  That is wrong: we have two separate services: Device Provisioning Service and IoT Hub. They are not the same and we shouldn't mix things up. The /iot folder will contain all IoT related clients.

CIPop: Oh, I take that back: It is called Azure IoT Hub Device Provisioning Service.

sdk/iot/provisioning/readme.md
```diff
@@ -1,4 +1,4 @@
-# Azure IoT Clients for Embedded C: Azure IoT Device Provisioning Client
+# Azure IoT Device Provisioning Client
```
CIPop, line 1: We should correct here as well: https://docs.microsoft.com/en-us/azure/iot-dps/about-iot-dps

ahsonkhan:
I am going to merge this now. Please follow up separately for any of the tag line changes.

PR #743, opened 2020-05-12
Rename BUILD_PAHO_TRANSPORT cmake option to TRANSPORT_PAHO.

ahsonkhan:
Follow up on https://github.com/Azure/azure-sdk-for-c/pull/732 for consistency of option names used for the transport policy.


danewalton:
Would curl be changed too?

PR #753, opened 2020-05-13
Fix remaining instances of BUILD_CURL_TRANSPORT => TRANSPORT_CURL

antkmsft:


antkmsft:
@ahsonkhan , just saw your PR, mine was just a bit sooner :) but we both made the same changes independently, which is a good sign. Once CI finishes, I will check this in, or feel free to complete it, or we can go with your PR - does not matter :)

PR #755, opened 2020-05-13
Proposal: CMake var renames: AZ_PLATFORM_IMPL => PLATFORM, USER => CUSTOM

antkmsft:
If I check this in, I plan to do so after the release is cut.

We recently made better naming for our cmake variables (BUILD_PRECONDITIONS => PRECONDITIONS, and so on).
I think instead of `AZ_PLATFORM_IMPL` Cmake variable, we should use just `PLATFORM`.
And other thing, I think `CUSTOM` looks better and clearer to me in naming/intent (to me it does), than `USER`. If you don't like this part, I can remove just this rename.
Let me know what you think!

Also, another question/proposal (not in this PR so far, see https://github.com/Azure/azure-sdk-for-c/pull/756):
Should we combine PLATFORM and USER_PLATFORM_NAME?
* If you set it to "WIN32", it is az_win32.
* If you set it to "POSIX", it is az_posix.
* If you set it to "NONE", it is az_noplatform.
* If you don't set it, it is "AUTO" (see below).
* If it is "AUTO", then it is "WIN32" on windows, and "POSIX" on Linux/MAC/BSD (we'll do our best effort to detect via CMake, and if we can't - it is "NONE")
* If it is anything else, it is equal to what today is "USER_PLATFORM_IMPL_NAME" ("CUSTOM_PLATFORM_NAME" in this PR) I.e. you specify it to CMake as
`-DPLATFORM=atmel_avr32_platform_lib`
(instead of
`-DPLATFORM=CUSTOM -DCUSTOM_PLATFORM_NAME=atmel_avr32_platform_lib`)


README.md
```diff
@@ -124,8 +124,8 @@ The following compiler options are available for adding/removing project feature
...
+<td>PLATFORM</td>
+<td>This option can be set to any of the next values:<br>- No_value: default value is used and no_platform library is used.<br>- "POSIX": Provides implementation for Linux and Mac systems.<br>- "WIN32": Provides platform implementation for Windows based system<br>- "CUSTOM": Tells cmake to use an specific implementation provided by user. When setting this option, user must provide an implementation library and set option `CUSTOM_PLATFORM_NAME` with the name of the library (i.e. <code>-DPLATFORM=CUSTOM -DCUSTOM_PLATFORM_NAME=user_platform_lib</code>). cmake will look for this library to link az_core</td>
```
ahsonkhan, line 128: an specific -> a specific

```suggestion
<td>This option can be set to any of the next values:<br>- No_value: default value is used and no_platform library is used.<br>- "POSIX": Provides implementation for Linux and Mac systems.<br>- "WIN32": Provides platform implementation for Windows based system<br>- "CUSTOM": Tells cmake to use a specific implementation provided by user. When setting this option, user must provide an implementation library and set option `CUSTOM_PLATFORM_NAME` with the name of the library (i.e. <code>-DPLATFORM=CUSTOM -DCUSTOM_PLATFORM_NAME=user_platform_lib</code>). cmake will look for this library to link az_core</td>

```

sdk/core/core/CMakeLists.txt
```diff
@@ -71,28 +71,28 @@ target_include_directories(${TARGET_NAME} PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRE
...
+if(PLATFORM STREQUAL "CUSTOM")
+  set(PLATFORM_CUSTOM ON)
+elseif(PLATFORM STREQUAL "POSIX")
+  set(PLATFORM_POSIX ON)
+elseif(PLATFORM STREQUAL "WIN32")
+  set(PLATFORM_WIN32 ON)
 else()
-  set(AZ_PLATFORM_IMPL_NONE ON)
+  set(PLATFORM_NONE ON)
 endif()
```
ahsonkhan, line 87: Since you are already modifying this, can we get rid of `PLATFORM_CUSTOM`, `PLATFORM_POSIX`, etc. and the if blocks below, and just move the  `target_link_libraries` instructions into this if-block? We can simplify the logic here.

For example:
```text
if(PLATFORM STREQUAL "CUSTOM")
   target_link_libraries(${TARGET_NAME} PRIVATE ${CUSTOM_PLATFORM_NAME})
elseif(PLATFORM STREQUAL "POSIX")
   target_link_libraries(${TARGET_NAME} PRIVATE az_posix)
...

It doesn't look like we use the `PLATFORM_*` names anywhere else anyway, so there is no point setting them.
```

sdk/core/core/CMakeLists.txt
```diff
@@ -71,28 +71,28 @@ target_include_directories(${TARGET_NAME} PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRE
...
 
-if(AZ_PLATFORM_IMPL_USER)
-  target_link_libraries(${TARGET_NAME} PRIVATE ${AZ_USER_PLATFORM_IMPL_NAME})
-elseif(AZ_PLATFORM_IMPL_POSIX)
+if(PLATFORM_CUSTOM)
+  target_link_libraries(${TARGET_NAME} PRIVATE ${CUSTOM_PLATFORM_NAME})
+elseif(PLATFORM_POSIX)
```
ahsonkhan, line 91: Can we simplify the logic and remove this if-block?

sdk/core/core/CMakeLists.txt
antkmsft: Thank you, good suggestion! Done.

sdk/core/core/CMakeLists.txt
antkmsft: Thank you, done :)

antkmsft:
Should we combine PLATFORM and USER_PLATFORM_NAME?
* If you set it to "WIN32", it is az_win32.
* If you set it to "POSIX", it is az_posix.
* If you set it to "NONE", it is az_noplatform.
* If you don't set it, it is "AUTO" (see below).
* If it is "AUTO", then it is "WIN32" on windows, and "POSIX" on Linux/MAC/BSD (we'll do our best effort to detect via CMake)
* If it is anything else, it is equal to what today is "USER_PLATFORM_IMPL_NAME" ("CUSTOM_PLATFORM_NAME" in this PR)

antkmsft:
@vhvb1989 is restructuring how we build/link platforms now, and even if we may want some ideas from this PR, they will be in another places, so I am closing this PR without merging.

PR #756, opened 2020-05-13
Proposal: Unify PLATFORM and CUSTOM_PLATFORM_NAME, Add auto-detection

antkmsft:
You probably should see https://github.com/Azure/azure-sdk-for-c/pull/755 first, but you don't have to.

In this PR:
* Unify PLATFORM (which is normally "WIN32", "POSIX", "NONE" or "USER") and CUSTOM_PLATFORM_NAME (which is supposed to be "user_pal_impl_lib").
* Add platform auto-detection (when no platform is specified or when explicitly requested).

This should make it easier for the new users to get started with the SDK.

I have not tested this  yet, but based on the info I found, we should be able to make this work (https://stackoverflow.com/questions/9160335/os-specific-instructions-in-cmake-how-to).

This PR is supposed to go into another PR, https://github.com/Azure/azure-sdk-for-c/pull/755, but should the ideas from that PR be rejected, it can be rebased against master, if neccessary.


antkmsft:
@vhvb1989 is restructuring how we build/link platforms now, and even if we may want some ideas from this PR, they will be in another places, so I am closing this PR without merging.

PR #761, opened 2020-05-13
Add missing doc strings

gilbertw:
Fixes a some of the places where doc strings are missing.  Does not address all the platform headers

### DIFF
--- a/sdk/core/core/inc/az_credentials.h
+++ b/sdk/core/core/inc/az_credentials.h
@@ -104,8 +104,9 @@ typedef struct
  * @param tenant_id an Azure tenant ID
  * @param client_id an Azure client ID
  * @param client_secret an Azure client secret
- * @return AZ_OK = Successfull initialization <br>
- * Other value = Initialization failed
+ * @retval An #az_result value indicating the result of the operation:
+ *         - #AZ_OK if successful
+ *         - Other value = Initialization failed
  */
 AZ_NODISCARD az_result az_credential_client_secret_init(
     az_credential_client_secret* self,
--- a/sdk/core/core/inc/az_credentials.h
+++ b/sdk/core/core/inc/az_credentials.h
@@ -106,7 +106,7 @@ typedef struct
  * @param client_secret an Azure client secret
  * @retval An #az_result value indicating the result of the operation:
  *         - #AZ_OK if successful
- *         - Other value = Initialization failed
+ *         - Other error code if initialization failed
  */
 AZ_NODISCARD az_result az_credential_client_secret_init(
     az_credential_client_secret* self,
--- a/sdk/core/core/inc/az_credentials.h
+++ b/sdk/core/core/inc/az_credentials.h
@@ -104,7 +104,7 @@ typedef struct
  * @param tenant_id an Azure tenant ID
  * @param client_id an Azure client ID
  * @param client_secret an Azure client secret
- * @retval An #az_result value indicating the result of the operation:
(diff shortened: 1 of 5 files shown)


sdk/platform/http_client/nohttp/src/az_nohttp.c
```diff
@@ -5,6 +5,14 @@
 
 #include <_az_cfg.h>
 
+/**
+ * @brief Provides no http support.
```
ahsonkhan, line 9:
```suggestion
 * @brief Provides no HTTP support.
```

sdk/platform/http_client/nohttp/src/az_nohttp.c
```diff
@@ -5,6 +5,14 @@
...
+ *
+ * @param p_request an internal http builder with data to build and send http request
```
ahsonkhan, line 11:
```suggestion
 * @param p_request An internal HTTP builder with data to build and send HTTP request.
```

sdk/platform/http_client/nohttp/src/az_nohttp.c
```diff
@@ -5,6 +5,14 @@
...
+ * @param p_response pre-allocated buffer where http response will be written
```
ahsonkhan, line 12:
```suggestion
 * @param p_response A pre-allocated buffer where the HTTP response will be written.
```

sdk/core/core/inc/az_credentials.h
```diff
@@ -104,8 +104,9 @@ typedef struct
  * @param tenant_id an Azure tenant ID
  * @param client_id an Azure client ID
  * @param client_secret an Azure client secret
- * @return AZ_OK = Successfull initialization <br>
- * Other value = Initialization failed
+ * @retval An #az_result value indicating the result of the operation:
```
ahsonkhan, line 107: For consistency, let's continue to use `@return`.
```suggestion
 * @return An #az_result value indicating the result of the operation:
```

See previous discussion on `@retval` vs `@return`.
https://github.com/Azure/azure-sdk-for-c/pull/622#discussion_r422412165

sdk/core/core/inc/az_credentials.h
```diff
@@ -104,8 +104,9 @@ typedef struct
...
+ *         - #AZ_OK if successful
+ *         - Other value = Initialization failed
```
ahsonkhan, line 109:
```suggestion
 *         - Other error code if initialization failed
```

sdk/core/core/inc/az_credentials.h
```diff
@@ -104,8 +104,9 @@ typedef struct
  * @param tenant_id an Azure tenant ID
  * @param client_id an Azure client ID
  * @param client_secret an Azure client secret
- * @return AZ_OK = Successfull initialization <br>
- * Other value = Initialization failed
+ * @return An #az_result value indicating the result of the operation:
```
antkmsft, line 107: I think we should be using `@retval` across all our docs. For this PR, it is fine (unless you want to solve this ultimately right now in this PR). But we should not forget this, let's discuss switching to `@retval` on our meetings.

ahsonkhan: @antkmsft - please file an issue and let's address it across the board if there is a strong desire. I am fine with what we have currently, but also like the mixed approach:
https://github.com/Azure/azure-sdk-for-c/pull/622#discussion_r422412165

PR #766, opened 2020-05-15
Update to use add_compile_definitions instead of C_FLAGS

danewalton:
Discussion here on another PR:
https://github.com/Azure/azure-sdk-for-c/pull/763#discussion_r424842003

We should try to avoid mucking with C_FLAGS if we can and use the CMake primitives to modify the compilation.


danewalton:
Closing since my man @vhvb1989 already did it in the other PR.

PR #767, opened 2020-05-15
Add subdirectory only for specified platform

danewalton:
@TiejunMS was using our SDK with arm-gcc and the way the current platform structure is set up in CMake causes POSIX code to get built even if `no_platform` is selected. It currently gets built
`if (!WIN32)` which in arm-gcc with `no_platform` selected renders true.

I think this change helps to only pull in code for that platform which was specified.

### DIFF
--- a/sdk/platform/posix/CMakeLists.txt
+++ b/sdk/platform/posix/CMakeLists.txt
@@ -1,8 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-if (NOT WIN32)
-
 cmake_minimum_required (VERSION 3.10)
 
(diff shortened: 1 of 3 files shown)


sdk/platform/posix/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-if (NOT WIN32)
```
antkmsft, line 4: can we change just this single line to "if (UNIX)"? It should work.

antkmsft: I am retracting this proposal, it would work, but I think checking AZ_PLATFORM_IMPL gives us more flexibility.

PR #769, opened 2020-05-15
Exclude unnecessary file from cmake

antkmsft:
The fact that there are quotes is a good sign that VS has automatically added it.
NoLogging commit introduced it, so it was not in the release. This change will go in now.


antkmsft:
The NoLogging commit introduced it, so it was not in the release. This change will go in now.

Issue #496, opened 2020-03-19
Review parameter names and generate guidelines for public APIs within the SDK


ahsonkhan:
Capture the feedback from https://github.com/Azure/azure-sdk-for-c/pull/492, apply it across the repo, and come up with guidelines.

Here are some issues to follow up on:
- self, this, context specific name
- request, ref_request, p_request (and same with response, etc.)
- out parameter naming
- use options instead of data for policies

> `policies, ref_request, ref_response`is the ideal naming. We don't use the `p_` convention anomore, we should change it to the other way.
> 
> ```
> somestruct const* options // means it is read-only and we will not modify it
> somestruct* ref_somestruct //means we read it and (may) modify it
> somestruct* out_somestruct // means we don't read it, the customer can provide an uninitialized struct, and we will fill it in as a returning result
> ``
> ```

antkmsft:
I'd add search and replace all instances of `hrb` with `request`.

gilbertw:
@ahsonkhan  Do you have suggestions beyond what is in the guidelines?
https://azure.github.io/azure-sdk/clang_design.html#functions
> YOU SHOULD use a meaningful name for parameters and local variable names. Parameters and local variable names should be named as all lower-case words, separated by underscores (snake-casing).

antkmsft:
Can I pile one more request here to keep track of? Consider documenting internal functions.
(Comes from one of the suggestions for #771)

antkmsft:
#819

antkmsft:
Lots of renames in our code, that part should be completely done. There are no more "self", or "hrb", no "_p", but are "ref_" and "out_" where appropriate. So now, to close this, I suppose, the only thing left is the guideline update?

antkmsft:
In #1057, I think, I've updated all the remaining parameter instances that should clearly be `ref_` or `out_` in the public API.
But in order to close this work item, IIRC, we wanted to also update guideline docs?

PR #771, opened 2020-05-15
Add thread safety for Client Secret Credential

antkmsft:
Notes:
* No more mutex functions, `az_platform_atomic_compare_exchange()` instead.
* Moved `internal/az_platform_internal.h` to be `inc/az_platform.h`. Reason: customers are supposed to implement PAL layer with signatures from this header, so it should be public.
* Rearranged fields in credential structs for better alignment.
* We take lock to set, AND to get (just read and then quickly release). Without locking when get(), we cannot read the buffer reliably and assume it won't change just as we read it, as well we can't be sure that expiration and the token string do match.
* compare-exchange swaps pointers, so we have these two pointers that we are swapping, `&_az_token_credential_locked` and `&_az_token_credential_unlocked`. They are locally defined in the newly formed `az_credential_token.c` (Corresponds to TokenCredential in SDKs for other languages).
* I pulled `az_token_expired()` and `az_token_set()` to `az_credential_token.c`, as well as added `az_token_get()` (thread-safe), and `az_credential_token_init()`, which initializes lock pointer with `&_az_token_credential_unlocked`.

I recommend to read file changes (https://github.com/Azure/azure-sdk-for-c/pull/771/files) from the bottom, as it starts with platform, then `az_credential_token_private.h`, and moves into more details.


sdk/core/core/inc/az_credentials.h
```diff
@@ -45,12 +46,21 @@ typedef struct
 {
   struct
   {
-    uint8_t token[_az_TOKEN_BUF_SIZE]; /*!< Base64-encoded token */
     int16_t token_length;
     int64_t expires_at_msec;
+    uint8_t token[_az_TOKEN_BUF_SIZE]; /*!< Base64-encoded token */
```
ahsonkhan, line 51: > Rearranged fields in credential structs for better alignment.

Can you please explain how this leads to better alignment?

sdk/core/core/inc/az_credentials.h
```diff
@@ -45,12 +46,21 @@ typedef struct
...
   } _internal;
 } _az_token;
 
+typedef struct
+{
+  struct
+  {
+    bool* volatile lock;
```
ahsonkhan, line 59: Do we have any atomic assignment operations in C? Is using `volatile` sufficient to implement a locking mechanism that is thread-safe?

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <az_result.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include <_az_cfg_prefix.h>
+
+/**
+ * @brief az_platform_clock_msec gets platform clock in milliseconds.
```
ahsonkhan, line 15: Do not re-state the method name within the docs and make sure they are complete sentences.
```suggestion
 * @brief Gets the platform clock in milliseconds.
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ * The moment of time where clock starts is undefined, but if this function is getting called twice
+ * with one second interval, the difference between the values returned should be equal to 1000.
```
ahsonkhan, line 17: The `@brief` should be brief, ideally 1 sentence.  The rest should be in `@details`, `@note`, or `@remark`

```suggestion
 * @remark The moment of time where clock starts is undefined, but if this function is getting called twice
 * with one second interval, the difference between the values returned should be equal to 1000.
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ * @return Platform clock in milliseconds.
+ */
+AZ_NODISCARD int64_t az_platform_clock_msec();
+
+/**
+ * @brief az_platform_sleep_msec tells platform to sleep for a given number of milliseconds.
```
ahsonkhan, line 23:
```suggestion
 * @brief Tells the platform to sleep for a given number of milliseconds.
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ * @param milliseconds number of milliseconds to sleep.
```
ahsonkhan, line 24:
```suggestion
 * @param milliseconds Number of milliseconds to sleep.
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ * @remarks For passing \p milliseconds with negative values and 0, behavior is undefined.
```
ahsonkhan, line 25: nit: Reword
```suggestion
 * @remarks The behavior is undefined when \p milliseconds is a non-positive value (0 or less than 0).
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+void az_platform_sleep_msec(int32_t milliseconds);
+
+/**
+ * @brief az_platform_atomic_compare_exchange Conditionally exchanges values of two pointers in a
```
ahsonkhan, line 30:
```suggestion
 * @brief Conditionally exchanges values of two pointers in a
```

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ * @param desired a value to assign to the value that is pointed by \p obj, if its value equals to
+ * \p expected.
+ *
+ * @return `true` if previous value of \p obj matches the \p expected.
+ *
+ * @remarks If `*obj` equals to \p expected, its value is being overwritten with \p desired, and no
+ * operation otherwise. Returns `true` if the value was overwritten, `false` otherwise.
+ */
+AZ_NODISCARD bool az_platform_atomic_compare_exchange(
+    void* volatile* obj,
```
ahsonkhan, line 44: `volatile*` - What does this mean?

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param obj a pointer to a pointer that needs to be changed.
+ * @param expected an expected value of a value that \p obj is pointing to, prior to exchange.
+ * @param desired a value to assign to the value that is pointed by \p obj, if its value equals to
+ * \p expected.
+ *
+ * @return `true` if previous value of \p obj matches the \p expected.
+ *
+ * @remarks If `*obj` equals to \p expected, its value is being overwritten with \p desired, and no
+ * operation otherwise. Returns `true` if the value was overwritten, `false` otherwise.
```
ahsonkhan, line 41: > Returns `true` if the value was overwritten, `false` otherwise.

This belongs in `@return`

sdk/platform/posix/src/az_posix.c
```diff
@@ -21,25 +20,10 @@ void az_platform_sleep_msec(int32_t milliseconds)
...
-
-AZ_NODISCARD az_result az_platform_mtx_lock(az_platform_mtx* mtx)
-{
-  return pthread_mutex_lock(&mtx->_internal.mutex) == 0 ? AZ_OK : AZ_ERROR_MUTEX;
-}
-
-AZ_NODISCARD az_result az_platform_mtx_unlock(az_platform_mtx* mtx)
-{
-  return pthread_mutex_unlock(&mtx->_internal.mutex) == 0 ? AZ_OK : AZ_ERROR_MUTEX;
+  return __sync_bool_compare_and_swap(obj, expected, desired);
```
ahsonkhan, line 28: Will this work for all supported compilers?

sdk/platform/win32/src/az_win32.c
```diff
@@ -1,34 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include <az_platform_internal.h>
+#include <az_platform.h>
+
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
+#include <windows.h>
```
ahsonkhan, line 8: Can you please add a comment explaining these?

sdk/core/core/src/az_credential_token_private.h
```diff
@@ -0,0 +1,23 @@
...
+#define _az_CREDENTIAL_TOKEN_PRIVATE_H
+
+#include <az_credentials.h>
+#include <az_result.h>
+
+#include <stdbool.h>
+
+#include <_az_cfg_prefix.h>
+
+AZ_NODISCARD az_result _az_credential_token_init(_az_credential_token* out_credential_token);
```
ahsonkhan, line 14: nit: I understand these are private headers, but consider adding docs.

sdk/core/core/inc/az_platform.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+void az_platform_sleep_msec(int32_t milliseconds);
+
+/**
+ * @brief az_platform_atomic_compare_exchange Conditionally exchanges values of two pointers in a
+ * thread-safe manner.
+ *
+ * @param obj a pointer to a pointer that needs to be changed.
+ * @param expected an expected value of a value that \p obj is pointing to, prior to exchange.
+ * @param desired a value to assign to the value that is pointed by \p obj, if its value equals to
```
ahsonkhan, line 35: Capitalize. 

nit: Also, I am having difficulty understanding this a bit due to the repetition. Maybe there is a clearer way to explain the parameters.

sdk/core/core/inc/az_credentials.h
antkmsft: Ah, I should move int64_t before int16, my eyes didn't catch that initially.
Basically, on some platforms, compilers would align data to a word binary. It is usually int32, but can be any reasonable number (1, 2, 4, 8). If you put an array of 9 int8's first, and then an int32, it goes as `[int8][int8][int8][int8] | [int8][int8][int8][int8] | [int8][padding][padding][padding]|[int32]`

sdk/core/core/inc/az_credentials.h
antkmsft: Oh do I wish it had!!! :) https://www.drdobbs.com/parallel/volatile-vs-volatile/212701484

sdk/core/core/inc/az_platform.h
antkmsft: Oh, we do it this way now? That's great, it is even better. I took a comment from another file as a template, we should fix it too.

sdk/core/core/inc/az_platform.h
antkmsft: I intended return to be short and factually/pedantically correct, while this section would better decipher would true/false means, in more simple words, but technically not as complete/accurate as @return section does.

sdk/platform/win32/src/az_win32.c
antkmsft: These are pretty much common knowledge in the Windows programming world, these macros are needed to be defined before including `windows.h`. Yet despite being pretty widely used, there isn't a small MSDN page I could point to - the data is spread across MSDN here and there - https://docs.microsoft.com/en-us/search/?search=WIN32_LEAN_AND_MEAN . I'll see what I can do.

sdk/platform/posix/src/az_posix.c
antkmsft: GCC and Clang for sure. One moment... It is a good question, I got a bit better idea.

sdk/core/core/src/az_credential_token_private.h
antkmsft: At one point, let's do it all the way across our docs, review everything, do the renames for parameter prefies, replace `@return` with `@retval`s.

sdk/core/core/inc/az_platform.h
antkmsft: So basically when you see a pointer, whatever is on the left side of `*` characterizes what it is pointing to. On the right - characterizes the pointer itself.
Remember this example?:
It is easier when you read pointers from right to left:
 `int const*` - non-constant pointer to constant int
`int* const` - constant pointer to non-constant int.
`int const* const` - constant pointer to constant int
`int*` - non-constant pointer to non-constant int.
... and so on, with pointers-to-pointers, it can take really monstrous shapes:
`int* const*** const* p` - p is a non-constant pointer to constant pointer to non-constant pointer to non-constant pointer to constant pointer to non-constant int.

A `const` is a modifier, and so is `volatile` (together, they are called "CV-modifiers"). Everywhere where you can use `const`, you can use `volatile`. You can use them both at the same time. Or just one. Or none of them.

And `void*` is "just" a pointer, to an unknown type. So, before you read or write to the value pointed by such pointer (`*p = 3`), you would need to cast it to a pointer of some type, other than `void`: *(int*)p = 3; *(double*)p = 2.0.

so, to read this, `void* volatile*` is a  (non-constant) pointer to a volatile pointer (to ???).

sdk/core/core/inc/az_platform.h
antkmsft: I'll try to see if I can rephrase this, maybe I'll borrow something from https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer. Ah, documenting the code you just written feels a bit like playing chess against yourself :)

sdk/platform/win32/src/az_win32.c
```diff
@@ -1,34 +1,22 @@
...
-
-AZ_NODISCARD az_result az_platform_mtx_unlock(az_platform_mtx* mtx)
+AZ_NODISCARD bool az_platform_atomic_compare_exchange(
+    void* volatile* obj,
+    void* expected,
+    void* desired)
 {
-  LeaveCriticalSection(&mtx->_internal.cs);
-  return AZ_OK;
+  return InterlockedCompareExchangePointer(obj, expected, desired) == expected;
```
antkmsft, line 21: @JeffreyRichter, I'll need to update this. https://docs.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchangepointer Says:
"On a 64-bit system, the parameters are 64 bits and must be aligned on 64-bit boundaries; otherwise, the function will behave unpredictably. On a 32-bit system, the parameters are 32 bits and must be aligned on 32-bit boundaries."
I think we need to store an array of 15 bytes, and then expect that whatever within this array ends up being aligned to a 64-bit boundary, that is our pointer. Because otherwise we need to require that the pointer is aligned, which means that the client needs to be aligned, which means a blob client, for example, can't be created on the stack.

JeffreyRichter: Don't worry about this; the compiler WILL ensure that the pointer is aligned even if allocated on the stack.

sdk/core/core/inc/az_credentials.h
antkmsft: Yes-yes, I acknowledged that right in my comment to Ahson above. I'll rearrange this.

sdk/platform/win32/src/az_win32.c
antkmsft: Currently, Linux 64 and only Linux 64 unit tests hang. It is hard to tell what's going on, because preconditions just sleep forever, and I don't know which one failed. But the symptoms are similar. Or maybe on Linux 64 + GCC, compiler does not do this alighnment, and fails into UB, that's another possibility. However `__sync_bool_compare_and_swap` do not mention requirement for alignment.

sdk/core/core/inc/az_credentials.h
antkmsft: Done.

sdk/core/core/inc/az_credentials.h
antkmsft: (resolving)

sdk/core/core/src/az_credential_token_private.h
antkmsft: (resolving, I updated #496 with the idea of documenting private headers)

sdk/platform/posix/src/az_posix.c
antkmsft: Updated, resolving.

sdk/platform/win32/src/az_win32.c
antkmsft: Done, to an extent

antkmsft:
I addressed all the requests and suggestions for this PR, I fixed the alignment, and commented the usage of the GNU C language extension for atomic compare-exchange. 
Next:implement suggestions that Jeffrey has sent me over email, and fix the test failure that onlu occurs on Linux64: `az_span_is_content_equal(AZ_SPAN_FROM_STR("Bearer AccessToken"), header.value) test_az_credential_client_secret.c:200`

sdk/platform/win32/src/az_win32.c
antkmsft: I have a suspicion that GCC on Linux64 does not align the pointer and ends up shooting into access token buffer memory.

JeffreyRichter: All compilers will align pointers. Plus, if a non-aligned value is accessed in a way that requires alignment, the system would not hang; it would generate an interrupt or exception instead. For example, on Windows, a EXCEPTION_DATATYPE_MISALIGNMENT exception is raised.

JeffreyRichter: To debug what's happening, run the app under the debugger and put a breakpoint on the failed preconditon function.  Or, override the precondition callback by calling https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/core/core/inc/az_precondition.h#L62 and have it to something else. For example, on Windows, you could call https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-debugbreak.  Or dump the stack somehow. It would be best to run in a debugger.

ahsonkhan: > To debug what's happening, run the app under the debugger and put a breakpoint on the failed preconditon function.

FWIW, I couldn't get this approach to work in VS. Let me know if you are able to, @antkmsft

JeffreyRichter: NOTE: You have the arguments in the wrong order: 
PVOID InterlockedCompareExchangePointer(
  PVOID volatile *Destination,
  PVOID          Exchange,
  PVOID          Comperand
);

requires "expected" to be the LAST argument and "desired" is the 2nd argument.

sdk/core/core/inc/az_credentials.h
ahsonkhan: Follow-up question: How are you figuring out how these struct fields are being laid out in memory to know that we are getting alignment? I understand the integer example, but what about array alignment, in C? Is that a pointer (word size) and int32 for the length?

antkmsft: Array elements are always getting placed sequentially right next to each other so that `&array[0] + sizeof(array[0]) == &array[1]`. In this case specifically, we have an array of bytes, so it is going to be `_az_TOKEN_BUF_SIZE` number of bytes placed sequentially.

(`_az_TOKEN_BUF_SIZE` is 2048 bytes and I think it additionally might be a good practice to round byte array sizes to "binary round" numbers for this reason, although I am not completely sure enough to make that recommendation decisively - nothing bad would happen if it would be just 2000 bytes, but certainly it won't hurt if it is 2048. I personally like it for aesthetic reasons, and also with "binary round" numbers you can sometimes do tricks like bit masking, but again, not enough evidence to recommend it for all cases, other than personal preference)

JeffreyRichter: The compiler will properly align fields (guaranteed) [unless you do something special like this: https://docs.microsoft.com/en-us/cpp/preprocessor/pack?view=vs-2019] - there is nothing we have to do. A byte will be on a byte boundary, a 16-bit field will be on a 16-bit boundary, a 32-bit field will be on a 32-bit boundary, and a 64-bit value will be on a 64-bit boundary. A pointer is either 32- or 64-bits depending on whether you compile for 32-bit or 64-bits. An array will be aligned as per the array type. So, in this case, a byte array is aligned on a byte boundary. For an array, it doesn't matter if its 2000 versus 2048 elements -- we should pick the number we want for the array; not a number that looks appealing in source code.

antkmsft:
@JeffreyRichter , this PR is ready to go. I don't have anything more to add to this PR, unless you have other requests. I fixed unit test and the bug we had, and I played with this code in debugger, both under Windows and Linux. If you review and everything looks good to you, please unblock this PR, and I'll check it in.

sdk/core/az_core/src/az_credential_token.c
```diff
@@ -0,0 +1,32 @@
...
+{
+  _az_spinlock_enter_reader(&self->_internal.lock);
+  *out_token = self->_internal.token;
+  _az_spinlock_exit_reader(&self->_internal.lock);
+  return AZ_OK;
+}
+
+AZ_NODISCARD bool _az_token_expired(_az_token const* token)
+{
+  int64_t const expires_at_msec = token->_internal.expires_at_msec;
```
JeffreyRichter, line 30: I think reading expires_at_msec should be done under the reader lock

sdk/platform/noplatform/src/az_noplatform.c
```diff
@@ -1,10 +1,21 @@
...
+#include <az_platform.h>
 
 #include <_az_cfg.h>
 
 AZ_NODISCARD int64_t az_platform_clock_msec() { return 0; }
 
 void az_platform_sleep_msec(int32_t milliseconds) { (void)milliseconds; }
+
+AZ_NODISCARD bool az_platform_atomic_compare_exchange(
+    uintptr_t volatile* obj,
```
JeffreyRichter, line 13: The default, not thread-safe implementation of this could be this:
if (*obj == expected) { *obj = desired; return true; }
return false;

sdk/core/az_core/src/az_credential_token.c
antkmsft: We get the token, with the locks. The token includes both the token string and the expiration time. Then you can read it safely to check expiration, or to include to headers. Read az_credential_client_secret.c in this PR, lines 55-71 is how we do it.
The token does not have a lock, token credential (_az_credential_token) does. _az_token_expired takes the _az_token, not the _az_credential_token.
This way, when you reuse-or-renew-and-use token in the `_az_credential_client_secret_apply()`, you get the token, with the lock, you check expiration, and if it is good, you apply it to the header. Otherwise, you send a request to obtain the new token (without locks), then you set it to the token credential (with locks), and use the token you just obtained and set, without locks to apply to the header.
It is just one read lock operation, instead of two. And you can be sure that as you read both the token string and the expiration, they two do match each other.

sdk/platform/noplatform/src/az_noplatform.c
antkmsft: Thank you, this is so much better. I updated it.

sdk/core/az_core/src/az_credential_token.c
antkmsft: Update: to discuss this, we talked over the phone, you are fine with the idea, and asked me to update comments, and I just pushed a commit with more comments. Once CI/CD completes, I am checking this PR in. Thank you! I am resolving this conversation thread.

sdk/platform/win32/src/az_win32.c
antkmsft: All is corrected, and works. I am resolving this conversation thread.

sdk/core/core/inc/az_platform.h
antkmsft: I capitalized and tried to rephrase, resolving the conversation thread.

PR #773, opened 2020-05-16
Make AZ_SPAN_INIT_FROM_BUFFER() to use the regular az_span_init()

antkmsft:
Make `AZ_SPAN_INIT_FROM_BUFFER()` to use the regular `az_span_init()` instead of always the lightweight one.

### DIFF
--- a/sdk/core/core/inc/az_span.h
+++ b/sdk/core/core/inc/az_span.h
@@ -72,15 +72,6 @@ AZ_NODISCARD AZ_INLINE az_span az_span_init(uint8_t* ptr, int32_t size)
 AZ_NODISCARD az_span az_span_init(uint8_t* ptr, int32_t size);
 #endif // AZ_NO_PRECONDITION_CHECKING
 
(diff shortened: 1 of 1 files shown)


sdk/core/core/inc/az_span.h
```diff
@@ -153,7 +144,7 @@ AZ_NODISCARD AZ_INLINE az_span _az_span_init(uint8_t* ptr, int32_t size)
  */
 // Force a division by 0 that gets detected by compilers for anything that isn't a byte array.
 #define AZ_SPAN_FROM_BUFFER(BYTE_BUFFER) \
-  _az_span_init( \
+  az_span_init( \
```
ahsonkhan, line 147: How would this work when calling the non-inlined az_span_init? Does this result in a `.h -> .c` cycle?

antkmsft: Yes - macro gets substituted by a function call.

Issue #686, opened 2020-05-05
Investigate and reconcile differences in how different langauges encode URLs and the C SDK


ahsonkhan:
Currently, we only URL encode (using percent encoding) the following characters, which follows RFC 3986 and only leaving unreserved characters un-encoded https://tools.ietf.org/html/rfc3986#section-2.3:
https://github.com/Azure/azure-sdk-for-c/blob/2b649d3d657e13cc8f9afbc8911afc4a93f65cd0/sdk/core/core/src/az_span.c#L695-L707

Other languages behave differently and we should make sure that either we have specific reasons for the discrepancy, OR we try to mimic what other languages do after figuring out what precedent to follow (since there isn't clear consensus).

**Things to consider:**
- What about following the IRI spec, i.e. RFC 3987 - https://tools.ietf.org/html/rfc3987#section-2.2
- ~Do we want to escape white space (' ') as '%20' or as '+'?~ No

C#: https://dotnetfiddle.net/TH7z7h
```C#
using System;
using System.Text.Encodings.Web;
					
public class Program
{
	public static void Main()
	{
		string query = "Hellö Wörld@Some lang$!*()+;";
		string encodedQuery = UrlEncoder.Default.Encode(query);
		// Hell%C3%B6%20W%C3%B6rld@Some%20lang$!*()%2B;
		Console.WriteLine(encodedQuery);
		
		// Hell%C3%B6%20W%C3%B6rld%40Some%20lang%24%21%2A%28%29%2B%3B
		Console.WriteLine(Uri.EscapeDataString(query));
	}
}
```

https://source.dot.net/#System.Text.Encodings.Web/System/Text/Encodings/Web/UrlEncoder.cs,123
https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapedatastring?view=netcore-3.1
https://docs.microsoft.com/en-us/dotnet/api/system.net.webutility.urlencode?view=netcore-3.1#remarks
https://github.com/Azure/azure-sdk-for-net/pull/11657

Golang: https://play.golang.org/p/BMrXEtoFR_K
```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	query := "Hellö Wörld@Some lang$!*()+;"
	// Hell%C3%B6+W%C3%B6rld%40Some+lang%24%21%2A%28%29%2B%3B
	fmt.Println(url.QueryEscape(query))
}
```

Java: https://repl.it/repls/VapidMealyKeygens
```java
[...]

PR #776, opened 2020-05-19
the DPS registration string for signing should not be escaped

ericwol-msft:


danewalton:
While it might be wrong to have it initially encoded for `az_iot_provisioning_client_sas_get_signature()`, it might be helpful to have it encoded here in [az_iot_provisioning_client_sas_get_password()](https://github.com/Azure/azure-sdk-for-c/blob/056aed5dadcf34a78ded1634acadd059bdd79fae/sdk/iot/provisioning/src/az_iot_provisioning_client_sas.c#L72): 

https://github.com/Azure/azure-sdk-for-c/blob/056aed5dadcf34a78ded1634acadd059bdd79fae/sdk/iot/provisioning/src/az_iot_provisioning_client_sas.c#L127

would it not?

ericwol-msft:
You can't say that it's not right because it works. 

> I'm not convinced this is right. Hub seems to be the other way.

You can't say that it's not right because it works.

danewalton:
> You can't say that it's not right because it works.

While that may be true I only say that because this is the python script which allows me to get a SAS Key which matches what the IoT Hub extension in VSCode generates. It doesn't match if the slashes are not escaped.

https://gist.github.com/danewalton/6013083f5d7f64b7f31de14bfaefafb8

sdk/iot/provisioning/src/az_iot_provisioning_client_sas.c
```diff
@@ -18,7 +18,7 @@
 #define AMPERSAND '&'
 #define EQUAL_SIGN '='
 #define STRING_NULL_TERMINATOR '\0'
-#define SCOPE_REGISTRATIONS_STRING "%2fregistrations%2f"
+#define SCOPE_REGISTRATIONS_STRING "/registrations/"
```
CIPop, line 21: To have a complete implementataion, the `sig`, `se` and `skn` must be encoded using semantics described here: https://docs.microsoft.com/en-us/dotnet/api/system.net.webutility.urlencode?view=netcore-3.1
Also, within the HMAC portion, the target (audience) needs to be encoded as well (this is commented in code but not implemented that way).

(from https://github.com/Azure/azure-iot-sdk-csharp/blob/master/common/src/device/provisioning/transport/ProvisioningSasBuilder.cs#L69)

sdk/iot/provisioning/tests/cmocka/test_az_iot_provisioning_client_sas.c
```diff
@@ -22,7 +22,7 @@
 
 #define TEST_REGISTRATION_ID_STR "myRegistrationId"
 #define TEST_ID_SCOPE "0neFEEDC0DE"
-#define TEST_URL_ENCODED_RESOURCE_URI "0neFEEDC0DE%2fregistrations%2fmyRegistrationId"
+#define TEST_URL_ENCODED_RESOURCE_URI "0neFEEDC0DE/registrations/myRegistrationId"
```
CIPop, line 25: Registration id cannot contain upper/lower characters. (https://docs.microsoft.com/en-us/azure/iot-dps/concepts-device#registration-id)
From past experience, having upper characters (or not allowed) was the top cause of issues (especially as certificates needed to be generated).
To set a good example, we should have UTs working with correct registration names.

sdk/iot/provisioning/src/az_iot_provisioning_client_sas.c
ahsonkhan: @CIPop are you sure you want the semantics described in S.N.WebUtility.UrlEncode? Or do you want S.T.E.W semantics?
https://docs.microsoft.com/en-us/dotnet/api/system.text.encodings.web.urlencoder?view=netcore-3.1

For example, are you sure ' ' should be encoded as a '+'?

I shared some of the differences in this issue along with an open question of what RFC we should support: What are your thoughts on it?

https://github.com/Azure/azure-sdk-for-c/issues/686

CIPop: This is what's used in the C# SDK that works. I am assuming that code is in-line with the code running on the server (didn't check because we had no complaints about the C# code so far).

PR #783, opened 2020-05-20
Switch to a single CI pipeline, instead of one per project

damirault:
Switch to a single CI pipeline, instead of one per project
Each project CI is already building the world, and they are being removed.


weshaggard:
Please ensure we delete the associated pipelines in our devops org as well, assuming you haven't already done so.

PR #788, opened 2020-05-21
Proposal: PLATFORM=(AUTO, WIN32, POSIX, NONE, <custom_lib>), HTTP_TRANSPORT=(NONE, CURL, <custom_lib>)

antkmsft:
Proposal:
* `PLATFORM`=`AUTO` | `WIN32` | `POSIX` | `NONE` | `<custom_lib>`
* `HTTP_TRANSPORT` : `NONE` | `CURL` | `<custom_lib>`

User downloads our repo and builds for Windows, Mac, Linux or FreeBSD (we look at the CMake target platform).
simple command line `cmake . && make` builds SDK with the corresponding `az_win32` or `az_posix`.
It works with minimum knowledge and minimum surprises.
If platform can't be detected (IoT), the implementation that gets picked if you provide no options is  `az_noplatform` - just as it does today.
Also, in the recent related commit we broke the custom PAL implementation scenario. Now, user provide `-DPLATFORM=mylib`, and it also works (no need to provide `-DPLATFORM=USER, -DUSER_PLATFORM_NAME=mylib`).

The default HTTP_TRANSPORT impl is `az_nohttp`, just as today. User can set it to CURL (`-DHTTP_TRANSPORT=CURL`) and we'll build with az_curl, just as today (currently: with `-DTRANSPORT_CURL=ON`).
User can provide their own implementation, the same way as with PAL: `-DHTTP_TRANSPORT=az_beast`.
Users can explicitly provide `NONE`, which is also the default, and it means `az_nohttp`.
Users can provide `AUTO` and then if the target platform is WIN32 or UNIX, we use `az_curl`, and `az_nohttp` otherwise.
I considered making `HTTP_TRANSPORT=AUTO` the default option, but CURL is an external dependency, and making it to build may require additional steps from the customer, and I don't want the default CMake build without any options to ever fail or require additional steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,12 +6,14 @@ cmake_minimum_required (VERSION 3.10)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake-modules")
 
 option(WARNINGS_AS_ERRORS "Treat compiler warnings as errors" ON)
-option(TRANSPORT_CURL "Build internal http transport implementation with CURL for HTTP Pipeline" OFF)
 option(UNIT_TESTING "Build unit test projects" OFF)
 option(UNIT_TESTING_MOCKS "wrap PAL functions with mock implementation for tests" OFF)
 option(TRANSPORT_PAHO "Build IoT Samples with Paho MQTT support" OFF)
 option(PRECONDITIONS "Build SDK with preconditions enabled" ON)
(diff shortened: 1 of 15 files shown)


ahsonkhan:
What are the next steps for this PR, @antkmsft? It is still a draft PR. Is it meant for the June milestone? cc @gilbertw 

I haven't looked at these renames too closely yet, but the intent and end goal is definitely beneficial to the end user. This PR makes it more clearer to me.

PR #790, opened 2020-05-22
Fix IAR warnings for enums

danewalton:
IAR gives warnings for ->

> Warning[Pe188]: enumerated type mixed with another type

So `0` initialized structs trigger the warnings. This explicitly sets the enums to default values.

### DIFF
--- a/sdk/iot/common/inc/az_iot_common.h
+++ b/sdk/iot/common/inc/az_iot_common.h
@@ -35,6 +35,9 @@ enum
  */
 typedef enum
 {
+  // Default, unset value
+  AZ_IOT_STATUS_UNSET = 0,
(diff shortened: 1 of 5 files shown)


sdk/iot/common/inc/az_iot_common.h
```diff
@@ -35,6 +35,9 @@ enum
  */
 typedef enum
 {
+  // Default, unset value
+  AZ_IOT_STATUS_UNSET = 0,
```
CIPop, line 39: I would call this STATUS_UNKNOWN instead.

danewalton: Done

PR #794, opened 2020-05-27
Add log classification for IoT Convenience layer

hihigupt:
Adding log classification for IoT Convenience layer e.g AZRTOS Convenience layer (https://github.com/azure-rtos/azure-iot-preview).

### DIFF
--- a/sdk/core/az_core/inc/az_log.h
+++ b/sdk/core/az_core/inc/az_log.h
@@ -55,6 +55,9 @@ typedef enum
 
   AZ_LOG_IOT_SAS_TOKEN
   = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 2), ///< IoT Client generated new SAS token.
+
+  AZ_LOG_IOT_AZURERTOS
+  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 3), ///< Azure IoT classification for Azure RTOS.
 } az_log_classification;
 
 /**


sdk/core/az_core/inc/az_log.h
```diff
@@ -55,6 +55,9 @@ typedef enum
 
   AZ_LOG_IOT_SAS_TOKEN
   = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 2), ///< IoT Client generated new SAS token.
+
+  AZ_LOG_IOT_CONVENIENCE_LAYER
```
gilbertw, line 59: I don't see a reason to expose this to the user.  Can you elaborate on the scenario

hihigupt: Convenience layer act as a platform specific sdk for Azure IoT. In the AZRTOS, we have AZRTOS convenience layer that depends upon azure-sdk-for-c as a submodule (https://github.com/azure-rtos/azure-iot-preview/tree/master/azure_iot). In this convenience layer we want to use the same infrastructure as sdk-iot-core uses, because from customer point of view it is sdk for IoT. To allow logging from convenience layer, I added this classification under IoT facility.

CIPop: I agree with @gilbertw - due to layering (convenience layers depend on emb SDK) I think the constant should be defined within the respective convenience layer, not within the SDK.
As I did offline, I argue that the convenience layer may want to use a different logging system that is more suited to the platform.

hihigupt: Few things to consider.
1. If I do not constant in sdk-core, I have no way of making sure that classification is unique. That is the sole reason to put this here.
2. If convenience layer has different logging system, that means now for customer we have to expose to mechanism for sdk logs (one for convenience and sdk). 

@JeffreyRichter Any suggestions.

JeffreyRichter: @hihigupt : Actually, each facility code does guarantee uniqueness for results and log classifications. If you simply use the IOT classification, then you can have conflicts within IoT but not won't conflict with other client libraries (like Storage or KeyVault).

If Azure RTOS does already have a logging mechanism, then it does make sense for you to integrate with that and to NOT produce log messages using the Azure Embedded C SDK. However, you could be a CONSUMER of the Azure SDK logs by registering a callback and then forwarding the SDK log messages to the Azure RTOS log mechanism.

As you say in #2, having multiple logging systems is a problem. If RTOS already has a log system, then there are already 2 and you definitely should not introduce a third. If RTOS doesn't have a log system, then you probably are better off using ours so that there is just 1; not 2.

hihigupt: In AZRTOS there is no logging infrastructure. But for convenience layer we created a logging mechanism by mapping our log to printf directly. Now we want to remove that logging mess, and align to sdk logging infrastructure.

CIPop: The SDK trace points should normally be available in production environments (enabled/disabled on request) to allow debugging. I don't think the same mechanism should be used for development-time tracing that should never make its way into production environments (I don't see a problem with those being `printf` as long as they don't interfere too much with application timings - e.g. a blocking write to a slow serial console will likely interfere).

For AZRTOS I thought TraceX can log events for production/performance analysis based on eventID. For a Windows application I would recommend ETW and for Linux something like LTTng.

hihigupt: For embedded system size is strong requirement which might not be true for other systems (Like Linux and Windows you mentioned). Having duplicate code that we can reuse, which does the same thing is going against this requirement. 

Regarding TraceX I will let @TiejunMS/@yuxin-azrtos to comment.

TiejunMS: @yuxin-azrtos, correct me If I was wrong. TraceX is designed for logging system events such as interrupt, thread scheduling and etc.  These events need to be simple enough so timing of system won't be affected. For convenience layer (CL), we need to log lots of literals for further debug purpose. It is not proper to use TraceX. Since log feature is provided by Azure SDK, we would like to use it in CL too. From user's pointer of view, CL and Azure SDK is one product. It makes no sense to define two log systems. So it would be great if we can have a unique classification in Azure SDK or share one with Azure SDK IoT.

hihigupt: @JeffreyRichter @CIPop  any comments?

JeffreyRichter: @gilbertw : Please unblock. Himanshu & I spoke and we think this is the best approach.

CIPop: @hihigupt, @TiejunMS for production-level debugging events need to be simple to avoid overheads. The two systems I mention are commonly used for both perf and bug investigations. On Windows, the ETW traces are collected automatically by Watson when a crash occurs. On large AzRTOS systems this should also be possible. On small systems the device should have the ability to turn things off and on. As I said before.

As you mention, your primary scenario is "Debug" in which case the logs should be compiled out as we do with contracts. Azure SDK logging _should not_ be compiled out and the application should have a way to enable/disable logging as well as sending the logs to the developer investigating the issue.

sdk/core/az_core/inc/az_log.h
```diff
@@ -55,6 +55,9 @@ typedef enum
...
+  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 3), ///< Azure IoT platform specific convenience layer.
```
ahsonkhan, line 60: Can you please share a sample usage code/scenario that requires this log classification, to help me understand the motivation?

sdk/core/az_core/inc/az_log.h
hihigupt: Discussed offline, @JeffreyRichter  is ok with this dependency in azure core.

sdk/core/az_core/inc/az_log.h
hihigupt: This will be used in Azure RTOS convenience layer implementation https://github.com/azure-rtos/azure-iot-preview/blob/master/azure_iot/nx_azure_iot_hub_client.c

sdk/core/az_core/inc/az_log.h
gilbertw: @JeffreyRichter do we want to expose the notion of convenience layers to the customer?  The distinction doesn't make sense as a developer in the track2 SDKs?  As a developer I would expect to turn on logging for "AZ_LOG_IOT".  
As a customer,
  What is the convenience layer?  Are we talking about the convenience paradigm of the old SDK?
Per the previous SDK:
`
Applications may also use the non \_LL\_ API set, which is also known as the "convenience layer."  
`

JeffreyRichter: @hihigupt : Rick & I spoke today and we think the name of this should change to AZ_LOG_IOT_AZURERTOS. This makes it clear to customers that it is specifically for Azure RTOS (and no other RTOS) and also does put the concept of "convenience layer" in the customer's face.

gilbertw, line 59:
```suggestion
  AZ_LOG_IOT_AZURERTOS
```

hihigupt: Ack.

sdk/core/az_core/inc/az_log.h
```diff
@@ -55,6 +55,9 @@ typedef enum
 
   AZ_LOG_IOT_SAS_TOKEN
   = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT, 2), ///< IoT Client generated new SAS token.
+
+  AZ_LOG_IOT_AZURERTOS
```
ahsonkhan, line 59: nit: Should this have an underscore between Azure and RTOS?
```suggestion
  AZ_LOG_IOT_AZURE_RTOS
```

JeffreyRichter: Rick and I decided against the additional underscore. We tried it and it made it look like azure and rtos were not related to each other but this is specifically for "azure rtos"

PR #800, opened 2020-06-02
Temporary measure for the header to point into token's memory

antkmsft:
Pointing to a volatile memory is a bit better category of a bug, than pointing to a random location.
#799

### DIFF
--- a/sdk/core/az_core/src/az_credential_client_secret.c
+++ b/sdk/core/az_core/src/az_credential_client_secret.c
@@ -67,8 +67,13 @@ static AZ_NODISCARD az_result _az_credential_client_secret_apply(
 
   int16_t const token_length = token._internal.token_length;
 
+  // Still bad in terms of cuncurrency, but at least the pointer is not pointing to an unknown memory.
+  // Temporary measure to improve https://github.com/Azure/azure-sdk-for-c/issues/799 a bit.
+  uint8_t* const token_buf_ptr
+      = (uint8_t*)credential->_internal.token_credential._internal.token._internal.token;
+
   AZ_RETURN_IF_FAILED(az_http_request_append_header(
-      ref_request, _az_auth_header_name, az_span_init(token._internal.token, token_length)));
+      ref_request, _az_auth_header_name, az_span_init(token_buf_ptr, token_length)));
 
   return AZ_OK;
 }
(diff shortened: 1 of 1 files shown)


antkmsft:
> do we want to create a new issue to track and don't forget about re-designing this? in terms of thread safety?
I was planning to keep the issue open.

sdk/core/az_core/src/az_credential_client_secret.c
```diff
@@ -67,8 +67,13 @@ static AZ_NODISCARD az_result _az_credential_client_secret_apply(
 
   int16_t const token_length = token._internal.token_length;
 
+  // Still bad in terms of concurrency, but at least the pointer is not pointing to an unknown
+  // memory. Temporary measure to improve https://github.com/Azure/azure-sdk-for-c/issues/799 a bit.
+  uint8_t* const token_buf_ptr
```
ahsonkhan, line 72: Can you please explain to me how this fixes the underlying issue, and maybe expand on the comment on why we still have a concurrency related bug here?

sdk/core/az_core/src/az_credential_client_secret.c
```diff
@@ -67,8 +67,13 @@ static AZ_NODISCARD az_result _az_credential_client_secret_apply(
...
+      = (uint8_t*)(uintptr_t)credential->_internal.token_credential._internal.token._internal.token;
+
   AZ_RETURN_IF_FAILED(az_http_request_append_header(
-      ref_request, _az_auth_header_name, az_span_init(token._internal.token, token_length)));
+      ref_request, _az_auth_header_name, az_span_init(token_buf_ptr, token_length)));
```
ahsonkhan, line 76: Can we add unit tests to capture the bug that is being fixed?

antkmsft: We'll enable live testing, it should catch that - this is how Victor caught it. Unit testing (at least in the compiler config it's built) did not catch this, because stack contents were still the same I assume.

sdk/core/az_core/src/az_credential_client_secret.c
antkmsft: We would get the token to a struct that is allocated on a stack, and would add a pointer to it to the http request. Once we exit, the memory is gone. Now we would initialize it with a pointer to credential's memory which is always a correct pointer, but should you read at the wrong time (a second once an hour, given that multiple threads are sending requests at the same time), you may end up reading inconsistent token - part of the old one, and part of the new one. But the memory location is correct. This is better category of a bug.

antkmsft:
I messed up with my fork's branching, and resubmitting the same changes. @vhvb1989 , could you please approve the new PR? - https://github.com/Azure/azure-sdk-for-c/pull/806 It is the same changes.

PR #808, opened 2020-06-04
Add sas key sample

danewalton:
This adds a SAS key sample for telemetry.
Included in this is a sample implementation of base64 encoding and decoding and HMAC256 using OpenSSL.

### DIFF
--- a/sdk/iot/hub/samples/CMakeLists.txt
+++ b/sdk/iot/hub/samples/CMakeLists.txt
@@ -65,7 +65,7 @@ target_link_libraries(
 
 # Telemetry Sample
 add_executable (
-  paho_iot_hub_telemetry_example 
+  paho_iot_hub_telemetry_example
   ${CMAKE_CURRENT_LIST_DIR}/src/paho_iot_hub_telemetry_example.c
   )
 
@@ -81,6 +81,30 @@ target_link_libraries(paho_iot_hub_telemetry_example PRIVATE ${PAL})
 
 target_link_libraries(
   paho_iot_hub_telemetry_example PRIVATE
+  eclipse-paho-mqtt-c::paho-mqtt3cs-static
+  OpenSSL::SSL
+  OpenSSL::Crypto
+  )
+
+# Telemetry Sample
+add_executable (
+  paho_iot_hub_sas_telemetry_example
+  ${CMAKE_CURRENT_LIST_DIR}/src/paho_iot_hub_sas_telemetry_example.c
+  ${CMAKE_CURRENT_LIST_DIR}/src/sample_base64.c
+  )
+
+target_link_libraries(
+  paho_iot_hub_sas_telemetry_example PRIVATE
+  az_iot_hub
+  az_iot_common
+  az_core
+  )
+
+# PAL
+target_link_libraries(paho_iot_hub_sas_telemetry_example PRIVATE ${PAL})
+
+target_link_libraries(
+  paho_iot_hub_sas_telemetry_example PRIVATE
   eclipse-paho-mqtt-c::paho-mqtt3cs-static
   OpenSSL::SSL
   OpenSSL::Crypto
--- a/sdk/iot/hub/samples/CMakeLists.txt
+++ b/sdk/iot/hub/samples/CMakeLists.txt
@@ -90,7 +90,7 @@ target_link_libraries(
 add_executable (
   paho_iot_hub_sas_telemetry_example
   ${CMAKE_CURRENT_LIST_DIR}/src/paho_iot_hub_sas_telemetry_example.c
-  ${CMAKE_CURRENT_LIST_DIR}/src/sample_base64.c
+  ${CMAKE_CURRENT_LIST_DIR}/src/sample_sas_utility.c
   )
 
 target_link_libraries(
--- a/sdk/iot/hub/samples/README.md
+++ b/sdk/iot/hub/samples/README.md
@@ -56,10 +56,18 @@ For all of these samples, the following environment variables will need to be se
 - `VCPKG_ROOT`: The full path to the VCPKG directory used to generate the triplet.
 - `AZ_IOT_DEVICE_ID`: Your device id.
 - `AZ_IOT_HUB_HOSTNAME`: The hostname of your IoT Hub.
-- `AZ_IOT_DEVICE_X509_CERT_PEM_FILE`: The full path to your device cert (in `.pem` format) concatenated
- with its private key.
 - `AZ_IOT_DEVICE_X509_TRUST_PEM_FILE`: The full path to the trusted server cert (in `.pem` format). This is usually
 not needed on Mac or Linux but might be needed for Windows.
+
+For samples using certificate authentication, the following environment variable needs to be set.
+
+- `AZ_IOT_DEVICE_X509_CERT_PEM_FILE`: The full path to your device cert (in `.pem` format) concatenated
+ with its private key.
+
+For samples using SAS Key authentication, the following environment variables need to be set.
+
+- `AZ_IOT_HUB_DEVICE_SAS_KEY`: The SAS key for your device (called "Primary Key" on the Azure Portal for your device).
+- `AZ_IOT_HUB_DEVICE_SAS_KEY_EXPIRATION`: Expiration (in hours) of the SAS key. If not set, this value will default to **2 hours.**
 
 Once these are set, you MUST compile the SDK with the `TRANSPORT_PAHO` option turned on. For example, on the command
 line, it might look like the following:
@@ -69,13 +77,16 @@ cmake -DTRANSPORT_PAHO=ON ..
 
 After the SDK is built, you are free to run any of the following samples.
 
-### [IoT Hub Telemetry][telemetry_sample]
-Send 5 telemetry messages using the IoT Hub Client.
+### [IoT Hub Telemetry (SAS Key)][telemetry_sample_sas]
+Send 5 telemetry messages using the IoT Hub Client with SAS key authentication.
 
-### [IoT Hub Twin][twin_sample]
+### [IoT Hub Telemetry (Certificates)][telemetry_sample_cert]
+Send 5 telemetry messages using the IoT Hub Client with certificate authentication.
+
+### [IoT Hub Twin (Certificates)][twin_sample]
 Use twin features such as updating reported properties, receiving the twin document, and receiving desired properties using the IoT Hub Client.
 
-### [IoT Hub Methods][methods_sample]
+### [IoT Hub Methods (Certificates)][methods_sample]
 Invoke methods from the cloud. The sample supports a method named "ping"
 which if successful will return back to you a json payload of the following:
 
@@ -85,7 +96,7 @@ which if successful will return back to you a json payload of the following:
 
 On failure, a status of `404` will be returned with an empty JSON payload.
 
-### [IoT Hub C2D][c2d_sample]
+### [IoT Hub C2D (Certificates)][c2d_sample]
 Receive and view incoming C2D messages using the IoT Hub Client.
 
 ## Troubleshooting
@@ -107,7 +118,8 @@ For extensive documentation on Azure IoT Hub, see the [API reference documentati
 [VCPKG_DIRECTIONS]:../../../../README.md#development-environment
 [c2d_sample]: src/paho_iot_hub_c2d_example.c
 [methods_sample]: src/paho_iot_hub_methods_example.c
-[telemetry_sample]: src/paho_iot_hub_telemetry_example.c
+[telemetry_sample_sas]: src/paho_iot_hub_sas_telemetry_example.c
+[telemetry_sample_cert]: src/paho_iot_hub_telemetry_example.c
 [twin_sample]: src/paho_iot_hub_twin_example.c
 [iot_hub_mqtt]: https://docs.microsoft.com/en-us/azure/iot-dps/iot-dps-mqtt-support
 [error_codes]: ../../doc/mqtt_state_machine.md#IoT-Service-Errors
--- a/sdk/iot/hub/samples/README.md
+++ b/sdk/iot/hub/samples/README.md
@@ -67,7 +67,7 @@ For samples using certificate authentication, the following environment variable
 For samples using SAS Key authentication, the following environment variables need to be set.
 
 - `AZ_IOT_HUB_DEVICE_SAS_KEY`: The SAS key for your device (called "Primary Key" on the Azure Portal for your device).
-- `AZ_IOT_HUB_DEVICE_SAS_KEY_EXPIRATION`: Expiration (in hours) of the SAS key. If not set, this value will default to **2 hours.**
+- `AZ_IOT_HUB_DEVICE_SAS_KEY_DURATION`: Expiration (in hours) of the SAS key. If not set, this value will default to **2 hours.**
 
 Once these are set, you MUST compile the SDK with the `TRANSPORT_PAHO` option turned on. For example, on the command
 line, it might look like the following:
--- a/sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
+++ b/sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
@@ -160,16 +160,16 @@ static az_result read_configuration_and_init_client()
       "X509 Trusted PEM Store File", ENV_DEVICE_X509_TRUST_PEM_FILE, "", false, trusted, &trusted));
 
   char iot_hub_sas_key_expiration_char[SAS_TOKEN_EXPIRATION_TIME_DIGITS];
-  az_span iot_hub_sas_expiration = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key_expiration_char);
+  az_span iot_hub_sas_expiration_span = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key_expiration_char);
   AZ_RETURN_IF_FAILED(read_configuration_entry(
       "IoT Hub Device SAS Key Expiration (Hours)",
       ENV_IOT_HUB_SAS_KEY_DURATION,
       "2",
       false,
-      iot_hub_sas_expiration,
-      &iot_hub_sas_expiration));
+      iot_hub_sas_expiration_span,
+      &iot_hub_sas_expiration_span));
 
-  AZ_RETURN_IF_FAILED(az_span_atou32(iot_hub_sas_expiration, &iot_hub_sas_key_expiration));
+  AZ_RETURN_IF_FAILED(az_span_atou32(iot_hub_sas_expiration_span, &iot_hub_sas_key_expiration));
 
   iot_hub_sas_key_span = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key);
   AZ_RETURN_IF_FAILED(read_configuration_entry(
--- a/sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
+++ b/sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
@@ -41,8 +41,8 @@
 // DO NOT MODIFY: IoT Hub SAS Key Environment Variable Name
 #define ENV_IOT_HUB_SAS_KEY "AZ_IOT_HUB_DEVICE_SAS_KEY"
 
-// DO NOT MODIFY: IoT Hub SAS Key Expiration Environment Variable Name (defaults to 2 hours)
-#define ENV_IOT_HUB_SAS_KEY_EXPIRATION "AZ_IOT_HUB_DEVICE_SAS_KEY_EXPIRATION"
+// DO NOT MODIFY: IoT Hub SAS Key Duration Environment Variable Name (defaults to 2 hours)
+#define ENV_IOT_HUB_SAS_KEY_DURATION "AZ_IOT_HUB_DEVICE_SAS_KEY_DURATION"
 
 // DO NOT MODIFY: the path to a PEM file containing the server trusted CA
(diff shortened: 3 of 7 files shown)


sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+  {
+    printf("Could not get the signature for SAS key: return code %d\n", res);
+    return res;
+  }
+
+  // HMAC the signature with the decoded key
+  az_span encoded_span = AZ_SPAN_FROM_BUFFER(sas_signature_hmac_encoded_buf);
+  if (az_failed(
+          res
+          = sample_hmac_encrypt(decoded_key_span, sas_signature_span, encoded_span, &encoded_span)))
```
ewertons, line 246: suggest renaming function to `sample_hmac_sha256_encrypt`, to make it clear to customers what they need to do.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+              sas_expiration,
+              AZ_SPAN_FROM_BUFFER(sas_signature_buf),
+              &sas_signature_span)))
+  {
+    printf("Could not get the signature for SAS key: return code %d\n", res);
+    return res;
+  }
+
+  // HMAC the signature with the decoded key
+  az_span encoded_span = AZ_SPAN_FROM_BUFFER(sas_signature_hmac_encoded_buf);
```
ewertons, line 243: everywhere that says "hmac encoded" should be "hmac sha256 encrypted".

ewertons: plus "encoded_span" is not very representative. 
Sorry Dane, I only have naming comments... :(

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+  if (az_failed(
+          res
+          = sample_hmac_encrypt(decoded_key_span, sas_signature_span, encoded_span, &encoded_span)))
+  {
+    printf("Could not encrypt the signature: return code %d\n", res);
+    return res;
+  }
+
+  // base64 encode the result of the HMAC
+  az_span encoded_span_b64;
```
ewertons, line 253: This should be more like "b64_encoded_hmac_sha256_encrypted_signature"

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+  // Get the resulting password, passing the base64 encoded, HMAC signed bytes
+  size_t sas_key_length;
+  if (az_failed(
+          res = az_iot_hub_client_sas_get_password(
+              &client,
+              encoded_span_b64,
+              sas_expiration,
+              AZ_SPAN_NULL,
+              sas_key_password,
+              sizeof(sas_key_password),
```
ewertons, line 270: So the correct name according to the spec would be "sas_token", not sas_key_password.
We should either call it "mqtt_password" (what was advocated by @CIPop ) or "sas_token".

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+          = az_iot_hub_client_get_user_name(&client, mqtt_username, sizeof(mqtt_username), NULL)))
+
+  {
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    return rc;
+  }
+
+  // This sample uses SAS authentication so the password field is set to the generated password
+  mqtt_connect_options.username = mqtt_username;
+  mqtt_connect_options.password = sas_key_password;
```
ewertons, line 303: Same comment for "sas_key_password". 
In this context it should be called "mqtt_password", as @CIPop has proposed.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
ewertons: Plus it matches the other variable `mqtt_username`, which is an easier mental model for customers to follow (people search for the combination of "username" and "password" when they are scanning code)

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
danewalton: ahhhhhh the return of the full sentence variables and functions

agreed though...it's better to be as clear as possible what the "state" is of the stuff the variable contains.

sdk/iot/hub/samples/CMakeLists.txt
```diff
@@ -86,6 +86,30 @@ target_link_libraries(
...
+  )
+
+target_link_libraries(
+  paho_iot_hub_sas_telemetry_example PRIVATE
+  az_iot_hub
+  az_iot_common
+  az_core
+  )
+
+# PAL
```
ericwol-msft, line 103: is this really a PAL? It's just a platform

sdk/iot/hub/samples/README.md
```diff
@@ -56,11 +56,19 @@ For all of these samples, the following environment variables will need to be se
...
 
+For samples using certificate authentication, the following environment variable needs to be set.
+
+- `AZ_IOT_DEVICE_X509_CERT_PEM_FILE`: The full path to your device cert (in `.pem` format) concatenated
+ with its private key.
+
+For samples using SAS Key authentication, the following environment variables need to be set.
+
+- `AZ_IOT_HUB_DEVICE_SAS_KEY`: The SAS key for your device (called "Primary Key" on the Azure Portal for your device).
+- `AZ_IOT_HUB_DEVICE_SAS_KEY_EXPIRATION`: Expiration (in hours) of the SAS key. If not set, this value will default to **2 hours.**
```
ericwol-msft, line 70: over kill?

danewalton: The expiration?

ericwol-msft: yes, does it really need to be an env var? env var should be used for secrets that you don't want in the code that you my check in my accident

sdk/iot/hub/samples/CMakeLists.txt
danewalton: platform vs (platform abstraction layer). I followed what the other `CMakeLists.txt` do here

sdk/iot/hub/samples/README.md
danewalton: Yea so I originally had it as just a `#define` to set in the file. But after some thinking I think it is better this way for a couple reasons:
1. Consistency with the other things that are used to get these samples working.
2. Having it as an environment variable allows the same value to be used in other samples should we implement them. Users wouldn't have to configure it for each sample.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+static uint32_t iot_hub_sas_key_expiration;
+static char iot_hub_sas_key[128];
+static az_span iot_hub_sas_key_span;
+static char x509_trust_pem_file[256];
+char telemetry_topic[128];
+
+static char mqtt_client_id[128];
+static char mqtt_username[128];
+static char mqtt_endpoint[128];
+static char sas_b64_decoded_key[64];
```
ericwol-msft, line 68: decoded device keys are 32 bytes

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+static char sas_signature_buf[128];
+static char sas_signature_hmac_encoded_buf[128];
+static char sas_signature_encoded_buf_b64[128];
+static char sas_key_password[256];
```
ericwol-msft, line 72: should be mqtt_password on line 68

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
danewalton: done

sdk/iot/hub/samples/README.md
ericwol-msft: at the cost of more knobs to turn to get it working

danewalton: Not exactly. It has a default value of 2 hours so it'll work if the user never touches it. I think it only really adds flexibility. There isn't any additional work the user has to do.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+static az_span mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("ssl://");
+static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+static const char* telemetry_message_payloads[NUMBER_OF_MESSAGES] = {
```
ericwol-msft, line 76: static const char* const telemetry_message_payloads[....

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+  "Message One", "Message Two", "Message Three", "Message Four", "Message Five",
+};
+
+static az_iot_hub_client client;
```
ericwol-msft, line 80: there are 2 clients... hub_client?

danewalton: valid point. I can go through all of the files after this PR and update them (I think they're all named the same thing). Same with the above one on the const

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+static MQTTClient mqtt_client;
+
+static uint32_t get_expiration_time(uint32_t hours)
```
ericwol-msft, line 83: get_epoch_expiration_time?

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+
+// Read the user environment variables used to connect to IoT Hub
+static az_result read_configuration_and_init_client()
+{
+  az_span trusted = AZ_SPAN_FROM_BUFFER(x509_trust_pem_file);
+  AZ_RETURN_IF_FAILED(read_configuration_entry(
+      "X509 Trusted PEM Store File", ENV_DEVICE_X509_TRUST_PEM_FILE, "", false, trusted, &trusted));
+
+  char iot_hub_sas_key_expiration_char[SAS_TOKEN_EXPIRATION_TIME_DIGITS];
+  az_span iot_hub_sas_expiration = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key_expiration_char);
```
ericwol-msft, line 163: if you are going to use _span, then it should be used here as well

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+  AZ_RETURN_IF_FAILED(az_iot_hub_client_init(
+      &client,
+      az_span_slice(iot_hub_hostname_span, 0, (int32_t)strlen(iot_hub_hostname)),
+      az_span_slice(device_id_span, 0, (int32_t)strlen(device_id)),
+      NULL));
+
+  return AZ_OK;
+}
+
+static int get_sas_key()
```
ericwol-msft, line 212: generate_sas_key  ?

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,418 @@
...
+#define ENV_DEVICE_ID "AZ_IOT_DEVICE_ID"
+
+// DO NOT MODIFY: IoT Hub Hostname Environment Variable Name
+#define ENV_IOT_HUB_HOSTNAME "AZ_IOT_HUB_HOSTNAME"
+
+// DO NOT MODIFY: IoT Hub SAS Key Environment Variable Name
+#define ENV_IOT_HUB_SAS_KEY "AZ_IOT_HUB_DEVICE_SAS_KEY"
+
+// DO NOT MODIFY: IoT Hub SAS Key Expiration Environment Variable Name (defaults to 2 hours)
+#define ENV_IOT_HUB_SAS_KEY_EXPIRATION "AZ_IOT_HUB_DEVICE_SAS_KEY_EXPIRATION"
```
ericwol-msft, line 45: ENV_IOT_HUB_SAS_KEY_EXPIRATION   -> ENV_IOT_HUB_SAS_KEY_DURATION

sdk/iot/hub/samples/src/sample_base64.c
```diff
@@ -0,0 +1,129 @@
...
+  {
+    result = AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
+  }
+
+  BIO_free_all(b64_encoder);
+
+  return result;
+}
+
+// HMAC256 an input span with an input key
```
ericwol-msft, line 103: HMAC-SHA256

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,424 @@
...
+              &sas_signature_span)))
+  {
+    printf("Could not get the signature for SAS key: return code %d\n", res);
+    return res;
+  }
+
+  // HMAC-SHA256 encrypt the signature with the decoded key
+  az_span hmac256_encrypted_span = AZ_SPAN_FROM_BUFFER(sas_signature_hmac_encoded_buf);
+  if (az_failed(
+          res = sample_hmac_sha256_encrypt(
```
hihigupt, line 245: >sample_hmac_sha256_encrypt [](start = 16, length = 26)

we are only doing hmac with sha2 hash .... can we remove encrypt? It make you think that you are encrypting sha2 hmac

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,424 @@
...
+              decoded_key_span,
+              sas_signature_span,
+              hmac256_encrypted_span,
```
hihigupt, line 248: >hmac256_encrypted_span [](start = 14, length = 22)

same with this variable.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -0,0 +1,424 @@
...
+
+  {
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    return rc;
+  }
+
+  // This sample uses SAS authentication so the password field is set to the generated password
+  mqtt_connect_options.username = mqtt_username;
+  mqtt_connect_options.password = mqtt_password;
+  printf("SAS Key: %s\n", mqtt_connect_options.password);
```
hihigupt, line 310: > printf("SAS Key: %s\n", mqtt_connect_options.password); [](start = 1, length = 56)

please do not print passwords

danewalton: ah thanks was using for debugging

sdk/iot/hub/samples/src/sample_base64.h
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+// Decode an input span from base64 to raw bytes
+az_result sample_base64_decode(az_span base64_encoded, az_span in_span, az_span* out_span);
+
+// Encode an input span of raw bytes to base64
+az_result sample_base64_encode(az_span bytes, az_span in_span, az_span* out_span);
+
+// HMAC256 an input span with an input key
+az_result sample_hmac_sha256_encrypt(az_span key, az_span bytes, az_span in_span, az_span* out_span);
```
hihigupt, line 19: >sample_hmac_sha256_encrypt [](start = 10, length = 26)

this file is doing more then base64... can we name it something else? Or do we even need it.

danewalton: I can rename it. I would like to keep it as a separate file because we will likely have other samples with SAS key auth and this prevents duplicating code.
How about `sample_sas_gen_helper.h` or something like that.

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
danewalton: how about `sample_hmac_sha256_sign`. Trying to be as descriptive as possible to make it easy for people to understand and port elsewhere.

sdk/iot/hub/samples/src/sample_base64.h
hihigupt: Utility functions.?

---
In reply to: [436076986](https://github.com/Azure/azure-sdk-for-c/pull/808#discussion_r436076986) [](ancestors = 436076986)

danewalton: I'd like to keep sas in there so how about 
`sample_sas_utility.h`

sdk/iot/hub/samples/src/sample_base64.c
```diff
@@ -0,0 +1,129 @@
...
+{
+  az_result result;
+
+  BIO* b64_decoder;
+  BIO* source_mem_bio;
+
+  memset(az_span_ptr(in_span), 0, (size_t)az_span_size(in_span));
+
+  // Create a BIO filter to process the bytes
+  b64_decoder = BIO_new(BIO_f_base64());
```
hihigupt: >b64_decoder [](start = 2, length = 11)

Should check for null.

hihigupt: we are not doing any checks for error, in any API. 

---
In reply to: [436091701](https://github.com/Azure/azure-sdk-for-c/pull/808#discussion_r436091701) [](ancestors = 436091701)

danewalton: updating

sdk/iot/hub/samples/src/paho_iot_hub_sas_telemetry_example.c
hihigupt: sample_hmac_sha256 is good enough

---
In reply to: [436079140](https://github.com/Azure/azure-sdk-for-c/pull/808#discussion_r436079140) [](ancestors = 436079140)

sdk/iot/hub/samples/src/sample_base64.h
hihigupt: One thing to think about is if we only want to share sas functions between samples then I it should be fine, but if we have bigger scope for this file then adding sas_utitlity might not make sesnse?

---
In reply to: [436085582](https://github.com/Azure/azure-sdk-for-c/pull/808#discussion_r436085582) [](ancestors = 436085582)

sdk/iot/hub/samples/src/sample_sas_utility.c
```diff
@@ -0,0 +1,157 @@
...
+  if (b64_decoder == NULL)
+  {
+    return AZ_ERROR_OUT_OF_MEMORY;
+  }
+
+  // Get the source BIO to push through the filter
+  source_mem_bio = BIO_new_mem_buf(az_span_ptr(base64_encoded), (int)az_span_size(base64_encoded));
+  if(source_mem_bio == NULL)
+  {
+    return AZ_ERROR_OUT_OF_MEMORY;
```
hihigupt, line 33: We should not leave leaks on error.

sdk/iot/hub/samples/src/sample_sas_utility.h
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#include "az_result.h"
+#include "az_span.h"
+
+/*
+ * This serves as an example for fundamental functionality needed to use SAS key authentication.
+ * This implementation uses OpenSSL.
```
hihigupt, line 9: this is header file.... Can we make this comment in c file?

sdk/iot/hub/samples/src/sample_base64.h
danewalton: The base64 encoding and decoding should be handled by the user. Same with the hmac256. So here, it is useful in making explicit what functionality would need to be implemented by the user in order to get the SAS functionality to work. The scope should be kept only to these samples since our public API's shouldn't touch/involve themselves with these. 

I believe there is broader effort to consolidate reused sample code into common source files so that we don't have to maintain multiple instances of the same thing. This aligns with that.

sdk/iot/hub/samples/src/sample_sas_utility.h
danewalton: moved

sdk/iot/hub/samples/src/sample_sas_utility.c
danewalton: updated

PR #810, opened 2020-06-04
Turn off clobbered warning for longjmp for UT'S

### LINKED ISSUE #801: SDK tests are broken for GCC 9
Uses of `ASSERT_PRECONDITION_CHECKED` is broken in GCC 9 when `AZ_SPAN_FROM...` macros are used as the function parameters for these checks.

Build output looks like the following:
```txt
 from /home/dwalton/repos/az-two/sdk/core/az_core/test/test_az_url_encode.c:6:
/home/dwalton/repos/az-two/sdk/core/az_core/test/test_az_url_encode.c: In function ‘test_url_encode_preconditions’:
/home/dwalton/repos/az-two/sdk/core/az_core/inc/az_span.h:103:3: error: variable ‘({anonymous})’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
  103 |   { \
      |   ^
/home/dwalton/repos/vcpkg/installed/x64-linux/include/cmocka.h:116:28: note: in definition of macro ‘cast_to_largest_integral_type’
  116 |     ((LargestIntegralType)(value))
      |                            ^~~~~
/home/dwalton/repos/az-two/sdk/core/az_core/test/test_az_url_encode.c:288:7: note: in expansion of macro ‘assert_true’
  288 |       assert_true(az_span_is_content_equal(buffer5, AZ_SPAN_FROM_STR("*****")));
      |       ^~~~~~~~~~~
/home/dwalton/repos/az-two/sdk/core/az_core/inc/az_span.h:122:52: note: in expansion of macro ‘AZ_SPAN_LITERAL_FROM_STR’
  122 | #define AZ_SPAN_FROM_STR(STRING_LITERAL) (az_span) AZ_SPAN_LITERAL_FROM_STR(STRING_LITERAL)
      |                                                    ^~~~~~~~~~~~~~~~~~~~~~~~
/home/dwalton/repos/az-two/sdk/core/az_core/test/test_az_url_encode.c:288:53: note: in expansion of macro ‘AZ_SPAN_FROM_STR’
  288 |       assert_true(az_span_is_content_equal(buffer5, AZ_SPAN_FROM_STR("*****")));
      |                                                     ^~~~~~~~~~~~~~~~
In file included from /home/dwalton/repos/az-two/sdk/core/az_core/test/inc/az_test_precondition.h:7,
                 from /home/dwalton/repos/az-two/sdk/core/az_core/test/test_az_url_encode.c:6:
```

This can be remedied by declaring local versions of those spans and passing those local versions in to the function. Don't currently have the cycles to do this just yet but just wanted to make this known if someone has the time.

danewalton:
Fixes #801 

From GCC docs
> Warn for variables that might be changed by longjmp or vfork. This warning is also enabled by -Wextra.

Full disclosure. There is another way to fix this that is more involved that doesn't suppress the warning. All function calls which have parameters with `AZ_SPAN_` literals as parameters would have to change to have local spans set to those values which are then passed on to the function. 

So this :arrow_down: 
```c
  ASSERT_PRECONDITION_CHECKED(
      az_http_request_append_header(&hrb, AZ_SPAN_FROM_STR(" \t\r"), AZ_SPAN_NULL));
```
becomes :arrow_down: 

```c
  az_span precondition_span = AZ_SPAN_FROM_STR(" \t\r");
  az_span precondition_null_span = AZ_SPAN_NULL;
  ASSERT_PRECONDITION_CHECKED(
      az_http_request_append_header(&hrb, precondition_span, precondition_null_span));
```
However, I think suppressing the warning also works since even if those variables get changed, it doesn't matter since they were valid values at the time that the precondition was evaluated. 

If others don't think that is correct thinking please let me know and I can update all the calls to `_PRECONDITION_CHECKED`.

### DIFF
--- a/cmake-modules/AddTestCMocka.cmake
+++ b/cmake-modules/AddTestCMocka.cmake
@@ -65,6 +65,11 @@ function(ADD_CMOCKA_TEST _TARGET_NAME)
 
     add_executable(${_TARGET_NAME} ${_add_cmocka_test_SOURCES})
 
+    # Suppress clobber warning for longjmp
+    if(CMAKE_C_COMPILER_ID MATHES "GNU")
+      target_compile_options(${_TARGET_NAME} -Wno-clobbered)
+    endif()
+
     if (DEFINED _add_cmocka_test_COMPILE_OPTIONS)
         target_compile_options(${_TARGET_NAME}
             PRIVATE ${_add_cmocka_test_COMPILE_OPTIONS}
--- a/cmake-modules/AddTestCMocka.cmake
+++ b/cmake-modules/AddTestCMocka.cmake
@@ -66,8 +66,8 @@ function(ADD_CMOCKA_TEST _TARGET_NAME)
     add_executable(${_TARGET_NAME} ${_add_cmocka_test_SOURCES})
 
     # Suppress clobber warning for longjmp
-    if(CMAKE_C_COMPILER_ID MATHES "GNU")
-      target_compile_options(${_TARGET_NAME} -Wno-clobbered)
+    if(CMAKE_C_COMPILER_ID MATCHES "GNU")
+      target_compile_options(${_TARGET_NAME} PRIVATE -Wno-clobbered)
     endif()
 
     if (DEFINED _add_cmocka_test_COMPILE_OPTIONS)
--- a/eng/cmake/global_compile_options.txt
+++ b/eng/cmake/global_compile_options.txt
@@ -15,5 +15,5 @@ else()
     set(WARNINGS_AS_ERRORS_FLAG "-Werror")
   endif()
 
-  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal)
+  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal -Wno-clobbered)
 endif()
--- a/eng/cmake/global_compile_options.txt
+++ b/eng/cmake/global_compile_options.txt
@@ -15,5 +15,5 @@ else()
     set(WARNINGS_AS_ERRORS_FLAG "-Werror")
   endif()
 
-  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal -Wno-clobbered)
+  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal)
 endif()


danewalton:
gcc9
Here is successful compilation with the warning turned off since we don't have a gate with GCC 9.

danewalton:
@barcharcraz do you have any opinions on this?

eng/cmake/global_compile_options.txt
```diff
@@ -15,5 +15,5 @@ else()
     set(WARNINGS_AS_ERRORS_FLAG "-Werror")
   endif()
 
-  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal)
+  add_compile_options(-Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wcast-qual -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal -Wno-clobbered)
```
ahsonkhan, line 18: How does GCC's `-Wall` compare to MSVC's `/W4`, that @antkmsft recently changed in https://github.com/Azure/azure-sdk-for-c/pull/816? Similarly, what about clang?

cc @barcharcraz, @BillyONeal 

Would we still need the no-clobbered flag if we changed -Wall to -W4?

ahsonkhan:
Are there any other types of warnings (outside of those related to setjmp/longjmp) that the `-Wno-clobbered` flag hides?

eng/cmake/global_compile_options.txt
antkmsft: Ah, I see. This is for GCC. We did not change (and probably shouldn't) switch from `-Wall` from GCC: `-Wall` for GCC and CLang is the equivalent of `/W4` for MSVC. `/Wall` for MSVC is an equivalent of `-Weverything` for GCC/Clang.

antkmsft:
If you want to avoid repetition/simplify future updates, make an additional CMakeFile in this directory, that is for use specifically by unit test CMakefiles, and include it from each unit test's CMakeFile.

ahsonkhan:
> To be on a safe side, I'd take only the unit tests' CMakeFiles and `add_compile_options(-Wno-clobbered)` there. I strongly suggest that, but the final decision is yours to make.

That's a good idea, if you are up for exploring it @danewalton 
Let us know if it becomes too tedious.

antkmsft:
Other possible option would be to try to add #pragmas for GCC around the AZ_SPAN_FROM_STR macro. See the syntax in _az_cfg.h - you can #ifdef for specifically GCC 9, then push warnings, then disable specifically this warning, then in the end of the macro you #ifdef for GCC 9 again, and pop the warnings.

danewalton:
@ahsonkhan @antkmsft I updated the PR to limit the scope of the suppression. Now it only applies to targets which are made via the `add_cmocka_test` cmake function.

danewalton:
I'll go ahead and merge but let me know if you guys were looking for something different.

cmake-modules/AddTestCMocka.cmake
```diff
@@ -65,6 +65,11 @@ function(ADD_CMOCKA_TEST _TARGET_NAME)
 
     add_executable(${_TARGET_NAME} ${_add_cmocka_test_SOURCES})
 
+    # Suppress clobber warning for longjmp
+    if(CMAKE_C_COMPILER_ID MATCHES "GNU")
```
ahsonkhan, line 69: Nice comment.

For reference:
https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html
> GNU = GNU Compiler Collection (gcc.gnu.org)

cmake-modules/AddTestCMocka.cmake
```diff
@@ -65,6 +65,11 @@ function(ADD_CMOCKA_TEST _TARGET_NAME)
...
+      target_compile_options(${_TARGET_NAME} PRIVATE -Wno-clobbered)
```
ahsonkhan, line 70: What does applying the `PRIVATE` keyword do in this case?

danewalton: With executables I don't think it really matters since it won't be consumed by anything else. But in theory with CMake you can make this public or private which makes them transitive (or not) if another lib/executable consumes them. Since this executable is the end of the road I'm almost positive it doesn't matter.

ahsonkhan:
> Here is successful compilation with the warning turned off since we don't have a gate with GCC 9.

@danewalton - did you test/verify that the warning disappears with the new approach?

I am asking because I am wondering if `MATCHES` is working as expected.
I have seen other places use `STREQUAL`.

Matches seems to be for regex:
https://cmake.org/cmake/help/v3.0/command/if.html
> if(<variable|string> MATCHES regex)
> True if the given string or variable’s value matches the given regular expression.

https://gitlab.kitware.com/cmake/cmake/-/issues/17831

danewalton:
> did you test/verify that the warning disappears with the new approach?

Yup verified on Ubuntu 2004 with GCC 9.something

For the CMake I matched what is here:
https://github.com/Azure/azure-sdk-for-c/blob/master/eng/cmake/global_compile_options.txt

cmake-modules/AddTestCMocka.cmake
```diff
@@ -65,6 +65,11 @@ function(ADD_CMOCKA_TEST _TARGET_NAME)
...
+    endif()
+
     if (DEFINED _add_cmocka_test_COMPILE_OPTIONS)
```
ahsonkhan, line 73: Unrelated to this PR, but @vhvb1989 what does `_add_cmocka_test_COMPILE_OPTIONS` do and used for? I don't see it defined anywhere.

https://github.com/Azure/azure-sdk-for-c/search?q=_add_cmocka_test_COMPILE_OPTIONS&unscoped_q=_add_cmocka_test_COMPILE_OPTIONS

PR #811, opened 2020-06-05
Add function to get user name with model id for pnp

danewalton:
Decision was made to add another function to work with PnP for now. Don't want to mess with the layering of the SDK and muddy it with PnP options right now.

### DIFF
--- a/sdk/iot/hub/src/az_iot_hub_client.c
+++ b/sdk/iot/hub/src/az_iot_hub_client.c
@@ -104,6 +104,7 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
   return AZ_OK;
 }
 
+// TODO: GH #812 (Revisit this temporary function for PnP GA)
 AZ_NODISCARD az_result az_iot_hub_client_get_user_name_with_model_id(
     az_iot_hub_client const* client,
     az_span model_id,
@@ -169,8 +170,11 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name_with_model_id(
     return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
   }
 
-  *out_mqtt_user_name_length
-      = mqtt_user_name_size - (size_t)az_span_size(remainder) - sizeof(null_terminator);
+  if (out_mqtt_user_name_length)
+  {
+    *out_mqtt_user_name_length
+        = mqtt_user_name_size - (size_t)az_span_size(remainder) - sizeof(null_terminator);
+  }
 
   return AZ_OK;
 }
--- a/sdk/iot/hub/src/az_iot_hub_client.c
+++ b/sdk/iot/hub/src/az_iot_hub_client.c
@@ -20,6 +20,7 @@ static const az_span hub_client_param_equals_span = AZ_SPAN_LITERAL_FROM_STR("="
 
 static const az_span hub_digital_twin_model_id = AZ_SPAN_LITERAL_FROM_STR("digital-twin-model-id");
 static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2018-06-30");
+static const az_span hub_service_preview_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2020-05-31-preview");
 static const az_span client_sdk_version
     = AZ_SPAN_LITERAL_FROM_STR("DeviceClientType=c%2F" AZ_SDK_VERSION_STRING);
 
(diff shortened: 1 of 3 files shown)


sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,77 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
   return AZ_OK;
 }
 
+AZ_NODISCARD az_result az_iot_hub_client_get_user_name_with_model_id(
```
jspaith, line 107: Please open a GitHub issue (and put // TODO linking back to GitHub # in the code) calling out need to long-term reconcile this once PnP goes GA.  

The options would be (A) leaving this API as is, but at minimum removing the preview API version stuff or (B) removing this completely and using a PnP SDK instead.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,77 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+  if(az_span_size(remainder) > 0)
+  {
+    remainder = az_span_copy_u8(remainder, null_terminator);
+  }
+  else
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
+  }
+
+  *out_mqtt_user_name_length
```
danewalton, line 172: Needs an if check here for `NULL`

sdk/iot/hub/src/az_iot_hub_client.c
danewalton: Opened #812

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,81 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+  const az_span* const user_agent = &(client->_internal.options.user_agent);
+
+  az_span mqtt_user_name_span
+      = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
+
+  int32_t required_length = az_span_size(client->_internal.iot_hub_hostname)
+      + (int32_t)sizeof(hub_client_forward_slash) + az_span_size(client->_internal.device_id)
+      + az_span_size(hub_service_preview_api_version)
+      + az_span_size(hub_client_param_separator_span) + az_span_size(hub_digital_twin_model_id)
+      + az_span_size(hub_client_param_equals_span);
```
ericwol-msft, line 129: model_id len needs to in here too?

danewalton: No its seems that way. But it is URL encoded later on line 163:
https://github.com/Azure/azure-sdk-for-c/blob/ba1bf8466a774b90c677ff27ef81ea7dd1f7a45b/sdk/iot/hub/src/az_iot_hub_client.c#L163
We don't know the size of the url encoded version so we have to do the validation there.

sdk/iot/hub/tests/test_az_iot_hub_client.c
```diff
@@ -880,6 +934,9 @@ int test_iot_hub_client()
     cmocka_unit_test(test_az_iot_hub_client_get_user_name_small_buffer_fail),
     cmocka_unit_test(test_az_iot_hub_client_get_user_name_user_options_succeed),
     cmocka_unit_test(test_az_iot_hub_client_get_user_name_user_options_small_buffer_fail),
+    cmocka_unit_test(test_az_iot_hub_client_get_user_name_with_model_id_succeed),
```
jspaith, line 937: could you add pre-condition checks and a withUsegAgentTest please?  Even though we think this is short-term code, it may end up in permanent SDK so would like a bit of extra guards.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,77 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+{
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_NOT_NULL(mqtt_user_name);
+  _az_PRECONDITION(mqtt_user_name_size > 0);
+
+  const az_span* const module_id = &(client->_internal.options.module_id);
+  const az_span* const user_agent = &(client->_internal.options.user_agent);
+
+  az_span mqtt_user_name_span
+      = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
```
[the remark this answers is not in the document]
danewalton: just ran the clang formatter on it.

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,77 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+AZ_NODISCARD az_result az_iot_hub_client_get_user_name_with_model_id(
+    az_iot_hub_client const* client,
+    az_span model_id,
+    char* mqtt_user_name,
+    size_t mqtt_user_name_size,
+    size_t* out_mqtt_user_name_length)
+{
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_NOT_NULL(mqtt_user_name);
+  _az_PRECONDITION(mqtt_user_name_size > 0);
```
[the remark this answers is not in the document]
danewalton: Yes on the first no on the second. The length check is optional and only is set if its a valid pointer

sdk/iot/hub/tests/test_az_iot_hub_client.c
danewalton: yea will do

sdk/iot/hub/src/az_iot_hub_client.c
```diff
@@ -101,6 +104,77 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+
+  const az_span* const module_id = &(client->_internal.options.module_id);
+  const az_span* const user_agent = &(client->_internal.options.user_agent);
+
+  az_span mqtt_user_name_span
+      = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
+
+  int32_t required_length = az_span_size(client->_internal.iot_hub_hostname)
+      + (int32_t)sizeof(hub_client_forward_slash) + az_span_size(client->_internal.device_id)
```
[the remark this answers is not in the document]
danewalton: I think they are. The ones that are guaranteed are done first and then optional after that.

sdk/iot/hub/tests/test_az_iot_hub_client.c
danewalton: done. four precondition and an additonal "max options" case

PR #814, opened 2020-06-08
Add documentation link to readme

danewalton:
I think it is important that we have the docs for the repo front and center in the readme. I would argue at the very least before the "get help" section to try and provide people with a resource to solve some of their questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ With this in mind, there are many tenants or principles that we follow in order 
 
 - We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](http://rtos.com)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, sleep, and an HTTP stack. We provide some platform layers, and more will be added over time.
 
+## Documentation
+
+We use [doxygen](https://www.doxygen.nl) to generate documentation for source code. You can find the generated, versioned documentation [here](https://azure.github.io/azure-sdk-for-c).
+
 ## The GitHub Repository
 
 To get help with the SDK:
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ With this in mind, there are many tenants or principles that we follow in order 
 
 - Unlike our other language SDKs, many things (such as composing an HTTP pipeline of policies) are done in source code as opposed to runtime. This reduces code size, improves execution speed and locks-in behavior, reducing the chance of bugs at runtime.
 
-- We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](http://rtos.com)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, sleep, and an HTTP stack. We provide some platform layers, and more will be added over time.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,10 @@ With this in mind, there are many tenants or principles that we follow in order
 
 - We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](http://rtos.com)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, sleep, and an HTTP stack. We provide some platform layers, and more will be added over time.
 
+## Documentation
```
ahsonkhan, line 21: nit: Consider moving this down to right above the `Contributing` section (and right after `SDK Architecture`).

The reason is because I think the GitHub repo and the getting started is more relevant/useful to customers than knowing where the docs link is.

Here's why:
1) On GitHub, at the top of the repo, we already broadcast the docs link
image

2) If already looking at the doxygen docs, via https://azure.github.io/azure-sdk-for-c, mentioning this link again, is redundant since the customer is already on that page.

Up to you though, just a nit.

JeffreyRichter: The new proper link for Azure rtos is this: https://azure.microsoft.com/en-us/services/rtos/

danewalton: @ahsonkhan I might be adding my bias here but it took me an embarrassing amount of time to find that link at the top; I never thought to look there. For almost every repo I use, I usually (for better or worse) scroll right to the Readme and try to find a documentation section. I was hoping to make that front and center and make it really easy to find. That way, people can understand what they are "getting started" with before actually "getting started".

README.md
```diff
@@ -16,7 +16,11 @@ With this in mind, there are many tenants or principles that we follow in order
 
 - Unlike our other language SDKs, many things (such as composing an HTTP pipeline of policies) are done in source code as opposed to runtime. This reduces code size, improves execution speed and locks-in behavior, reducing the chance of bugs at runtime.
 
-- We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](http://rtos.com)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, sleep, and an HTTP stack. We provide some platform layers, and more will be added over time.
+- We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](https://azure.microsoft.com/en-us/services/rtos/)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, sleep, and an HTTP stack. We provide some platform layers, and more will be added over time.
```
ahsonkhan, line 19: I don't think the platform layer (specifically) requires an HTTP stack, if we want to be precise. @JeffreyRichter - do we remove  the mention of HTTP stack from this sentence or leave it?

```suggestion
- We support microcontrollers with no operating system, microcontrollers with a real-time operating system (like [Azure RTOS](https://azure.microsoft.com/en-us/services/rtos/)), Linux, and Windows. Customers can implement their own "platform layer" to use our SDK on devices we don’t support out-of-the-box. The platform layer requires minimal functionality such as a clock, a mutex, and thread sleep. We provide some platform layers, and more will be added over time.
```

danewalton: This seems logical to me

danewalton: For timing reasons I will leave this open but will merge. We can look to update later.

PR #817, opened 2020-06-08
Remove disabling warnings that are no longer needed

antkmsft:


antkmsft:
#564

danewalton:
> #564

I think if you put "Fixes #564" then it will automatically close the issue when this is merged.

PR #822, opened 2020-06-09
Update console output values for sample environment vars

danewalton:
It does little from a debugging perspective to have the strings which are output from the samples mismatch the environment vars which they represent. Updating to have the output to console match the env var names.


danewalton:
> This is great stuff!!! Keep it up!!!

@momuno is my partner in crime for most of this :partying_face:

PR #823, opened 2020-06-10
Add to top level IoT readme

danewalton:
- Added more wording to set user expectations on how to use the SDK
- Linked to the generated doxygen docs.
- Specified which CMake lib targets are generated for users to link to.

### DIFF
--- a/sdk/docs/iot/README.md
+++ b/sdk/docs/iot/README.md
@@ -1,10 +1,10 @@
 # Azure IoT Clients
 
-Azure SDK for Embedded C official IoT client library.
+Azure SDK for Embedded C official IoT client libraries.
 
 ## Getting Started
 
-The IoT Client library is created to facilitate connectivity to the Azure IoT Hub and Azure IoT Provisioning Service alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true library.
+The IoT Client library is created to facilitate connectivity to Azure IoT services alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true library.
 
 From a functional perspective, this means that the user's application code (not the SDK) calls directly to the MQTT stack of their choice. The SDK provides utilities (in the form of functions, default values, etc) which help make the connection and feature set easier. Some examples of those utilities include:
 - Publish topics to which messages can be sent and subscription topics to which users can subscribe for incoming messages.
--- a/sdk/docs/iot/README.md
+++ b/sdk/docs/iot/README.md
@@ -4,7 +4,7 @@ Azure SDK for Embedded C official IoT client libraries.
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


sdk/iot/README.md
```diff
@@ -1,14 +1,32 @@
 # Azure IoT Clients
 
-Azure SDK for Embedded C official IoT Hub client library.
+Azure SDK for Embedded C official IoT client library.
```
CIPop, line 3: IoT client_s_ (plural)?

sdk/iot/README.md
```diff
@@ -1,14 +1,32 @@
 # Azure IoT Clients
 
-Azure SDK for Embedded C official IoT Hub client library.
+Azure SDK for Embedded C official IoT client library.
 
 ## Getting Started
 
-- **Docs**: If you're new to the Azure SDK for Embedded C take a look at the [State Machine diagram](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/iot/doc/mqtt_state_machine.md) that explains the high-level architecture, SDK components, and a clear view of SDK x Application responsibilities. 
+The IoT Client library is created to facilitate connectivity to the Azure IoT Hub and Azure IoT Provisioning Service alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true library.
```
CIPop, line 7: We have many clients: hub, provisioning. We'll likely add more in the future.
Maybe we should be less specific here and just say it's to facilitate devices connecting to Azure IoT services?

CIPop: We should also be careful about protocols: likely we'll use protocols other than MQTT - I'd hate to be the one changing everywhere when that happens.

sdk/iot/README.md
danewalton: how about instead `libraries`

sdk/iot/README.md
danewalton: Yes to the first point, the second: ehhhhhh maybe but then we can do a find-all MQTT. I think for now I'd like to be that specific since right now MQTT is all we support. No use saying "a transport and tls of your choice" and leave them guessing which transports are allowed.

sdk/iot/README.md
```diff
@@ -1,10 +1,10 @@
 # Azure IoT Clients
 
-Azure SDK for Embedded C official IoT client library.
+Azure SDK for Embedded C official IoT client libraries.
 
 ## Getting Started
 
-The IoT Client library is created to facilitate connectivity to the Azure IoT Hub and Azure IoT Provisioning Service alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true library.
+The IoT Client library is created to facilitate connectivity to Azure IoT services alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true library.
```
ericwol-msft, line 7:
```suggestion
The Azure IoT Client library is created to facilitate connectivity to Azure IoT services alongside an MQTT and TLS stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true SDK library.

```

PR #835, opened 2020-06-15
update comment for result makeup

danewalton:
@antkmsft I believe this should update it
It currently says successful results don't involve a bit shifted facility code. Updated to reflect the implementation.

### DIFF
--- a/sdk/inc/azure/core/az_result.h
+++ b/sdk/inc/azure/core/az_result.h
@@ -61,7 +61,7 @@ enum
   } while (0)
 
 /**
- * The type represents error conditions.
+ * The type represents success and error conditions.
  * Bits:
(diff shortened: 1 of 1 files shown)


sdk/inc/azure/core/az_result.h
```diff
@@ -61,14 +61,12 @@ enum
   } while (0)
 
 /**
- * The type represents error conditions.
+ * The type represents success and error conditions.
  * Bits:
  * - 31 Severity (0 - success, 1 - failure).
- * - if failure then
+ * - then
```
antkmsft, line 67: I'd remove "then" line at all, but either way looks good!

danewalton: removing

PR #845, opened 2020-06-17
Update sample file structures to be consistent

momuno:
* function declarations appear first. Organized by type and order.
* parallel commenting
* return rc instead of 0 for MQTTClient_SUCCESS

### DIFF
--- a/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
+++ b/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
@@ -56,11 +56,10 @@ static az_iot_hub_client client;
 static MQTTClient mqtt_client;
 
 //
-// Configuration and connection functions - uses MQTT API
+// Configuration and connection functions
 //
 static az_result read_configuration_and_init_client();
 static az_result read_configuration_entry(
-    const char* name,
     const char* env_name,
     char* default_value,
     bool hide_value,
@@ -82,7 +81,8 @@ int main()
   // Read in the necessary environment variables and initialize the az_iot_hub_client
   if (az_failed(rc = read_configuration_and_init_client()))
   {
-    printf("Failed to read configuration from environment variables, return code %d\n", rc);
+    printf(
+        "Failed to read configuration from environment variables, az_result return code %04x\n", rc);
     return rc;
   }
 
@@ -92,7 +92,7 @@ int main()
           rc = az_iot_hub_client_get_client_id(
               &client, mqtt_client_id, sizeof(mqtt_client_id), &client_id_length)))
   {
-    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    printf("Failed to get MQTT clientId, az_result return code %04x\n", rc);
     return rc;
   }
 
@@ -101,7 +101,7 @@ int main()
            &mqtt_client, mqtt_endpoint, mqtt_client_id, MQTTCLIENT_PERSISTENCE_NONE, NULL))
       != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to create MQTT client, return code %d\n", rc);
+    printf("Failed to create MQTT client, MQTTClient return code %d\n", rc);
     return rc;
   }
 
@@ -109,18 +109,18 @@ int main()
   if ((rc = MQTTClient_setCallbacks(mqtt_client, NULL, NULL, on_received, NULL))
       != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to set MQTT callbacks, return code %d\n", rc);
+    printf("Failed to set MQTT callbacks, MQTTClient return code %d\n", rc);
     return rc;
   }
 
   // Connect to IoT Hub
-  if ((rc = connect_device()) != 0)
+  if ((rc = connect_device()) != MQTTCLIENT_SUCCESS)
   {
     return rc;
   }
 
   // Subscribe to the C2D topic to receive cloud-to-device messages
-  if ((rc = subscribe()) != 0)
+  if ((rc = subscribe()) != MQTTCLIENT_SUCCESS)
   {
     return rc;
   }
@@ -128,16 +128,16 @@ int main()
   printf("Subscribed to topics.\n");
 
   // Wait for any incoming C2D messages
-  printf("Waiting for activity. [Press ENTER to abort]\n");
+  printf("Waiting for activity. [Press any key to abort]\n\n");
   (void)getchar();
 
   // Gracefully disconnect: send the disconnect packet and close the socket
   if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to disconnect MQTT client, return code %d\n", rc);
+    printf("Failed to disconnect MQTT client, MQTTClient return code %d\n", rc);
     return rc;
   }
-  printf("Disconnected.\n");
+  printf("Disconnected\n");
 
   // Clean up and release resources allocated by the mqtt client
   MQTTClient_destroy(&mqtt_client);
@@ -149,30 +149,20 @@ int main()
 static az_result read_configuration_and_init_client()
 {
   az_span cert = AZ_SPAN_FROM_BUFFER(x509_cert_pem_file);
-  AZ_RETURN_IF_FAILED(read_configuration_entry(
-      ENV_DEVICE_X509_CERT_PEM_FILE, ENV_DEVICE_X509_CERT_PEM_FILE, NULL, false, cert, &cert));
+  AZ_RETURN_IF_FAILED(
+      read_configuration_entry(ENV_DEVICE_X509_CERT_PEM_FILE, NULL, false, cert, &cert));
 
   az_span trusted = AZ_SPAN_FROM_BUFFER(x509_trust_pem_file);
-  AZ_RETURN_IF_FAILED(read_configuration_entry(
-      ENV_DEVICE_X509_TRUST_PEM_FILE,
-      ENV_DEVICE_X509_TRUST_PEM_FILE,
-      "",
-      false,
-      trusted,
-      &trusted));
+  AZ_RETURN_IF_FAILED(
+      read_configuration_entry(ENV_DEVICE_X509_TRUST_PEM_FILE, "", false, trusted, &trusted));
 
   az_span device_id_span = AZ_SPAN_FROM_BUFFER(device_id);
-  AZ_RETURN_IF_FAILED(read_configuration_entry(
-      ENV_DEVICE_ID, ENV_DEVICE_ID, "", false, device_id_span, &device_id_span));
+  AZ_RETURN_IF_FAILED(
+      read_configuration_entry(ENV_DEVICE_ID, NULL, false, device_id_span, &device_id_span));
 
   az_span iot_hub_hostname_span = AZ_SPAN_FROM_BUFFER(iot_hub_hostname);
   AZ_RETURN_IF_FAILED(read_configuration_entry(
-      ENV_IOT_HUB_HOSTNAME,
-      ENV_IOT_HUB_HOSTNAME,
-      "",
-      false,
-      iot_hub_hostname_span,
-      &iot_hub_hostname_span));
+      ENV_IOT_HUB_HOSTNAME, NULL, false, iot_hub_hostname_span, &iot_hub_hostname_span));
 
   // Paho requires that the MQTT endpoint be of the form ssl://<HUB ENDPOINT>:8883
   AZ_RETURN_IF_FAILED(
@@ -190,31 +180,27 @@ static az_result read_configuration_and_init_client()
 
 // Read OS environment variables using stdlib function
 static az_result read_configuration_entry(
-    const char* name,
     const char* env_name,
     char* default_value,
     bool hide_value,
     az_span buffer,
     az_span* out_value)
 {
-  printf("%s = ", name);
-  char* env = getenv(env_name);
+  printf("%s = ", env_name);
+  char* env_value = getenv(env_name);
 
-  if (env != NULL)
+  if (env_value == NULL && default_value != NULL)
   {
-    printf("%s\n", hide_value ? "***" : env);
-    az_span env_span = az_span_from_str(env);
+    env_value = default_value;
+  }
+
+  if (env_value != NULL)
+  {
+    printf("%s\n", hide_value ? "***" : env_value);
+    az_span env_span = az_span_from_str(env_value);
     AZ_RETURN_IF_NOT_ENOUGH_SIZE(buffer, az_span_size(env_span));
     az_span_copy(buffer, env_span);
     *out_value = az_span_slice(buffer, 0, az_span_size(env_span));
-  }
-  else if (default_value != NULL)
-  {
-    printf("%s\n", default_value);
-    az_span default_span = az_span_from_str(default_value);
-    AZ_RETURN_IF_NOT_ENOUGH_SIZE(buffer, az_span_size(default_span));
-    az_span_copy(buffer, default_span);
-    *out_value = az_span_slice(buffer, 0, az_span_size(default_span));
   }
   else
   {
@@ -262,7 +248,7 @@ static int connect_device()
           rc
           = az_iot_hub_client_get_user_name(&client, mqtt_username, sizeof(mqtt_username), NULL)))
   {
-    printf("Failed to get MQTT clientId, return code %d\n", rc);
+    printf("Failed to get MQTT clientId, az_result return code %04x\n", rc);
     return rc;
   }
 
@@ -282,11 +268,11 @@ static int connect_device()
   // Connect to IoT Hub
   if ((rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options)) != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to connect, return code %d\n", rc);
+    printf("Failed to connect, MQTTClient return code %d\n", rc);
     return rc;
   }
 
-  return rc;
+  return MQTTCLIENT_SUCCESS;
 }
 
 static int subscribe()
@@ -296,16 +282,17 @@ static int subscribe()
   if ((rc = MQTTClient_subscribe(mqtt_client, AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC, 1))
       != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to subscribe to C2D topic filter, return code %d\n", rc);
+    printf("Failed to subscribe to the C2D topic filter, MQTTClient return code %d\n", rc);
     return rc;
   }
 
-  return rc;
+  return MQTTCLIENT_SUCCESS;
 }
 
 static int on_received(void* context, char* topicName, int topicLen, MQTTClient_message* message)
 {
   (void)context;
+  int rc;
 
   printf("Received a message from service.\n");
   printf("Topic: %s\n", topicName);
@@ -320,14 +307,12 @@ static int on_received(void* context, char* topicName, int topicLen, MQTTClient_
 
   // Parse the incoming message topic and check to make sure it is a C2D message
   az_iot_hub_client_c2d_request c2d_request;
-  if (az_failed(
-          az_iot_hub_client_c2d_parse_received_topic(&client, topic_span, &c2d_request)))
+  if (az_failed(rc = az_iot_hub_client_c2d_parse_received_topic(&client, topic_span, &c2d_request)))
   {
-    printf("Topic is not a C2D message.\n");
+    printf("Topic is not a C2D message, az_result return code %04x\n", rc);
   }
   else
   {
-    printf("Topic is a C2D message.\n");
     if (message->payloadlen)
     {
       printf("Received payload:\n%.*s\n", message->payloadlen, (char*)message->payload);
--- a/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
+++ b/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
@@ -82,7 +82,8 @@ int main()
   if (az_failed(rc = read_configuration_and_init_client()))
   {
     printf(
-        "Failed to read configuration from environment variables, az_result return code %04x\n", rc);
+        "Failed to read configuration from environment variables, az_result return code %04x\n",
+        rc);
     return rc;
   }
 
--- a/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
+++ b/sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
@@ -15,9 +15,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include <azure/iot/az_iot_hub_client.h>
 #include <azure/core/az_result.h>
 #include <azure/core/az_span.h>
+#include <azure/iot/az_iot_hub_client.h>
 
 #ifdef _MSC_VER
 // "'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead."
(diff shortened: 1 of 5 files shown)


momuno:
These changes reorganize the files, so may be difficult to follow using GitHub code viewer.

momuno:
Testing of individual samples has not been completed yet.  Will not merge until that is validated.

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -55,60 +55,94 @@ static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
...
+  {
+    printf("Failed to disconnect MQTT client, return code %d\n", rc);
+    return rc;
+  }
+  printf("Disconnected.\n");
+
+  // Clean up and release resources allocated by the mqtt client
+  MQTTClient_destroy(&mqtt_client);
+
+  return 0;
```
gilbertw, line 145:
```suggestion
  return AZ_OK;
```

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -55,60 +55,94 @@ static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
...
+  // Subscribe to the C2D topic to receive cloud-to-device messages
+  if ((rc = subscribe()) != 0)
+  {
+    return rc;
+  }
+
+  printf("Subscribed to topics.\n");
+
+  // Wait for any incoming C2D messages
+  printf("Waiting for activity. [Press ENTER to abort]\n");
```
ericwol-msft, line 131: This is really "any key", not just "enter"

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
ericwol-msft: this is not an AZ_RESULT function. It's a "int main()" function

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -149,35 +188,62 @@ static az_result read_configuration_and_init_client()
   return AZ_OK;
 }
 
-static int on_received(void* context, char* topicName, int topicLen, MQTTClient_message* message)
+// Read OS environment variables using stdlib function
+static az_result read_configuration_entry(
+    const char* name,
```
ericwol-msft, line 193: do we really need "name" and "env_name"? Quick search shows these are always the same.

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -149,35 +188,62 @@ static az_result read_configuration_and_init_client()
...
+    const char* env_name,
+    char* default_value,
+    bool hide_value,
+    az_span buffer,
+    az_span* out_value)
 {
-  (void)context;
+  printf("%s = ", name);
+  char* env = getenv(env_name);
```
ericwol-msft, line 201:
```suggestion
  char* env_value = getenv(env_name);
```

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -149,35 +188,62 @@ static az_result read_configuration_and_init_client()
...
-    topicLen = (int)strlen(topicName);
+    printf("%s\n", hide_value ? "***" : env);
+    az_span env_span = az_span_from_str(env);
+    AZ_RETURN_IF_NOT_ENOUGH_SIZE(buffer, az_span_size(env_span));
+    az_span_copy(buffer, env_span);
+    *out_value = az_span_slice(buffer, 0, az_span_size(env_span));
+  }
+  else if (default_value != NULL)
+  {
+    printf("%s\n", default_value);
```
ericwol-msft, line 213: this looks like dup code

if (env == NULL && default_value != NULL)
{
   env = default_value;
}

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -149,35 +188,62 @@ static az_result read_configuration_and_init_client()
...
 
-  // Parse the incoming message topic and check to make sure it is a c2d message
-  az_iot_hub_client_c2d_request c2d_request;
-  if (az_succeeded(az_iot_hub_client_c2d_parse_received_topic(
-          &client, az_span_init((uint8_t*)topicName, topicLen), &c2d_request)))
+  return AZ_OK;
+}
+
+// Create mqtt endpoint e.g: ssl//contoso.azure-devices.net:8883
+static az_result create_mqtt_endpoint(char* destination, int32_t destination_size, az_span iot_hub)
```
ericwol-msft, line 229: hummm?

```suggestion
static az_result get_mqtt_endpoint(char* destination, int32_t destination_size, az_span iot_hub)
```

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -230,79 +296,47 @@ static int subscribe()
   if ((rc = MQTTClient_subscribe(mqtt_client, AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC, 1))
       != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to subscribe, return code %d\n", rc);
+    printf("Failed to subscribe to C2D topic filter, return code %d\n", rc);
```
ericwol-msft, line 299:
```suggestion
    printf("Failed to subscribe to the C2D topic filter, return code %d\n", rc);
```

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -230,79 +296,47 @@ static int subscribe()
...
-  printf("Waiting for activity. [Press ENTER to abort]\n");
-  (void)getchar();
-
-  // Gracefully disconnect: send the disconnect packet and close the socket
-  if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
+  else
   {
-    printf("Failed to disconnect MQTT client, return code %d\n", rc);
-    return rc;
+    printf("Topic is a C2D message.\n");
```
ericwol-msft, line 330: delete, too verbose

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
-    printf("Failed to set MQTT callbacks, return code %d\n", rc);
-    return rc;
+    if (az_span_is_content_equal(ping_method_name_span, method_request.name))
+    {
+      // Invoke Method
+      az_span response = ping_method();
+
+      // Send a response
+      int rc;
+      if ((rc = send_method_response(&method_request, 200, response)) != 0)
```
ericwol-msft, line 350: 200 is defined in az_iot_common.h

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
+      {
+        printf("Unable to send %d response, status %d\n", 200, rc);
+      }
+    }
+    else
+    {
+      // Unsupported Method
+      int rc;
+      if ((rc = send_method_response(&method_request, 404, ping_method_fail_response)) != 0)
```
ericwol-msft, line 359: same, 404 is defined in az_iot_common.h

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
+      // Send a response
+      int rc;
+      if ((rc = send_method_response(&method_request, 200, response)) != 0)
+      {
+        printf("Unable to send %d response, status %d\n", 200, rc);
+      }
+    }
+    else
+    {
+      // Unsupported Method
```
ericwol-msft, line 357: printf("Method %.*s not found\n", az_span_len(method_request.name), az_span_ptr(method_request.name));

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
-  if ((rc = connect_device()) != 0)
+  MQTTClient_freeMessage(&message);
+  MQTTClient_free(topicName);
+
+  return 1;
+}
+
+static az_span ping_method(void)
+{
+  // Nothing to do. Good to go.
```
ericwol-msft, line 374: printf("PING!\n");

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
+  if (az_failed(az_iot_hub_client_methods_response_get_publish_topic(
+          &client,
+          request->request_id,
+          status,
+          methods_response_topic,
+          sizeof(methods_response_topic),
+          NULL)))
   {
-    return rc;
+    printf("Unable to get twin document publish topic");
```
ericwol-msft, line 392:
```suggestion
    printf("Unable to get method response publish topic");
```

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -308,79 +307,117 @@ static int subscribe()
...
   }
 
-  // Subscribe to the methods topic to receive method invocations
-  if ((rc = subscribe()) != 0)
+  printf("Status: %u\tPayload:", status);
+  char* payload_char = (char*)az_span_ptr(response);
+  for (int32_t i = 0; i < az_span_size(response); i++)
   {
-    return rc;
+    putchar(*(payload_char + i));
```
ericwol-msft, line 400:
```suggestion
    putchar(*payload_char++);
```

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -82,89 +82,109 @@ static const char* telemetry_message_payloads[NUMBER_OF_MESSAGES] = {
...
+  if (az_failed(
+          rc = az_iot_hub_client_get_client_id(
+              &client, mqtt_client_id, sizeof(mqtt_client_id), &client_id_length)))
   {
-    printf("%s\n", default_value);
-    az_span default_span = az_span_from_str(default_value);
-    AZ_RETURN_IF_NOT_ENOUGH_SIZE(buffer, az_span_size(default_span));
-    az_span_copy(buffer, default_span);
-    *out_value = az_span_slice(buffer, 0, az_span_size(default_span));
+    printf("Failed to get MQTT clientId, return code %d\n", rc);
```
ericwol-msft, line 124:
```suggestion
    printf("Failed to get MQTT client id, return code %d\n", rc);
```

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -82,89 +82,109 @@ static const char* telemetry_message_payloads[NUMBER_OF_MESSAGES] = {
...
+  (void)getchar();
 
-  return AZ_OK;
+  // Gracefully disconnect: send the disconnect packet and close the socket
+  if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
+  {
+    printf("Failed to disconnect MQTT client, return code %d\n", rc);
+    return rc;
+  }
+  printf("Disconnected.\n");
```
ericwol-msft, line 164: some messages has a trailing period, some do not. Most do not.

```suggestion
  printf("Disconnected\n");
```

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -283,6 +358,11 @@ static int generate_sas_key()
   return AZ_OK;
 }
 
+static uint32_t get_epoch_expiration_time(uint32_t hours)
```
ericwol-msft, line 361:
```suggestion
static uint32_t get_epoch_expiration_time_from_hours(uint32_t hours)
```

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -357,67 +436,11 @@ static int send_telemetry_messages()
   return rc;
 }
 
-int main()
+static void sleep_seconds(uint32_t seconds)
```
ericwol-msft, line 439:
```suggestion
static void sleep_for_seconds(uint32_t seconds)
```

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
momuno: Also, AZ_OK is not 0.

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
momuno: nope!

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
momuno: almost.  one uses env_value, while the other uses default_value.  But it's unclear to me why sometimes the default value passed in is NULL and other times it's an empty string literal. ... We don't have a legit default value in either case, so this param doesn't seem useful.  If the environment val is NULL, I think we should just make the user set it.

danewalton: In some cases the default value can be useful as a string. DPS is a good example. There is a default endpoint for DPS that the user can overwrite if they would like. But we pass the default one in case they don't care to change it.

In the case that we require someone to supply something, the default value should be `NULL`. If it is optional then an empty string literal. If I am reading correctly then the `device_id` and `hostname` should be changed to `NULL` and not an empty string. The device cert should remain a `NULL` and the trusted cert should remain an empty string.

That's another thing on the list of things to do is to consolidate the reused code like "getting environment variables". We should look to merge these into a single `.c` since we have five similar versions which could grow out of sync.

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -61,62 +61,99 @@ static az_span ping_method_fail_response = AZ_SPAN_LITERAL_FROM_STR("{}");
...
     bool hide_value,
     az_span buffer,
-    az_span* out_value)
+    az_span* out_value);
+static az_result create_mqtt_endpoint(char* destination, int32_t destination_size, az_span iot_hub);
+static int connect_device();
+static int subscribe();
+
+//
+// Messaging functions
```
danewalton, line 80: Yea same here like a previous comment. This does use MQTT API so we might need to be more specific here.

I think for now, it might be enough to remove the "uses MQTT API" part until we structure this a little better (to distinguish between MQTT stack functionality and IoT Hub client stuff.

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
danewalton, line 131: After running this I have a nit:
```suggestion
  printf("Waiting for activity. [Press ENTER to abort]\n\n");
```

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
```diff
@@ -61,62 +61,99 @@ static az_span ping_method_fail_response = AZ_SPAN_LITERAL_FROM_STR("{}");
...
-  az_span destination_span = az_span_init((uint8_t*)destination, destination_size);
-  az_span remainder = az_span_copy(destination_span, mqtt_url_prefix);
-  remainder = az_span_copy(remainder, az_span_slice(iot_hub, 0, iot_hub_length));
-  remainder = az_span_copy(remainder, mqtt_url_suffix);
-  az_span_copy_u8(remainder, null_terminator);
+  printf("Subscribed to topics.\n");
 
-  return AZ_OK;
+  // Wait for any incoming method invocations
+  printf("Waiting for activity. [Press ENTER to abort]\n");
```
danewalton, line 142:
```suggestion
  printf("Waiting for activity. [Press ENTER to abort]\n\n");
```

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -175,7 +195,7 @@ static az_result read_configuration_and_init_client()
 
   iot_hub_sas_key_span = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key);
   AZ_RETURN_IF_FAILED(read_configuration_entry(
-      "IoT Hub Device SAS Key",
+      ENV_IOT_HUB_SAS_KEY,
```
danewalton, line 198: :+1:

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -82,89 +82,109 @@ static const char* telemetry_message_payloads[NUMBER_OF_MESSAGES] = {
...
+      "",
+      false,
+      trusted,
+      &trusted));
 
   char iot_hub_sas_key_expiration_char[SAS_TOKEN_EXPIRATION_TIME_DIGITS];
   az_span iot_hub_sas_expiration_span = AZ_SPAN_FROM_BUFFER(iot_hub_sas_key_expiration_char);
   AZ_RETURN_IF_FAILED(read_configuration_entry(
-      "IoT Hub Device SAS Key Expiration (Hours)",
+      ENV_IOT_HUB_SAS_KEY_DURATION,
```
danewalton, line 187: :+1:

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
momuno: I view this as creating more than getting, in that its constructing the endpoint rather than retrieving something stored in memory.

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
momuno: yay, i already started changing the device_id and hostname to NULL before I saw your comment!

momuno: Yes, there is a lot of duplicated code between the samples.  Perfect time for a class structure... if this were C++. I will settle for a C solution :)

ericwol-msft: My issue was not dup code across samples, but dup code *in* the samples

sdk/samples/iot/hub/src/paho_iot_hub_c2d_example.c
```diff
@@ -55,60 +55,94 @@ static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
...
     bool hide_value,
     az_span buffer,
-    az_span* out_value)
+    az_span* out_value);
+static az_result create_mqtt_endpoint(char* destination, int32_t destination_size, az_span iot_hub);
+static int connect_device();
+static int subscribe();
+
+//
+// Messaging functions
```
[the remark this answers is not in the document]
momuno: Yes, I think a cleaner interface would be helpful for the user.  Let's add to the task list for sample improvements.

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
momuno: haha already had this change made for my next push :D yas! 💯

momuno: nm last comment. i got myself confused.

sdk/samples/iot/hub/src/paho_iot_hub_methods_example.c
momuno: sounds like a plan

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -69,13 +69,7 @@ static MQTTClient mqtt_client;
...
 //
 static az_result read_configuration_and_init_client();
-static az_result read_configuration_entry(
-    const char* name,
-    const char* env_name,
-    char* default_value,
-    bool hide_value,
-    az_span buffer,
-    az_span* out_value);
+static az_result read_configuration_entry(const char* env_name, char* default_value, bool hide_value, az_span buffer, az_span* out_value);
```
danewalton: This doesn't seem clang-formatted. Could you run the formatter on this file?

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -190,30 +184,16 @@ int main()
...
-      "",
-      false,
-      trusted,
-      &trusted));
+  AZ_RETURN_IF_FAILED(read_configuration_entry(ENV_DEVICE_X509_TRUST_PEM_FILE, "", false, trusted, &trusted));
 
   az_span device_id_span = AZ_SPAN_FROM_BUFFER(device_id);
-  AZ_RETURN_IF_FAILED(read_configuration_entry(
-      ENV_DEVICE_ID, ENV_DEVICE_ID, "", false, device_id_span, &device_id_span));
+  AZ_RETURN_IF_FAILED(read_configuration_entry(ENV_DEVICE_ID, "", false, device_id_span, &device_id_span));
```
danewalton: This and the iot_hub one below should have `NULL` passed instead of the empty string.

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -231,20 +211,19 @@ static az_result read_configuration_and_init_client()
 
 // Read OS environment variables using stdlib function
 static az_result read_configuration_entry(
-    const char* name,
```
danewalton, line 190: good change

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -231,20 +211,19 @@ static az_result read_configuration_and_init_client()
...
     const char* env_name,
     char* default_value,
     bool hide_value,
     az_span buffer,
     az_span* out_value)
 {
-  printf("%s = ", name);
-  char* env = getenv(env_name);
+  printf("%s = ", env_name);
+  char* env_var = getenv(env_name);
```
danewalton: nit: I would call this `env_value`. `var` still seems like it could be the name or value.

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -373,14 +360,13 @@ static int on_received(void* context, char* topicName, int topicLen, MQTTClient_
 
   // Parse the incoming message topic and check to make sure it is a twin message
   az_iot_hub_client_twin_response twin_response;
-  if (az_failed(az_iot_hub_client_twin_parse_received_topic(&client, topic_span, &twin_response)))
+  if (az_failed(
+          rc = az_iot_hub_client_twin_parse_received_topic(&client, topic_span, &twin_response)))
   {
-    printf("Topic is not a twin message.\n");
+    printf("Topic is not a twin message, az_result return code %04x\n", rc);
```
danewalton, line 366: `%04x`
Is this just for readability's sake?

momuno: Yes.  
So for the error AZ_ERROR_ITEM_NOT_FOUND, the print out is 8001004 (which follows the logic in az_result.h) instead of -2147418108 which is cryptic at face value.

PR #850, opened 2020-06-19
fix typo

danewalton:


ahsonkhan:
How did you come across these typos? Just by chance?

danewalton:
> How did you come across these typos? Just by chance?

I was checking the file for something and I have the spell check VSCode extension so it stood out. Definitely not a spidey-sense for misspelled words haha.

ahsonkhan:
> Definitely not a spidey-sense for misspelled words haha.

Someone with such a spidey-sense would say that. Is there a way to use this superpower you have across the whole repo? Maybe we can run the generated doxygen docs under a spell checker.

danewalton:
> Someone with such a spidey-sense would say that. Is there a way to use this superpower you have across the whole repo? Maybe we can run the generated doxygen docs under a spell checker.

Possibly. It’s hard because it will give a lot of false positives because we have a lot of acronyms and jargon. But it might be possible.

PR #852, opened 2020-06-23
update platform file names

danewalton:
Preliminary smaller work on updating the PAL layer to make it a little clearer. Open to suggestions for the name which I have here as `stub`. I remember @JeffreyRichter didn't like `none` so `stub` is my next suggestion. Other options might be `default` (which I don't necessarily like since it connotes that it is a default working version) or `basic`, `bare`, `empty`.


CMakeLists.txt
```diff
@@ -61,13 +61,13 @@ file(WRITE ${CMAKE_BINARY_DIR}/coverage_targets.txt "")
...
-  set(PAL az_win32)
+  set(PAL az_platform_win32)
 elseif(AZ_PLATFORM_IMPL STREQUAL "POSIX")
   # build linux platform
-  set(PAL az_posix)
+  set(PAL az_platform_posix)
 else()
   #noplatform
-  set(PAL az_noplatform)
+  set(PAL az_platform_stub)
```
CIPop, line 70: There should be some form of warning or error to inform the users that there are actions required here.

README.md
```diff
@@ -134,10 +134,10 @@ By default, when building the project with no options, the following static libr
     -  iot_provisioning, iot_hub, etc.
   - az_storage_blobs
     -  Storage SDK blobs client.
-  - az_noplatform
+  - az_platform_stub
     - Library that provides a basic returning error for platform abstraction as AZ_NOT_IMPLEMENTED. This ensures the project can be compiled without the need to provide any specific platform implementation. This is useful if you want to use az_core without platform specific functions like `mutex` or `time`.
```
CIPop, line 138: I think returning AZ_NOT_IMPLEMENTED at runtime is not a good idea, especially for boards where it may be hard to debug and find this. See above for one idea of breaking the build if the stub is included. 

I also think we should avoid PALs unless they are used (e.g. IoT should have no dependency on the PAL layer).

danewalton: Both of these points I think are part of a larger discussion which I am aggregating information for in parallel.

> (e.g. IoT should have no dependency on the PAL layer)

Definitely something we will be discussing. Not for this PR though.

CMakeLists.txt
danewalton: I think it should build and be usable but I can output a message on CMake that lets the user know it is only using stubs?

README.md
gilbertw: I prefer none.  What was Jeffrey's reason against none?

danewalton: paging @JeffreyRichter :)

CMakeLists.txt
CIPop: Last I've tried it two weeks ago (drag&drop the entire folder struct) it didn't work because of preconditions requiring the time PAL implemented.

I do agree with you: it should just work (and no warning in that case).

danewalton: Yes we will talk about the specifics of what changes we might want. I'm looking to send an email today and have a short meeting with our team tomorrow to formalize/discuss what we might want to change.

README.md
danewalton: Could also be `az_platform_no_op`

danewalton:
other PR's are tackling this

PR #853, opened 2020-06-24
make all links in readme's to other readme's relative

danewalton:
Good suggestion from @CIPop 
All links to README's should be relative so that if you are working on a branch they resolve to your branch's version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ With this in mind, there are many tenets or principles that we follow in order t
   - [Table of Contents](#table-of-contents)
   - [Documentation](#documentation)
   - [The GitHub Repository](#the-github-repository)
-    - [Services](#service)
+    - [Services](#services)
     - [Structure](#structure)
     - [Master Branch](#master-branch)
     - [Release Branches and Release Tagging](#release-branches-and-release-tagging)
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -462,8 +462,8 @@ For more information see the [Code of Conduct FAQ](https://opensource.microsoft.
 Many people all over the world have helped make this project better.  You'll want to check out:
 
 * [What are some good first issues for new contributors to the repo?](https://github.com/azure/azure-sdk-for-c/issues?q=is%3Aopen+is%3Aissue+label%3A%22up+for+grabs%22)
-* [How to build and test your change](https://github.com/Azure/azure-sdk-for-c/blob/master/CONTRIBUTING.md#developer-guide)
-* [How you can make a change happen!](https://github.com/Azure/azure-sdk-for-c/blob/master/CONTRIBUTING.md#pull-requests)
+* [How to build and test your change](./CONTRIBUTING.md#developer-guide)
```
ahsonkhan, line 465: The unfortunate side effect of such a change is that more links behave unexpectedly in doxygen.

See this issue which might suggest we go in the opposite direction and make all links absolute:
https://github.com/Azure/azure-sdk-for-c/issues/746

I understand that not having links work in a particular branch is a contributor pain point, but broken links hurts the customer experience. If we can resolve both issues, great, but if not, I would focus on improving user experience and readability of docs.

danewalton: Replied in the GH issue

PR #854, opened 2020-06-24
Update DPS example

momuno:
* Update to match other samples' structures
* Fix memory leaks in get_operation_status. Memory was not freed when return early due to failure.

### DIFF
--- a/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
+++ b/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
@@ -22,9 +22,9 @@
 #include <unistd.h>
 #endif
 
-#include <azure/iot/az_iot_provisioning_client.h>
 #include <azure/core/az_result.h>
 #include <azure/core/az_span.h>
+#include <azure/iot/az_iot_provisioning_client.h>
 
 #ifdef _MSC_VER
 // "'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead."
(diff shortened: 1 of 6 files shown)


sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -373,80 +445,28 @@ static int get_operation_status()
...
-  printf("Disconnected.\n");
 
-  // Clean up and release resources allocated by the mqtt client
-  MQTTClient_destroy(&mqtt_client);
-
-  return 0;
-}
+  putchar('\n');
+  return;
+}
```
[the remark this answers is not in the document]
danewalton: [Interesting reason why](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline). I think this is causing the Mac failure.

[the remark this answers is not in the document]
momuno: Oh interesting.  Never realized that!

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -63,84 +63,127 @@ static char x509_trust_pem_file[256] = { 0 };
...
+    return rc;
   }
 
-  return AZ_OK;
+  // Subscribe to the necessary provisioning topic to receive provisioning responses
+  if ((rc = subscribe()) != MQTTCLIENT_SUCCESS)
+  {
+    return rc;
+  }
+
```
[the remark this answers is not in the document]
momuno: I did.  Also with "Connect to."  None of the other samples that called the subscribe and connect_device functions have these printfs.  If we do want them here, I would think we should also include them in the other samples.  I can def put them back in if it is preferred. 

  printf("Connected to %s.\n", global_provisioning_endpoint);

  // Subscribe to the necessary provisioning topic to receive provisioning responses
  if ((rc = subscribe()) != 0)
  {
    return rc;
  }

  printf("Subscribed.\n");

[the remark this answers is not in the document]
danewalton: Not a blocker just curious. Might be useful in the future to have these print out those stages so people can see the flow.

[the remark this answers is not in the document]
momuno: Makes sense.  I'll leave them in.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -63,84 +63,129 @@ static char x509_trust_pem_file[256] = { 0 };
...
+    printf(
+        "Failed to read configuration from environment variables, az_result return code %04x\n",
+        rc);
+    return rc;
   }
-  else if (default_value != NULL)
+
+  // Get the MQTT client id used for the MQTT connection
+  if (az_failed(
+          rc = az_iot_provisioning_client_get_client_id(
```
CIPop, line 102: Nit: formatting - would az_failed and the func call fit on a single line?

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -63,84 +63,129 @@ static char x509_trust_pem_file[256] = { 0 };
...
+    return rc;
   }
 
-  return AZ_OK;
+  // Connect to the Provisioning Service
+  if ((rc = connect_device()) != MQTTCLIENT_SUCCESS)
+  {
+    return rc;
+  }
+  printf("Connected to %s\n", global_provisioning_endpoint);
```
CIPop, line 127: Nit: NL after closing } (here and everywhere else).

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -150,6 +195,39 @@ static az_result read_configuration_and_init_client()
   return AZ_OK;
 }
 
+// Read OS environment variables using stdlib function
+static az_result read_configuration_entry(
```
CIPop, line 199: I think we need to decide on extracting this as common code for all samples at some time.
/cc @ahsonkhan @gilbertw I think you were considering this before: was there any progress on adding a common way for samples to read from environment?

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -186,11 +264,11 @@ static int connect_device()
   // Connect to the Provisioning Service
   if ((rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options)) != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to connect, return code %d\n", rc);
+    printf("Failed to connect, MQTTClient return code %d\n", rc);
     return rc;
   }
 
-  return 0;
+  return MQTTCLIENT_SUCCESS;
```
CIPop, line 271: I don't think our code should return Paho return codes. We could return something new (specific for samples) instead.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -233,23 +311,11 @@ static int register_device()
...
-static void print_az_span(az_span span)
-{
-  char* buffer = (char*)az_span_ptr(span);
-  for (int32_t i = 0; i < az_span_size(span); i++)
-  {
-    putchar(*buffer++);
-  }
-
-  putchar('\n');
+  return MQTTCLIENT_SUCCESS;
```
CIPop, line 318: Same here and everywhere else: our samples should not return Paho codes.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -268,48 +334,52 @@ static int get_operation_status()
...
             rc = az_iot_provisioning_client_parse_received_topic_and_payload(
-                &provisioning_client,
-                az_span_init((uint8_t*)topic, topic_len),
-                az_span_init((uint8_t*)message->payload, message->payloadlen),
-                &response)))
+                &provisioning_client, topic_span, message_span, &response)))
     {
-      printf("Message from unknown topic, return code %d\n", rc);
+      printf("Message from unknown topic, az_result return code %0x4\n", rc);
+      MQTTClient_freeMessage(&message);
```
CIPop, line 361: Instead of having the free everywhere, wrap the processing function in a separate function. That should allow less duplicated code.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -63,84 +63,129 @@ static char x509_trust_pem_file[256] = { 0 };
...
+  // Begin the registration process by sending a registration request message
+  if ((rc = register_device()) != MQTTCLIENT_SUCCESS)
+  {
+    return rc;
+  }
+
+  // Receive the status of the registration request
+  if ((rc = get_operation_status()) != MQTTCLIENT_SUCCESS)
+  {
+    return rc;
```
CIPop, line 145: The sample is very strict on unknown errors: on errors that cannot be handled, it will exit the process (thus deallocating, disconnecting, etc). The way one would use such a program is within a service that will automatically restart the process if it stopped (logs also being preserved for an eventual investigation).

If we deallocate in `get_operation_status` we should also search for all other places where deallocations are required and add retries. 
We should be careful to de-allocate only when the error is non-critical. Paho has some instances where it checks for structure corruption by verifying the signature of a `struct`. If the error is indicative of memory corruption, out of memory, etc, we should immediately stop the process (also creating a memory dump) to avoid worse things from happening (e.g. memory corruption can be indicative of a security breach).

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: clang format likes it on separate :/

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: Let's chat.  I don't like us returning a non-explained integer if an enum or defined macro is available.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: I had considered putting it all together, something like "return rc_free(rc, &message, topic)" so that there is just one line.  But not sure if that would be harder on the sample reader to jump around.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
ericwol-msft: return rc;  ??

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: will refactor method in future PR.

PR #857, opened 2020-06-25
Update SECURITY.md

joshfree:


SECURITY.md
```diff
@@ -1,8 +1,41 @@
...
 
+## Reporting Security Issues
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+Instead, please report them to the Microsoft Security Response Center (MSRC) at [https://msrc.microsoft.com/create-report](https://msrc.microsoft.com/create-report).
+
+If you prefer to submit without logging in, send email to [secure@microsoft.com](mailto:secure@microsoft.com).  If possible, encrypt your message with our PGP key; please download it from the [Microsoft Security Response Center PGP Key page](https://www.microsoft.com/en-us/msrc/pgp-key-msrc).
+
+You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your original message. Additional information can be found at [microsoft.com/msrc](https://www.microsoft.com/msrc).
```
ahsonkhan, line 17: nit: Keep the whole link as is?
```suggestion
You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your original message. Additional information can be found at https://www.microsoft.com/msrc. 
```
OR use a named link. Something like:
```suggestion
You should receive a response within 24 hours. If for some reason you do not, please follow up via email to ensure we received your original message. Additional information can be found at the [Coordinated Vulnerability Disclosure](https://www.microsoft.com/msrc) page. 
```

joshfree: this is the verbatim text all new repos are getting seeded with (there's a comment tag at the top of the file indicated the text version); I'm just going to keep this as-is, since we're catching the C repo up to the latest text.  Note the CPP repo already has this version of the security.md.

PR #866, opened 2020-06-29
Switch from Sleep to SpinWait

RickWinter:
I've verified on gcc, clang, msvc.  The resulting code on each compiler is an endless loop.

I verified with /01, /02 varieties.

### DIFF
--- a/sdk/src/azure/core/az_precondition.c
+++ b/sdk/src/azure/core/az_precondition.c
@@ -9,11 +9,7 @@
 static void az_precondition_failed_default()
 {
   /* By default, when a precondition fails the calling thread spins forever */
-  static uint16_t value = 0;
-  while (value >= 0)
-  {
-    value++;
-  }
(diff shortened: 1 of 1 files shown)


sdk/src/azure/core/az_precondition.c
```diff
@@ -1,19 +1,18 @@
...
 #include <stdint.h>
 
 #include <azure/core/_az_cfg.h>
 
 static void az_precondition_failed_default()
 {
-  /* By default, when a precondition fails the calling thread is suspended forever */
-  while (true)
+  /* By default, when a precondition fails the calling thread spins forever */
+  static uint16_t value = 0;
```
danewalton, line 12: Technical reason for not just a `while(1)`?

RickWinter: clang -o2 results in the loop optimized away completely.

Issue #612, opened 2020-04-22
Provide a float-free JSON parser


CIPop:
**Is your feature request related to a problem? Please describe.**
az_json currently requires floating point which is not always available on low-end MCUs, including CortexM4 (optional FPU co-processor).
See: https://github.com/Azure/azure-sdk-for-c/pull/590#pullrequestreview-397711310

**Describe the solution you'd like**
Provide `az_json_token_get_number(az_json_token, int32_t)` which should not depend on any internal floating-point operations.
The API should fail if the token contains a floating-point number.

**Describe alternatives you've considered**
Alternatively, provide an API able to always extract an `az_span` from a token (regardless of type).

**Additional context**
IoT protocols do not rely on json documents that contain floating point numbers.

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [x] Description Added
- [x] Expected solution specified

Issue #608, opened 2020-04-22
Provide ability to do JSON string comparison that is resilient to escaped characters


ahsonkhan:
See https://github.com/Azure/azure-sdk-for-c/pull/590#discussion_r412524964

When it comes to JSON, "e" and "\u0065" are considered equal strings. We should provide some mechanism for the user to either:
- get the unescaped string so they can do direct byte-by-byte comparison with a known literal they are looking for (and/or)
- provide a helper API that does the comparison for them.

We don't want components that process/parse JSON and search for a particular property to behave differently between strings that have escaped characters vs. those that don't.

cc @JeffreyRichter, @gilbertw

JeffreyRichter:
Better to provide an equals function that does the correct comparison because this would be non-allocating

CIPop:
Using this feature should be optional. Azure service protocols come with some guarantees around what text is acceptable.
No issue having it enabled by default.

ahsonkhan:
Fixed in https://github.com/Azure/azure-sdk-for-c/pull/871

PR #871, opened 2020-07-01
Redesign JSON parser and JSON token APIs with lazy evaluation of tokens, proper unescaping support, and hardened validation

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/612, https://github.com/Azure/azure-sdk-for-c/issues/608, https://github.com/Azure/azure-sdk-for-c/issues/348, 

**Changes:**
- Updates `az_json_parser` and `az_json_token` APIs and type definitions
  - Materializes the JSON tokens into actual types lazily by deferring the actual creation of ints, strings, bools, until the caller actually needs it.
  - When the  caller needs to get back a printable string, `az_json_token_get_string` should be called, which unescapes and copies the slice, and null-terimates it.
  - When the  caller needs to compare a JSON property name/string text value to a literal lookup string, `az_json_token_is_text_equal` should be called since it unescapes the slice, if needed to do proper semantic comparison.
  - Added a default parser option, which is currently empty, mirroring the JSON builder API, to avoid needing breaking changes in the future.
  - Removes use of double from the parser, so folks who don't have or need floating point support, can still use it. It becomes pay-for-play.
  - Simplify the public API to avoid having the caller figure out the distinction between arrays, objects, and token members.
  - Hardened the parser implementation so it is more resilient to invalid JSON, particularly edge cases, and add several invalid JSON payload tests.
  - General implementation refactoring and clean up, including unifying on a singular stack (and removing duplicate logic) - addresses previous feedback https://github.com/Azure/azure-sdk-for-c/pull/664#discussion_r418622076, https://github.com/Azure/azure-sdk-for-c/pull/664#discussion_r418622584.
  - Removes `az_json_token` creation APIs, since only the `az_json_parser` should be creating these.
  - Removes `az_json_parser_done` since it isn't necessary and no one was calling it. The usage pattern of the new API (in a loop based on token kind or AZ_ERROR_EOF) circumvents the need for this validation API that the user had to remember to call.
- Removes `az_json_pointer` support since it isn't necessary (not used within the SDK other than AAD), nor is it optimal for repeated look-ups of property names since it does redundant parsing. If a critical scenario comes up that cannot be solved by `az_json_parser`, we can add this functionality back with more design and testing.
  - Also remove the `az_json_string_private` helpers as they are no longer needed.

**Coding pattern and usage sample:**
```c
az_json_parser jp;
AZ_RETURN_IF_FAILED(az_json_parser_init(&jp, body, NULL));

AZ_RETURN_IF_FAILED(az_json_parser_move_to_next_token(&jp));
if (jp.token.kind != AZ_JSON_TOKEN_BEGIN_OBJECT)
{
return AZ_ERROR_PARSER_UNEXPECTED_CHAR;
}

bool found_expires_in = false;
bool found_access_token = false;

while (jp.token.kind != AZ_JSON_TOKEN_END_OBJECT)
{
if (az_json_token_is_text_equal(&jp.token, AZ_SPAN_FROM_STR("expires_in")))
{
    AZ_RETURN_IF_FAILED(az_json_parser_move_to_next_token(&jp));
    AZ_RETURN_IF_FAILED(az_json_token_get_uint64(&jp.token, expires_in_seconds));
    found_expires_in = true;
}
else if (az_json_token_is_text_equal(&jp.token, AZ_SPAN_FROM_STR("access_token")))
{
    AZ_RETURN_IF_FAILED(az_json_parser_move_to_next_token(&jp));
    *json_access_token = jp.token;
    found_access_token = true;
}
else
{
    // ignore other tokens
    AZ_RETURN_IF_FAILED(az_json_parser_skip_children(&jp));
}
AZ_RETURN_IF_FAILED(az_json_parser_move_to_next_token(&jp));
}

if (!found_expires_in || !found_access_token)
{
}

return AZ_OK;
```

**TODO:**
- [ ] Add back tests that validate az_json_token_get_double results when double support is enabled, along with az_json_token_get APIs for all other numeric types (atox).
- [ ] Add support for characters escaped in the form of \uXXXX where XXXX is the UTF-16 code point.
- [ ] Additional testing to improve code coverage, including parsing valid and incomplete JSON payloads.
- [ ] Should az_json_parser be renamed to az_json_tokenizer or az_json_reader, since that seems more accurate to what it actually does (reads token, forward-only)?


CMakeSettings.json
```diff
@@ -10,7 +10,13 @@
       "cmakeCommandArgs": "",
       "buildCommandArgs": "-v",
       "ctestCommandArgs": "",
-      "variables": []
+      "variables": [
+        {
+          "name": "UNIT_TESTING",
+          "value": "False",
```
ahsonkhan, line 16: I think we should keep this JSON change to avoid VS adding it back every time. I would also be OK with turning Unit testing on by default, if this is only used by IDEs like VS.

Let me know if there is a reason to revert this change.

sdk/samples/keyvault/keyvault/samples/src/keys_client_example.c
```diff
@@ -40,11 +40,11 @@
  *
  */
 
+#include <az_keyvault.h>
```
ahsonkhan, line 43: Even though this is going away (in https://github.com/Azure/azure-sdk-for-c/pull/867), it is an interesting usage scenario, so I updated it based on the new APIs. Please skim, @vhvb1989.

sdk/src/azure/core/az_aad.c
```diff
@@ -121,25 +168,15 @@ AZ_NODISCARD az_result _az_aad_request_token(_az_http_request* request, _az_toke
...
 
-  // Expiration
-  az_json_token json_token;
-  AZ_RETURN_IF_FAILED(az_json_parse_by_pointer(body, AZ_SPAN_FROM_STR("/expires_in"), &json_token));
-
-  double expires_in_seconds = 0;
-  AZ_RETURN_IF_FAILED(az_json_token_get_number(&json_token, &expires_in_seconds));
+  uint64_t expires_in_seconds = 0;
+  az_json_token json_access_token = { 0 };
+  AZ_RETURN_IF_FAILED(_az_parse_json_payload(body, &expires_in_seconds, &json_access_token));
```
ahsonkhan, line 173: Please take a look at the AAD changes, @antkmsft

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -251,15 +251,13 @@ Documented at
     "etag":"IjYxMDA4ZDQ2LTAwMDAtMDEwMC0wMDAwLTVlOGZlM2QxMDAwMCI="}}
 */
 AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
-    az_json_token_member* tm,
+    az_json_parser* jp,
```
ahsonkhan, line 254: Please review the provisioning client changes, @CIPop, @danewalton.

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
```diff
@@ -369,7 +369,7 @@ test_az_iot_provisioning_client_received_topic_and_payload_parse_invalid_result_
   az_iot_provisioning_client_register_response response;
   az_result ret = az_iot_provisioning_client_parse_received_topic_and_payload(
       &client, received_topic, received_payload, &response);
-  assert_int_equal(AZ_ERROR_ITEM_NOT_FOUND, ret);
+  assert_int_equal(AZ_ERROR_EOF, ret);
```
ahsonkhan, line 372: I believe `AZ_ERROR_EOF` makes more sense here given the payload is an incomplete JSON. Please let me know if this needs to be `AZ_ERROR_ITEM_NOT_FOUND` for some reason.

Was this the intention of the test, or should a complete JSON payload be provided (but one that doesn't have the properties we are a looking for).

cc @CIPop, @danewalton

sdk/src/azure/core/az_json_parser.c
```diff
@@ -1,558 +1,693 @@
...
+AZ_NODISCARD az_result az_json_parser_move_to_next_token(az_json_parser* json_parser)
 {
   _az_PRECONDITION_NOT_NULL(json_parser);
 
-  switch (token.kind)
+  az_span json = _az_json_parser_skip_white_space(json_parser);
+
+  // An empty JSON payload is invalid.
+  // TODO: Should this return invalid state instead?
+  if (az_span_size(json) < 1)
```
ahsonkhan, line 619: @JeffreyRichter - what are your thoughts on the error `az_result` we should return in this case?

sdk/inc/azure/core/az_json.h
```diff
@@ -422,119 +369,127 @@ AZ_NODISCARD az_result az_json_builder_append_end_array(az_json_builder* json_bu
...
-    az_json_token_member* out_token_member);
+    az_span json_buffer,
+    az_json_parser_options const* options)
+{
+  // TODO: Should this be a precondition? If so, there is an inlining issue here.
+  // An empty JSON payload is invalid.
+  if (az_span_size(json_buffer) < 1)
+  {
+    return AZ_ERROR_EOF;
+  }
```
ahsonkhan, line 445: @JeffreyRichter - should this check during init be a precondition?

sdk/inc/azure/core/az_json.h
```diff
@@ -422,119 +369,127 @@ AZ_NODISCARD az_result az_json_builder_append_end_array(az_json_builder* json_bu
...
+ * @brief Reads the next token in the JSON text and updates the parser state.
+ *
+ * @param json_parser A pointer to an #az_json_parser instance containing the JSON to parse.
  *
- * @param json_parser A pointer to an az_json_parser instance containing the JSON to parse.
- * @param out_token A pointer to an az_json_token containing the next parsed JSON array item.
  * @return AZ_OK if the token was parsed successfully.<br>
  *         AZ_ERROR_EOF when the end of the JSON document is reached.<br>
  *         AZ_ERROR_PARSER_UNEXPECTED_CHAR when an invalid character is detected.<br>
  *         AZ_ERROR_ITEM_NOT_FOUND when no more items are found.
```
ahsonkhan, line 474: We don't ever return `AZ_ERROR_ITEM_NOT_FOUND` anymore. I am going to remove this from the docs unless someone can come up with a case I might have missed.

sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -535,30 +535,43 @@ static int send_reported_property(az_span reported_property_payload)
 static az_result update_property(az_span desired_payload)
 {
   // Parse desired property payload
-  az_json_parser json_parser;
-  az_json_token token;
-  az_json_token_member token_member;
+  az_json_parser json_parser = { 0 };
```
ahsonkhan, line 538: @momuno, please review this change to the sample. I tried to follow the same behavior from https://github.com/Azure/azure-sdk-for-c/pull/828, but this could benefit from a second look.

sdk/src/azure/core/az_json_parser.c
```diff
@@ -1,558 +1,693 @@
...
+AZ_NODISCARD static az_result _az_json_parser_process_string(az_json_parser* json_parser)
 {
-  *out = (double)p->value * _ten_to_exp(p->exp) * (double)p->sign;
-  return AZ_OK;
-}
+  // Move past the first '"' character
+  json_parser->_internal.bytes_consumed++;
+
+  az_span token = _get_remaining_json(json_parser);
+  uint8_t* const token_ptr = az_span_ptr(token);
```
ahsonkhan, line 128: @antkmsft, there were a bunch of parameter renaming conflicts in this file based on https://github.com/Azure/azure-sdk-for-c/pull/819. I kept most of the const changes where possible, but none of the `self to ref_span` or `p_reader to  reader` renames apply anymore.

Also, I didn't fully understand why you changed some `json_parser` parameters to `ref_json_parser`, within the helper methods. The public surface area wasn't changed, so I kept things as `json_parser`. What was the rationale for only applying that parameter rename, selectively?

Please take a look to make sure I didn't accidentally undo some of your changes that I should add back.

sdk/inc/azure/core/az_json.h
```diff
@@ -58,141 +58,88 @@ typedef struct
...
+ * is ignored.
+ * @return AZ_OK if the string is returned.<br>
+ * AZ_ERROR_JSON_INVALID_STATE if the kind != AZ_JSON_TOKEN_STRING.<br>
+ * AZ_ERROR_INSUFFICIENT_SPAN_SIZE if \p destination does not have enough size.
  */
-AZ_NODISCARD AZ_INLINE az_json_token az_json_token_begin_array()
-{
-  return (az_json_token){ .kind = AZ_JSON_TOKEN_BEGIN_ARRAY, ._internal = { 0 } };
-}
+AZ_NODISCARD az_result az_json_token_get_string(
```
ahsonkhan, line 121: @JeffreyRichter - I know we discussed adding such an API on span directly, rather than az_json_token, but 
having the `az_json_token_get_boolean` API felt odd to add on span. Also, all the call sites ended up needing to fetch the slice field from the token every time, which impacts usability.

Most calling code doesn't first check the token kind before calling these APIs, and want to fail for mismatched/unexpected token kind anyway.

Let me know what you think.

ahsonkhan: Also, I decided against adding an AZ_JSON_TOKEN_ESCAPED_STRING token kind since we would also need to add a AZ_JSON_TOKEN_ESCAPED_PROPERTY_NAME token kind and that exposes too much implementation detail complexity to the caller.

sdk/tests/core/test_az_json.c
```diff
@@ -791,276 +618,271 @@ az_result write_str(az_span span, az_span s, az_span* out, int32_t* written)
...
+    TEST_EXPECT_SUCCESS(az_json_parser_init(&parser, json, NULL)); \
+    az_result result = AZ_OK; \
+    while (result == AZ_OK) \
+    { \
+      result = az_json_parser_move_to_next_token(&parser); \
+    } \
+    assert_int_equal(result, expected_result); \
+  } while (0)
+
+static void test_json_parser_invalid(void** state)
```
ahsonkhan, line 636: I added a whole bunch of invalid JSON test cases by reasoning through the various states (note: invalid, not incomplete). Let me know if folks come up with some more cases.

cc @vhvb1989

sdk/src/azure/core/az_json_parser.c
```diff
@@ -1,558 +1,693 @@
...
 }
 
-AZ_NODISCARD az_result
-az_json_parser_skip_children(az_json_parser* json_parser, az_json_token token)
+AZ_NODISCARD az_result az_json_parser_move_to_next_token(az_json_parser* json_parser)
 {
   _az_PRECONDITION_NOT_NULL(json_parser);
 
-  switch (token.kind)
+  az_span json = _az_json_parser_skip_white_space(json_parser);
```
antkmsft, line 615: Sorry for the non-thorough review, but it is the only thing that I my eyes catch as I was scanning through the code changes. I tini it should be "whitespace", and not the "white_space". https://en.wikipedia.org/wiki/Whitespace_character

sdk/src/azure/core/az_json_parser.c
antkmsft: spans with negative sizes are non-valid, they violate the contract, correct? Unless there are exception and we are putting new meanings to it. if so, then the code should act as if the size is never negative, i.e. do not check it in the normal `if`s, but to additionally ensure that and detect if our assumptions are incorrect, we put a precondition that he span is valid.

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
antkmsft: do you want to rename EOF to "UNEXPECTED_END"? I think it is the better naming.

sdk/src/azure/core/az_json_parser.c
ahsonkhan: This check isn't about negative size'd spans, but rather empty spans, essentially equivalent to `if (az_span_size(json) == 0)`

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
ahsonkhan: Yep, that is much better.

sdk/src/azure/core/az_json_parser.c
antkmsft: Oh, sorry, I've read it as < 0

sdk/src/azure/core/az_json_parser.c
antkmsft: It is easy to remember, same as in C#: `ref_` is for in/out parameters. `out_` is for out parameters. If it is read only, even though it is a pointer, no prefix is needed,

antkmsft: And normally, if a parameter is `T*`, it is at the very least `ref_`, and sometimes `out_`, because a function can write to the location it points to.
If it is `T const*`, I don't see how it can be `ref_` or `out_`.

antkmsft: Please try not to undo my renaming changes where applicable, because I've correctly addressed 99% of the namings, put tome in it, and unified it :)

sdk/src/azure/core/az_json_parser.c
```diff
@@ -1,558 +1,693 @@
...
 }
 
-AZ_NODISCARD az_result az_json_parser_init(az_json_parser* json_parser, az_span json_buffer)
+// A hex digit is valid if it is in the range: [0..9] | [A..F] | [a..f]
+// Otherwise, return false.
+AZ_NODISCARD static bool _az_is_hex_digit(uint8_t byte)
 {
-  *json_parser = (az_json_parser){ ._internal = { .reader = json_buffer, .stack = 1 } };
-  return AZ_OK;
+  return (byte >= '0' && byte <= '9') || (byte >= 'a' && byte <= 'f')
```
antkmsft, line 103:
```
#include <ctype.h>
isxdigit()
```
https://en.cppreference.com/w/c/string/byte/isxdigit

antkmsft: unlike some other functions, such as isalnum(), this one is not affected by locale, so it is safe to use and expect it works the same on all locales.

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
ahsonkhan: > do you want to rename EOF to "UNEXPECTED_END"? I think it is the better naming.

Deferring that to outside this PR:
https://github.com/Azure/azure-sdk-for-c/issues/894

sdk/src/azure/core/az_json_parser.c
ahsonkhan: > Please try not to undo my renaming changes where applicable, because I've correctly addressed 99% of the namings, put tome in it, and unified it :)

I did not. The ones in your previous PR that affects `az_json_parse`r, no longer apply to this PR anymore since the code changed substantially. I will revisit this again in the future, to make sure the parameter naming rules you described are followed.

https://github.com/Azure/azure-sdk-for-c/pull/819/files#diff-36e2a004054a75fbdbed6861091ae0d3R46

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
CIPop: @ahsonkhan the original item not found is the right error: we've scanned for "status" and could not find it. The code has no expectation that the complete JSON is received.
Please revert to the previous behavior and ensure that none of the existing UTs change behaviors or error codes.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -270,46 +268,69 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
...
+
+  while (jp->token.kind != AZ_JSON_TOKEN_END_OBJECT)
   {
-    if (az_span_is_content_equal(AZ_SPAN_FROM_STR("assignedHub"), tm->name))
+    if (az_json_token_is_text_equal(&jp->token, AZ_SPAN_FROM_STR("assignedHub")))
     {
+      AZ_RETURN_IF_FAILED(az_json_parser_next_token(jp));
+      if (jp->token.kind != AZ_JSON_TOKEN_STRING)
+      {
+        return AZ_ERROR_ITEM_NOT_FOUND;
```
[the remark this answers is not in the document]
ahsonkhan: > Was this the error that the previous function was returning?

Yes, I didn't change that. The call to `az_json_token_get_string` used to fail with `AZ_ERROR_ITEM_NOT_FOUND` if token type wasn't string.

https://github.com/Azure/azure-sdk-for-c/blob/f31482c03a12f462bea8590f56bd0dd4d1b02717/sdk/src/azure/core/az_json_token.c#L23-L31

> I think "not found" is less informative than something like "malformed JSON" which would be the case here when the key exists but the string value doesn't.

Makes sense. Let me know if you still want me to update the `az_result` for this case specifically in `_az_iot_provisioning_client_payload_registration_result_parse`, or leave it as is,

Keep in mind though, it could also be that the key exists, but instead of it being a string, it got changed to a JSON number or boolean true/false, or `null` instead.

> Token not found could be interpreted as a protocol change bug

So, what `az_result` are you thinking that should translate to?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -270,46 +268,69 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
...
-    else if (tm->token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT)
+    else if (az_succeeded(_az_iot_provisioning_client_parse_payload_error_code(jp, out_state)))
     {
-      AZ_RETURN_IF_FAILED(az_json_parser_skip_children(jp, tm->token));
+      // Do nothing
     }
     else
     {
-      (void)_az_iot_provisioning_client_parse_payload_error_code(tm, out_state);
+      // ignore other tokens
```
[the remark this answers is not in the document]
ahsonkhan: I don't believe this logic really change. The parser was still being advanced before, like this:
```C
else if (tm->token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT)
{
   AZ_RETURN_IF_FAILED(az_json_parser_skip_children(jp, tm->token));
}
```

The reason why it got moved to the else block rather than a check on `AZ_JSON_TOKEN_BEGIN_OBJECT` is because the parser no longer returns key/value pair `az_json_token_member`, but rather moves one `az_json_token` at a time.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -326,63 +347,89 @@ AZ_INLINE az_result az_iot_provisioning_client_parse_payload(
...
     else if (az_succeeded(_az_iot_provisioning_client_parse_payload_error_code(
-                 &tm, &out_response->registration_result)))
+                 &jp, &out_response->registration_result)))
     {
       found_error = true;
     }
+    else
+    {
+      // ignore other tokens
+      AZ_RETURN_IF_FAILED(az_json_parser_skip_children(&jp));
```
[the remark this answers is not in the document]
ahsonkhan: Essentially, yes.

Unless a JSON property is pointing to an object or array, `skip_children` just calls `next_token` once, under the covers.

For primitive tokens such as JSON strings, numbers, etc. (i.e. tokens that are not property name, object, array), `skip_children`  does nothing. So, the only cost is a couple of O(1) if-checks.

ahsonkhan:
This PR improved the line and branch coverage of the parser, but there is still some more room to improve (particularly branch coverage), outside of preconditions.

Before:
image

After:
image

PR #886, opened 2020-07-06
Update azure_iot_hub_telemetry.ino

ksaye:
receiveCallback was losing the last character.

without 'mqtt_client.loop();', C2D were not being processed

### DIFF
--- a/sdk/samples/iot/hub/esp8266nodemcu/azure_iot_hub_telemetry.ino
+++ b/sdk/samples/iot/hub/esp8266nodemcu/azure_iot_hub_telemetry.ino
@@ -92,7 +92,7 @@ void receivedCallback(char* topic, byte* payload, unsigned int length)
   Serial.print("Received [");
   Serial.print(topic);
   Serial.print("]: ");
-  for (int i = 0; i < length; i++)
+  for (int i = 0; i <= length; i++)
   {
     Serial.print((char)payload[i]);
   }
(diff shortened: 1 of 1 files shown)


sdk/samples/iot/hub/esp8266nodemcu/azure_iot_hub_telemetry.ino
```diff
@@ -92,7 +92,7 @@ void receivedCallback(char* topic, byte* payload, unsigned int length)
   Serial.print("Received [");
   Serial.print(topic);
   Serial.print("]: ");
-  for (int i = 0; i < length; i++)
+  for (int i = 0; i <= length; i++)
```
ewertons, line 95: This is not correct, it would lead to reading beyond the payload size, resulting in a memory access failure.

sdk/samples/iot/hub/esp8266nodemcu/azure_iot_hub_telemetry.ino
```diff
@@ -281,4 +281,6 @@ void loop()
 
     next_telemetry_send_time_ms = millis() + TELEMETRY_FREQUENCY_MILLISECS;
   }
+  
+  mqtt_client.loop();
```
ewertons, line 285: This indeed seems to be missing from our code. We will add that. Thanks.

ewertons: Closing this PR as the main fix (lack of .loop() calls) was introduced by https://github.com/Azure/azure-sdk-for-c/pull/899

ewertons:
Thanks for contributing to the Azure Embedded Client SDK.

PR #888, opened 2020-07-06
Diagnosing Windows CI failure with unit tests failing

ahsonkhan:
Follow up to https://github.com/Azure/azure-sdk-for-c/pull/885 with more debugging statements to narrow down CI failure root cause that can't be reproduced locally.


ahsonkhan:
An unrelated `errno` being set in CI was impacting the behavior and result of the parsing APIs. Making sure to reset it to 0 at the beginning of the method, fixed the issue.
Also, using `assert_return_code` was sometimes problematic so using `assert_int_equal` where possible.

PR #909, opened 2020-07-08
Rename JSON parser/builder APIs to reader/writer and update all call sites.

ahsonkhan:
The JSON APIs we provide for reading and processing the JSON payload can more accurately be described as a forward-only reader (or tokenizer), since the term parser indicates random access and building up an object graph/DOM under the covers (which we don't do in the reader).

Once we rename the parser to the reader, it follows (for symmetry), that the JSON builder becomes JSON writer. Those names capture the semantics more clearly and convey the intent of the APIs.

I updated some of the docs, local variable names, and comments to use the following verbs instead of parsing/building:
reading/writing, reader/writer, reads/writes

**Rename:**
`az_json_parser` -> `az_json_reader`
`az_json_builder` -> `az_json_writer`

I believe I updated all instances, but if I missed something (possibly in comments or variable names, or samples), please let me know.


ahsonkhan:
cc @TiejunMS, @amih90 - this rename (along with recent API changes) will impact upstream callers

PR #911, opened 2020-07-09
DPS Sample: printf updates

momuno:
* checks punctuation
* adds (void)

### DIFF
--- a/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
+++ b/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
@@ -101,7 +101,7 @@ int main()
   if (az_failed(rc = read_configuration_and_init_client()))
   {
     (void)printf(
-        "Failed to read configuration from environment variables: az_result return code %04x .\n",
+        "Failed to read configuration from environment variables: az_result return code 0x%04x .\n",
         rc);
     return rc;
   }
@@ -111,7 +111,7 @@ int main()
           rc = az_iot_provisioning_client_get_client_id(
               &provisioning_client, mqtt_client_id, sizeof(mqtt_client_id), NULL)))
   {
-    (void)printf("Failed to get MQTT client id: az_result return code %04x .\n", rc);
+    (void)printf("Failed to get MQTT client id: az_result return code 0x%04x .\n", rc);
     return rc;
   }
 
@@ -253,7 +253,7 @@ static int connect_device()
           rc = az_iot_provisioning_client_get_user_name(
               &provisioning_client, mqtt_username, sizeof(mqtt_username), NULL)))
   {
-    (void)printf("Failed to get MQTT username: az_result return code %04x .\n", rc);
+    (void)printf("Failed to get MQTT username: az_result return code 0x%04x .\n", rc);
     return rc;
   }
 
@@ -306,7 +306,7 @@ static int register_device()
           rc = az_iot_provisioning_client_register_get_publish_topic(
               &provisioning_client, register_publish_topic, sizeof(register_publish_topic), NULL)))
   {
-    (void)printf("Failed to get MQTT register publish topic: az_result return code %04x .\n", rc);
+    (void)printf("Failed to get MQTT register publish topic: az_result return code 0x%04x .\n", rc);
     return rc;
   }
 
@@ -356,7 +356,7 @@ static int get_operation_status()
     if (az_failed(
             rc = parse_operation_message(topic, topic_len, message, &response, &operation_status)))
     {
-      (void)printf("Failed to parse operation message: az_result return code %04x .\n", rc);
+      (void)printf("Failed to parse operation message: az_result return code 0x%04x .\n", rc);
       MQTTClient_freeMessage(&message);
       MQTTClient_free(topic);
       return rc;
@@ -433,7 +433,7 @@ static az_result parse_operation_message(
           rc = az_iot_provisioning_client_parse_received_topic_and_payload(
               &provisioning_client, topic_span, message_span, response)))
   {
-    (void)printf("Message from unknown topic: az_result return code %04x .\n", rc);
+    (void)printf("Message from unknown topic: az_result return code 0x%04x .\n", rc);
     return rc;
   }
   print_az_span("Received payload:\n", message_span);
@@ -442,7 +442,7 @@ static az_result parse_operation_message(
   // Parse the operation status from a string to an enum
   if (az_failed(rc = az_iot_provisioning_client_parse_operation_status(response, operation_status)))
   {
-    (void)printf("Failed to parse operation_status: az_result return code %04x .\n", rc);
+    (void)printf("Failed to parse operation_status: az_result return code 0x%04x .\n", rc);
     return rc;
   }
 
@@ -461,7 +461,7 @@ static int send_operation_query_message(
           rc = az_iot_provisioning_client_query_status_get_publish_topic(
               &provisioning_client, response, query_topic, sizeof(query_topic), NULL)))
   {
-    (void)printf("Unable to get query status publish topic: az_result return code %04x .\n", rc);
+    (void)printf("Unable to get query status publish topic: az_result return code 0x%04x .\n", rc);
     return rc;
   }
 
--- a/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
+++ b/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
@@ -382,19 +382,19 @@ static int get_operation_status()
   if (operation_status == AZ_IOT_PROVISIONING_STATUS_ASSIGNED)
   {
     (void)printf("SUCCESS - Device provisioned:\n");
-    (void)print_az_span("\tHub Hostname: ", response.registration_result.assigned_hub_hostname);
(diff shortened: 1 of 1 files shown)


sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -491,16 +491,11 @@ static void sleep_for_seconds(uint32_t seconds)
 }
 
 // Print an az_span to the console
-static void print_az_span(const char* str, az_span span)
+static int printf_az_span(const char* str, az_span span)
 {
-  printf("%s", str);
+  char buffer[az_span_size(span) + 1];
+  memcpy(buffer, az_span_ptr(span), (uint32_t)az_span_size(span));
+  buffer[az_span_size(span)] = '\0';
```
ahsonkhan, line 498: This helper shouldn't be necessary and is redundant.

I would consider using the existing, provided API, which essentially does what you are trying to do here:
`az_span_to_str(...)`

https://github.com/Azure/azure-sdk-for-c/blob/eb836ef1af71515b5644f40cb2a163389a250801/sdk/inc/azure/core/az_span.h#L238

One reason to avoid your own custom implementations is that use of `memcpy` isn't resilient to overlapping buffers, which is why `az_span_to_str` using `memmove`.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -289,7 +289,7 @@ static int subscribe()
        = MQTTClient_subscribe(mqtt_client, AZ_IOT_PROVISIONING_CLIENT_REGISTER_SUBSCRIBE_TOPIC, 1))
       != MQTTCLIENT_SUCCESS)
   {
-    printf("Failed to subscribe to the register topic filter, MQTTClient return code %d\n", rc);
+    (void)printf("Failed to subscribe to the register topic filter; MQTTClient return code %d .\n", rc);
```
ahsonkhan, line 292: Why do we need all these `(void)` casts for every `printf` statement?

momuno: Per Azure SDK C guidelines.  https://azure.github.io/azure-sdk/clang_implementation.html#clang-check-syscall-errors
image

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: I just found az_span_to_str haha, but didn't know about it when I wrote this :)  Actually, I talked to Himanshu and he explained to me that because some az_spans could include multi-byte characters, we want to print out bytes one by one.  I am going to revert to the previous putchar approach Cris had.  The helper helps in that case.

ahsonkhan: Does putchar support printing out multi-byte characters? Aren't we only giving it one byte anyway so how is it's behavior different from printf?

ahsonkhan: For example, suppose we have a chinese character string: 汉
The is 3 bytes in UTF-8: 0xE6 0xB1 0x89, so imagine you have an az_span of size 3 containing those bytes.
https://www.fileformat.info/info/unicode/char/6c49/index.htm

What does printf output when turning that span to a string versus calling putchar 3 times? I would have thought both end up with the same output.

momuno: If I recall correctly, @hihigupt explained that printf can drop bytes, such as if there are multiple 0s next to each other.  Himanshu, can you please clarify for me?  Thanks.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -306,7 +306,7 @@ static int register_device()
           rc = az_iot_provisioning_client_register_get_publish_topic(
               &provisioning_client, register_publish_topic, sizeof(register_publish_topic), NULL)))
   {
-    printf("Failed to get MQTT register publish topic, az_result return code %04x\n", rc);
+    (void)printf("Failed to get MQTT register publish topic; az_result return code %04x .\n", rc);
```
danewalton, line 309: nit: being a bit of a grammar nazi here but I would prefer everywhere where semicolons are used that we use colons:
```suggestion
    (void)printf("Failed to get MQTT register publish topic: az_result return code %04x .\n", rc);
```

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -426,23 +426,23 @@ static az_result parse_operation_message(
...
-  print_az_span("Topic:\n", topic_span);
+  print_az_span("Topic: ", topic_span);
 
   // Parse the incoming message and payload
   if (az_failed(
           rc = az_iot_provisioning_client_parse_received_topic_and_payload(
               &provisioning_client, topic_span, message_span, response)))
   {
-    printf("Message from unknown topic, az_result return code %0x4\n", rc);
+    (void)printf("Message from unknown topic; az_result return code %0x4 .\n", rc);
```
danewalton, line 436: Should these all be
```suggestion
    (void)printf("Message from unknown topic; az_result return code %04x .\n", rc);
```

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -426,23 +426,23 @@ static az_result parse_operation_message(
...
   }
   print_az_span("Received payload:\n", message_span);
-  printf("Response status is %d.\n", response->status);
+  (void)printf("Response status: %d\n", response->status);
 
   // Parse the operation status from a string to an enum
   if (az_failed(rc = az_iot_provisioning_client_parse_operation_status(response, operation_status)))
   {
-    printf("Failed to parse operation_status, az_result return code %0x4\n", rc);
+    (void)printf("Failed to parse operation_status; az_result return code %0x4 .\n", rc);
```
danewalton, line 445: here too

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -253,7 +253,7 @@ static int connect_device()
           rc = az_iot_provisioning_client_get_user_name(
               &provisioning_client, mqtt_username, sizeof(mqtt_username), NULL)))
   {
-    printf("Failed to get MQTT username, az_result return code %0x4\n", rc);
+    (void)printf("Failed to get MQTT username; az_result return code %0x4 .\n", rc);
```
danewalton, line 256: here as well 
```suggestion
    (void)printf("Failed to get MQTT username; az_result return code %04x .\n", rc);
```

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: great catch. thank you!

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
hihigupt: @ahsonkhan  Reason we have putchar API is if span constrains binary data.. and we use printf then data might be truncated .... so to print everything using putchar.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -111,7 +111,7 @@ int main()
           rc = az_iot_provisioning_client_get_client_id(
               &provisioning_client, mqtt_client_id, sizeof(mqtt_client_id), NULL)))
   {
-    printf("Failed to get MQTT clientId, az_result return code %04x\n", rc);
+    (void)printf("Failed to get MQTT client id: az_result return code %04x .\n", rc);
```
CIPop, line 114: Nit: Will %04x print something like 0xDEADC0DE or do we need to add the 0x like `0x%04x`? 
I remember a nightmare debugging session a long time ago where my mistake was not adding the "h" descriptor in an assembly DD instruction. Took me a week to find the bug so I am very careful now.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -381,24 +381,24 @@ static int get_operation_status()
...
   if (operation_status == AZ_IOT_PROVISIONING_STATUS_ASSIGNED)
   {
-    printf("SUCCESS - Device provisioned:\n");
+    (void)printf("SUCCESS - Device provisioned:\n");
     print_az_span("\tHub Hostname: ", response.registration_result.assigned_hub_hostname);
     print_az_span("\tDevice Id: ", response.registration_result.device_id);
   }
   else // Unsuccesful assignment (unassigned, failed or disabled states)
   {
-    printf("ERROR - Device Provisioning failed:\n");
```
CIPop, line 390: Nit: We could extract methods for the success vs failure cases (maybe in the future?).

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: i think you're right.  I need to add 0x in front of %
Thanks

PR #915, opened 2020-07-09
Instructions on how to upgrade installed packages

CIPop:
In certain cases, VCPKG packages need to be upgraded. Adding documentation on how to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,8 @@ vcpkg.exe install --triplet x64-windows-static curl[winssl] cmocka paho-mqtt
 # VCPKG_ROOT=PATH_TO_VCPKG (replace PATH_TO_VCPKG for where vcpkg is installed)
 ```
 
+If you previously installed VCPKG and dependencies, you may need to run `.\vcpkg.exe upgrade --no-dry-run` to upgrade to the latest packages.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -291,6 +291,8 @@ vcpkg.exe install --triplet x64-windows-static curl[winssl] cmocka paho-mqtt
 # VCPKG_ROOT=PATH_TO_VCPKG (replace PATH_TO_VCPKG for where vcpkg is installed)
 ```
 
+If you previously installed VCPKG and dependencies, you may need to run `.\vcpkg.exe upgrade --no-dry-run` to upgrade to the latest packages.
```
ahsonkhan, line 294: Is this something we want customers to do regularly, or is it up to them to decide to move to newer versions, as needed?

CIPop: This is based on your comment here: https://github.com/Azure/azure-sdk-for-c/issues/912#issuecomment-655872488

You mentioned it could be due to my libs being old (which could have been the case). If someone has issues with dependencies I expect them to read the dev guide again. If they do, simply running the install steps will not cause any of the libs to be updated. This clarifies how to update.

Maybe there are better solution (something like install or update) that I don't know. I think this will work until we find something better.

PR #917, opened 2020-07-09
change pnp api to use option instead of standalone method

### LINKED ISSUE #812: Remove/update PnP temporary function once PnP GA's
A temporary function has been added to the IoT SDK which enables PnP model id support. This should be revisited once PnP GA's.

danewalton:
Closes #812 

Pardon some of the clang-format changes.

### DIFF
--- a/sdk/inc/azure/iot/az_iot_hub_client.h
+++ b/sdk/inc/azure/iot/az_iot_hub_client.h
@@ -46,6 +46,8 @@ typedef struct az_iot_hub_client_options
   az_span module_id; /**< The module name (if a module identity is used). */
   az_span user_agent; /**< The user-agent is a formatted string that will be used for Azure IoT
                          usage statistics. */
+  az_span model_id; /**< The model id used to identify the capabilities of a device based on the
+                       Digital Twin document */
 } az_iot_hub_client_options;
 
 /**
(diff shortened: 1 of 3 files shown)


sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -74,6 +79,11 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
   {
     required_length += az_span_size(*user_agent) + az_span_size(hub_client_param_separator_span);
   }
+  if (az_span_size(*model_id) > 0)
+  {
+    required_length += az_span_size(hub_client_param_separator_span)
+        + az_span_size(hub_client_param_equals_span);
```
ericwol-msft, line 85: Missing + az_span_size(*model_id) ?

jspaith: I think that's up by line 76 but I'm not seeing adding the size for `hub_digital_twin_model_id` anywhere during initial calc.  If(!!) this is indeed a bug would be good to understand how it got past tests.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -46,6 +46,8 @@ typedef struct az_iot_hub_client_options
   az_span module_id; /**< The module name (if a module identity is used). */
   az_span user_agent; /**< The user-agent is a formatted string that will be used for Azure IoT
                          usage statistics. */
+  az_span model_id; /**< The model id used to identify the capabilities of a device based on the
```
jspaith, line 49: `model id` is usually referred to as `ModelId` in English at least from what I've tracked.

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -59,13 +61,16 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
       = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
 
   int32_t required_length = az_span_size(client->_internal.iot_hub_hostname)
       + az_span_size(client->_internal.device_id) + az_span_size(hub_service_api_version)
       + (int32_t)sizeof(hub_client_forward_slash);
+  required_length += az_span_size(*model_id) > 0 ? az_span_size(hub_service_preview_api_version)
+                                                 : az_span_size(hub_service_api_version);
   if (az_span_size(*module_id) > 0)
   {
     required_length += az_span_size(*module_id) + (int32_t)sizeof(hub_client_forward_slash);
```
jspaith, line 76: Is this the right check?  Because we url_encode later on and it can be bigger, shouldn't we canculate from what encoded length would be?

sdk/src/azure/iot/az_iot_hub_client.c
danewalton: added comment for why it's missing

danewalton: added comment for why it's missing

sdk/src/azure/iot/az_iot_hub_client.c
danewalton: this one specifically is determining whether we use the preview API string or the regular API string. We have to skip the calculation of the model id since we url encode it later and don't know how long it will be

sdk/inc/azure/iot/az_iot_hub_client.h
danewalton: unless reallllly necessary I'll probably keep as is to match the above module name

PR #920, opened 2020-07-11
update vscode instructions

danewalton:


CIPop:
This is important because VSCode defaults to x86 (if one chooses auto-detect CMake) and we default to x64 for prerequisites.

PR #922, opened 2020-07-13
remove double length include for api version

danewalton:


sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -67,8 +67,7 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
       = az_span_init((uint8_t*)mqtt_user_name, (int32_t)mqtt_user_name_size);
 
   int32_t required_length = az_span_size(client->_internal.iot_hub_hostname)
-      + az_span_size(client->_internal.device_id) + az_span_size(hub_service_api_version)
-      + (int32_t)sizeof(hub_client_forward_slash);
+      + az_span_size(client->_internal.device_id) + (int32_t)sizeof(hub_client_forward_slash);
```
ahsonkhan, line 70: We should consider adding a unit test  that would have failed due to this bug (and this generally applies to any bug we find and fix). A unit test will make sure we don't regress this code again. Is that feasible to do?

I am **guessing** we could test that this method is expected to succeed for a certain input, but in master, it wouldn't because required_size would be too high, and would instead return the "not enough size" error. Or, we can test something about the value of `out_mqtt_user_name_length`. Not sure.

danewalton: Yea I had a feeling this was going to get posted. So one of the problems is that with many elements, the number of unit tests that would account for each element individually would get out of hand. The unit tests currently will test to make sure that if a buffer of `length - 1` is passed then it should return `not enough space`. Anything of `length` or greater should be fine.

ahsonkhan: I think a targeted unit test is good enough for this particular issue to capture future regression.

Have `model_id` and `user_agent` point to `AZ_SPAN_NULL`, and then have `mqtt_user_name_size` be large enough:
`az_span_size(client->_internal.iot_hub_hostname) + az_span_size(client->_internal.device_id) + az_span_size(hub_service_api_version)`

In current master, this test will fail with `not enough space` (i.e. bug), but with your PR, that test should succeed, set `out_mqtt_user_name_length` to the right amount and return `AZ_OK`.

The goal is to get CI to prevent me from reverting your commit, accidentally. Taken to its extreme, a code logic change (not new code/refactoring) should require more unit tests, or an existing test to break, which applies to this change.

CIPop: +1 on UT but I agree with @danewalton we would be missing out a lot of other places and there is some urgency to fix this bug.
We could add this and other missing UTs in a future PR.

PR #924, opened 2020-07-13
Adding IoT Hub WebSockets APIs, fixing Y2k38 issue

### LINKED ISSUE #916: Can't use VSCode with CMake
**Describe the bug**
CMake from VSCode isn't working anymore.

***Exception or Stack Trace***
The cmake extension doesn't work (previously working fine) in Windows.
(original stack/error fixed - see comments for reference).

**To Reproduce**
On Windows + cmake version 3.17.2 + Powershell (Windows Terminal) or cmd.exe:
1. follow instructions in https://github.com/Azure/azure-sdk-for-c#development-environment.
2. follow instructions in https://github.com/Azure/azure-sdk-for-c#vscode

***Code Snippet***
1. Set all required environment variables.

In Windows I do this with `Invoke-BatchScript`, on Linux I dot-source a bash script.

2. Follow the cmake steps in command-prompt. In the `\build` folder:

`cmake ..` or `cmake-gui ..` or `ccmake ..`

3. Build using your preferred tool-chain 

`msbuild /m ./az.sln` or `make -j`

4. Open Visual Studio Code (with the CMake plugin installed) from the same command-prompt. (code.exe will inherit the command-line environment, still in `\build`).

`code ..`

5. Select "Configure All Projects". 
image

(Make sure to select an x64 toolset on Windows - autodetect will select the x86 toolset instead.)

**Expected behavior**
VSCode auto-detects and uses the configuration I've specified in command-line.

_Actual behavior:_

VSCode replaces the configuration specified in command-line.

**Screenshots**
CMake-Gui option-buttons are broken by VSCode, configuration is replaced by something else:

My configuration:
image

After VSCode configured cmake:
image

None of the samples are getting built or can be debugged.

 - OS: [e.g. iOS] Windows, PWSH command-line, cmake-gui
 - IDE : [e.g. IntelliJ] Visual Studio Code.
 - Version of the Library used: master branch

**Additional context**

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [X] Bug Description Added
- [X] Repro Steps Added
- [X] Setup information Added

CIPop:
* Public API change: Adding IoT Hub WebSockets APIs + Sample.
* Public API change: Fixing Y2k38 issue for SAS tokens. (@hihigupt this is impacting AzRTOS).
* Moving samples configuration at the top of the file, above `#define`s that should never be modified in code.
* Moving `.vscode` template files into `vscode_config` - these files are overwriting command-line `cmake` options (#916) and updating readme.
* Making SAS Sample environment variable names different from X509 devices to allow using the same environment context for both at the same time. Making SAS token time more granular (minutes instead of hours). (For security reasons, SAS tokens should not be long lived - 2hr or lower recommended.)

Fixes #916.


sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
```diff
@@ -34,8 +34,15 @@
...
+// Comment to use MQTT without WebSockets.
+#define USE_WEB_SOCKET
+#define TIMEOUT_MQTT_DISCONNECT_MS (10 * 1000)
+#define TELEMETRY_SEND_INTERVAL 1
+#define NUMBER_OF_MESSAGES 5
+#define SAS_TOKEN_EXPIRATION_TIME_DIGITS 4
+
 // DO NOT MODIFY: Device ID Environment Variable Name
-#define ENV_DEVICE_ID "AZ_IOT_DEVICE_ID"
+#define ENV_DEVICE_ID "AZ_IOT_DEVICE_ID_SAS"
```
danewalton, line 45: I think this should be left as is? Or is this to differentiate in an env so you can run both cert and sas devices in the same env.

sdk/samples/iot/hub/src/paho_iot_hub_telemetry_example.c
```diff
@@ -26,13 +26,17 @@
...
+#define NUMBER_OF_MESSAGES 5
+
 #ifdef _MSC_VER
 // "'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead."
 #pragma warning(disable : 4996)
 #endif
 
 // DO NOT MODIFY: Device ID Environment Variable Name
-#define ENV_DEVICE_ID "AZ_IOT_DEVICE_ID"
+#define ENV_DEVICE_ID "AZ_IOT_DEVICE_ID_SAS"
```
danewalton: this is a cert sample

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -376,10 +376,11 @@ static void receive_registration_status()
     if (!is_operation_complete)
     {
       LOG("Operation is still pending.");
-      MQTTClient_freeMessage(&message);
-      MQTTClient_free(topic);
 
       send_operation_query_message(&response);
+
+      MQTTClient_freeMessage(&message);
+      MQTTClient_free(topic);
```
momuno, line 383: This looks good to me.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -44,11 +44,11 @@
 
 // DO NOT MODIFY: the path to a PEM file containing the device certificate and
 // key as well as any intermediate certificates chaining to an uploaded group certificate.
-#define ENV_DEVICE_X509_CERT_PEM_FILE_PATH "AZ_IOT_DEVICE_X509_CERT_PEM_FILE_PATH"
+#define ENV_DEVICE_X509_CERT_PEM_FILE_PATH "AZ_IOT_DEVICE_X509_CERT_PEM_FILE"
 
 // DO NOT MODIFY: the path to a PEM file containing the server trusted CA
 // This is usually not needed on Linux or Mac but needs to be set on Windows.
-#define ENV_DEVICE_X509_TRUST_PEM_FILE_PATH "AZ_IOT_DEVICE_X509_TRUST_PEM_FILE_PATH"
+#define ENV_DEVICE_X509_TRUST_PEM_FILE_PATH "AZ_IOT_DEVICE_X509_TRUST_PEM_FILE"
```
momuno, line 52: I do want to update the environment name later on to include "_PATH" for consistency, but I realize that will involve updating a bunch of documentation and the generate_cert script.

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
momuno: I'm unclear on the need for SAS here as well.

sdk/samples/iot/hub/src/paho_iot_hub_telemetry_example.c
CIPop: Yeah - found it myself running the samples. Fixed.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
CIPop: We need to be consistent. I think we should have a batch file, powershell file or bash file (or all 3) checked in as templates. That way, it's so much easier to run tests on a dev box.
In that case, we need all of these to work for all samples (I don't want to change the env variables to run telemetry vs provisioning).

sdk/samples/iot/hub/src/paho_iot_hub_sas_telemetry_example.c
CIPop: Exactly for that reason: we need to be able to run all samples within a single command prompt. This isn't possible today because my x509 device is interfering with the one that does SAS.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: I agree with consistency, hence why I'm fine with you changing it back now.  Later, I would like to update them all to include "_PATH" because that is more precise.

CIPop:
> Can the VSCode stuff be tackled in a different PR? I'm not sure of the top of my head how that works and I think it should be it's own.

Sorry, I don't have the time. I can revert and leave the bug opened.

The reasons against having a `.vscode` folder:

1. The checked-in config doesn't work: you need to edit the file which marks it changed and then it'll be pushed with my changes.
1. The `.vscode` folder is interfering with our supported documentation for build (which is cmd-line based): `cmake ..`; `cmake --build`; This needs to work and VSCode should be used only to edit or debug applications. The CMake Extension works fine if `cmake ..` (or `ccmake ..` or `cmake-gui ..`) was previously called in command prompt and will maintain the configuration. 
1. The `.vscode` folder is interfering with my personal preferences for the workspace which I don't wish to accidentally check-in.

Also, I've just renamed the folder. If someone wants to use it today, they need to edit the files. With my change, they will need to rename the folder as .vscode and edit the files. The advantage is that the files won't be marked as "changed" by git and so cannot be accidentally checked in.

danewalton:
@CIPop totally hear you. I'd just like to see if there's another way to do it. I'll personally try and tackle it if you don't have the time.

CIPop:
@danewalton I'll remove that part of the change and will assign the bug to you for analysis. I'll also clarify the way I'm using VSCode there.

CIPop:
@JeffreyRichter @RickWinter please let me know if there are any concerns for the public APIs.

JeffreyRichter:
You mostly just added some time parameters, right? Is this what you're asking about? If so, I'm OK with these parameters.
Other than this, I see new #defines and some new internal functions (which I don't care about)

CIPop:
> You mostly just added some time parameters, right? Is this what you're asking about? If so, I'm OK with these parameters.
> Other than this, I see new #defines and some new internal functions (which I don't care about)

@JeffreyRichter you're right: I've added two additional #defines part of the public API and the uint32_t to uint64_t.
Thanks for your sign-off!

PR #942, opened 2020-07-21
Make c2d and methods samples single threaded

momuno:
This PR is for two of the samples (hub_c2d and hub_methods), which used MQTT_setCallbacks and were asynchronous.  This change makes these samples single threaded by using MQTTClient_recieve.

This PR does NOT include the hub_twin sample.  While this sample also needs to be made single-threaded, to do so requires work in another task: removing user command prompt.  MQTTClient_receive for hub_twin sample will naturally be incorporated into that task.

I intentionally did not go through and refactor other methods, nor update all the printfs to LOG_ERROR, etc.  This is because I wanted to keep this PR focused on the threading change required.  Consistency cleanup for all samples will happen in a separate task.


momuno:
Rewriting sections to receive up to 5 messages.

PR #946, opened 2020-07-21
DPS Samples Update

momuno:
- update logging (added direction if TRUST PEM variable is not set)
- update consistencies

### DIFF
--- a/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
+++ b/sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
@@ -82,25 +82,27 @@
     (void)printf("\n"); \
   }
 
-// Buffers
-static char x509_cert_pem_file_path_buffer[256];
-static char x509_trust_pem_file_path_buffer[256];
+// Store environment variables
 static char global_provisioning_endpoint_buffer[256];
 static char id_scope_buffer[16];
 static char registration_id_buffer[256];
-static char mqtt_client_id_buffer[128];
-static char mqtt_username_buffer[128];
-static char register_publish_topic_buffer[128];
-static char query_topic_buffer[256];
+static char x509_cert_pem_file_path_buffer[256];
+static char x509_trust_pem_file_path_buffer[256];
 
 // Clients
 static az_iot_provisioning_client provisioning_client;
 static MQTTClient mqtt_client;
+static char mqtt_client_id_buffer[128];
+static char mqtt_username_buffer[128];
+
+// Topics
+static char register_publish_topic_buffer[128];
+static char query_topic_buffer[256];
 
 // Functions
 static void create_and_configure_client();
 static az_result read_environment_variables(
-    az_span* endpoint,
+    az_span* global_provisioning_endpoint,
     az_span* id_scope,
     az_span* registration_id);
 static az_result read_configuration_entry(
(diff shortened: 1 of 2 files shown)


sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
```diff
@@ -82,25 +82,27 @@
...
+static char x509_trust_pem_file_path_buffer[256];
 
 // Clients
 static az_iot_provisioning_client provisioning_client;
 static MQTTClient mqtt_client;
+static char mqtt_client_id_buffer[128];
+static char mqtt_username_buffer[128];
+
+// Topics
+static char register_publish_topic_buffer[128];
```
ahsonkhan, line 99: Can you please explain why we chose 128 as the size for some of these, while others have 256?

sdk/samples/iot/provisioning/src/paho_iot_provisioning_sas_example.c
```diff
@@ -29,11 +29,22 @@
 #include <azure/core/az_span.h>
 #include <azure/iot/az_iot_provisioning_client.h>
 
+#define TIMEOUT_MQTT_RECEIVE_MS (60 * 1000)
+#define TIMEOUT_MQTT_DISCONNECT_MS (10 * 1000)
```
ahsonkhan, line 33: Same here. I am curious, do these values hold some special semantic meaning? Why 60 seconds for the timeout for receive, but 10 for disconnect?

sdk/samples/iot/provisioning/src/paho_iot_provisioning_sas_example.c
```diff
@@ -29,11 +29,22 @@
...
+
 #ifdef _MSC_VER
 // "'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead."
 #pragma warning(disable : 4996)
 #endif
 
+// DO NOT MODIFY: Service information
+#define ENV_GLOBAL_PROVISIONING_ENDPOINT_DEFAULT "ssl://global.azure-devices-provisioning.net:8883"
```
ahsonkhan, line 41: Is this something the user should never change, even if the endpoint running in a specific region or was private?

sdk/samples/iot/provisioning/src/paho_iot_provisioning_example.c
momuno: I'm not sure.  I have not investigated that far. @CIPop might know.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_sas_example.c
momuno: The receive message timeout is longer so that the user can go to the service and send a message for the client to receive.  That is not necessary for the disconnect.  Why those exact values?   Likely estimates.

sdk/samples/iot/provisioning/src/paho_iot_provisioning_sas_example.c
momuno: from Hub code comment
//Paho requires that the MQTT endpoint be of the form ssl://<HUB ENDPOINT>:8883

momuno: // Paho requires that the MQTT endpoint be of the form ssl://\<HUB ENDPOINT\>:8883

ahsonkhan: Right, so couldn't the `<HUB ENDPOINT>` be  different based on the user's needs? Why does the  `// DO NOT MODIFY: Service information` comment apply to it?

PR #951, opened 2020-07-23
Twin sample: Synchronous and no user prompt

momuno:


sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c
```diff
@@ -141,37 +171,9 @@ int main()
...
-      }
-      if (input == 'q')
-      {
-        break;
-      }
-    }
-  }
+  get_twin_document();
+  send_reported_property();
+  get_desired_property();
```
danewalton, line 176:
```suggestion
  receive_desired_property();
```
"get" sounds as though it is actively asking for it and expecting something in return. "receive" makes it more explicit it is waiting for something.

momuno: totally.

danewalton:
that took a bit of effort man haha

Issue #459, opened 2020-03-13
Markdown Lint


danewalton:
Should we look to have a markdown lint?
[Extension here for VSCode](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint)

PR #953, opened 2020-07-24
Sample code restructure part 1

momuno:
This PR focuses on the directory restructuring.
As a result, some files needed to be combined, such as the readme.md, cmakelists.txt. and the generate_cert scripts.
Also renames samples to *sample.c instead of *example.c
Tested on Windows and Ubuntu.

Further refactoring of the samples to pull common code into a single header file will happen in part 2.


sdk/samples/iot/readme.md
```diff
@@ -1,29 +1,234 @@
...
+
+More detailed step-by-step guides on how to run an IoT Hub Client sample from scratch can be found below:
+  - Linux: [How to setup and run Azure SDK for Embedded C IoT Hub Samples on Linux](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/docs/how_to_iot_hub_samples_linux.md)
+  - Windows: [How to setup and run Azure SDK for Embedded C IoT Hub Samples on Microsoft Windows](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/docs/how_to_iot_hub_samples_windows.md)
+  - ESP8266: [How to Setup and Run Azure SDK for Embedded C IoT Hub Client on Esp8266 NodeMCU](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/docs/how_to_iot_hub_esp8266_nodemcu.md)
+
+## Sample Descriptions
+This section provides an overivew of the different samples available to run and what to expect from each.
```
ahsonkhan, line 16:
```suggestion
This section provides an overview of the different samples available to run and what to expect from each.
```

sdk/samples/iot/readme.md
```diff
@@ -1,29 +1,234 @@
...
+    * Navigate on the Solution Explorer panel to the sample project you would like to run.
+    * Make it the default startup project (right-click on the sample project, then click on Set as Startup Project).
+    * Build and run the project (F5 on most installations).
+
+
+## Next Steps and Additional Documentation
+
+Start using the IoT Provisioning Client in your solutions!
+
+* A general overivew of the Embedded C SDK and additional background on running samples can be found in the [Azure SDK for Embedded C README](https://github.com/Azure/azure-sdk-for-c#azure-sdk-for-embedded-c).
```
ahsonkhan, line 215:
```suggestion
* A general overview of the Embedded C SDK and additional background on running samples can be found in the [Azure SDK for Embedded C README](https://github.com/Azure/azure-sdk-for-c#azure-sdk-for-embedded-c).
```

sdk/samples/iot/readme.md
momuno: thank you

sdk/samples/iot/generate_certificate.cmd
```diff
@@ -21,30 +21,29 @@ type device_ec_cert.pem > device_cert_store.pem
...
 
 echo.
+echo IMPORTANT:
 echo It is NOT recommended to use OpenSSL on Windows or OSX. Recommended TLS stacks are:
 echo Microsoft Windows SChannel: https://docs.microsoft.com/en-us/windows/win32/com/schannel
 echo OR
 echo Apple Secure Transport : https://developer.apple.com/documentation/security/secure_transport
 echo If using OpenSSL, it is recommended to use the OpenSSL Trusted CA store configured on your system.
 echo.
-echo If required (for example on Windows), download the Baltimore PEM CA from https://www.digicert.com/digicert-root-certificates.htm to the current folder.
```
ewertons, line 30: Why this is not needed anymore? This seems specific to Windows OS.

momuno: Good question.  I took it out because the download instructions in this script are for ALL samples running on Windows, but cert samples are the only samples which will run this generate_certificate script.  I think it is most appropriate then to put this info in a common location, such as the README.md, where it can be referenced for all samples.  

I have included logging in the provisioning samples to alert the user if the env variable AZ_IOT_DEVICE_X509_TRUST_PEM_FILE is not set when it is needed.  (The other samples will be updated later to match.)

sdk/samples/iot/docs/how_to_iot_hub_esp8266_nodemcu.md
```diff
@@ -28,7 +28,7 @@ _The following was run on an Ubuntu Desktop 18.04 environment, with Arduino IDE
 1. Create an Arduino library for Azure Embedded SDK for C
 
     ```shell
-    $ wget https://raw.githubusercontent.com/Azure/azure-sdk-for-c/master/sdk/samples/iot/hub/aziot_esp8266/generate_arduino_zip_library.sh
+    $ wget https://raw.githubusercontent.com/Azure/azure-sdk-for-c/master/sdk/samples/iot/aziot_esp8266/generate_arduino_zip_library.sh
```
ewertons, line 31: Thanks. That's one I was concerned about.

sdk/samples/iot/generate_certificate.cmd
ewertons: That's a good idea to have a common doc for setting the certs. The only thing is how the information is getting scattered again. The feedback from customers in the past is that they need to take many hops to get all the info they need. I rather have to maintain more docs than to make the dev experience worse with not having all info in the same place.

ewertons: If the info you removed is applicable to just some samples, maybe instead of removing we should add a note explaining that. Makes it clearer, and the information is all there available at a single doc.

sdk/samples/iot/readme.md
```diff
@@ -1,29 +1,234 @@
...
+  ```
+* Select Save to send the message. The device will store the value locally and report the updated property to the service.
+
+**IoT Hub PnP Sample**
+* *Executable:* `paho_iot_hub_pnp_sample`
+
+  Connect a PnP enabled device with the Digital Twin Model ID (DTMI) detailed [here](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/Thermostat.json). X509 self-certification is used.
+
+  In short, the capabilities are listed here:
+* **Methods**: Invoke a method called `getMaxMinReport` with JSON payload value `"since"` with an [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) value for start time for the report. The method sends a response containing the following JSON payload:
```
danewalton, line 66: note there was a PR for a readme linter which means we are going to use `-` for unordered lists instead of `*`

danewalton: https://github.com/Azure/azure-sdk-for-c/pull/959

danewalton: Could you open the README in VSCode with [this](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) linter to match what Ahson merged?

ahsonkhan: Another is `Lists should be surrounded by blank lines`

We may want to advocate use of such a linter team-wide. I was reluctant to add it to CI because of the false positives and exceptions, but maybe that's the way to go, to "stay green". This is part of this feature request: https://github.com/Azure/azure-sdk-for-c/issues/459

momuno: Thanks all

ahsonkhan: > note there was a PR for a readme linter which means we are going to use `-` for unordered lists instead of `*`

To clarify, I think the linter enforces local consistency. So, if the doc starts using `*` in lists, then it expects the whole doc to use it. If you start with `-`, it expects that going forward. Other than consistency, I don't think there is any intrinsic benefit of using one vs the other.

sdk/samples/iot/CMakeLists.txt
```diff
@@ -22,13 +22,8 @@ if(NOT OpenSSL_FOUND)
...
 
 # Azure IoT Samples Library
-add_library (${TARGET_NAME}
-  ${CMAKE_CURRENT_LIST_DIR}/src/sample_sas_utility.c
-)
-
-target_include_directories (${TARGET_NAME}
-  PUBLIC
-  ${CMAKE_CURRENT_LIST_DIR}/inc
+add_library (az_iot_samples_common
```
ahsonkhan, line 25: This change overwrote parts of the changes from https://github.com/Azure/azure-sdk-for-c/pull/961

Specifically, using `{TARGET_NAME}`, from `set(TARGET_NAME "az_iot_samples_common")`

ahsonkhan: Please take a look at this file (from the above linked PR) and retain pieces that make sense still with your change: https://github.com/ahsonkhan/azure-sdk-for-c/blob/7ad50097b2eb8708345970440306f4b59b3551b6/sdk/samples/iot/CMakeLists.txt

sdk/samples/iot/generate_certificate.cmd
```diff
@@ -21,30 +21,29 @@ type device_ec_cert.pem > device_cert_store.pem
...
+echo Use the following fingerprint when creating your device in IoT Hub.
+echo (The fingerprint has also been placed in fingerprint.txt for future reference.)
 echo.
-echo Use the following fingerprint when creating your device in IoT Hub:
 
 FOR /F "tokens=*" %%a in ('openssl x509 -noout -fingerprint -in device_ec_cert.pem') do SET output=%%a
 set fingerprint=%output::=%
 
-echo %fingerprint%
+echo 	%fingerprint%
```
ahsonkhan, line 47: nit: remove extra space

sdk/samples/iot/CMakeLists.txt
momuno: looks like my commit from yesterday did this.  fixing.

PR #968, opened 2020-07-29
update vscode directory

### LINKED ISSUE #916: Can't use VSCode with CMake

danewalton:
closes #916 

Even if we ignore the directory, because we added the files to the index, it'll keep tracking changes if users make changes. With this, users must copy the config files over to their `.vscode` directory where changes will not be tracked.


ahsonkhan:
This is certainly an improvement, but it begs the question, what is preventing me from modifying the files in `.vscode-config` and accidentally pushing a commit with my sas key or client_secret?

danewalton:
@ahsonkhan its a good question. In theory, nothing. However in this case, because the files are useless in this new directory, it's less likely to happen; someone has to copy them over to get any use out of them. An alternative might be to take what is in these JSON files and put them in a README somewhere where someone could copy and paste from there, but it adds a bit of friction.

ahsonkhan:
> An alternative might be to take what is in these JSON files and put them in a README somewhere where someone could copy and paste from there, but it adds a bit of friction.

That might be better given the potential info disclosure issue, but I don't have a good sense of how much friction doing so would add, so don't have a strong preference one way or another atm.

Issue #636, opened 2020-04-28
Support national cloud authentication urls.


gilbertw:
This needs to be a configuration setting.  It can change based on which cloud the application is running.  By default we would use the above.
https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud

_Originally posted by @gilbertw in https://github.com/Azure/azure-sdk-for-c/pull/633/review_comment/create_

antkmsft:
Thank you Rick. Yes, I think it must not only be a config setting, but the client secret credential should be configurable for that, i.e. you should be able to create credentials that use different auth URLs.

gilbertw:
Needs to be addressed before GA

antkmsft:
Based on offline conversations, we can move this one for the next sprint.

PR #973, opened 2020-07-30
Add support for national cloud auth URLs

antkmsft:
Closes https://github.com/Azure/azure-sdk-for-c/issues/636

### DIFF
--- a/sdk/inc/azure/core/az_credentials.h
+++ b/sdk/inc/azure/core/az_credentials.h
@@ -104,6 +104,7 @@ typedef struct
     az_span client_id;
     az_span client_secret;
     az_span scopes;
+    az_span authority;
   } _internal;
 } az_credential_client_secret;
 
@@ -111,10 +112,11 @@ typedef struct
  * @brief az_credential_client_secret_init initializes an az_credential_client_secret instance
  * with the specified tenant ID, client ID and client secret.
  *
- * @param out_credential reference to a az_credential_client_secret instance to initialize
+ * @param out_credential reference to an az_credential_client_secret instance to initialize
  * @param tenant_id an Azure tenant ID
  * @param client_id an Azure client ID
  * @param client_secret an Azure client secret
+ *
  * @return An #az_result value indicating the result of the operation:
  *         - #AZ_OK if successful
  *         - Other error code if initialization failed
@@ -124,6 +126,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
     az_span tenant_id,
     az_span client_id,
     az_span client_secret);
+
+/**
+ * @brief sets authentication authority URL for the client secret credential.
+ *
+ * @param ref_credential reference to an az_credential_client_secret instance to set the authority
+ * URL for
+ * @param authority authentication authority URL to set
+ *
+ * @return An #az_result value indicating the result of the operation:
+ *         - #AZ_OK if successful
+ *         - Other error code if initialization failed
+ */
+AZ_NODISCARD az_result
+az_credential_set_authority(az_credential_client_secret* ref_credential, az_span authority);
 
 #include <azure/core/_az_cfg_suffix.h>
 
--- a/sdk/inc/azure/core/az_credentials.h
+++ b/sdk/inc/azure/core/az_credentials.h
@@ -128,10 +128,13 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
     az_span client_secret);
 
 /**
- * @brief Sets authentication authority URL for the client secret credential.
+ * @brief Initializes an az_credential_client_secret instance with the specified tenant ID, client
+ * ID, client secret, and Azure AD authority URL.
  *
- * @param ref_credential Reference to an #az_credential_client_secret instance to set the authority
- * URL for.
+ * @param out_credential Reference to an #az_credential_client_secret instance to initialize,
+ * @param tenant_id An Azure tenant ID.
+ * @param client_id An Azure client ID.
+ * @param client_secret An Azure client secret.
  * @param authority Authentication authority URL to set.
  *
  * @note #authority example: "https://login.microsoftonline.us/".
@@ -142,8 +145,12 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
  *         - #AZ_OK if successful.
  *         - Other error code if initialization failed.
  */
-AZ_NODISCARD az_result
-az_credential_set_authority(az_credential_client_secret* ref_credential, az_span authority);
+AZ_NODISCARD az_result az_credential_client_secret_init_with_authority(
+    az_credential_client_secret* out_credential,
+    az_span tenant_id,
+    az_span client_id,
+    az_span client_secret,
+    az_span authority);
 
 #include <azure/core/_az_cfg_suffix.h>
 
--- a/sdk/inc/azure/core/az_credentials.h
+++ b/sdk/inc/azure/core/az_credentials.h
@@ -109,17 +109,17 @@ typedef struct
 } az_credential_client_secret;
 
 /**
- * @brief az_credential_client_secret_init initializes an az_credential_client_secret instance
- * with the specified tenant ID, client ID and client secret.
+ * @brief Initializes an az_credential_client_secret instance with the specified tenant ID, client
+ * ID and client secret.
  *
- * @param out_credential reference to an az_credential_client_secret instance to initialize
- * @param tenant_id an Azure tenant ID
- * @param client_id an Azure client ID
- * @param client_secret an Azure client secret
+ * @param out_credential Reference to an #az_credential_client_secret instance to initialize,
+ * @param tenant_id An Azure tenant ID.
+ * @param client_id An Azure client ID.
+ * @param client_secret An Azure client secret.
  *
  * @return An #az_result value indicating the result of the operation:
- *         - #AZ_OK if successful
- *         - Other error code if initialization failed
+ *         - #AZ_OK if successful.
+ *         - Other error code if initialization failed.
  */
 AZ_NODISCARD az_result az_credential_client_secret_init(
     az_credential_client_secret* out_credential,
@@ -128,15 +128,19 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
     az_span client_secret);
 
 /**
- * @brief sets authentication authority URL for the client secret credential.
+ * @brief Sets authentication authority URL for the client secret credential.
  *
- * @param ref_credential reference to an az_credential_client_secret instance to set the authority
- * URL for
- * @param authority authentication authority URL to set
+ * @param ref_credential Reference to an #az_credential_client_secret instance to set the authority
+ * URL for.
+ * @param authority Authentication authority URL to set.
+ *
+ * @note #authority example: "https://login.microsoftonline.us/".
+ * See national clouds' Azure AD authentication endpoints:
+ * https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud.
  *
  * @return An #az_result value indicating the result of the operation:
- *         - #AZ_OK if successful
- *         - Other error code if initialization failed
+ *         - #AZ_OK if successful.
+ *         - Other error code if initialization failed.
  */
 AZ_NODISCARD az_result
 az_credential_set_authority(az_credential_client_secret* ref_credential, az_span authority);
--- a/sdk/src/azure/core/az_credential_client_secret.c
+++ b/sdk/src/azure/core/az_credential_client_secret.c
@@ -5,8 +5,9 @@
 #include "az_credential_token_private.h"
 #include <azure/core/az_credentials.h>
 #include <azure/core/az_http.h>
-#include <azure/core/internal/az_http_internal.h>
 #include <azure/core/az_http_transport.h>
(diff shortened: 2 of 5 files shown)


sdk/inc/azure/core/az_credentials.h
```diff
@@ -104,17 +104,19 @@ typedef struct
...
+    az_span authority;
   } _internal;
 } az_credential_client_secret;
 
 /**
  * @brief az_credential_client_secret_init initializes an az_credential_client_secret instance
  * with the specified tenant ID, client ID and client secret.
  *
- * @param out_credential reference to a az_credential_client_secret instance to initialize
+ * @param out_credential reference to an az_credential_client_secret instance to initialize
```
ahsonkhan, line 115: nit:
```suggestion
 * @param out_credential reference to an #az_credential_client_secret instance to initialize
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -125,6 +127,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
     az_span client_id,
     az_span client_secret);
 
+/**
+ * @brief sets authentication authority URL for the client secret credential.
```
ahsonkhan, line 131:
```suggestion
 * @brief Sets authentication authority URL for the client secret credential.
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -125,6 +127,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
...
+ *
+ * @param ref_credential reference to an az_credential_client_secret instance to set the authority
```
ahsonkhan, line 133: nit:
```suggestion
 * @param ref_credential reference to an #az_credential_client_secret instance to set the authority
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -125,6 +127,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
...
+ * URL for
+ * @param authority authentication authority URL to set
```
ahsonkhan, line 135: nit: Ideally these should be complete sentences (start with capital letter and end with period).

sdk/src/azure/core/az_credential_client_secret.c
```diff
@@ -88,6 +91,14 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
     az_span client_id,
     az_span client_secret)
 {
+  _az_PRECONDITION_NOT_NULL(out_credential);
+  _az_PRECONDITION_VALID_SPAN(tenant_id, 1, false);
+  _az_PRECONDITION_VALID_SPAN(client_id, 1, false);
+  _az_PRECONDITION_VALID_SPAN(client_secret, 1, false);
+
+  static az_span const microsoft_aad_authority
+      = AZ_SPAN_LITERAL_FROM_STR("https://login.microsoftonline.com/");
```
ahsonkhan, line 100: Can we define this once outside the method, similar to these:
https://github.com/Azure/azure-sdk-for-c/blob/355cfe5900f5186a4c94e1ad9ff7f0c5555b78be/sdk/src/azure/iot/az_iot_hub_client.c#L21-L22

sdk/src/azure/core/az_credential_client_secret.c
```diff
@@ -101,8 +112,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
...
   return AZ_OK;
 }
+
+AZ_NODISCARD az_result
+az_credential_set_authority(az_credential_client_secret* ref_credential, az_span authority)
+{
+  _az_PRECONDITION_NOT_NULL(ref_credential);
+  _az_PRECONDITION_VALID_SPAN(authority, 1, false);
+
+  ref_credential->_internal.authority = authority;
```
ahsonkhan, line 128: Should we try to validate the contents of the user-provided custom authority in some way?

sdk/tests/core/test_az_credential_client_secret.c
```diff
@@ -17,22 +17,34 @@
 
 #include <azure/core/_az_cfg.h>
 
+static az_span authority = AZ_SPAN_LITERAL_FROM_STR("https://login.microsoftonline.com/");
```
ahsonkhan, line 20:
```suggestion
static az_span msft_authority = AZ_SPAN_LITERAL_FROM_STR("https://login.microsoftonline.com/");
```

sdk/tests/core/test_az_credential_client_secret.c
```diff
@@ -17,22 +17,34 @@
...
+
 static void test_credential_client_secret(void** state)
 {
   (void)state;
+  az_span const authorities[] = { AZ_SPAN_FROM_STR("https://login.microsoftonline.com/"),
+                                  AZ_SPAN_FROM_STR("https://somesite.contoso.com/") };
```
ahsonkhan, line 26: Can we structure the tests so we can also validate that the user-provided authority is retained correctly. I don't see anywhere we are asserting that we still have `"https://somesite.contoso.com/"` in the `request_url`.

sdk/tests/core/test_az_credential_client_secret.c
```diff
@@ -41,63 +53,65 @@ static void test_credential_client_secret(void** state)
...
+    // invoked (it does not return previously set value).
+    int const clock_nrequests[] = {
+      2, // wait to retry, set token expiration
+      1, // check if token has expired
+      3, // check if token has expired, wait to retry, set token expiration
+      1, // check if token has expired
+    };
+
+    // Some value that is big enough so that when you add 3600000 milliseconds to it (1 hour),
+    // and while in debuger, the vallue you see is seen as "103600000", which is easy to debug.
```
ahsonkhan, line 73:
```suggestion
    // and while in debugger, the value you see is seen as "103600000", which is easy to debug.
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -125,6 +127,20 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
...
+ *
+ * @return An #az_result value indicating the result of the operation:
+ *         - #AZ_OK if successful
+ *         - Other error code if initialization failed
+ */
+AZ_NODISCARD az_result
+az_credential_set_authority(az_credential_client_secret* ref_credential, az_span authority);
```
ahsonkhan, line 142: It could be useful to the caller to know from the docs when to call this and what possible values to pass in. For example, maybe explain and add a link to the national cloud doc.

sdk/tests/core/test_az_credential_client_secret.c
antkmsft: We do -- see `assert_true(az_span_is_content_equal(az_auth_url, request_url));` - and this is why we should not be renaming `authority` to `msft_authority`.

sdk/tests/core/test_az_credential_client_secret.c
antkmsft: It is an initial value. Later, we set it to other value.

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: I can't think of anything, and also I haven't seen other SDKs doing any validation.

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: I don't think we should do it, if it is not used outside. If/when it is used outside, we absolutely should do it. If used locally, the more local definition is - the better.

sdk/inc/azure/core/az_credentials.h
antkmsft: Updated.

sdk/tests/core/test_az_credential_client_secret.c
antkmsft: Updated.

sdk/inc/azure/core/az_credentials.h
antkmsft: Fixed.

sdk/inc/azure/core/az_credentials.h
antkmsft: Fixed.

sdk/inc/azure/core/az_credentials.h
antkmsft: Fixed.

sdk/inc/azure/core/az_credentials.h
antkmsft: Fixed.

sdk/src/azure/core/az_credential_client_secret.c
ahsonkhan: Does that end up creating the span repeatedly each time the caller calls this method, or only once, as if it was declared outside as a static?

sdk/tests/core/test_az_credential_client_secret.c
ahsonkhan: Why do we have a global static that is being modified by one test, and then used in another? If the order in which those tests run were flipped, wouldn't that potentially break the test?

sdk/tests/core/test_az_credential_client_secret.c
```diff
@@ -121,9 +143,21 @@ az_result send_request(_az_http_request const* request, az_http_response* respon
...
+      az_span az_auth_url = AZ_SPAN_FROM_BUFFER(auth_url_buf);
+
+      {
+        az_span auth_url_remainder = az_span_copy(az_auth_url, authority);
+
+        auth_url_remainder
+            = az_span_copy(auth_url_remainder, AZ_SPAN_FROM_STR("TenantID/oauth2/v2.0/token"));
+
+        az_auth_url = az_span_slice(
+            az_auth_url, 0, (int32_t)(az_span_ptr(auth_url_remainder) - az_span_ptr(az_auth_url)));
```
ahsonkhan, line 157: nit: Correct me if I am mistaken, but I think this should use `az_span_size` rather than `az_span_ptr`.
```suggestion
        az_auth_url = az_span_slice(
            az_auth_url, 0, az_span_size(az_auth_url) - az_span_size(auth_url_remainder));
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -104,27 +104,54 @@ typedef struct
...
     az_span client_secret;
     az_span scopes;
+    az_span authority;
   } _internal;
 } az_credential_client_secret;
 
 /**
- * @brief az_credential_client_secret_init initializes an az_credential_client_secret instance
- * with the specified tenant ID, client ID and client secret.
+ * @brief Initializes an az_credential_client_secret instance with the specified tenant ID, client
```
ahsonkhan, line 112: Missed this:
```suggestion
 * @brief Initializes an #az_credential_client_secret instance with the specified tenant ID, client
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -104,27 +104,54 @@ typedef struct
...
+ * ID and client secret.
+ *
+ * @param out_credential Reference to an #az_credential_client_secret instance to initialize,
```
ahsonkhan, line 115:
```suggestion
 * @param out_credential Reference to an #az_credential_client_secret instance to initialize.
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -104,27 +104,54 @@ typedef struct
...
  *
- * @param out_credential reference to a az_credential_client_secret instance to initialize
- * @param tenant_id an Azure tenant ID
- * @param client_id an Azure client ID
- * @param client_secret an Azure client secret
  * @return An #az_result value indicating the result of the operation:
- *         - #AZ_OK if successful
- *         - Other error code if initialization failed
+ *         - #AZ_OK if successful.
+ *         - Other error code if initialization failed.
```
ahsonkhan, line 122: nit: since these are sentence fragments in lists, they don't need to end in a period. At least that's what I have seen and been following across all the other API docs.
```suggestion
 *         - #AZ_OK if successful
 *         - Other error code if initialization failed
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -104,27 +104,54 @@ typedef struct
...
 
+/**
+ * @brief Initializes an az_credential_client_secret instance with the specified tenant ID, client
+ * ID, client secret, and Azure AD authority URL.
+ *
+ * @param out_credential Reference to an #az_credential_client_secret instance to initialize,
+ * @param tenant_id An Azure tenant ID.
+ * @param client_id An Azure client ID.
+ * @param client_secret An Azure client secret.
+ * @param authority Authentication authority URL to set.
```
ahsonkhan, line 138: To confirm, is that what this is called and referred to as (`authentication authority URL`)? Search results don't seem to refer to it that way:
https://www.google.com/search?q=Authentication+authority+URL&rlz=1C1GCEU_enUS820US821&oq=Authentication+authority+URL&aqs=chrome..69i57&sourceid=chrome&ie=UTF-8

Maybe it should be something like the following (granted, I am not the domain expert):
- Authority URL to set, used for authentication.
-  Authenticated authority URL to set.

Issue #934, opened 2020-07-16
Add json api to add generic, prebuilt JSON span to a document


danewalton:
@ahsonkhan for eyes.

Imagine we have a span which already contains valid JSON that we want to add to a document. We currently can only add per element. It would be nice to have an api to append something already built up. Especially useful in IoT where we might have some "envelope" for a JSON payload and we just want to let the user build up their "value" and append it in our JSON envelope.

Something like the following where the "value" could be any of the three:
```json
{
  "component": {
    "name": "my_component",
    "value": 45
  }
}
```
```json
{
  "component": {
    "name": "my_component",
    "value": "my_value"
  }
}
```
```json
{
  "component": {
    "name": "my_component",
    "value": {
       "foo": 45,
        "bar": "example",
    }
  }
}
```

The API for us would be simplified with the following:
```c
az_result pnp_helper_create_reported_property(
    az_span component_name,
    az_span property_name,
    az_span property_json_value);
```
instead of needing a separate API for each type (double, int, string, object, etc).

hihigupt:
We are also hitting issue when we are trying to write complex object using the json writer.

PR #978, opened 2020-08-03
add pnp component sample

danewalton:


sdk/samples/iot/hub/src/sample_pnp_component_mqtt.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+#ifndef SAMPLE_PNP_COMPONENT_MQTT_H
```
danewalton, line 3: prefix with `_az`

sdk/samples/iot/hub/src/sample_pnp_device_info_component.h
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
danewalton, line 3: needs `#ifdef` guards

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.h
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifndef SAMPLE_PNP_THERMOSTAT_COMPONENT_H
```
danewalton, line 4: prefix with `_az`

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.h
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_SAMPLE_PNP_THERMOSTAT_COMPONENT_H
+#define _az_SAMPLE_PNP_THERMOSTAT_COMPONENT_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
ahsonkhan, line 9: Why do we have this?

danewalton: might just be muscle memory from before. seems that `_az_cfg_prefix.h` has it too but didn't think to add that

danewalton: I mean should we be adding `_az_cfg_prefix.h` to sample headers?

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+
+      if ((len == (int32_t)az_span_size(sample_iot_hub_twin_desired_version))
+          && (memcmp(
+                  (void*)az_span_ptr(sample_iot_hub_twin_desired_version),
+                  (void*)scratch_buf,
+                  (size_t)az_span_size(sample_iot_hub_twin_desired_version))
+              == 0))
+      {
+        continue;
+      }
```
ahsonkhan, line 389: You should not do direct memcmp comparing byte-by-byte for property name look-ups within the JSON.
Use the built-in comparison API: `az_json_token_is_text_equal`

Other than correctness, there is an added benefit that you don't need to copy every token into your scratch buffer, even if they would be ignored.

sdk/samples/iot/hub/README.md
```diff
@@ -162,6 +162,16 @@ with correct values substituted for each field.
 - **Telemetry**: Device sends a JSON message with the field name `temperature` and the `double` value of the temperature.
 - **Twin**: Desired property with the field name `targetTemperature` and the `double` value for the desired temperature. Reported property with the field name `maxTempSinceLastReboot` and the `double` value for the highest temperature.Note that part of the PnP spec is a response to a desired property update from the service. The device will send back a reported property with a similarly named property and a set of "ack" values: `ac` for the HTTP-like ack code, `av` for ack version of the property, and an optional `ad` for an ack description.
 
+### [IoT Hub PnP Component (Certificates)][pnp_component_sample]
```
ericwol-msft, line 165:
```suggestion
### [IoT Hub PnP Multiple Components (Certificates)][pnp_component_sample]
```
??

sdk/samples/iot/hub/README.md
```diff
@@ -162,6 +162,16 @@ with correct values substituted for each field.
...
+
+Use a PnP device composed of multiple components. This temperature controller is made up of the following sub-components
```
ericwol-msft, line 167:
```suggestion
Use a PnP device which is composed of multiple components. This temperature controller is made up of the following sub-components
```

sdk/samples/iot/hub/README.md
```diff
@@ -162,6 +162,16 @@ with correct values substituted for each field.
...
+
+- Temperature Sensor 1
+- Temperature Sensor 2
+- Device Info
+
+Link to the component DTMI can be found [here](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json).
```
ericwol-msft, line 173:
```suggestion
Link to the device component DTMI can be found [here](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json).
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+
+#define TIMEOUT_WAIT_FOR_RECEIVE_MESSAGE_MS (8 * 1000)
+#define TIMEOUT_WAIT_FOR_COMPLETION_MS 1000
+#define TIMEOUT_MQTT_DISCONNECT_MS (10 * 1000)
+#define DEVICE_DO_WORK_SLEEP_MS 2
+#define TELEMETRY_SEND_INTERVAL 1
+#define DEFAULT_START_TEMP_CELSIUS 22.0
+#define DOUBLE_DECIMAL_PLACE_DIGITS 2
+
+bool device_operational = true;
```
ericwol-msft, line 87: verb? is_device_operational

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+static const uint8_t null_terminator = '\0';
+static char start_time_str[32];
+static az_span start_time_span;
+
+// * PnP Values *
+// The model id is the JSON document (also called the Digital Twins Model Identifier or DTMI)
```
ericwol-msft, line 93:
```suggestion
// The model id is a JSON document (also called the Digital Twins Model Identifier or DTMI)
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+// the functionality would need to match the DTMI and you would need to update the below 'model_id'.
+// Please see the sample README for more information on this DTMI.
+static const az_span model_id
+    = AZ_SPAN_LITERAL_FROM_STR("dtmi:com:example:TemperatureController;1");
+static sample_pnp_thermostat_component sample_thermostat_1;
+static const az_span sample_thermostat_1_component_name = AZ_SPAN_LITERAL_FROM_STR("thermostat1");
+static sample_pnp_thermostat_component sample_thermostat_2;
+static const az_span sample_thermostat_2_component_name = AZ_SPAN_LITERAL_FROM_STR("thermostat2");
+static const az_span sample_device_info_component = AZ_SPAN_LITERAL_FROM_STR("deviceInformation");
+static const az_span* sample_components[] = { &sample_thermostat_1_component_name,
```
ericwol-msft, line 105:
```suggestion
static const az_span* sample_pnp_components[] = { &sample_thermostat_1_component_name,
```

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.h
ahsonkhan: No, I don't think we should add non-public headers in samples.

I think my question should be directed to folks working on core. @antkmsft, do you recall why we have this and do we still need it?

https://github.com/Azure/azure-sdk-for-c/blob/f50b90403b8b370b55c14cfc60f6ebe013a32464/sdk/inc/azure/core/_az_cfg_prefix.h#L15-L18

Looks like it was added in https://github.com/Azure/azure-sdk-for-c/pull/31/files#diff-9bea604bc146230b0c86a481fa559384R9-R11 and then moved to the config files: https://github.com/Azure/azure-sdk-for-c/pull/27/files#diff-ed9d3d869305cdb374fc7243b8c9c5c8R4

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+static az_span mqtt_url_suffix = AZ_SPAN_LITERAL_FROM_STR(":8883");
+
+// Reuse topic and payload buffers since API's are synchronous
+static char publish_topic[128];
+static char publish_payload[512];
+static sample_pnp_mqtt_message publish_message;
+
+// IoT Hub Command
+static const az_span reboot_command_name = AZ_SPAN_LITERAL_FROM_STR("reboot");
+static const az_span empty_json_payload = AZ_SPAN_LITERAL_FROM_STR("{}");
```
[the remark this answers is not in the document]
ericwol-msft: looks empty to me 
https://www.bing.com/search?q=empty+json+object&lvl=1&FORM=PMETIS&filters=ans%3a%22cvns%22+level%3a%220%22+mcid%3a%22bc110c9eff99431ea34099cf303b4803%22&idx=2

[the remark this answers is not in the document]
ahsonkhan: I think we are overloading the term `object`. In JSON, that specifically means 0 or more key-value pairs within curly braces.

There isn't really a general purpose concept of an "empty JSON payload". An empty JSON payload, i.e. a string of length 0, is considered invalid JSON.

Why isn't this an empty array `[]`, or `null`, or an empty string`""`? Potentially those too could be considered an "empty JSON payload".

So, if we really need a sentinel, could `AZ_SPAN_NULL work`? Essentially, I am asking, why do we know that the JSON is guaranteed to be a "JSON object" and hence we can start off with an empty "JSON object"?

If we want to stick with it, maybe rename the variable:
```suggestion
static const az_span empty_json_object = AZ_SPAN_LITERAL_FROM_STR("{}");
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+  // Connect to IoT Hub
+  connect_device();
+
+  // Subscribe to the necessary twin and commands topics to receive twin updates and responses
+  subscribe();
+
+  // Initialize PnP Components
+  components_init();
+
+  // Send device info once on start up
```
ericwol-msft, line 229:
```suggestion
  // On start up, send device info 
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+  send_device_info();
+
+  // Receive response for device info publish
+  mqtt_receive_message();
```
ericwol-msft, line 233: ok not to retry this if this fails?

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+    mqtt_receive_message();
+
+    // Send max temp for each component since boot if needed
+    if (sample_pnp_thermostat_get_max_temp_report(&client, &sample_thermostat_1, &publish_message))
+    {
+      mqtt_publish_message(publish_message.topic, publish_message.out_payload_span, 0);
+    }
+    if (sample_pnp_thermostat_get_max_temp_report(&client, &sample_thermostat_2, &publish_message))
+    {
+      mqtt_publish_message(publish_message.topic, publish_message.out_payload_span, 0);
```
ericwol-msft, line 251: #define the QOS?

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+
+    // Send a telemetry message
+    send_telemetry_message();
+  }
+
+  // Gracefully disconnect: send the disconnect packet and close the socket
+  if ((rc = MQTTClient_disconnect(mqtt_client, TIMEOUT_MQTT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
+  {
+    LOG_ERROR("Failed to disconnect MQTT client, return code %d", rc);
+    return rc;
```
ericwol-msft, line 262: return rc or exit rc?

ericwol-msft: or fall thru and call destroy?

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
ericwol-msft: move this into send_device_info()?

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+{
+  (void)user_context_callback;
+  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
+  {
+    LOG_ERROR(
+        "Property=%.*s arrived for Control component itself.  This does not support\
+                writeable properties on it (all properties are on subcomponents)",
+        az_span_size(property_name),
+        az_span_ptr(property_name));
+  }
```
ericwol-msft, line 461: Send a response so that the hub does not keep retrying and spamming us?

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+  az_json_reader json_reader;
+  az_span twin_payload_span;
+
+  if (message->payloadlen)
+  {
+    LOG_SUCCESS("Payload: %.*s", message->payloadlen, (char*)message->payload);
+    twin_payload_span = az_span_init((uint8_t*)message->payload, (int32_t)message->payloadlen);
+    if ((result = az_json_reader_init(&json_reader, twin_payload_span, NULL)) != AZ_OK)
+    {
+      LOG_ERROR("Could not initialize JSON reader");
```
ericwol-msft, line 516: return;

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+    if (az_failed(
+            result = az_iot_hub_client_methods_response_get_publish_topic(
+                &client,
+                command_request->request_id,
+                200,
+                mqtt_message->topic,
+                mqtt_message->topic_length,
+                NULL)))
+    {
+      LOG_ERROR("Unable to get twin document publish topic");
```
ericwol-msft, line 581:
```suggestion
      LOG_ERROR("Unable to get twin response publish topic");
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+      return result;
+    }
+
+    mqtt_message->out_payload_span = empty_json_payload;
+  }
+  else
+  {
+    result = AZ_ERROR_ITEM_NOT_FOUND;
```
ericwol-msft, line 589: LOG_ERROR("Unknown command %s", command_name);

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+      == AZ_OK)
+  {
+    LOG_SUCCESS(
+        "Successfully executed command %.*s on thermostat 2",
+        az_span_size(command_name),
+        az_span_ptr(command_name));
+
+    mqtt_publish_message(publish_message.topic, publish_message.out_payload_span, 0);
+
+    LOG_SUCCESS("Sent response");
```
ericwol-msft, line 651:
```suggestion
    LOG_SUCCESS("Sent command response");
```

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_PNP_HELPER_H
```
ericwol-msft, line 4:
```suggestion
#ifndef PNP_HELPER_H
```
Not an SDK header

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * @brief Callback which is called for each property found by the #pnp_helper_process_twin_data()
+ * API.
+ */
+typedef az_result (*pnp_append_property_callback)(az_json_writer* json_writer, void* context);
+
+/**
+ * @brief Gets the MQTT topic that must be used for device to cloud telemetry messages.
+ * @remark Telemetry MQTT Publish messages must have QoS At least once (1).
+ * @remark This topic can also be used to set the MQTT Will message in the Connect message.
```
ericwol-msft, line 37: hub supports will messages?

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+ */
+az_result pnp_helper_parse_command_name(
+    az_span component_command,
+    az_span* component_name,
+    az_span* command_name);
+
+/**
+ * @brief Build a reported property
+ *
+ * @param[in] json_buffer The span into which the json payload will be placed.
```
ericwol-msft, line 74: ??

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+ * @param[in] component_name The name of the component for the reported_property.
```
ericwol-msft, line 75:
```suggestion
 * @param[in] component_name The name of the component for the reported property.
```

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+ * @param[in] property_name The name of the property to which to send an update.
+ * @param[in] append_callback The user callback to invoke to add the property value.
+ * @param[in] context The user context which is passed to the callback.
+ * @param[in] ack_code The return value for the reported property.
+ * @param[in] ack_version The ack version for the reported property.
+ * @param[in] ack_description The optional description for the reported property.
+ * @param[out] out_span The #az_span pointer to the output json payload.
+ */
+az_result pnp_helper_create_reported_property_with_status(
```
ericwol-msft, line 102:
```suggestion
az_result pnp_create_reported_property_with_status(
```
Do we really need to call these helper? It's not like our SDK API names have helper in it.

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+    az_span property_name,
+    pnp_append_property_callback append_callback,
+    void* context,
+    az_span* out_span);
+
+/**
+ * @brief Build a reported property
+ *
+ * @param[in] json_buffer The span into which the json payload will be placed.
+ * @param[in] component_name The name of the component for the reported_property.
```
ericwol-msft, line 93:
```suggestion
 * @param[in] component_name The name of the component for the reported property.
```

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+az_result pnp_helper_create_reported_property(
+    az_span json_buffer,
+    az_span component_name,
+    az_span property_name,
+    pnp_append_property_callback append_callback,
+    void* context,
+    az_span* out_span);
+
+/**
+ * @brief Build a reported property
```
ericwol-msft, line 90:
```suggestion
 * @brief Build a reported property with the ack status
```

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,137 @@
...
+ * `sample_components_ptr`.
+ * @param[in] scratch_buf Scratch buffer to place parsed JSON values.
+ * @param[in] scratch_buf_len Length of `scratch_buf`.
+ * @param[in] property_callback The callback which is called on each twin property.
+ * @param[in] context_ptr Pointer to user context.
+ */
+az_result pnp_helper_process_twin_data(
+    az_json_reader* json_reader,
+    bool is_partial,
+    const az_span** sample_components_ptr,
```
ericwol-msft, line 130: or ?
```suggestion
    const az_span* sample_components_ptr[],
```

sdk/samples/iot/hub/src/sample_pnp_component_mqtt.c
```diff
@@ -0,0 +1,22 @@
...
+#include <stdint.h>
+
+#include <azure/core/az_span.h>
+
+#include "sample_pnp_component_mqtt.h"
+
+static int32_t request_id_int;
+static char request_id_buf[8];
+
+// Create request id span which increments source int each call. Capable of holding 8 digit number.
```
ericwol-msft, line 13:
```suggestion
// Create request id span which increments source integer each call. Capable of holding 8 digit number.
```

sdk/samples/iot/hub/src/sample_pnp_component_mqtt.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+#ifndef _az_SAMPLE_PNP_COMPONENT_MQTT_H
+#define _az_SAMPLE_PNP_COMPONENT_MQTT_H
```
ericwol-msft, line 4:
```suggestion
#define SAMPLE_PNP_COMPONENT_MQTT_H
```

sdk/samples/iot/hub/src/sample_pnp_component_mqtt.h
```diff
@@ -0,0 +1,22 @@
...
+typedef struct sample_pnp_mqtt_message_tag
+{
+  char* topic;
+  size_t topic_length;
+  size_t* out_topic_length;
+  az_span payload_span;
+  az_span out_payload_span;
+} sample_pnp_mqtt_message;
+
+// Create request id span which increments source int each call. Capable of holding 8 digit number.
```
ericwol-msft, line 19:
```suggestion
// Create a new request id span. which increments the source integer each call. Capable of holding 8 digit number.
```

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.h
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifndef _az_SAMPLE_PNP_THERMOSTAT_COMPONENT_H
+#define _az_SAMPLE_PNP_THERMOSTAT_COMPONENT_H
```
ericwol-msft, line 5:
```suggestion
#define SAMPLE_PNP_THERMOSTAT_COMPONENT_H
```

sdk/samples/iot/hub/src/sample_pnp_component_mqtt.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+#ifndef _az_SAMPLE_PNP_COMPONENT_MQTT_H
+#define _az_SAMPLE_PNP_COMPONENT_MQTT_H
+
```
hihigupt, line 5: vertical space?

sdk/samples/iot/hub/src/pnp_helper.h
```diff
@@ -0,0 +1,135 @@
...
+
+#include <stdint.h>
+
+#include <azure/core/az_json.h>
+#include <azure/core/az_result.h>
+#include <azure/core/az_span.h>
+#include <azure/iot/az_iot_hub_client.h>
+
+#include "sample_pnp_thermostat_component.h"
+
```
hihigupt, line 15: this is wrong dependency .. helper should not depend upon sample

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include <azure/core/az_json.h>
+#include <azure/core/az_result.h>
+#include <azure/core/az_span.h>
+#include <azure/iot/az_iot_hub_client.h>
+
+#include "pnp_helper.h"
```
hihigupt, line 12: generally prefered to be first line.

kartben:
Existential question of the day: shouldn't pnp_helper (or at least something very close to it) become part of the IoT SDK? 

It is completely platform-agnostic (except maybe it should LOG_XXX rather than printf?) and my guess is folks will end up copy-pasting a good chunk of it in their own apps, which might not be ideal. Or to put it differently, how can we (and do we want to) help folks not have to implement PnP conventions from scratch, esp. when they have multiple components and hence must start playing with `$.sub`, `__t: c`, etc.

sdk/samples/iot/hub/src/pnp_helper.h
kartben: Apparently it does! Today I learned something :) 
See https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#using-the-mqtt-protocol-directly-as-a-device

> The device app can specify a Will message in the CONNECT packet. The device app should use `devices/{device_id}/messages/events/ or devices/{device_id}/messages/events/{property_bag}` as the Will topic name to define Will messages to be forwarded as a telemetry message. In this case, if the network connection is closed, but a DISCONNECT packet was not previously received from the device, then IoT Hub sends the Will message supplied in the CONNECT packet to the telemetry channel. The telemetry channel can be either the default Events endpoint or a custom endpoint defined by IoT Hub routing. The message has the `iothub-MessageType` property with a value of `Will` assigned to it.

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+// IoT Hub Commands Values
+static const az_span report_command_name_span = AZ_SPAN_LITERAL_FROM_STR("getMaxMinReport");
+static const az_span report_max_temp_name_span = AZ_SPAN_LITERAL_FROM_STR("maxTemp");
+static const az_span report_min_temp_name_span = AZ_SPAN_LITERAL_FROM_STR("minTemp");
+static const az_span report_avg_temp_name_span = AZ_SPAN_LITERAL_FROM_STR("avgTemp");
+static const az_span report_start_time_name_span = AZ_SPAN_LITERAL_FROM_STR("startTime");
+static const az_span report_end_time_name_span = AZ_SPAN_LITERAL_FROM_STR("endTime");
+static const az_span report_error_payload = AZ_SPAN_LITERAL_FROM_STR("{}");
+static char end_time_buffer[32];
+static char incoming_since_value[32];
```
ericwol-msft, line 34: why does the one above has _buffer and this one does not?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+
+// Twin Values
+static const az_span desired_temp_property_name = AZ_SPAN_LITERAL_FROM_STR("targetTemperature");
```
ericwol-msft, line 37: why do you sometimes add _span at the end like above, and something not?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+  AZ_RETURN_IF_FAILED(az_json_writer_append_end_object(json_builder));
+
+  *response_payload = az_json_writer_get_bytes_used_in_destination(json_builder);
+
+  return AZ_OK;
+}
+
+// Invoke the command requested from the service. Here, it generates a report for max, min, and avg
+// temperatures.
+static az_result invoke_getMaxMinReport(
```
ericwol-msft, line 79:
```suggestion
static az_result invoke_getMaxMinReport_command(
```

ericwol-msft: ?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+  az_json_writer json_builder;
+  AZ_RETURN_IF_FAILED(az_json_writer_init(&json_builder, response, NULL));
+  AZ_RETURN_IF_FAILED(build_command_response_payload(
+      handle, &json_builder, start_time_span, end_time_span, out_response));
+
+  return AZ_OK;
+}
+
+static az_result build_telemetry_message(
+    sample_pnp_thermostat_component* handle,
```
ericwol-msft, line 120: Humm Handle? Context?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+
+  az_json_writer json_builder;
+  AZ_RETURN_IF_FAILED(az_json_writer_init(&json_builder, response, NULL));
+  AZ_RETURN_IF_FAILED(build_command_response_payload(
+      handle, &json_builder, start_time_span, end_time_span, out_response));
+
+  return AZ_OK;
+}
+
+static az_result build_telemetry_message(
```
ericwol-msft, line 119:
```suggestion
static az_result build_temperature_telemetry_message(
```
?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+  AZ_RETURN_IF_FAILED(az_json_writer_append_property_name(&json_builder, telemetry_name));
+  AZ_RETURN_IF_FAILED(az_json_writer_append_double(
+      &json_builder, handle->current_temperature, DOUBLE_DECIMAL_PLACE_DIGITS));
+  AZ_RETURN_IF_FAILED(az_json_writer_append_end_object(&json_builder));
+  *out_payload = az_json_writer_get_bytes_used_in_destination(&json_builder);
+
+  return AZ_OK;
+}
+
+static az_result append_double(az_json_writer* json_writer, void* context)
```
ericwol-msft, line 137:
```suggestion
static az_result append_double(az_json_writer* json_writer, void* value)
```

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+{
+  double value = *(double*)context;
+
+  return az_json_writer_append_double(json_writer, value, DOUBLE_DECIMAL_PLACE_DIGITS);
+}
+
+az_result sample_pnp_thermostat_init(
+    sample_pnp_thermostat_component* handle,
```
ericwol-msft, line 145:
```suggestion
    sample_pnp_thermostat_component* context,
```

ericwol-msft: ... just not a big fan of using handle

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+      return false;
+    }
+
+    handle->send_max_temp_property = false;
+
+    return true;
+  }
+  else
+  {
+    return false;
```
ericwol-msft, line 201: move to line 173 to avoid nesting?

sdk/samples/iot/hub/src/pnp_helper.h
danewalton: I forgot "helper" needs to be removed from this entirely. The V1 has it removed and needs to be done here

danewalton:
@kartben

> Existential question of the day: shouldn't pnp_helper (or at least something very close to it) become part of the IoT SDK?

Yes we have plans to formalize some of the functionality here. This serves to get PnP devices off the ground and show how non-component and component samples might work without PnP SDK intervention.

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    if (handle->current_temperature < handle->min_temperature)
+    {
+      handle->min_temperature = handle->current_temperature;
+    }
+
+    /* Increment the avg count, add the new temp to the total, and calculate the new avg */
+    handle->device_temperature_avg_count++;
+    handle->device_temperature_avg_total += handle->current_temperature;
```
ericwol-msft, line 283: possible overflow?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+    sample_pnp_thermostat_component* handle,
+    az_iot_hub_client_method_request* command_request,
+    az_span component_name,
+    az_span command_name,
+    az_span command_payload,
+    sample_pnp_mqtt_message* mqtt_message)
+{
+  az_result result;
+
+  if (az_span_is_content_equal(handle->component_name, component_name)
```
ericwol-msft, line 324: should this also send an error back to the hub service?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,363 @@
...
+    if (az_failed(
+            result = az_iot_hub_client_methods_response_get_publish_topic(
+                client,
+                command_request->request_id,
+                return_code,
+                mqtt_message->topic,
+                mqtt_message->topic_length,
+                mqtt_message->out_topic_length)))
+    {
+      printf("Unable to get twin document publish topic\n");
```
ericwol-msft, line 352:
```suggestion
      printf("Unable to get method response publish topic\n");
```

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+    az_json_token* property_value,
+    int32_t version,
+    void* user_context_callback)
+{
+  (void)user_context_callback;
+  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
+  {
+    LOG_ERROR(
+        "Property=%.*s arrived for Control component itself.  This does not support\
+                writeable properties on it (all properties are on subcomponents)",
```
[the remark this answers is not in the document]
danewalton: fun fact. writeable can be written with or without an `e` :) Sub-components i agree

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+#ifdef _MSC_VER
+// warning C4201: nonstandard extension used: nameless struct/union
+#pragma warning(disable : 4201)
```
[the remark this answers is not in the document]
danewalton: Changed to push and pop. I could push a commit to the build system to see what triggers it but I don't remember off the top of my head.

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+#include <azure/iot/az_iot_hub_client.h>
+
+#include "pnp_helper.h"
+#include "sample_pnp_component_mqtt.h"
+#include "sample_pnp_device_info_component.h"
+#include "sample_pnp_thermostat_component.h"
+
+#ifdef _MSC_VER
+// "'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead."
+#pragma warning(disable : 4996)
```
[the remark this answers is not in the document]
danewalton: I do not believe we do in the samples

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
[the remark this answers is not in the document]
danewalton: The service expects a JSON document. The convention we have used is the empty curly braces. [Example](https://github.com/Azure/azure-iot-sdk-c/blob/b3a3fe39aa95b4527cea691201f3e0518559a27c/iothub_client/samples/pnp/pnp_simple_thermostat/pnp_simple_thermostat.c#L94)

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+// Callback invoked by pnp functions each time it finds a property in the twin document
+static void sample_property_callback(
+    az_span component_name,
+    az_span property_name,
+    az_json_token* property_value,
+    int32_t version,
+    void* user_context_callback)
+{
+  (void)user_context_callback;
+  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
```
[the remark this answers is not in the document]
danewalton: In case a user passes a span that is has a valid address and a length of zero. We wouldn't want a component name with that property.

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+  {
+    LOG_ERROR(
+        "Property=%.*s arrived for Control component itself.  This does not support\
+                writeable properties on it (all properties are on subcomponents)",
+        az_span_size(property_name),
+        az_span_ptr(property_name));
```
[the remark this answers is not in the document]
danewalton: No. It's a static function so those are checked before getting here. It is possible for the component name to be NULL which is an unexpected state which we log.

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: updated

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: updated

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: good call

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: changed to exit

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: yup thanks

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: updated to `methods response`

sdk/samples/iot/hub/src/pnp_helper.h
danewalton: true

sdk/samples/iot/hub/src/pnp_helper.h
danewalton: 🎉

sdk/samples/iot/hub/src/pnp_helper.h
danewalton: by convention elsewhere (v1 etc) it's usually been a double pointer. i'll probably keep as is for now unless someone has a good reason to switch

sdk/samples/iot/hub/src/pnp_helper.h
danewalton: you're right thank you

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: done

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: because I'm inconsistent :)

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: removed the `_span`

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: ehh other samples have it as this so I'm partial to leave it for now

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+  if ((result = sample_pnp_thermostat_get_telemetry_message(&client, &sample_thermostat_2, &publish_message))
+      != AZ_OK)
+  {
+    LOG_ERROR("Error getting message and topic for thermostat 2");
+    exit(result);
+  }
+
+  LOG_SUCCESS("Sending Telemetry Message for thermostat 2");
+
+  mqtt_publish_message(publish_message.topic, publish_message.out_payload_span, 0);
```
hihigupt, line 855: What about working set telemetry? 
[https://github.com/Azure/opendigitaltwins-dtdl/blob/39ad8b17812e973e69a3605041545e79f0cd34f9/DTDL/v2/samples/TemperatureController.json#L12](https://github.com/Azure/opendigitaltwins-dtdl/blob/39ad8b17812e973e69a3605041545e79f0cd34f9/DTDL/v2/samples/TemperatureController.json#L12)

danewalton: Crap yea. I think I also need the serial number

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+        return AZ_ERROR_UNEXPECTED_CHAR;
+      }
+
+      if (az_failed(az_json_reader_next_token(json_reader)))
+      {
+        printf("Failed to get next token\r\n");
+        return AZ_ERROR_UNEXPECTED_CHAR;
+      }
+
+      if (memcmp(
```
hihigupt, line 60: >memcmp [](start = 10, length = 6)

can we read more then what is valid in scratch buffer?

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+    *component_name = az_span_slice(component_command, 0, index);
+    *pnp_command_name
+        = az_span_slice(component_command, index + 1, az_span_size(component_command));
+  }
+  else
+  {
+    *component_name = AZ_SPAN_NULL;
+    *pnp_command_name = component_command;
+  }
+  return AZ_OK;
```
hihigupt, line 227: add vertical space

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+    az_span component_name,
+    az_span property_name,
+    pnp_append_property_callback append_callback,
+    void* context,
+    az_span* out_span)
+{
+  az_result result;
+
+  az_json_writer json_writer;
+  result = az_json_writer_init(&json_writer, json_buffer, NULL);
```
hihigupt, line 242: >result [](start = 2, length = 6)

what if it fails?

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+    void* context,
+    int32_t ack_code,
+    int32_t ack_version,
+    az_span ack_description,
+    az_span* out_span)
+{
+  az_result result;
+
+  az_json_writer json_writer;
+  result = az_json_writer_init(&json_writer, json_buffer, NULL);
```
hihigupt, line 287: >result [](start = 2, length = 6)

same

sdk/samples/iot/hub/src/pnp_helper.c
```diff
@@ -0,0 +1,438 @@
...
+    pnp_append_property_callback append_callback,
+    void* context,
+    az_span* out_span)
+{
+  az_result result;
+
+  az_json_writer json_writer;
+  result = az_json_writer_init(&json_writer, json_buffer, NULL);
+
+  bool has_component = az_span_ptr(component_name) != NULL;
```
hihigupt, line 244: >has_component [](start = 7, length = 13)

use sample convention as in line 289.

sdk/samples/iot/hub/src/sample_pnp_device_info_component.h
```diff
@@ -0,0 +1,21 @@
...
+#include <azure/core/az_result.h>
+#include <azure/core/az_span.h>
+#include <azure/iot/az_iot_hub_client.h>
+
+/**
+ * @brief Get the payload to send for device info
+ *
+ * @param client The `az_iot_hub_client` pointer to the hub client.
+ * @param mqtt_message The `sample_pnp_mqtt_message` pointer to the struct to be populated.
+ */
```
hihigupt, line 16: @ret?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,368 @@
...
+
+static az_result build_command_response_payload(
+    sample_pnp_thermostat_component* handle,
+    az_json_writer* json_builder,
+    az_span start_time_span,
+    az_span end_time_span,
+    az_span* response_payload)
+{
+  double avg_temp = handle->device_temperature_avg_total / handle->device_temperature_avg_count;
+  // Build the command response payload
```
hihigupt, line 54: vertical space

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,368 @@
...
+static const char iso_spec_time_format[] = "%Y-%m-%dT%H:%M:%S%z";
+
+static az_result build_command_response_payload(
+    sample_pnp_thermostat_component* handle,
+    az_json_writer* json_builder,
+    az_span start_time_span,
+    az_span end_time_span,
+    az_span* response_payload)
+{
+  double avg_temp = handle->device_temperature_avg_total / handle->device_temperature_avg_count;
```
hihigupt, line 53: >device_temperature_avg_count [](start = 67, length = 28)

total update?

hihigupt, line 53: >device_temperature_avg_total [](start = 28, length = 28)

allTemperatures

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,368 @@
...
+  }
+}
+
+az_result sample_pnp_thermostat_get_telemetry_message(
+    az_iot_hub_client* client,
+    sample_pnp_thermostat_component* handle,
+    sample_pnp_mqtt_message* mqtt_message)
+{
+  az_result result;
+  if (az_failed(
```
hihigupt, line 214: >if (az_failed( [](start = 2, length = 14)

should we use RETURN_IF_FAILED?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,368 @@
...
+              NULL,
+              handle->component_name,
+              mqtt_message->topic,
+              mqtt_message->topic_length,
+              NULL)))
+  {
+    return result;
+  }
+
+  if (az_failed(
```
hihigupt, line 226: >az_failed [](start = 6, length = 9)

same

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,368 @@
...
+    if (response != AZ_OK)
+    {
+      return_code = 400;
+    }
+    else
+    {
+      return_code = 200;
+    }
+
+    if (az_failed(
```
hihigupt, line 348: >az_failed [](start = 8, length = 9)

RETURN_IF_FAILED

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
danewalton: done

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: In theory yes but for the purposes of, unlikely.

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: done

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: good point wrapped in `AZ_RETURN_IF_FAILED`

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: wrapped

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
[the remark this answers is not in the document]
ahsonkhan: Yep, I understand, your call. Looks like this gets discussed often: https://github.com/seqan/seqan3/pull/1933#discussion_r446508506

I brought that up because we have existing usages of `writable` in the SDK but no usage of `writeable`.
https://github.com/Azure/azure-sdk-for-c/search?q=writable&unscoped_q=writable

https://github.com/Azure/azure-sdk-for-c/blob/43e4de8dcac97b144c24e0693028c88a17b98b91/sdk/src/azure/platform/az_curl.c#L101

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: done

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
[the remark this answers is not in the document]
ahsonkhan: Maybe we should, unless there is a reason for certain warnings following the push/pop model and this one we leave disabled?

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: I believe the other pnp sample in embedded has it this way so I will likely keep it

danewalton: same as above. other pnp sample is named similarly

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: added a log message

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: wouldn't have the log message printed if we did

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.c
danewalton: the return code would not be printed if we did that

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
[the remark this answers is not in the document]
ahsonkhan: That makes sense, which is why checking for `az_span_size(...) == 0` should be able to capture that case. I don't think we need the `ptr == NULL` check.

The only expected time when that is true, size should also be 0. Ideally, you as the caller, shouldn't need to make a distinction between span containing a null ptr and an empty span (other than rare corner cases). An empty span check is sufficient to cover both cases.

Also, I have tried to advocate reducing calls to `az_span_ptr` and `az_span_init` as much as possible, to only when necessary, to reduce the number of places where we have pointer access and moves, making reviewing code and auditing call sites easier.

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: Alright changed it to check with `az_json_token_get_string` first. If matches, it advances and `continue`. Else it `_get_string`'s

sdk/samples/iot/hub/src/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,859 @@
...
+      + az_span_size(mqtt_url_suffix) + (int32_t)sizeof(null_terminator);
+
+  if (required_size > destination_size)
+  {
+    return AZ_ERROR_INSUFFICIENT_SPAN_SIZE;
+  }
+
+  az_span destination_span = az_span_init((uint8_t*)destination, destination_size);
+  az_span remainder = az_span_copy(destination_span, mqtt_url_prefix);
+  remainder = az_span_copy(remainder, az_span_slice(iot_hub, 0, iot_hub_length));
```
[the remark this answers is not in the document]
danewalton: gonna be honest do not remember the logic there (why it was done that way). This is copied from the other samples so they should probably be updated as well

sdk/samples/iot/hub/src/pnp_helper.c
danewalton: updated logic to use `az_json_token_is_text_equal`

sdk/samples/iot/hub/src/sample_pnp_thermostat_component.h
antkmsft: `extern "C"` allows customers to include our headers from C++ compilers and build/link them.
This is C SDK, but I don't think we should drop having the ability to be interact with C++. Sometimes phave C++ compilers, but use it as "advanced C", or even just use the C89 subset of features.

ahsonkhan: Thanks @antkmsft. Have we recently tried to build/link the SDK using a C++ compiler? We should check if that actually works if that is a use case we want to continue to support.

sdk/samples/iot/sample_pnp.c
```diff
@@ -0,0 +1,428 @@
...
+
+  az_iot_hub_client_properties pnp_properties;
+
+  if (properties == NULL)
+  {
+    AZ_RETURN_IF_FAILED(az_iot_hub_client_properties_init(
+        &pnp_properties, AZ_SPAN_FROM_BUFFER(pnp_properties_buffer), 0));
+  }
+
+  if ((result = az_iot_hub_client_properties_append(
```
ahsonkhan, line 182: Use `az_failed` and `az_succeeded` methods rather than direct comparison to AZ_OK. We may add more success conditions in the future (in fact, we currently have one that is unused, AZ_CONTINUE).

This applies across the board and is the documented recommendation.
https://github.com/Azure/azure-sdk-for-c/blob/df4b7e032abc8adda8e1c27026d2800fbd76c178/sdk/docs/core/README.md#function-results

sdk/samples/iot/sample_pnp_thermostat_component.c
```diff
@@ -0,0 +1,369 @@
...
+  AZ_RETURN_IF_FAILED(az_json_writer_append_property_name(&json_builder, telemetry_name));
+  AZ_RETURN_IF_FAILED(az_json_writer_append_double(
+      &json_builder, thermostat_component->current_temperature, DOUBLE_DECIMAL_PLACE_DIGITS));
+  AZ_RETURN_IF_FAILED(az_json_writer_append_end_object(&json_builder));
+  *out_payload = az_json_writer_get_bytes_used_in_destination(&json_builder);
+
+  return AZ_OK;
+}
+
+static az_result append_double(az_json_writer* json_writer, void* value)
```
ahsonkhan, line 141: Where are we getting doubles as `void*` from?

sdk/samples/iot/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,870 @@
...
+  az_result result;
+
+  az_json_reader json_reader;
+  az_span twin_payload_span;
+
+  if (message->payloadlen)
+  {
+    LOG_SUCCESS("Payload: %.*s", message->payloadlen, (char*)message->payload);
+    twin_payload_span = az_span_init((uint8_t*)message->payload, (int32_t)message->payloadlen);
+    if ((result = az_json_reader_init(&json_reader, twin_payload_span, NULL)) != AZ_OK)
```
ahsonkhan, line 519: Same here/elsewhere. Don't do direct comparison to a particular `az_result` if all you care about is whether the operation succeeded or failed.

sdk/samples/iot/sample_pnp.c
```diff
@@ -0,0 +1,428 @@
...
+        printf("Failed to get next token\r\n");
+        return AZ_ERROR_UNEXPECTED_CHAR;
+      }
+
+      property_callback(
+          component_name,
+          az_span_init((uint8_t*)scratch_buf, len),
+          &(json_reader->token),
+          version,
+          context_ptr);
```
ahsonkhan, line 77: I am having trouble understanding the flow here. Do we need to give the unescaped JSON string token as a span back to the user, or is giving them just the unmodified slice of the input they provided, good enough?

If it's the latter, just save and return back the token OR the span from the token, `&(json_reader->token.slice`.

sdk/samples/iot/sample_pnp.c
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  while (index < sample_components_num)
+  {
+    if ((az_span_size(component_name) == az_span_size(*sample_components_ptr[index]))
+        && (memcmp(
+                (void*)az_span_ptr(component_name),
+                (void*)az_span_ptr(*sample_components_ptr[index]),
+                (size_t)az_span_size(component_name))
+            == 0))
```
ahsonkhan, line 151: Try to limit uses of memcmp and other ptr based methods as much as possible and try to stay in the "span world" where possible.

In this case, can you use `az_span_is_content_equal`?
This is essentially re-implementing parts of the existing API, so we should avoid that.

sdk/samples/iot/sample_pnp.c
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  copy_json_reader = *json_reader;
+  if ((sample_json_child_token_move(&copy_json_reader, sample_iot_hub_twin_desired_version)
+       != AZ_OK)
+      || az_failed(az_json_token_get_int32(&(copy_json_reader.token), (int32_t*)&version)))
+  {
+    printf("Failed to get version\r\n");
+    return AZ_ERROR_UNEXPECTED_CHAR;
+  }
```
ahsonkhan, line 352: This is interesting. We are theoretically doing two passes over the data. I wonder if this can be written differently, but don't fully understand what the intended goal is yet.

sdk/samples/iot/sample_pnp.h
```diff
@@ -0,0 +1,135 @@
...
+
+#define PNP_STATUS_SUCCESS 200
+#define PNP_STATUS_BAD_FORMAT 400
+#define PNP_STATUS_NOT_FOUND 404
+#define PNP_STATUS_INTERNAL_ERROR 500
+
+typedef void (*pnp_property_callback)(
+    az_span component_name,
+    az_span property_name,
+    az_json_token* property_value,
```
hihigupt, line 22: >az_json_token [](start = 4, length = 13)

How about sharing the copy of reader with caller, rather then token. In this way if callback can also parse complex objects.

sdk/samples/iot/sample_pnp.c
ahsonkhan: Follow up snippet from offline conversation.

Returning `az_span property_name` in the callback to the user (currently in PR):
1) az_span_is_content_equal(property_name, AZ_SPAN_FROM_STR("MY LITERAL"));
2) Not needed, because you, the pnp sample, already did it for me.

Returning `az_json_token property_name` in the callback to the user (proposal):
1) az_json_token_text_equal(property_name, AZ_SPAN_FROM_STR("MY LITERAL"))
2) az_json_token_get_string(..., scratch buffer)

sdk/samples/iot/sample_pnp_thermostat_component.c
danewalton: Callback from the user. The assumption here is if someone passes the `append_double` callback as their callback of choice, then their callback context will be a double pointer.

sdk/samples/iot/sample_pnp.c
danewalton: I think I had a discussion with Cris about this before. Whether we do a direct comparison or not, were an additional success case to be added, the resulting behavior might not be as the user expects. Lets assume this flow expects an `AZ_OK` and were `AZ_CONTINUE` to be added, it would want to take a different action: `az_succeeded` would cause unintended behavior. I'm not necessarily disagreeing about changing it to `az_succeeded`, but as long as we recognize that this doesn't necessarily give "more correct" behavior.

sdk/samples/iot/sample_pnp_thermostat_component.c
ahsonkhan: Then, what's stopping us from changing this to be a follows?
```suggestion
static az_result append_double(az_json_writer* json_writer, double* value)
```

sdk/samples/iot/sample_pnp.c
ahsonkhan: > but as long as we recognize that this doesn't necessarily give "more correct" behavior.

Fair. Do you agree that using `az_failed` make sense over comparing to a specific failure condition if the only intention is to check whether something failed or succeeded?

In theory, `az_succeeded` is just `!az_failed` and vice versa.

Also, for the sake of argument, let's say we add AZ_SUCCESS `az_result` tomorrow and change all APIs to return that to mean success (not that we would), now the samples that relied on these hard coded comparisons would break.

sdk/samples/iot/sample_pnp_thermostat_component.c
danewalton: because the callback is typedef'd to have an `az_json_writer*` and a `void*` as parameters. That way anything can be appended by the user. Say they want to append some complex object. They could do that as long as they passed something like `append_object` and their object as the void pointer. The callback would then know how to handle that.

TLDR: it makes the callback generic enough that the user can append anything instead of just a double.

sdk/samples/iot/sample_pnp.c
danewalton: Oh I patched them up. I agree it's cleaner and probably the better way to go. I just wanted to make sure we are aware of the side effects going one way or another.

sdk/samples/iot/sample_pnp.h
danewalton: For the purposes of, I'd say that might bring in more complex situations than we're looking for. If you disagree please let me know.

sdk/samples/iot/sample_pnp_component_mqtt.c
```diff
@@ -0,0 +1,23 @@
...
+
+static int32_t request_id_int;
+static char request_id_buf[8];
+
+// Create request id span which increments request id integer each call. Capable of holding 8 digit
+// number.
+az_span get_request_id(void)
+{
+  az_span remainder;
+  az_span out_span = az_span_init((uint8_t*)request_id_buf, sizeof(request_id_buf));
```
ahsonkhan, line 18: Can this span be a static created once rather than on every `get_request_id` call?

sdk/samples/iot/sample_pnp_component_mqtt.c
```diff
@@ -0,0 +1,23 @@
...
+  az_result result = az_span_i32toa(out_span, request_id_int++, &remainder);
+  (void)remainder;
+  (void)result;
```
ahsonkhan, line 21: What happens if `request_id_int` becomes too large and won't fit in a buffer of size 8? We are ignoring `AZ_ERROR_INSUFFICIENT_SPAN_SIZE`.

`2147483647` and `-2147483648` are length 10 and 11 respectively.

Maybe add a comment on why it is OK for us to ignore remainder and result.

Writing 100, and then 1, without clearing/accounting for reminder would have unexpected results. Our usage here works only because we are always incrementing request_id by one.

sdk/samples/iot/sample_pnp_component_mqtt.c
danewalton: In theory yes but we trade temporary stack space for static, permanent storage.

sdk/samples/iot/sample_pnp_component_mqtt.c
danewalton: Yea I can add some comments on the assumptions we are making. In this case, if this overflows, this sample has been running for a LONGGGGGGG time.

And yes about the last point but incrementing by one each time gets around that.

sdk/samples/iot/sample_pnp.c
hihigupt: how will this will look if property name is in discontinuous memory? 

---
In reply to: [465991913](https://github.com/Azure/azure-sdk-for-c/pull/978#discussion_r465991913) [](ancestors = 465991913)

sdk/samples/iot/sample_pnp_component_mqtt.c
danewalton: actually I'm going to change it to a `uint32_t` so if it does overflow it doesn't matter. If it does overflow then the message with `rid = 0` is long gone.

Also, how does 
```c
return az_span_slice(out_span, 0, (int32_t)(az_span_ptr(remainder) - az_span_ptr(out_span)));
```
sound to make sure that just the digits are returned?

sdk/samples/iot/sample_pnp.c
hihigupt: I guess your token API will take care of it.. right?

---
In reply to: [466077550](https://github.com/Azure/azure-sdk-for-c/pull/978#discussion_r466077550) [](ancestors = 466077550,465991913)

sdk/samples/iot/sample_pnp.c
hihigupt: Here goal is to read the version of the document and then pass this version to the callback with each property in the document.
e.g:

{
   "thermostat1": {
    "__t": "c",
    "targetTemperature": 6662
   },
   "thermostat2": {
    "__t": "c",
    "targetTemperature": 20022
   },
   "$version": 69
  }

Now we want to call user callback with following calls
callback(thermostat1, targetTemperature, 6662, 69)
callback(thermostat2, targetTemperature, 20022, 69)



---
In reply to: [465978572](https://github.com/Azure/azure-sdk-for-c/pull/978#discussion_r465978572) [](ancestors = 465978572)

sdk/samples/iot/sample_pnp.h
danewalton: updated to pass the json reader

sdk/samples/iot/sample_pnp.c
danewalton: I believe the problem is the inability to go backwards. Some data has different meanings depending on what follows it. So we have to look ahead to determine what to do with it.

sdk/samples/iot/sample_pnp_component_mqtt.c
ahsonkhan: That sounds good. But use `az_span_size`.

sdk/samples/iot/sample_pnp_component_mqtt.c
```diff
@@ -0,0 +1,25 @@
...
+az_span get_request_id(void)
+{
+  az_span remainder;
+  az_span out_span = az_span_init((uint8_t*)request_id_buf, sizeof(request_id_buf));
+
+  // Note that if left to run for a long time, this will overflow and reset back to 0.
+  az_result result = az_span_u32toa(out_span, request_id_int++, &remainder);
+  (void)result;
+
+  return az_span_slice(out_span, 0, (int32_t)(az_span_ptr(remainder) - az_span_ptr(out_span)));
```
ahsonkhan, line 24: Use `az_span_size`, just like `_az_span_diff`:
https://github.com/Azure/azure-sdk-for-c/blob/43e4de8dcac97b144c24e0693028c88a17b98b91/sdk/inc/azure/core/internal/az_span_internal.h#L18-L20

```suggestion
  return az_span_slice(out_span, 0, az_span_size(out_span) - az_span_size(remainder));
```

sdk/samples/iot/sample_pnp_component_mqtt.c
```diff
@@ -0,0 +1,25 @@
...
+static char request_id_buf[8];
+
+// Create request id span which increments request id integer each call. Capable of holding 8 digit
+// number.
+az_span get_request_id(void)
+{
+  az_span remainder;
+  az_span out_span = az_span_init((uint8_t*)request_id_buf, sizeof(request_id_buf));
+
+  // Note that if left to run for a long time, this will overflow and reset back to 0.
```
ahsonkhan, line 20: Do we want to increase `request_id_buf` to be of size 10 or 16 to cover all the bases (need 10, but a power of 2 might make sense)? `1_000_000_000` won't fit.

Or just add a comment saying this is for illustration purposes and we expect `request_id_int` to be a small number (for example, less than 1_000).

sdk/samples/iot/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,1004 @@
...
+
+  switch (value_token.kind)
+  {
+    case AZ_JSON_TOKEN_NUMBER:
+      AZ_RETURN_IF_FAILED(az_json_token_get_double(&value_token, &value_as_double));
+      AZ_RETURN_IF_FAILED(
+          az_json_writer_append_double(json_writer, value_as_double, DOUBLE_DECIMAL_PLACE_DIGITS));
+      break;
+    case AZ_JSON_TOKEN_STRING:
+      AZ_RETURN_IF_FAILED(az_json_writer_append_string(json_writer, value_token.slice));
```
ahsonkhan, line 466: As a follow up to our conversation, until [this](https://github.com/Azure/azure-sdk-for-c/issues/934) feature is in, I would first call `az_json_token_get_string(...)` before passing it to `az_json_writer_append_string`.

`az_json_writer_append_string` does escaping, so passing in a raw token slice which hasn't been unescaped, would change the output JSON.

For example: `"foo\u0061"` will become `"foo\\u0061"` when passed in to the writer. Embedded quotes will be a bigger problem as they would result in invalid JSON: `"Hi \"Dane\""` becomes `"Hi \\"Dane\\""` and now the string ends right after Hi.

On the other hand: getting the unescaped string `fooa` and then writing it out as `fooa` is OK because those are semantically equivalent strings.

sdk/samples/iot/paho_iot_hub_pnp_component_example.c
```diff
@@ -0,0 +1,1004 @@
...
+    az_json_token* property_name,
+    az_json_reader* property_value,
+    int32_t version,
+    void* user_context_callback)
+{
+  (void)user_context_callback;
+  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
+  {
+    LOG("Property=%.*s arrived for Control component itself. This does not support writable "
+        "properties on it(all properties are on sub - components) ",
```
ahsonkhan, line 540: strange spacing in `sub - components`

Was this intentional?

sdk/samples/iot/sample_pnp.h
```diff
@@ -0,0 +1,134 @@
...
+#define PNP_STATUS_INTERNAL_ERROR 500
+
+/**
+ * @brief Callback which is called for each property found by the #pnp_process_twin_data()
+ * API.
+ */
+typedef void (*pnp_property_callback)(
+    az_span component_name,
+    az_json_token* property_name,
+    az_json_reader* property_value,
```
ahsonkhan, line 26: Change the parameter name. `property_value` for a `az_json_reader` doesn't fit.

sdk/samples/iot/sample_pnp.c
ahsonkhan: Yes, but atm, this code is only working on a contiguous span buffer, and it owns the creation of the reader, so that isn't a concern here.

sdk/samples/iot/sample_pnp_component_mqtt.c
danewalton: makes sense increased to 10

sdk/samples/iot/sample_pnp_component_mqtt.c
```diff
@@ -21,5 +21,5 @@ az_span get_request_id(void)
   az_result result = az_span_u32toa(out_span, request_id_int++, &remainder);
   (void)result;
 
-  return az_span_slice(out_span, 0, (int32_t)(az_span_ptr(remainder) - az_span_ptr(out_span)));
+  return az_span_slice(out_span, 0, (int32_t)(az_span_size(out_span) - az_span_size(remainder)));
```
ahsonkhan, line 24: The int32_t should no longer be necessary.

sdk/samples/iot/sample_pnp.c
```diff
@@ -0,0 +1,388 @@
...
+
+      az_json_token property_name = json_reader->token;
+
+      if (az_failed(az_json_reader_next_token(json_reader)))
+      {
+        printf("Failed to get next token\r\n");
+        return AZ_ERROR_UNEXPECTED_CHAR;
+      }
+
+      copy_json_reader = *json_reader;
```
hihigupt, line 64: > copy_json_reader = *json_reader; [](start = 5, length = 33)

Just pass this as copy rather then pointer.

sdk/samples/iot/CMakeLists.txt
```diff
@@ -83,6 +83,29 @@ target_link_libraries(paho_iot_hub_pnp_sample
     az::iot::samples::common
 )
 
+# PnP Component Sample
+add_executable (paho_iot_hub_pnp_component_example
```
ahsonkhan, line 87: Change `example` to `sample` like the rest?
```suggestion
add_executable (paho_iot_hub_pnp_component_sample
```

danewalton: ahhh yes thank you
https://github.com/Azure/azure-sdk-for-c/pull/1018

PR #985, opened 2020-08-04
Update dropped changes from sample directory restructure

momuno:
add back in Wellington's changes from dropped readme files: pnp to IoT Plug and Play
add back in Ahson's changes from dropped CMakeList files: LNK 4098 workaround


ahsonkhan:
Are we sure this was the only change that was impacted by https://github.com/Azure/azure-sdk-for-c/pull/953? I am surprised GitHub didn't show this as a merge conflict, preventing the merge. Did the pull/merge from master just unintentionally override this change?

momuno:
> Are we sure this was the only change that was impacted by #953? I am surprised GitHub didn't show this as a merge conflict, preventing the merge. Did the pull/merge from master just unintentionally override this change?

I had a merge conflict and went through the files in conflict and resolved the merge.  These lines were missed somehow.  So yes, the merge from master unintentionally overrode this change.

sdk/samples/iot/CMakeLists.txt
```diff
@@ -133,4 +133,20 @@ target_link_libraries(paho_iot_provisioning_sas_sample
...
+        paho_iot_hub_sas_telemetry_sample
+        paho_iot_hub_twin_sample
+        paho_iot_provisioning_sample
+        paho_iot_provisioning_sas_sample
+        PROPERTIES LINK_FLAGS
+        "/NODEFAULTLIB:libcmtd.lib"
+    )
+endif()
+
+endif() # TRANSPORT_PAHO
```
ahsonkhan, line 152: nit: Needs a new line at eof.

momuno: In this same file, you initially removed the last new line at eof, so I left it as you had it. See: https://github.com/Azure/azure-sdk-for-c/commit/df4b7e032abc8adda8e1c27026d2800fbd76c178#diff-08e085d86a6c448fc1ded13223c00a1f
You have also previously done a nit on their being a new line at the eof. I actually kept it off here because I figured you would nit it.
Moving forward, I plan to add a new line at the eof, regardless of what kind of file it is.

ahsonkhan: Apologies, let me try to clear up the confusion.
We want **one** new line character at the end of the file, **not zero, not two**.

This file initially had **two**, so I removed **one of the two** in the diff you linked above, so we have **one**.

This PR removed another one and now we have **zero**.

You can see this if you look at the files view on GitHub, where we see the warning for the lack of newline:
https://github.com/Azure/azure-sdk-for-c/pull/985/files
image

> Moving forward, I plan to add a new line at the eof, regardless of what kind of file it is.

Right, this makes sense. Just don't add two :)

You don't want to have an extra new line beyond just ending the file with a **single** `\n` character. GitHub knows not to show that extra line in the file view. For example, open this file up in GitHub and IDEs and though there is **one** new line character at eof, which is what we want, IDEs show an extra line number and GitHub doesn't:
https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/src/azure/core/CMakeLists.txt#L86

**What editor/IDE are you using?**

In VSCode for example, file endings should look like this:
image

And **not** like either of these:
image
image

PR #987, opened 2020-08-05
Sample directory restructure - with linting

momuno:
Linted readme.md


ahsonkhan:
nit: Can we update the PR title to describe in more detail the intent and result of the change (rather than just the branch name)? It helps with looking back at the history of changes, and also to create the changelog updates later.

Some recent examples of titles that I think were good (tried to pick a variety of changes):
- Use custom, portable implementation of IEEE 754 compliant isfinite since some embedded platforms don't have it.
- Update MD files with new PnP naming convention
- Scope doxygen generation to include only *.h and *.md files 
- Update spacing and formatting of README markdowns by using a linter.
- Adding IoT Hub WebSockets APIs, fixing Y2k38 issue

sdk/samples/iot/readme.md
```diff
@@ -73,162 +84,198 @@ This section provides an overview of the different samples available to run and
...
+- File an issue via [Github Issues](https://github.com/Azure/azure-sdk-for-c/issues/new/choose).
+- Check [previous questions](https://stackoverflow.com/questions/tagged/azure+c) or ask new ones on StackOverflow using
   the `azure` and `c` tags.
 
 ## Contributing
 
 This project welcomes contributions and suggestions. Find more contributing details [here](https://github.com/Azure/azure-sdk-for-c/blob/master/CONTRIBUTING.md).
 
-### License
+## License
```
ahsonkhan, line 279: All the other markdown files have `License` at "h3", as a subheading of `Contributing` rather than a sibling heading.

https://github.com/Azure/azure-sdk-for-c/search?l=Markdown&q=license

I am wondering if we should leave this change as is for now, and fix it up separately across the board.

@RickWinter what do you think, should it be `## License`, where it is its own heading, or should it be `### License` and have it be a subheading under `## Contributing`?

momuno:
> nit: Can we update the PR title to describe in more detail the intent and result of the change (rather than just the branch name)? It helps with looking back at the history of changes, and also to create the changelog updates later. Something to keep in mind going forward. Bonus point for providing additional context or details in the PR description (where applicable, especially for larger/more complex changes that may not be backed by an existing issue).
> 
> Some recent examples of titles that I think were good (tried to pick a variety of changes):
> 
> * Use custom, portable implementation of IEEE 754 compliant isfinite since some embedded platforms don't have it.
> * Update MD files with new PnP naming convention
> * Scope doxygen generation to include only *.h and *.md files
> * Update spacing and formatting of README markdowns by using a linter.
> * Adding IoT Hub WebSockets APIs, fixing Y2k38 issue

yes, I usually do but missed this one.  Thanks.

sdk/samples/iot/readme.md
momuno: good find.  we can update across.
Also, I was able to apply all linter rules.

ahsonkhan: > I am wondering if we should leave this change as is for now, and fix it up separately across the board.

Sorry, let me clarify what I intended to say.

I was proposing leaving the heading as `### License`, unchanged, and only change it to `## License` in a separate PR which also has changes for all the license headings across the repo. That way we could discuss whether we want to take that change or not independently, while master remains in a consistent state.

Now, we are inconsistent. NBD though.

PR #959, opened 2020-07-25
Update spacing and formatting of README markdowns by using a linter.

ahsonkhan:
- **Updated uses of `pre-requisites` to the more common `prerequisites` for consistency.**
- **Removed unnecessary extra new-lines and added a single new-line at the end of files that needed it.**
- **Fixed up inconsistent numbering issue (where certain numbers were duplicated in lists) and references to those step numbers in text.**
- **Used [this](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) markdown linter extension in VS Code to fix up spacing and other consistency issues.**

Most of these changes should have little to no impact on how the markdown is rendered on GitHub.

[MD004](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md004) - Unordered list style
[MD007](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md007) - Unordered list indentation
[MD010](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md010---hard-tabs) - Hard tabs
[MD022](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md022) - Headings should be surrounded by blank lines
[MD029](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md029---ordered-list-item-prefix) - Ordered list item prefix
[MD030](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md030) - Spaces after list markers
[MD031](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md031) - Fenced code blocks should be surrounded by blank lines
[MD032](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md032) - Lists should be surrounded by blank lines

**Left to do:**
[MD046](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md046) - Code block style
[MD014](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md014) - Dollar signs used before commands without showing output
[MD026](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md026) - Trailing punctuation in heading
[MD033](https://github.com/DavidAnson/markdownlint/blob/v0.20.4/doc/Rules.md#md033) - Inline HTML
[...]

PR #988, opened 2020-08-05
API review feedback

### LINKED ISSUE #989: Tweak according to API review feedback (AntK's part)

antkmsft:
Closes #989 

 - API breaking changes:
  - `az_context`:
    - `key` and `value` are `const`.
    - `az_context_with_expiration()` is renamed to `az_context_create_with_expiration()`.
    - `az_context_with_value()` is renamed to `az_context_create_with_value()`.
    - `az_context_app` is renamed to `az_context_application`.
  - `az_credential_client_secret_init()` now takes fourth parameter, `authority`.
  - `az_http_policy_retry_options`:
    - `status_codes` now should be terminated by `AZ_HTTP_STATUS_CODE_END_OF_LIST`.
    - `max_retries` is now `int32_t` instead of `int16_t`.
  - `az_config.h`:
    -  `AZ_HTTP_REQUEST_URL_BUF_SIZE` renamed to `AZ_HTTP_REQUEST_URL_BUFFER_SIZE`.
    - `AZ_HTTP_REQUEST_BODY_BUF_SIZE` renamed to `AZ_HTTP_REQUEST_BODY_BUFFER_SIZE`.
    - `AZ_LOG_MSG_BUF_SIZE` renamed to `AZ_LOG_MESSAGE_BUFFER_SIZE`.


sdk/docs/core/README.md
```diff
@@ -143,10 +143,10 @@ Note however that cancellation is performed as a best effort; it is not guarante
 
    ```C
    // Some function creates a child with a 10-second expiration:
-   az_context child = az_context_with_expiration(&az_context_app, 10);
+   az_context child = az_context_create_with_expiration(&az_context_app, 10);
```
ahsonkhan, line 146: Looking at this again, this might warrant an update to the guidelines, because it recommends the format:
`az_<shortname>_<objname>_create_<noun>`
https://azure.github.io/azure-sdk/clang_design.html#clang-apisurface-standardized-verbs

That format works when the `noun` we are creating is different from the obj creating it, like `az_iot_create_client`
or `az_keyvault_keyclient_create_key`

In our case here, for us to follow the guidelines, we would need something like:
`az_context_create_context_with_expiration`

And that might be OK too, though seeing `context` twice in the method name is a bit cumbersome.

cc @barcharcraz

sdk/inc/azure/core/az_config.h
```diff
@@ -21,10 +21,10 @@
 
 enum
 {
-  AZ_HTTP_REQUEST_URL_BUF_SIZE = 2 * 1024, ///< Default maximum length a URL can have.
-  AZ_HTTP_REQUEST_BODY_BUF_SIZE = 1024, ///< Default maximum buffer size for an HTTP request body.
+  AZ_HTTP_REQUEST_URL_BUFFER_SIZE = 2 * 1024, ///< Maximum length a URL can have.
```
ahsonkhan, line 24: nit:
```suggestion
  AZ_HTTP_REQUEST_URL_BUFFER_SIZE = 2 * 1024, ///< The maximum length a URL can have.
```

sdk/inc/azure/core/az_config.h
```diff
@@ -21,10 +21,10 @@
...
+  AZ_HTTP_REQUEST_BODY_BUFFER_SIZE = 1024, ///< Maximum buffer size for an HTTP request body.
```
ahsonkhan, line 25: nit:
```suggestion
  AZ_HTTP_REQUEST_BODY_BUFFER_SIZE = 1024, ///< The maximum buffer size for an HTTP request body.
```

sdk/inc/azure/core/az_config.h
ahsonkhan: Also, let's be consistent in the docs.
```suggestion
  AZ_HTTP_REQUEST_URL_BUFFER_SIZE = 2 * 1024, ///< The maximum buffer size for a URL.
```

sdk/inc/azure/core/az_config.h
```diff
@@ -21,10 +21,10 @@
...
 
-  AZ_LOG_MSG_BUF_SIZE = 1024, ///< The maximum size of a log message.
+  AZ_LOG_MESSAGE_BUFFER_SIZE = 1024, ///< The maximum size of a log message.
```
ahsonkhan, line 27: nit: consistent wording across all three enum docs.
```suggestion
  AZ_LOG_MESSAGE_BUFFER_SIZE = 1024, ///< The maximum buffer size for a log message.
```

sdk/inc/azure/core/az_context.h
```diff
@@ -52,24 +52,24 @@ struct az_context
 extern az_context az_context_app;
 
 /**
- * @brief az_context_with_expiration creates a newexpiring az_context node that is a child of the
- * specified parent.
+ * @brief az_context_create_with_expiration creates a newexpiring az_context node that is a child of
```
ahsonkhan, line 55: I just noticed this while doing a repo-wide/docs spell check. The comment is weirdly worded and should be fixed. For example, what is `newexpiring`?

sdk/inc/azure/core/az_context.h
```diff
@@ -52,24 +52,24 @@ struct az_context
...
+az_context_create_with_expiration(az_context const* parent, int64_t expiration)
 {
   return (az_context){ ._internal = { .parent = ((parent != NULL) ? parent : &az_context_app),
                                       .expiration = expiration } };
 }
 
 /**
- * @brief az_context_with_value creates a new key/value az_context node that is a child of the
- * specified parent.
+ * @brief az_context_create_with_value creates a new key/value az_context node that is a child of
```
ahsonkhan, line 71: Since you are already modifying parts of the docs with this API and file, can you also just remove the redundant method name in the comments (limit it to just the code you are touching)? They aren't necessary.

See https://github.com/Azure/azure-sdk-for-c/issues/589

sdk/inc/azure/core/az_credentials.h
```diff
@@ -135,17 +123,18 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
  * @param tenant_id An Azure tenant ID.
  * @param client_id An Azure client ID.
  * @param client_secret An Azure client secret.
- * @param authority Authentication authority URL to set.
+ * @param authority Authentication authority URL to set. Passing AZ_SPAN_NULL initializes credential
```
ahsonkhan, line 126: nit:
```suggestion
 * @param authority Authentication authority URL to set. Passing #AZ_SPAN_NULL initializes credential
```

sdk/inc/azure/core/az_credentials.h
```diff
@@ -135,17 +123,18 @@ AZ_NODISCARD az_result az_credential_client_secret_init(
...
+ * with default authority (Azure AD global authority - "https://login.microsoftonline.com/")
  *
- * @note #authority example: "https://login.microsoftonline.us/".
+ * @note non-NULL #authority example: "https://login.microsoftonline.us/".
```
ahsonkhan, line 129: authority isn't a type, function or macro. It's a parameter name.

```suggestion
 * @note Example of a non-NULL \p authority string: "https://login.microsoftonline.us/".
```

sdk/inc/azure/core/az_http.h
```diff
@@ -31,8 +31,6 @@
  */
 typedef enum
 {
-  AZ_HTTP_STATUS_CODE_NONE = 0,
```
ahsonkhan, line 34: I thought we decided to keep this because it was needed for something.

sdk/inc/azure/core/az_http.h
```diff
@@ -184,10 +185,10 @@ AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* respons
  */
 typedef struct
 {
+  az_span reason_phrase;
```
ahsonkhan, line 188: Can you add a comment explaining why we have these fields in this order, so folks who might want to change this later know about the rationale?

sdk/inc/azure/core/az_http.h
```diff
@@ -114,10 +115,10 @@ typedef enum
  */
 typedef struct
 {
-  int16_t max_retries;
+  az_http_status_code const* status_codes;
   int32_t retry_delay_msec;
   int32_t max_retry_delay_msec;
-  az_http_status_code const* status_codes;
+  int16_t max_retries;
```
ahsonkhan, line 121:
```suggestion
  int32_t max_retries;
```

sdk/inc/azure/core/internal/az_spinlock_internal.h
```diff
@@ -4,7 +4,7 @@
 #ifndef _az_SPINLOCK_INTERNAL_H
 #define _az_SPINLOCK_INTERNAL_H
 
-#include <azure/core/_az_spinlock.h>
+#include <azure/core/az_credentials.h>
```
ahsonkhan, line 7: Hmm, I think we may want to revisit the move of spinlock to az_credential.h. I didn't realize it was used in internal headers like this.

Now, this seems odd, having spinlock seemingly rely on credentials.

sdk/src/azure/core/az_credential_client_secret.c
```diff
@@ -110,7 +96,10 @@ AZ_NODISCARD az_result az_credential_client_secret_init_with_authority(
   _az_PRECONDITION_VALID_SPAN(tenant_id, 1, false);
   _az_PRECONDITION_VALID_SPAN(client_id, 1, false);
   _az_PRECONDITION_VALID_SPAN(client_secret, 1, false);
-  _az_PRECONDITION_VALID_SPAN(authority, 1, false);
+  _az_PRECONDITION_VALID_SPAN(authority, 0, true);
+
+  static az_span const azure_ad_global_authority
+      = AZ_SPAN_LITERAL_FROM_STR("https://login.microsoftonline.com/");
```
ahsonkhan, line 102: Do we want to make this public?

Also, does this name make more sense?
```suggestion
  static az_span const default_azure_ad_global_authority
      = AZ_SPAN_LITERAL_FROM_STR("https://login.microsoftonline.com/");
```

sdk/src/azure/core/az_credential_client_secret.c
```diff
@@ -125,7 +114,7 @@ AZ_NODISCARD az_result az_credential_client_secret_init_with_authority(
         .client_id = client_id,
         .client_secret = client_secret,
         .scopes = { 0 },
-        .authority = authority,
+        .authority = az_span_size(authority) > 0 ? authority : azure_ad_global_authority,
```
ahsonkhan, line 117: Nice!

sdk/src/azure/platform/az_curl.c
```diff
@@ -305,8 +305,7 @@ static AZ_NODISCARD az_result _az_http_client_curl_send_get_request(CURL* ref_cu
 /**
  * handles DELETE request
  */
-static AZ_NODISCARD az_result
-_az_http_client_curl_send_delete_request(CURL* ref_curl)
+static AZ_NODISCARD az_result _az_http_client_curl_send_delete_request(CURL* ref_curl)
```
ahsonkhan, line 308: nit: It's fine here, but ideally, such formatting changes on files that you would otherwise not be included in a PR should be done separately.

sdk/docs/core/README.md
antkmsft: It also means `az_span_create_span_from_str`. I'd add a rule so that we can omit object name when it is a constructor, i.e. the object it creates is the same. But I'd go as we decide. If we don't reach a decision here within a day, I'd remove this rename from this PR.

sdk/inc/azure/core/az_context.h
antkmsft: @RickWinter , what should we rename it to?

sdk/inc/azure/core/internal/az_spinlock_internal.h
antkmsft: I knew we run into this, and didn't see an issue in this. Let's ask @JeffreyRichter?

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: Not sure about this. Default *is* Azure AD global authority, Azure AD global authority is default. Azure AD global authority is the only one, there is no non-default Azure AD global authority. So I'd pick either `default_authority`, or `azure_ad_global_authority`, but not both. I think the way it is used, and from the comments is obvious it is a default value, so I'dpick the more descriptive name from the two.

sdk/src/azure/platform/az_curl.c
antkmsft: It's just whenever I modify a file, before saving, I run auto-format (ctrl+K-D). I think separate PR is fine, but we always should be auto-formatting.

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: Terminology, see: https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-national-cloud

sdk/src/azure/platform/az_curl.c
ahsonkhan: I agree, and that makes sense (I do the same). But this PR isn't otherwise modifying `az_curl.c`.

ahsonkhan:
Just in case you missed it, there are some more comments (but they are hidden/collapsed):
image

sdk/docs/core/README.md
antkmsft: I think we shouldn't go that route, because should we apply that naming rule, `az_span_create` would be named `az_span_create_span`, and it wont be possible to name a function `az_X_create` - it will always have to be `az_X_create_X`. And later overloads will be `az_X_create_X_with_Y` / `_from_Y`.

sdk/inc/azure/core/az_context.h
antkmsft: TBH, having obsessive idealist tendencies, I would much rather just run through all comments and do it all at once, including `@retval`s `#AZ_OK`s, `[in]/[out]`s etc. But I'll do this one file, or otherwise the PR is going to be too big and out of scope.

sdk/inc/azure/core/internal/az_config_internal.h
```diff
@@ -24,7 +24,7 @@ enum
  */
 enum
 {
-  _az_INT64_AS_STR_BUF_SIZE = 20,
+  _az_INT64_AS_STR_BUFFER_SIZE = 20,
```
JeffreyRichter, line 27: Please add comment as to how 20 was determined

sdk/inc/azure/core/internal/az_spinlock_internal.h
JeffreyRichter: It's fine to keep the _az_spinlock.h public header file.

sdk/src/azure/core/az_aad_private.h
```diff
@@ -17,10 +17,10 @@
 
 enum
 {
-  _az_AAD_REQUEST_URL_BUF_SIZE = AZ_HTTP_REQUEST_URL_BUF_SIZE,
-  _az_AAD_REQUEST_HEADER_BUF_SIZE = 10 * sizeof(az_pair),
-  _az_AAD_REQUEST_BODY_BUF_SIZE = AZ_HTTP_REQUEST_BODY_BUF_SIZE,
-  _az_AAD_RESPONSE_BUF_SIZE = 3 * 1024,
+  _az_AAD_REQUEST_URL_BUFFER_SIZE = AZ_HTTP_REQUEST_URL_BUFFER_SIZE,
+  _az_AAD_REQUEST_HEADER_BUFFER_SIZE = 10 * sizeof(az_pair),
```
JeffreyRichter, line 21: Note: When we get rid of az_pair, this will have to change.

sdk/src/azure/core/az_credential_client_secret.c
JeffreyRichter: How about az_aad_global_authority.
I think we should prefix with az_ because it is in Azure Core.
aad is a well-known acronym for Azure AD
this is the global authority.
It is only a "default" in the context of when AZ_SPAN_NULL is passed when creating the credential. This string is not a default by itself; only in the context of something else.

sdk/src/azure/core/az_credential_client_secret.c
JeffreyRichter: Exactly, this is the context when az_aad_global_authority becomes the default.

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: I agree with aad_global_authority. We don't necessarily have to preix it with az, because it is a local static, you can see it is defined in the function scope, it is not even visible in other places in this file, and not available to linker.

sdk/inc/azure/core/internal/az_spinlock_internal.h
antkmsft: Hmm, ok, I'll undo this change.

sdk/inc/azure/core/az_context.h
ahsonkhan: Makes sense.

sdk/inc/azure/core/az_context.h
ahsonkhan: It's fine to do this separately :)

sdk/inc/azure/core/internal/az_config_internal.h
antkmsft: The comment about that is right above :)

sdk/src/azure/core/az_credential_client_secret.c
ahsonkhan: > We don't necessarily have to preix it with az, because it is a local static, you can see it is defined in the function scope, it is not even visible in other places in this file, and not available to linker.

Sure, for now. And your suggestion is fine :)

But part of the discussion was whether we want to make it public and if so, what  would we call it.

sdk/docs/core/README.md
antkmsft: @barcharcraz , are you fine with `az_X_create`/`az_X_create_with_Y` instead of `az_X_create_X`/`az_X_create_X_with_Y`?
If we don't reach a conclusion soon, I'll leave the naming as it is, so it conforms to neither naming scheme :)

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: I actually don't have that strong of opinion here. So, `az_aad_global_authority` looks good?
I can declare it as static global, which still make it visible only inside this only `.c` file, and not visible to linker, but should we ever want to make it public/visible, we'll need to remove `static`, and add `extern` declaration to the header file.

sdk/inc/azure/core/az_context.h
antkmsft: I think it is meant to be `new expiring`, that makes sense, I'll do the update.

sdk/inc/azure/core/az_http.h
antkmsft: I'll add comment, but I think optimizing for packing should be the default, and we should only comment when we make an exception from that :)

sdk/inc/azure/core/az_http.h
antkmsft: It is not used anywhere. I thought it was used for the scenario when the cable gets unplugged, but it was the old design, now az_http_send_reqyes would return non-AZ_OK result.
But... You know what, let's keep it. If you have response default-initialized, this is what you re going to have there. Not that we have to have a constant defined for the 0 value (enums still can take "undefined" random integer values), but I think it is better to have it.

sdk/docs/core/README.md
barcharcraz: I think that was the original intent of the wording actually. I'll respond with more detail after reviewing the rest of the PR.

antkmsft: Thank you! You don't have to review the entire PR, but certainly that would be super awesome! :)

sdk/docs/core/CHANGELOG.md
```diff
@@ -2,10 +2,20 @@
...
 
 ## 1.0.0-preview.4 (Unreleased)
-
+- `az_context`:
+  - `key` and `value` are `const`.
+  - `az_context_with_expiration()` is renamed to `az_context_create_with_expiration()`.
+  - `az_context_with_value()` is renamed to `az_context_create_with_value()`.
+- `az_credential_client_secret_init()` now takes fourth parameter, `authority`.
+- `az_http_policy_retry_options`:
+  - `status_codes` now should be terminated by `AZ_HTTP_STATUS_CODE_END_OF_LIST`.
```
ahsonkhan, line 11: I don't think this is a breaking change to the end user, we are just adding another value.

If we were removing `none` I would have mentioned it, but I think we can remove this as it is minor implementation detail.

sdk/inc/azure/core/az_context.h
```diff
@@ -27,58 +27,56 @@ typedef struct az_context az_context;
...
 /**
- * @brief az_context_app is the ultimate root of all az_context instances. It allows you to cancel
- * your entire application. The az_context_app never expires but you can explicitly cancel it by
- * passing its address to az_context_cancel which effectively cancels all the az_context child
+ * @brief The ultimate root of all #az_context instances. It allows you to cancel
+ * your entire application. The #az_context_app never expires but you can explicitly cancel it by
+ * passing its address to #az_context_cancel which effectively cancels all the #az_context child
  * nodes.
  */
 extern az_context az_context_app;
```
ahsonkhan, line 52: Missed this one:
```suggestion
extern az_context az_context_application;
```

sdk/inc/azure/core/az_context.h
```diff
@@ -27,58 +27,56 @@ typedef struct az_context az_context;
...
 
 /**
- * @brief az_context_with_expiration creates a newexpiring az_context node that is a child of the
- * specified parent.
+ * @brief Creates a new expiring az_context node that is a child of the specified parent.
```
ahsonkhan, line 55: Reads much better :)

sdk/inc/azure/core/az_http.h
```diff
@@ -182,12 +185,13 @@ AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* respons
  *
  * @see https://tools.ietf.org/html/rfc7230#section-3.1.2
  */
+// Member order is optimized for alignment.
```
ahsonkhan, line 188: Just double check that this isn't visible in the generated docs.

sdk/inc/azure/core/az_http.h
```diff
@@ -182,12 +185,13 @@ AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* respons
...
 typedef struct
 {
+  az_span reason_phrase;
```
ahsonkhan, line 191: Outside the scope of this PR, but these are public fields and need documentation.
https://azuresdkdocs.blob.core.windows.net/$web/c/docs/1.0.0-preview.3/structaz__http__response__status__line.html

sdk/inc/azure/core/az_http.h
ahsonkhan: Do you think it is worth documenting it to make clear that is what it's purpose is?

I think overloading the 0 HTTP_STATUS_CODE might confuse folks into thinking there is an actual status code of value zero. We may want to avoid that and remove it.

sdk/docs/core/CHANGELOG.md
antkmsft: I think it is - if you used to have a list terminated by NONE, everything worked correctly, now you HAVE to replace NONE with END_OF_LIST, because otherwise it is going to read the entire memory until it manages to read -1.

sdk/inc/azure/core/az_http.h
ahsonkhan: >  I think optimizing for packing should be the default

If you feel strongly, we should first fix up the packing of all our existing structs and then I agree :)

Keeping it aligned/packed is difficult though and could be hard to correctly maintain.

sdk/inc/azure/core/az_context.h
antkmsft: Thank you for bringing it up, I accidentally noticed that we missed `#` here :)

sdk/inc/azure/core/az_http.h
antkmsft: No, it is not, and that's intentional.

sdk/src/azure/core/az_credential_client_secret.c
ahsonkhan: > add `extern` declaration to the header file.

Is the other option to move it into the `.h` file from the `.c` file rather than using `extern` feasible as well (if we wanted to do it in the future)?

sdk/inc/azure/core/az_http.h
antkmsft: I created an issue to not lose this: https://github.com/Azure/azure-sdk-for-c/issues/994

sdk/docs/core/CHANGELOG.md
ahsonkhan: Oh right, this list of status codes is user provided. Gotcha.

sdk/inc/azure/core/az_http.h
antkmsft: To be fair, AZ_HTTP_STATUS_CODE_END_OF_LIST doesn't map to a HTTP status code either.

sdk/src/azure/core/az_credential_client_secret.c
antkmsft: If we declare it in the header file, as extern in the future, an extra symbol will become visible to the user. Currently it is not.
We should not move *declarations* (az_span X = "ABC") to header files, because that results in the symbol being moved to customer variables and would be likely included multiple times to the binary.
So, we are good, we're compatible with doing the things in idiomatic way. For example, see how `az_context_application` is defined - it is `extern` in header and global in .c.

sdk/docs/core/README.md
barcharcraz: ah, the reason for the confusion is because these "create" function are constructors not member functions. There are separate guidelines for constructors that do what you suggest, there should be a cross-referencing link, however.

antkmsft: So, to clarify more specifically, the name `az_context_create_with_expiration()` is ok. Correct?

ahsonkhan: Can you share a link? I couldn't find the separate guideline on constructors.

antkmsft: Given the answer is "yes", I crated a work item to update guidelines: https://github.com/Azure/azure-sdk-for-c/issues/999

sdk/docs/core/CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
...
+  - `az_http_policy_retry_options`:
+    - `status_codes` now should be terminated by `AZ_HTTP_STATUS_CODE_END_OF_LIST`.
+    - `max_retries` is now `int32_t` instead of `int16_t`.
+  - `az_config.h`:
+    -  `AZ_HTTP_REQUEST_URL_BUF_SIZE` renamed to `AZ_HTTP_REQUEST_URL_BUFFER_SIZE`.
+    - `AZ_HTTP_REQUEST_BODY_BUF_SIZE` renamed to `AZ_HTTP_REQUEST_BODY_BUFFER_SIZE`.
+    - `AZ_LOG_MSG_BUF_SIZE` renamed to `AZ_LOG_MESSAGE_BUFFER_SIZE`.
 
 ## 1.0.0-preview.3 (2020-07-20)
-
```
ahsonkhan, line 8: Leave this new line please. The markdown linter suggested it. See https://github.com/Azure/azure-sdk-for-c/pull/959 for more info.

sdk/docs/core/CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
 
 
 ## 1.0.0-preview.4 (Unreleased)
-
+- API breaking changes:
```
ahsonkhan, line 5: Add a new line after the heading.

sdk/docs/core/README.md
antkmsft: I just pinged Charlie via Teams, the answer is yes. I am resolving this comment thread. We have a work item to track guidelines update.

ahsonkhan: > I crated a work item to update guidelines: #999

One more to a thousand! 4 digit issues and PRs :)

sdk/src/azure/core/az_credential_client_secret.c
```diff
@@ -34,8 +37,8 @@ static AZ_NODISCARD az_result _az_credential_client_secret_request_token(
       credential->_internal.client_secret,
       &body));
 
-  uint8_t header_buf[_az_AAD_REQUEST_HEADER_BUF_SIZE];
-  az_http_request request = { 0 };
+  uint8_t header_buf[_az_AAD_REQUEST_HEADER_BUFFER_SIZE];
+  _az_http_request request = { 0 };
```
antkmsft, line 41:
```suggestion
  az_http_request request = { 0 };
```

ahsonkhan:
@antkmsft - when do you plan to merge this?

sdk/inc/azure/core/az_context.h
```diff
@@ -122,17 +119,18 @@ AZ_NODISCARD AZ_INLINE bool az_context_has_expired(az_context const* context, in
...
+ * @param[in] context The #az_context node in the tree where checking starts.
+ * @param[in] key A pointer to the key to be scanned for.
+ * @param[out] out_value A pointer to a `void const*` that will receive the key's associated value
+ * if the key is found.
+ * @return  #AZ_OK if the key is found.
  *          #AZ_ERROR_ITEM_NOT_FOUND if no nodes are found with the specified key.
  */
-AZ_NODISCARD az_result az_context_get_value(az_context const* context, void* key, void** out_value);
+AZ_NODISCARD az_result
+az_context_get_value(az_context const* context, void const* key, void const** out_value);
```
ahsonkhan, line 133: @RickWinter - please review this const-ness change with other `az_context` changes you might have, as part of your follow up.

sdk/inc/azure/core/az_http.h
```diff
@@ -114,10 +117,10 @@ typedef enum
  */
 typedef struct
 {
-  int16_t max_retries;
+  az_http_status_code const* status_codes;
```
ahsonkhan, line 120: From the API review:
> Is the const here in the right place? The status codes should not change.

antkmsft: Yes, it is. It should't be in the other place - it would be the same as if `max_retries` was declared as `int const max_retries`.

antkmsft:
Ideally,I wanted all comments to be resolved. We'll follow up after the PR is complete.

sdk/inc/azure/core/az_http.h
antkmsft: Jeffrey told me in Teams: `Sure, let's keep both. Before a response comes back, the status code will be NONE. And, if some network error occurs, then the status code will also be NONE.`

ahsonkhan: Wouldn't a network error set the code to some 5xx error like 503, 523, or 598? Why would it be `NONE` in that case?

PR #991, opened 2020-08-05
update TARGET_NAME

momuno:


sdk/samples/iot/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ if(NOT OpenSSL_FOUND)
 endif()
 
 # Azure IoT Samples Library
-add_library (az_iot_samples_common
+add_library (${TARGET_NAME}
```
danewalton, line 25: Because there are multiple targets defined in this `CMakeLists`, we shouldn't be using `${TARGET_NAME}`. This is one of the reasons I personally don't like using `${TARGET_NAME}` pervasively in CMake since you have to refer elsewhere in the file to see what the value actually is. Everything in this file should have an explicit name.

TLDR: `TARGET_NAME` should be removed from this file.

momuno: Ahson has changed all these to TARGET_NAME.  I am only reverting back a change I inadvertently made to one of his line changes.  Let's address this in C99 meeting.

ahsonkhan: I am not sure I agree @danewalton 

Where are the multiple targets defined here? I see only one `set` and only two `add_library` calls, both using the same  `az_iot_samples_common` target name.

https://github.com/Azure/azure-sdk-for-c/blob/917738f1c6479184e9679fdeea5fdabf3c5a5aef/sdk/samples/iot/CMakeLists.txt#L8
https://github.com/Azure/azure-sdk-for-c/blob/917738f1c6479184e9679fdeea5fdabf3c5a5aef/sdk/samples/iot/CMakeLists.txt#L52

We have a bunch of `target_link_libraries` calls, but only one target definition.

We have 5 different usages of `${TARGET_NAME}` in this CMakeLists.txt file. Having a common target name makes it easy to update the name in one place, if we wanted to, and re-use that variable, which is the main benefit.

I think this is also consistent with the rest of the `CMakeLists.txt` files in the repo that all have a single declaration of `${TARGET_NAME}` used in multiple places.
https://github.com/Azure/azure-sdk-for-c/search?q=TARGET_NAME&unscoped_q=TARGET_NAME

danewalton: A CMake target is an executable or a library. So in this file, there are many targets. I personally would rather we change this behavior throughout the whole repo. It opens us up for bad habits which is detailed here in *Professional CMake: A Practical Guide* by Craig Scott who is a maintainer of CMake itself

Screenshot 2020-08-05 150931

ahsonkhan: TIL. Thanks for sharing.

So, how do we solve the "we need to update the project name" problem? It becomes annoying to chase down and find all copies of a particular string, since it is repeated, having it in one place is much nicer, just like how we use named constants in actual C code.

danewalton: Understandable. But I don't think it would be as much of a pain as you might think. Lets say I have to change `az_iot_hub` to `az_iot_hub_two`. I would change the name of the library in it's respective CMake list but I would still have to do a find and replace all in the rest of the CMakeList.txt files for things that link to it. This just adds a few more instances of it but it would still be one click using VSCode.

Not to mention, this actually already was done when we flattened the repo src files. See here:
https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/src/azure/iot/CMakeLists.txt
We have one CMakeList.txt for a given service and create a lib for each feature. In that file, we can't have a variable with a target name and reuse it everywhere since there are multiple libs.

Also, funny enough I just found a bug related to `${TARGET_NAME}` in that CMake at the top lol.

danewalton: https://github.com/Azure/azure-sdk-for-c/pull/997
to fix

PR #997, opened 2020-08-05
fix iot CMake project name

danewalton:


sdk/src/azure/iot/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 
 cmake_minimum_required (VERSION 3.10)
 
-project (${TARGET_NAME} LANGUAGES C)
+project (az_iot LANGUAGES C)
```
ahsonkhan, line 6: Just double checking, is that what it was prior to https://github.com/Azure/azure-sdk-for-c/pull/830?

danewalton: Not exactly, since before each feature had it's own CMake, the project name was the feature name. Example:
https://github.com/Azure/azure-sdk-for-c/blob/9bd3125d20976205d0e60435c73e8831cb4de911/sdk/iot/hub/CMakeLists.txt#L6-L8
Changing it to this seems reasonable though for all of the features/libs

PR #1001, opened 2020-08-06
update project name on iot CMake

danewalton:
Anything in a CMake project gets grouped in IDE integrations. Currently it looks like this in VSCode:
Screenshot 2020-08-05 174051

The title `az_iot_samples_common` isn't exactly right so updated the name to better reflect the grouping.

### DIFF
--- a/sdk/samples/iot/CMakeLists.txt
+++ b/sdk/samples/iot/CMakeLists.txt
@@ -7,7 +7,7 @@ cmake_minimum_required (VERSION 3.10)
 
 set(TARGET_NAME "az_iot_samples_common")
 
-project (${TARGET_NAME} LANGUAGES C)
+project (az_iot_samples LANGUAGES C)
 
 set(CMAKE_C_STANDARD 99)
 


sdk/samples/iot/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ cmake_minimum_required (VERSION 3.10)
 
 set(TARGET_NAME "az_iot_samples_common")
```
ahsonkhan, line 8: Other than our earlier discussion, why not rename this to `az_iot_samples` instead?

sdk/samples/iot/CMakeLists.txt
```diff
@@ -149,4 +149,4 @@ if (MSVC)
     )
 endif()
 
-endif() # TRANSPORT_PAHO
```
ahsonkhan, line 152: Nice :)

ahsonkhan:
Good catch!

Issue #935, opened 2020-07-16
Be consistent and clear about what "platform" means within the docs.


ahsonkhan:
This README doc:
https://github.com/Azure/azure-sdk-for-c/blob/7f435f0eaffd11a8e71bfd659be4b6bc70be44a8/sdk/docs/platform/README.md#azure-sdk-platform

implies that "the platform" contains the platform abstraction (along with the HTTP transport layer adapter) which is confusing. 

We should consider re-wording/re-titling this doc and change whatever else is necessary within it, since (as far as I can tell) this is the **only** README that conflates/groups together both the HTTP adapter and the actual platform into one thing also called `Azure SDK Platform`. Everywhere else, we have a conceptual separation of az_platform/no platform, and az_http/no http/adapter.

When we think of and use the term `platform`, we mean the abstraction layer between the SDK and the underlying hardware, which provides the method implementations for things like clock and sleep. For example, see this:
https://github.com/Azure/azure-sdk-for-c/blob/43e4de8dcac97b144c24e0693028c88a17b98b91/sdk/inc/azure/core/az_platform.h#L21
https://github.com/Azure/azure-sdk-for-c/tree/master/sdk/docs/core#porting-the-azure-sdk-to-another-platform

cc @JeffreyRichter, @RickWinter

PR #1005, opened 2020-08-06
Update the storage readme with concepts

RickWinter:


sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
 Use the Azure SDK Storage Blobs Library for Embedded C to work with Azure Storage Blobs.
 
 * Upload blobs
-* Create and modify blob metadata
 
 ## Getting Started
 
-### Install the Package
+The Azure Storage Blobs Client library is created to facilitate connectivity to Azure Storage Blobs services using the HTTP stack of the user's choice. This means that this SDK is **NOT** a platform but instead is a true SDK library.
```
ahsonkhan, line 11: I don't understand what we are trying to say with that last sentence. Seems odd to bring up.

```suggestion
The Azure Storage Blobs Client library is created to facilitate connectivity to Azure Storage Blobs services using the HTTP stack of the user's choice.
```

sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
 Use the Azure SDK Storage Blobs Library for Embedded C to work with Azure Storage Blobs.
 
 * Upload blobs
```
ahsonkhan, line 7: This list just appears and the preamble sentence above it doesn't lead up to it. Reword?

sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
...
+### Build
 
-### Get Credentials
+The Azure Storage Blobs library is compiled following the same steps listed on the root [README][azure_sdk_for_c_readme] documentation, under ["Getting Started Using the SDK"][azure_sdk_for_c_readme_getting_started].
 
-TODO
+The library targets made available via CMake are the following:
 
-### Create Client
+- `az::storage::blobs` - For Azure Storage Blobs features ([API documentation here][azure_sdk_for_c_doxygen_docs])
```
ahsonkhan, line 23: Why are we linking to our root README for azure storage blobs features? What section (if any) is relevant to the user here?

sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
...
+### Authentication
+
+The embedded C SDK currently supports SAS based authentication.  See [this page][storage_access_control_sas] for information on creating SAS tokens.
+The client credential should be set to `AZ_CREDENTIAL_ANONYMOUS` when using SAS tokens.
 
-## Examples
+### Creating the Storage Client
 
-TODO
+To use the storage client the first action is to initialize the client with `az_storage_blobs_blob_client_init`.
```
ahsonkhan, line 45:
```suggestion
To use the storage client, the first action is to initialize the client with `az_storage_blobs_blob_client_init`.
```

sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
...
+```C
+  az_storage_blobs_blob_client client;
+  az_storage_blobs_blob_client_options options = az_storage_blobs_blob_client_options_default();
 
-## Troubleshooting
+  az_storage_blobs_blob_client_init(
+          &client, az_span_from_str(getenv(URI_ENV)), AZ_CREDENTIAL_ANONYMOUS, &options);
```
ahsonkhan, line 51: We should just pass in `NULL` for options, to keep the getting started simple and "one-liner".

sdk/docs/storage/README.md
```diff
@@ -5,49 +5,67 @@ Azure Storage Blobs library (`az_storage_blobs`) provides abstractions, and help
...
+```
 
-### General
+### Uploading a blob
 
-TODO
+Once the client is created it can be used to upload blobs.
+```C
+  az_result const blob_upload_result = az_storage_blobs_blob_upload(
+      &client, &az_context_app, content_to_upload, NULL, &http_response)
```
ahsonkhan, line 59:
```suggestion
  az_result const blob_upload_result = az_storage_blobs_blob_upload(
      &client, &az_context_application, content_to_upload, NULL, &http_response)
```

ahsonkhan: Do we want to mention here a little bit about context and http response (maybe link out), or is the intention to detail this further in samples? I was wondering where the response came from.

ahsonkhan: FYI @antkmsft, keep an eye out for merge conflicts here related to https://github.com/Azure/azure-sdk-for-c/pull/988.

sdk/docs/storage/README.md
RickWinter: Its not linking to the root readme.  It points to the root of the docs.  Unfortunately there isn't a way to reference into the docs w/o placing version specific information into the paths.  This is the best we can do for the time.

sdk/docs/storage/README.md
RickWinter: I prefer using the explicit value.  It makes it explicit what we are doing.

sdk/docs/storage/README.md
RickWinter: It helps the user understand that they can control the underlying behavior.  That is distinct difference from the previous C SDKs which had implementations which could not be changed

sdk/docs/storage/README.md
ahsonkhan: Oh I see. I clicked the wrong link.

I am assuming your intended link would be this (if we didn't have the version issue), correct?
https://azuresdkdocs.blob.core.windows.net/$web/c/docs/1.0.0-preview.3/az__storage__blobs_8h.html

sdk/docs/storage/README.md
ahsonkhan: I think not allowing `NULL`, as in keeping the options, optional, is a usability concern.
We have the pattern of passing in `NULL` to mean `default` in a whole bunch of existing SDK APIs. I think we should make sure both work, but that is related to the API and not docs (unrelated to this PR).

But you are right, leaving the option in the code snippet in the docs and being explicit is reasonable.

Remove the `NULL` parameter below in `az_storage_blobs_blob_upload`?

sdk/docs/storage/README.md
ahsonkhan: Makes sense. I don't know how to word that well, but using overloaded terms like "it is not a platform", and "true SDK library" are confusing. I start asking these questions: What's a platform in this context? What's a true SDK library supposed to be?

Maybe...
```suggestion
The Azure Storage Blobs Client library is created to facilitate connectivity to Azure Storage Blobs services using the HTTP stack of the user's choice. This capability differentiates the Embedded SDK from the previous C SDKs which had underlying implementations that the user could not be changed.
```

> underlying transport (platform) 

Also, we have been a few conversations (with @JeffreyRichter) about terminology around platform, and HTTP transport. Let's make sure to keep those two distinct (i.e. platform != http transport adapter)
See this issue: https://github.com/Azure/azure-sdk-for-c/issues/935

sdk/docs/storage/README.md
RickWinter: We can't remove that NULL parameter.  That is the options for upload which are optional

sdk/docs/storage/README.md
JeffreyRichter: I suggest this:
The Azure Storage Blobs Client library facilitates connectivity to the Azure Storage Blobs service. Since this service uses HTTP, you must plug in an HTTP adapter capable of sending the HTTP request to your device's HTTP stack. For information on how to do this, please see (link goes here).

(I'm assuming that we'll have more HTTP clients in the future and so the HTTP adapter stuff should be in 1 place and then simply referred to by the various clients)

sdk/docs/storage/README.md
```diff
@@ -62,20 +80,19 @@ If you'd like to contribute to this library, please read the [contributing guide
 Azure SDK for Embedded C is licensed under the [MIT][azure_sdk_for_c_license] license.
 
 <!-- LINKS -->
+[azure_pattern_circuit_breaker]: https://docs.microsoft.com/azure/architecture/patterns/circuit-breaker
```
JeffreyRichter, line 83: I would remove any references to "circuit breaker pattern" from this document as our SDK does not have support for it built in.

sdk/docs/storage/README.md
ahsonkhan: By remove, I mean state that parameter explicitly. The same rationale you are applying here.

sdk/docs/storage/README.md
RickWinter: We don't have full circuit breakers but we do have some aspects with retry.  Also the article is a good read for handling network error patterns

sdk/docs/storage/README.md
```diff
@@ -2,52 +2,70 @@
...
 
-### Create Client
+- `az::storage::blobs` - For Azure Storage Blobs features ([API documentation here][azure_sdk_for_c_doxygen_docs])
 
-TODO
+### Samples
+
+This [page][samples_storage_blobs_readme] explains samples for the Azure Embedded C SDK Storage Blobs Client.
+
+
```
ahsonkhan, line 29: You should consider using the markdown linter: https://github.com/Azure/azure-sdk-for-c/issues/459, https://github.com/Azure/azure-sdk-for-c/pull/959

nit: Remove extra new line.

sdk/inc/azure/core/az_context.h
```diff
@@ -89,7 +89,7 @@ az_context_create_with_value(az_context const* parent, void const* key, void con
  * @brief Cancels the specified #az_context node; this cancels all the child nodes as well.
  *
  * @param[in] context A pointer to the #az_context node to be canceled; passing `NULL` cancels the
- * root #az_context_app.
+ * root #az_context.
```
ahsonkhan, line 92: Why did we change this from `az_context_application` to `az_context`?

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -90,6 +90,7 @@ AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
  */
 typedef struct
 {
+  az_context* context;
```
ahsonkhan, line 93: Public field needs documentation.

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -137,7 +138,6 @@ az_storage_blobs_blob_upload_options_default()
  */
 AZ_NODISCARD az_result az_storage_blobs_blob_upload(
     az_storage_blobs_blob_client* client,
-    az_context* context,
```
ahsonkhan, line 140: We need a changelog entry with this change since this is a breaking change.

sdk/src/azure/storage/az_storage_blobs_blob_client.c
```diff
@@ -163,7 +162,7 @@ AZ_NODISCARD az_result az_storage_blobs_blob_upload(
   az_http_request request;
   AZ_RETURN_IF_FAILED(az_http_request_init(
       &request,
-      context,
+      opt.context,
```
ahsonkhan, line 165: We no longer need the `(void)` on line 147 now:
https://github.com/RickWinter/azure-sdk-for-c/blob/0cd5306e9193dcdd133ef0a2ce102d045d7526c9/sdk/src/azure/storage/az_storage_blobs_blob_client.c#L147

sdk/src/azure/storage/az_storage_blobs_blob_client.c
```diff
@@ -130,7 +130,6 @@ AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
 
 AZ_NODISCARD az_result az_storage_blobs_blob_upload(
     az_storage_blobs_blob_client* client,
-    az_context* context,
```
ahsonkhan, line 133: nit: Ideally, such API and code changes would be done in a separate PR, given this PR was meant for just readme/docs, so that it could be reviewed in more detail.

ahsonkhan: Also, I am surprised this didn't result in any changes to tests. Do we have unit tests for this?

sdk/inc/azure/storage/az_storage_blobs.h
ahsonkhan: Fixed in https://github.com/Azure/azure-sdk-for-c/pull/1028

PR #1008, opened 2020-08-06
Address API review feedback related to APIs in az_span.h.

ahsonkhan:
Addressing API review feedback related to `az_core` from https://apiview.dev/Assemblies/Review/e29157a38d8e43949fae98a53381f5a5

- API breaking changes:
  - `az_span.h`:
    - `az_span_init(...)` is renamed to `az_span_create(...)`
    - `az_span_from_str(...)` is renamed to `az_span_create_from_str(...)`
    - `az_allocator_context` is renamed to `az_span_allocator_context`
    - Removed `az_pair_from_str(...)`

Also went and formatted (spacing/etc.) of the files I modified, in VS, using Ctrl+K, Ctrl+D.

Since IoT uses `az_pair` atm in their public surface area, leaving it for now:
https://github.com/Azure/azure-sdk-for-c/blob/cb8694901af8ff23c81e05444557e1940a239869/sdk/src/azure/iot/az_iot_hub_client.c#L268

We should consider revisiting that and then remove the `az_pair` type.


sdk/docs/core/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 
 ## 1.0.0-preview.4 (Unreleased)
 
+- API breaking changes:
+  - `az_span.h`:
+    - `az_span_init(...)` is renamed to `az_span_create(...)`
```
antkmsft, line 8: What I've seen somewhat idiomatically - in books and docs, is when talking about functions, you write `f()` and not `f(...)`.

sdk/docs/core/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
...
+    - `az_span_from_str(...)` is renamed to `az_span_create_from_str(...)`
```
antkmsft, line 9: I used periods at the end of each item. Please make them to match after merging my changes. I slightly prefer periods at the end, because if we add a 2-sentence item, we end up with 2 periods, ant there's a slight mismatch with 1-sentence items. But I've seen everywhere (not only our repo) it done both ways, and it is minor..

sdk/docs/core/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 ## 1.0.0-preview.4 (Unreleased)
 
 - API breaking changes:
+  - `az_span.h`:
```
antkmsft, line 7: In this case, I'd name it as `az_span`, not `az_span.h`.

sdk/docs/core/CHANGELOG.md
ahsonkhan: I slight prefer the `...`, but will change it.

sdk/docs/core/CHANGELOG.md
ahsonkhan: The reason why have the header file is because APIs/methods other than just `az_span` also changed (such as the allocator context).

antkmsft: In my mind, it is still super closely related to `az_span` itself, but I see the reason to do it either way.

ahsonkhan: We are distributed by source, which makes me lean towards leaving it as is. Folks do `#include <azure/core/az_span.h>`, so I think it is fine to use the same terminology when talking about particular header file changes.

Besides, we also have az_config.h below :)

sdk/tests/core/inc/az_test_precondition.h
```diff
@@ -4,9 +4,12 @@
 #ifndef _az_TEST_PRECONDITION_H
 #define _az_TEST_PRECONDITION_H
 
-#include <assert.h>
+// These headers must be included prior to including cmocka.h.
 #include <setjmp.h>
 #include <stdarg.h>
+#include <stddef.h>
+
```
antkmsft, line 11: I don't think we need this specific empty line. All are headers from standard library, but in adjacent include blocks, there's probably no reason.

ahsonkhan: I wanted to make it clear what headers specifically need to be included before cmocka.h and hence the comment.

We have this type of grouping in a bunch of places. Merging them together causes VS auto-format to re-order them, which I don't want since commenting the three headers becomes more cumbersome. That's why I split them up.

If you feel strongly about it, please feel free to fix it up across the repo and let's be consistent. For this PR, I am going to limit the change and leave it as is. It doesn't have a material impact on correctness, so low priority, imo.

sdk/docs/core/CHANGELOG.md
antkmsft: az_config.h is special, because it is the only common thing about 3 constants.
I am fine with az_span.h.

sdk/tests/core/inc/az_test_precondition.h
antkmsft: If so, then sure, it is exactly a valid reason to create a group of includes, the change is not needed in this case.
(I know that grouping would lead to sorting).

ahsonkhan:
In the interest of time, merging this. Others who haven't looked at it, please feel free to review and leave other feedback if it comes up.

PR #1024, opened 2020-08-08
Replace bash and cmd cert gen files with powershell

danewalton:
@jspaith for eyes since you are the powershell master

I tested on WSL and Windows. If anyone else would like to double check please feel free.


sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
+openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
+openssl x509 -noout -text -in device_ec_cert.pem
+
+if(Test-Path device_cert_store.pem)
```
ahsonkhan, line 8: nit: 
```suggestion
if (Test-Path device_cert_store.pem)
```

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
...
+Write-Output "It is NOT recommended to use OpenSSL on Windows or OSX. Recommended TLS stacks are:"
+Write-Output "Microsoft Windows SChannel: https://docs.microsoft.com/en-us/windows/win32/com/schannel"
+Write-Output "OR"
+Write-Output "Apple Secure Transport : https://developer.apple.com/documentation/security/secure_transport"
+Write-Output "If using OpenSSL, it is recommended to use the OpenSSL Trusted CA store configured on your system."
+
+Write-Output "`nSAMPLE CERTIFICATE GENERATED:"
+Write-Output "Use the following command to set the environment variable for the samples:"
+if ($IsWindows) {
+  Write-Output "set AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
```
ahsonkhan, line 25: Add new line and tab, like before?
```suggestion
  Write-Output "`n`tset AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
```

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
...
+}
+else {
+  Write-Output "export AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
```
ahsonkhan, line 28: Same here.
```suggestion
  Write-Output "`n`texport AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
```

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
+openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
+openssl x509 -noout -text -in device_ec_cert.pem
+
+if (Test-Path device_cert_store.pem)
```
jspaith, line 8: A new line for "{" is not consistent with what you have below for $isWindows.

sdk/samples/iot/README.md
```diff
@@ -138,6 +138,7 @@ To run the samples, ensure you have the following programs or tools installed on
     ```
 
   - For non-Linux based systems, download the [OpenSSL 1.1.1 LTS](https://www.openssl.org/source/openssl-1.1.1g.tar.gz) command line utility and follow the downloaded INSTALL document.
+- To use the certificate generation script (`generate_certificate.ps1`), you must have Powershell 7 installed. Please follow the installation instructions for your platform [here](https://github.com/PowerShell/PowerShell/tree/v7.0.3#get-powershell).
```
jspaith, line 141: Nit: What if PowerShell 8 comes out?  Presumambly this would keep working?  What happens 10 years off when PS7 is deprecated but we still link to it?  

Could we just have people go to latest PowerShell core or whatever right nomenclature is?

sdk/samples/iot/README.md
```diff
@@ -138,6 +138,7 @@ To run the samples, ensure you have the following programs or tools installed on
     ```
 
   - For non-Linux based systems, download the [OpenSSL 1.1.1 LTS](https://www.openssl.org/source/openssl-1.1.1g.tar.gz) command line utility and follow the downloaded INSTALL document.
```
jspaith, line 140: Nit: What happens if OpenSSL 1.1.1 is deprecated because whatever?  Unlike PS1 where you should always be fine on latest as they're pretty good on compat, I actually don't know if we should link to 1.1.1 no matter what or latest for fear of latest breaking something?

jspaith: Also even though openssl last time I checked doesn't hand out binaries, a lot of customers might go to a site they trust anyway (though MS should NOT be in business of suggesting sites).  We might consider being a bit more generic on instructions and letting customer find this on their own since there's so many ways/versions that could change.

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
```
jspaith, line 4: It's helpful but not required to test prerequisites before starting something that has them.  Specifically I might recommend something like

```script
if (-Not (Get-Command "openssl" -ErrorAction SilentlyContinue))
{
  throw "openssl is not availabe.  You will need to install it or add it to the PATH to proceed"
}
```

sdk/samples/iot/generate_certificate.ps1
jspaith, line 8: I'd recommend doing cleanup before you start invoking `openssl`, not after.

sdk/samples/iot/README.md
jspaith, line 140: There are many issues in original implementation of this the C SDK you're cleaning up, which is great.  But one thing I'd like to see carried forward is the warning https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md#warning, or some such after review with Azure SDK team.

This sample is needed at getting people going but basing a deployment off of keys generated on a dev's prototype box isn't something we should be encouraging.  I recommend you reach out to Eustace and Chrissie (I can't find them on GitHub) as IoT Security && DPS might have a better production quality guidance we should point folks to for productizing.

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
...
+openssl x509 -noout -text -in device_ec_cert.pem
+
+if (Test-Path device_cert_store.pem)
+{
+  Remove-Item device_cert_store.pem
+}
+
+Get-Content device_ec_cert.pem, device_ec_key.pem | Set-Content device_cert_store.pem
+
+Write-Output "IMPORTANT:"
```
jspaith, line 15: Nit: If someone is using Linux I don't think they need to be told this.  Could you wrap with an if ($IsWindows) and if ($IsMac) (if there is such a thing)?

jspaith: Also if you can put into OS specific, I'd recommend going with a full on `Write-Warning` that'll make this pop even more instead `Write-Output`

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey
+openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"
```
jspaith, line 5: Please check $LASTEXITCODE on each openssl call and doing a "throw" on any failure.  And any other exe we may call.

There's a real danger that if a ton of output here (either now or in future if you add more) then an `openssl` call failing could be buried in console history a developer would never see and they'd get some Frankenstein setup.  That would be worst of all worlds because it might even succeed but not be secure.

You can do a helper function to call openssl with arbitrary command like arguments.

There might be better ways to code this up but from our release scripts.

```script
function Invoke-IoTExecutable {

    param(
        [string]$Name = $(throw "Must provide executable name"),
        [string]$Command = $(throw "Must provide executable command string"),
        [string]$Path = (Get-Item .)
    )

    Push-Location $Path
    Invoke-Expression "$Name $Command"
    Pop-Location

    if ($LastExitCode -ne 0) {
        throw "'$Name' invocation failed with exit code ($LastExitCode)"
    }
}
```

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,39 @@
...
+Write-Output "Use the following command to set the environment variable for the samples:"
+if ($IsWindows) {
+  Write-Output "`n`tset AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
+}
+else {
+  Write-Output "`n`texport AZ_IOT_DEVICE_X509_CERT_PEM_FILE=$(Resolve-Path device_cert_store.pem)"
+}
+
+Write-Output "`nDPS SAMPLE:"
+Write-Output "Upload device_ec_cert.pem when enrolling your device with the Device Provisioning Service."
```
jspaith, line 32: Are there instructions (that DPS team owns, NOT use) that are somewhat permalinked where we can explain where and how to upload?

sdk/samples/iot/README.md
momuno: We could change this to the generic https://www.openssl.org/source/ and then say 1.1.1 is recommended.  I had trouble with the most recent 3.0.0-alpha6 and found 1.1.1 worked.

momuno: The very clear explicit call out is nice.  I had boldened warnings in this readme, but your suggestion is MUCH more clear.

sdk/samples/iot/README.md
momuno: I would reorder the sentence to match the others in this section:
"Have \[Powershell\](https://github.com/PowerShell/PowerShell/tree/v7.0.3#get-powershell) installed if running certificate samples.  This is required to run the certificate generation script `generate_certificate.ps1`. "

sdk/samples/iot/generate_certificate.ps1
momuno: There are instructions in the Readme: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/README.md#iot-hub-dps-certificate-sample

And in that subsection is a link to further instruction: https://docs.microsoft.com/en-us/azure/iot-dps/quick-create-simulated-device-x509#create-a-device-enrollment-entry-in-the-portal

I don't want to put too much of that instruction in the generation file, to limit all the areas we need to update info if something was to change.

sdk/samples/iot/README.md
danewalton: Technically the API's we are using won't work with the latest (I believe they are deprecated). So here I think we should specify that this LTS is required and we would have to revisit if a new one comes out (someone might open a GH issue saying "Hey please update to the latest).

danewalton:
@jspaith thank you Powershell master

sdk/samples/iot/generate_certificate.ps1
danewalton: ahh thank you this is the kind of stuff I didn't know was available with Powershell. Will update

sdk/samples/iot/README.md
danewalton: I put it in the script itself with a `Write-Warning` so hopefully people pay attention to that.

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
```
jspaith, line 3: Also on top of this please do a `Set-StrictMode on` or whatever syntax is.  This will help catch errors now and in future for us.

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/set-strictmode?view=powershell-7#:~:text=Description%20The%20Set-StrictMode%20cmdlet%20configures%20strict%20mode%20for,or%20script%20block%20violates%20basic%20best-practice%20coding%20rules.

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,69 @@
...
+Write-Warning "Certificates created by this script MUST NOT be used for production."
```
jspaith, line 4: Also while we're being careful please add a `$errorActionPreference="stop"` (assuming PowerShell core supports which I think it would).  This will make it so that PowerShell native commands like the Get-Content, if they fail, will cause the entire script to stop execution.

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,69 @@
...
+Invoke-IoTCommand -Command 'openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"'
+Invoke-IoTCommand -Command 'openssl x509 -noout -text -in device_ec_cert.pem'
+
+Get-Content device_ec_cert.pem, device_ec_key.pem | Set-Content device_cert_store.pem
+
+if ($IsWindows -or $IsMacOS) {
+  Write-Output "`n"
+  Write-Warning "IMPORTANT:"
+  Write-Warning "It is NOT recommended to use OpenSSL on Windows or OSX. Recommended TLS stacks are:"
+  Write-Warning "Microsoft Windows SChannel: https://docs.microsoft.com/en-us/windows/win32/com/schannel"
```
jspaith, line 42: Pretty sure it's `Schannel`, not `SChannel`.  E.G. https://docs.microsoft.com/en-us/windows/win32/com/schannel

sdk/samples/iot/generate_certificate.ps1
```diff
@@ -0,0 +1,69 @@
...
+
+  Invoke-Expression "$Command"
+
+  if ($LastExitCode -ne 0) {
+    throw "'$Command' invocation failed with exit code ($LastExitCode)"
+  }
+}
+
+Invoke-IoTCommand -Command 'openssl ecparam -out device_ec_key.pem -name prime256v1 -genkey'
+Invoke-IoTCommand -Command 'openssl req -new -days 365 -nodes -x509 -key device_ec_key.pem -out device_ec_cert.pem -config x509_config.cfg -subj "/CN=paho-sample-device1"'
```
jspaith, line 33: You're setting `days 365` but at top of file you say it expires in 30 days.  Are these the same things?

danewalton: good point those are different. updating

sdk/samples/iot/generate_certificate.ps1
danewalton: Verbiage looks like its:
```powershell
Set-StrictMode -Version 3.0
```
where the number can be a certain version. I set it to 3.0

Issue #1030, opened 2020-08-10
Investigate and fix issue with paho_iot_hub_twin_sample related to JSON writer


ahsonkhan:
For a repro:
> Use the updated sample documentation
README: https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/README.md
The executable will be `paho_iot_hub_twin_sample`

Investigate the changes in this commit.

dc533e1 : Add support for non-contiguous buffers within JSON build using a ca…
https://github.com/Azure/azure-sdk-for-c/commit/dc533e1db8f030dd94161e78a155ad66134a0d8a#diff-73d7c292a70e269a4e1660e924301ec4
 
@momuno tracked the issue to this commit, in which the `sdk/samples/iot/paho_iot_hub_twin_sample` gives the following error when run:
```text
ERROR:          /mnt/c/Repos/momuno_azure-sdk-for-c/sdk/samples/iot/hub/src/paho_iot_hub_twin_example.c:send_reported_property():398: Unable to build reported property payload to send, az_result return code 80010002
(80010002 = AZ_ERROR_INSUFFICIENT_SPAN_SIZE)
```

The code in the twin sample being called is the following:
https://github.com/Azure/azure-sdk-for-c/blob/14ca09fca5bd7785044dcbd41070baf049a09ac9/sdk/samples/iot/paho_iot_hub_twin_sample.c#L424-L438
 
And the error is occurring on this line:
```C
 AZ_RETURN_IF_FAILED(az_json_writer_append_property_name(&json_writer, reported_property_name));
 ```

When I looked into the `az_json_writer.c` file, it looks like `append_property_name()` is calling a new function called `append_property_name_chunked()`.  And somewhere in this chunked function the `AZ_ERROR_INSUFFICIENT_SPAN_SIZE` is occurring. 
 
FYI, the size for the buffer is the same as before:
```C
static char reported_property_buffer[64];
```

ahsonkhan:
This is by-design because the JSON writer pessimistically requires at least 64-bytes of space when writing any string larger than 10 characters because it tries to write in 64 byte chunks (10 character * 6 if all need to be escaped into the unicode form \uXXXX). The expected fix is to add 64 bytes of slack in the output JSON buffer.

Removing `blocking-release`. If necessary, we can re-evaluate that design, to see whether it needs to change, but for now, it doesn't seem like a blocker to me.

```C
[...]

PR #1034, opened 2020-08-10
Fix paho_iot_hub_twin sample by using a slightly larger buffer to hold the reported buffer JSON.

ahsonkhan:
See https://github.com/Azure/azure-sdk-for-c/issues/1030

cc @RickWinter


momuno:
I will test with this change.

momuno:
> I will test with this change.

Tested with static char reported_property_buffer[128]; 
Passes :)

PR #1036, opened 2020-08-10
Update az_span_scan_util logic and behavior.

ahsonkhan:
Follow-up to feedback and conversation from https://github.com/Azure/azure-sdk-for-c/pull/1031

cc @vhvb1989

### DIFF
--- a/sdk/src/azure/core/az_span_private.h
+++ b/sdk/src/azure/core/az_span_private.h
@@ -65,7 +65,7 @@ AZ_NODISCARD az_result _az_span_replace(
     int32_t end,
     az_span replacement);
 
-typedef bool (*_az_predicate)(az_span slice);
+typedef bool (*_az_predicate)(uint8_t next_byte);
 
 // PRIVATE. read until condition is true on character.
 // Then return number of positions read with output parameter
--- a/sdk/src/azure/core/az_http_request.c
+++ b/sdk/src/azure/core/az_http_request.c
@@ -15,10 +15,7 @@
 
 #include <azure/core/_az_cfg.h>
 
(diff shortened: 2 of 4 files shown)


sdk/src/azure/core/az_span_private.h
```diff
@@ -65,7 +65,7 @@ AZ_NODISCARD az_result _az_span_replace(
     int32_t end,
     az_span replacement);
 
-typedef bool (*_az_predicate)(az_span slice);
+typedef bool (*_az_predicate)(uint8_t next_byte);
```
ahsonkhan, line 68: This change results in code size wins.

**Before:**
az_span.o 176

**After:**
az_span.o 84

```C
AZ_NODISCARD AZ_INLINE bool _az_is_char(az_span slice, uint8_t c)
{
  return az_span_ptr(slice)[0] == c;
}

static AZ_NODISCARD bool _az_is_new_line(az_span slice) { return _az_is_char(slice, '\n'); }

int B(void)
{
  az_span response_span = AZ_SPAN_FROM_STR( //
        "HTTP/1.2 404 We removed the\tpage!\r\n"
        "\r\n"
        "But there is somebody. :-)");
    az_result result = az_http_response_init(&response, response_span);
    
   int32_t offset = 0;
   result = _az_span_scan_until(response_span, _az_is_new_line, &offset);
   printf("%d\n", result);
   
  return 0;
}
```

ahsonkhan:
I'd like to stage the changes and merge this in first before https://github.com/Azure/azure-sdk-for-c/pull/1041, in case we want to bring this functionality back.

PR #1054, opened 2020-08-15
Sample update: includes and defines

momuno:
iot_samples_common.c only adds to the includes already specified in iot_samples_common.h. It does not replicate the includes.

all other sample.c files include all header files associated with content.

### DIFF
--- a/sdk/samples/iot/iot_samples_common.c
+++ b/sdk/samples/iot/iot_samples_common.c
@@ -1,7 +1,28 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include "iot_sample_foundation.h"
+#ifdef _MSC_VER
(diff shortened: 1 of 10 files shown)


sdk/samples/iot/iot_samples_common.c
```diff
@@ -1,7 +1,28 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include "iot_sample_foundation.h"
+#ifdef _MSC_VER
```
danewalton, line 4: this file should include all the includes from the header. I think ones like:
```c
#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
```
are missing.

momuno: So I think this is where the question is that @ahsonkhan had, and I see his point: should the c file duplicate what the h file already has, given the c and h file have the same name?  I think this becomes redundant.

PR #1056, opened 2020-08-15
Sample update: functions and variables

momuno:
-iot prefix naming
-static vs non static
-PRECONDITIONS
-(void)
-az_span pass by value
-az_span destination vs az_span* out_span

### DIFF
--- a/sdk/samples/iot/paho_iot_hub_sas_telemetry_sample.c
+++ b/sdk/samples/iot/paho_iot_hub_sas_telemetry_sample.c
@@ -9,7 +9,7 @@
 #define TELEMETRY_SEND_INTERVAL_SEC 1
 #define TELEMETRY_NUMBER_OF_MESSAGES 5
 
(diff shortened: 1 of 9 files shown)


sdk/samples/iot/paho_iot_hub_sas_telemetry_sample.c
```diff
@@ -151,7 +151,7 @@ void connect_mqtt_client_to_iot_hub()
   if (*az_span_ptr(env_vars.x509_trust_pem_file_path)
```
ahsonkhan, line 151: Is it feasible for this to use `az_span_size(...) != 0`? If so, that is the preferred check.

Otherwise, nit, use the `NULL` keyword, rather than comparing to the null byte.

momuno: making change in new PR

PR #1057, opened 2020-08-15
Core and Storage docs formatting tweaks

### LINKED ISSUE #1037: Use @returns instead of formatted @return

antkmsft:
Closes #1037.

* Use `@retval` to describe concrete return values where applicable.
* Mark parameters as `[in]`, `[out]` and `[in,out]`.
* Update the remaining places where `ref_` and `out_` convention naming scheme was not used.
* Add some minor missing documentation entries.
* Remove repetition of the function name in the description (Doxygen does it).
* Remove file name from the `@file` section - Doxygen still adds it to documentation perfectly, only that now if you rename a file, you don't need to update the comment, the name will get picked up automatically.


sdk/inc/azure/core/_az_cfg.h
```diff
@@ -68,6 +70,7 @@
   } while (0)
 #endif // __GNUC__ >= 7
 
+/// @brief Enforce that the return value get handled (if compiler supports the feature).
```
RickWinter, line 73:
```suggestion
/// @brief Enforce that the return value is handled (only applicable on supported compilers).
```

sdk/inc/azure/core/_az_cfg_prefix.h
```diff
@@ -2,9 +2,10 @@
 // SPDX-License-Identifier: MIT
 
 /**
- * @file _az_cfg_prefix.h
+ * @file
  *
- * @brief Open "extern C" and save warnings state.
+ * @brief Opens "extern C" and saves warnings state. Do not include this file directly. This file
```
RickWinter, line 7: This is repetitive.  The same statement is made is the note about using '_' items.

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -68,29 +66,30 @@ typedef struct
...
  *
- * @param client The blob client instance to initialize.
- * @param endpoint A url to a blob storage account.
- * @param credential The object used for authentication.
- *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS.
- * @param options A reference to an #az_storage_blobs_blob_client_options structure which defines
- * custom behavior of the client.
+ * @param[out] out_client The blob client instance to initialize.
+ * @param[in] endpoint A URL to a blob storage account.
+ * @param credential The object used for authentication. #AZ_CREDENTIAL_ANONYMOUS should be used for
```
RickWinter, line 71: [in]

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -108,38 +108,38 @@ typedef struct
...
+                                                 ._internal = { .unused = AZ_SPAN_NULL } };
 }
 
 /**
  * @brief Uploads the contents to blob storage.
  *
- * @param client A storage blobs client structure.
- * @param content The blob content to upload.
- * @param options __[nullable]__ A reference to an #az_storage_blobs_blob_upload_options
+ * @param client An #az_storage_blobs_blob_client structure.
```
RickWinter, line 128: [in]

sdk/tests/core/inc/az_test_span.h
```diff
@@ -27,12 +27,13 @@
 #include <azure/core/_az_cfg_prefix.h>
 
 /**
- * @brief az_span_for_test_init returns a span over a byte buffer, directly invoking
- * #az_span_create.  The buffer is initialized with 0xcc to help tests check against buffer overflow.
+ * @brief Returns a span over a byte buffer, directly invoking #az_span_create. The buffer is
+ * initialized with `0xCC` to help tests check against buffer overflow.
```
RickWinter, line 31: Nit, we should fill with something that is highly unlikely to exist.  i.e. "0xBAADF00D"

sdk/inc/azure/storage/az_storage_blobs.h
ahsonkhan: I was about to say that the client shouldn't be marked with `[in]`.

But, for some reason, GitHub is confusing what line this comment corresponds to.
image

In the "Files Changed" view, you can see this comment is about the credential parameter, not the client.
https://github.com/Azure/azure-sdk-for-c/pull/1057/files
image

sdk/tests/core/inc/az_test_span.h
ahsonkhan: This was the rationale, from @CIPop, for choosing 0xCC:
https://github.com/Azure/azure-sdk-for-c/pull/534#discussion_r401269100

> Nit: Can we actually set this to 0xCC (Assembly INT3) instead?
Would immediately stop in a debugger if the filler somehow gets executed on x86.

@jspaith added this capability.

Re-iterating feedback from https://github.com/Azure/azure-sdk-for-c/pull/534#discussion_r401875781
> Makes sense. Can we still capture @CIPop feedback to make it 0xcc in a comment in one of the two places just so the rationale of why we picked this lives in the code? Otherwise, I would have just used 0xFF and it's good to know that 0xCC has more benefit.

antkmsft: I agree, `0xCC` dosn't look bad in debugger either, and has `int3` benefits. Thank you Ahson for bringing the context!

sdk/inc/azure/core/_az_cfg_prefix.h
antkmsft: But on the other hand, the statement does not explicitly mention files I'd double-mention in this case, but I don't have strong opinion, of others think I should remove this note.

sdk/inc/azure/storage/az_storage_blobs.h
antkmsft: The reason I didn't mark it is that logically it does not mutate, but factually it does. But since all the mutations are in the `_internal`, these can be marked as `[in]` I think.

sdk/inc/azure/storage/az_storage_blobs.h
antkmsft: The reason I didn't mark it is that logically it does not mutate, but factually it does. But since all the mutations are in the `_internal`, these can be marked as `[in]` I think.

ahsonkhan: I would think about what contract we want to document to the customer and what expectations we want them to have, especially as it changes in the future. If we are mutating the internal, and that is necessary, can we decide to make those fields public and not go against the "documented contract"?

If we want the flexibility, and do mutate it today, I would lean towards not marking it as `[in]` to leave that option open. If we don't ever foresee mutating it in the future, and we want to give that guarantee to the user, only then should you mark it as `[in`].

ahsonkhan:
Please wait until I have reviewed this PR before merging. I will make an effort to get to it before eod tomorrow.

sdk/inc/azure/core/az_http_transport.h
```diff
@@ -154,54 +142,55 @@ AZ_NODISCARD az_result az_http_request_get_url(az_http_request const* request, a
...
+ * @retval #AZ_ERROR_HTTP_RESPONSE_OVERFLOW There was any issue while trying to write into \p
  * ref_response. It might mean that there was not enough space in \p ref_response to hold the entire
  * response from the network.
- *         - #AZ_ERROR_HTTP_RESPONSE_COULDNT_RESOLVE_HOST if the url from \p ref_request can't be
- * resolved by the http stack and the request was not sent.
- *         - #AZ_ERROR_HTTP_PLATFORM any other issue from the transport layer.
- *
+ * @retval #AZ_ERROR_HTTP_RESPONSE_COULDNT_RESOLVE_HOST The URL from \p ref_request can't be
+ * resolved by the HTTP stack and the request was not sent.
+ * @retval #AZ_ERROR_HTTP_ADAPTER Any other issue from the transport adapter layer.
```
ahsonkhan, line 193: Nice catch.

sdk/inc/azure/core/az_json.h
```diff
@@ -24,9 +24,7 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
-/**
- * @brief Defines symbols for the various kinds of JSON tokens that make up any JSON text.
- */
+/// @brief Defines symbols for the various kinds of JSON tokens that make up any JSON text.
```
ahsonkhan, line 27: Same here/elsewhere. Revert these doc format changes.

sdk/inc/azure/core/az_json.h
```diff
@@ -96,102 +95,111 @@ typedef struct
...
- * @remarks If \p token doesn't contain any text, this function will just return \p destination.
+ * of \p json_token.
+ * @remarks If \p json_token doesn't contain any text, this function will just return \p
+ * destination.
  */
 az_span az_json_token_copy_into_span(az_json_token const* json_token, az_span destination);
 
 /**
- * @brief Returns the JSON token's boolean.
+ * @brief Gets (as an output parameter) the JSON token's boolean.
```
ahsonkhan, line 113: I don't think we need to state the `(as an output parameter)`.

```suggestion
 * @brief Gets the JSON token's boolean.
```

sdk/inc/azure/core/az_json.h
```diff
@@ -278,42 +283,43 @@ typedef struct
...
  *
- * @param[out] json_writer A pointer to an #az_json_writer instance to initialize.
- * @param[in] destination_buffer An #az_span over the byte buffer where the JSON text is to be
- * written.
+ * @param[out] out_json_writer A pointer to an #az_json_writer instance to initialize.
+ * @param destination_buffer An #az_span over the byte buffer where the JSON text is to be written.
  * @param[in] options __[nullable]__ A reference to an #az_json_writer_options
  * structure which defines custom behavior of the #az_json_writer. If `NULL` is passed, the writer
- * will use the default options (i.e. #az_json_writer_options_default()).
+ * will use the default options (i.e. #az_json_writer_options_default).
```
ahsonkhan, line 290: `az_json_writer_options_default` is a method.

```suggestion
 * will use the default options (i.e. #az_json_writer_options_default()).
```

sdk/inc/azure/core/az_json.h
```diff
@@ -278,42 +283,43 @@ typedef struct
...
  *
- * @return An #az_result value indicating the result of the operation:
- *         - #AZ_OK if the az_json_writer is initialized successfully
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK #az_json_writer is initialized successfully.
+ * @retval other Initialization failure.
  */
 AZ_NODISCARD az_result az_json_writer_init(
-    az_json_writer* json_writer,
+    az_json_writer* out_json_writer,
```
ahsonkhan, line 297: I don't think we should call this `out_json_writer`.

Revert?
```suggestion
    az_json_writer* json_writer,
```

sdk/inc/azure/core/az_json.h
```diff
@@ -555,40 +562,42 @@ typedef struct
...
- * @param[out] json_reader A pointer to an #az_json_reader instance to initialize.
+ * @param[out] out_json_reader A pointer to an #az_json_reader instance to initialize.
  * @param[in] json_buffers An array of non-contiguous byte buffers, as spans, containing the JSON
  * text to read.
  * @param[in] number_of_buffers The number of buffer segments provided, i.e. the length of the \p
  * json_buffers array.
  * @param[in] options __[nullable]__ A reference to an #az_json_reader_options
  * structure which defines custom behavior of the #az_json_reader. If `NULL` is passed, the reader
- * will use the default options (i.e. #az_json_reader_options_default()).
+ * will use the default options (i.e. #az_json_reader_options_default).
```
ahsonkhan, line 596: Same here/elsewhere. It's a method.

sdk/inc/azure/core/az_log.h
```diff
@@ -2,10 +2,14 @@
...
- * @file az_log.h
+ * @file
  *
- * @brief This header defines the types and functions your application uses
- *        to be notified of Azure SDK client library log messages.
+ * @brief This header defines the types and functions your application uses to be notified of Azure
+ * SDK client library log messages.
+ *
+ * If you define the `AZ_NO_LOGGING` symbol when compiling the SDK code (or adding option
+ * `-DLOGGING=OFF` with cmake), all of the Azure SDK logging checking will be excluded making the
```
ahsonkhan, line 11:
```suggestion
 * `-DLOGGING=OFF` with cmake), all of the Azure SDK logging checks will be excluded making the
```

sdk/inc/azure/core/az_log.h
```diff
@@ -27,13 +31,13 @@
   ((int32_t)((uint32_t)(facility) << 16) | (uint32_t)(code))
 
 /**
- * @brief az_log_classification identify the classifications of log messages
- *        that the Azure SDK client libraries produce.
+ * @brief Identifies the classifications of log messages that the Azure SDK Core or Azure SDK client
+ * libraries produce.
```
ahsonkhan, line 35:
```suggestion
 * @brief Identifies the classifications of log messages produced by the SDK.
```

sdk/inc/azure/core/az_log.h
```diff
@@ -61,22 +65,21 @@ typedef enum
...
- * @brief az_log_set_classifications allows the application to specify which
- * az_log_classification types it is interested in receiving. If no classifications are set
- * (NULL), the application will receive log messages for all az_log_classification values.
+ * @brief Allows the application to specify which #az_log_classification types it is interested in
+ * receiving. If no classifications are set (`NULL`), the application will receive log messages for
+ * all #az_log_classification values.
  *
- * @param classifications An array of az_log_classification values.
- *                        The last element of the array must be AZ_LOG_END_OF_LIST.
+ * @param[in] classifications __[nullable]__ An array of az_log_classification values, terminated by
```
ahsonkhan, line 81: What does nullable mean here? Consider moving the sentence from the brief  to the parameter description.

sdk/inc/azure/core/az_log.h
```diff
@@ -88,12 +91,12 @@ AZ_INLINE void az_log_set_classifications(az_log_classification const classifica
 #endif // AZ_NO_LOGGING
 
 /**
- * @brief az_log_set_callback sets the function that will be invoked to report an Azure SDK
- * client library log message.
+ * @brief Sets the function that will be invoked to report an Azure SDK Core or client library log
+ * message.
```
ahsonkhan, line 95:
```suggestion
 * @brief Sets the function that will be invoked to report an SDK log message.
```

sdk/inc/azure/core/az_log.h
```diff
@@ -88,12 +91,12 @@ AZ_INLINE void az_log_set_classifications(az_log_classification const classifica
...
  *
- * @param az_log_message_callback A pointer to the function that will be invoked when an Azure
- * SDK client library report a log message matching one of the az_log_classifications passed to
- * az_log_set_classifications.
+ * @param[in] az_log_message_callback __[nullable]__ A pointer to the function that will be invoked
```
ahsonkhan, line 97: What does passing null mean?

sdk/inc/azure/core/az_platform.h
```diff
@@ -13,37 +24,40 @@
...
 /**
- * @brief Conditionally exchanges values of two pointers in a
- * thread-safe manner.
+ * @brief Conditionally exchanges values of two pointers in a thread-safe manner.
  *
- * @param obj A pointer to a pointer that needs to be changed.
- * @param expected An expected value of a value that \p obj is pointing to, prior to exchange.
- * @param desired A value to assign to the value that is pointed by \p obj, if its value equals to
- * \p expected.
+ * @param[in,out] ref_obj A pointer to a pointer that needs to be changed.
```
ahsonkhan, line 49:
```suggestion
 * @param[in,out] ref_obj A pointer to a volatile pointer that needs to be changed.
```

OR is it:
```suggestion
 * @param[in,out] ref_obj A volatile pointer to a pointer that needs to be changed.
```

sdk/inc/azure/core/az_result.h
```diff
@@ -60,17 +60,12 @@ enum
...
-    - 31 Severity (0 - success, 1 - failure).
-    - 16..30 Facility.
-    - 0..15 Code.
-*/
+// az_result Bits:
+//   - 31 Severity (0 - success, 1 - failure).
+//   - 16..30 Facility.
+//   - 0..15 Code.
+
+/// @brief The type represents success and error conditions.
```
ahsonkhan, line 68:
```suggestion
/// @brief The type represents the various success and error conditions.
```

sdk/inc/azure/core/az_result.h
```diff
@@ -101,30 +96,46 @@ typedef enum
...
 
   // Platform
   AZ_ERROR_OUT_OF_MEMORY = _az_RESULT_MAKE_ERROR(
       _az_FACILITY_PLATFORM,
       1), ///< Dynamic memory allocation request was not successful.
 
   // JSON error codes
+
+  /// The kind of the token being read is not compatible with the type of the value attempted being
+  /// read.
```
ahsonkhan, line 109:
```suggestion
  /// The kind of the token being read is not compatible with the expected type of the value.
```

sdk/inc/azure/core/az_result.h
```diff
@@ -133,26 +144,32 @@ typedef enum
       8), ///< Generic error in the HTTP transport adapter implementation.
 
   // IoT error codes
+
+  /// The IoT topic is not matching the expected format.
```
ahsonkhan, line 148: Added in https://github.com/Azure/azure-sdk-for-c/pull/565. @danewalton - review the doc comment please.

sdk/inc/azure/core/az_result.h
```diff
@@ -133,26 +144,32 @@ typedef enum
...
 } az_result;
 
-/// Checks whether the \a result provided indicates a failure.
-///
-/// @param result Result value to check for failure.
-///
-/// @retval true \a result indicates a failure.
-/// @retval false \a result is successful.
+/**
+ * @brief Checks whether the \a result provided indicates a failure.
```
ahsonkhan, line 153: We use `\p` elsewhere. Keep using that?

```suggestion
 * @brief Checks whether the \p result provided indicates a failure.
```

sdk/inc/azure/core/az_result.h
```diff
@@ -133,26 +144,32 @@ typedef enum
...
+ *
+ * @param[in] result Result value to check for failure.
+ *
+ * @retval true \a result indicates a failure.
+ * @retval false \a result is successful.
+ */
 AZ_NODISCARD AZ_INLINE bool az_failed(az_result result)
```
ahsonkhan, line 160: @JeffreyRichter - do we need both `az_failed` and `az_succeeded` or can we just have `az_succeeded`?

Semantically, isn't one just a negation of the other. Why have both?

sdk/inc/azure/core/az_span.h
```diff
@@ -34,21 +34,23 @@ typedef struct
...
   } _internal;
 } az_span;
 
 /********************************  SPAN GETTERS */
 
 /**
  * @brief Returns the #az_span byte buffer's starting memory address.
- *
+ * @param[in] span An #az_span to get the buffer's starting memory address of.
+ * @return Starting memory address of \p span buffer.
```
ahsonkhan, line 46: Is having a `@return` required? I feel like repeating the same information three times would unnecessarily clutter up the docs and force users to read more.

sdk/inc/azure/core/az_span.h
```diff
@@ -57,8 +59,9 @@ AZ_NODISCARD AZ_INLINE int32_t az_span_size(az_span span) { return span._interna
 /**
  * @brief Returns an #az_span over a byte buffer.
  *
- * @param[in] ptr The memory address of the 1st byte in the byte buffer.
- * @param[in] size The number of total bytes in the byte buffer.
+ * @param[in] ptr Memory address of the first byte in the byte buffer.
+ * @param[in] size Total number of bytes in the byte buffer.
```
ahsonkhan, line 63: Why did we reword this? I prefer the original.

sdk/inc/azure/core/az_span.h
```diff
@@ -225,15 +231,15 @@ AZ_NODISCARD bool az_span_is_content_equal_ignoring_case(az_span span1, az_span
...
  * \p destination.
  * @param[in] source The #az_span containing the not-0-terminated string to copy into \p
  destination.
  *
  * @remarks The buffer referred to by \p destination must have a size that is at least 1 byte bigger
  * than the \p source #az_span for the \p destination string to be zero-terminated.
- * Content is copied from the \p source buffer and then \0 is added at the end.
+ * Content is copied from the \p source buffer and then `\0` is added at the end.
  */
 void az_span_to_str(char* destination, int32_t destination_max_size, az_span source);
```
ahsonkhan, line 244: This could be const and we can keep `[in]`.

sdk/inc/azure/core/az_span.h
```diff
@@ -243,10 +249,12 @@ void az_span_to_str(char* destination, int32_t destination_max_size, az_span sou
...
- * @return The position of \p target in \p source if \p source contains the \p target within it:
- *         - 0 if \p target is empty (if its size is equal zero)
- *         - -1 if \p source is empty (if its size is equal zero) and \p target is non-empty
- *         - -1 if \p target is not found in `source`
+ *
+ * @return The position of \p target in \p source if \p source contains the \p target within it.
+ * @retval 0 \p target is empty (if its size is equal zero).
+ * @retval -1 \p source is empty (if its size is equal zero) and \p target is non-empty.
+ * @retval -1 \p target is not found in `source`.
+ * @retval other The position of \p target in \p source.
```
ahsonkhan, line 257:
```suggestion
 * @retval non-zero positive number The position of \p target in \p source.
```

sdk/inc/azure/core/az_span.h
```diff
@@ -243,10 +249,12 @@ void az_span_to_str(char* destination, int32_t destination_max_size, az_span sou
...
  * @param[in] target The #az_span containing the tokens to be searched within \p source.
- * @return The position of \p target in \p source if \p source contains the \p target within it:
- *         - 0 if \p target is empty (if its size is equal zero)
- *         - -1 if \p source is empty (if its size is equal zero) and \p target is non-empty
- *         - -1 if \p target is not found in `source`
+ *
+ * @return The position of \p target in \p source if \p source contains the \p target within it.
+ * @retval 0 \p target is empty (if its size is equal zero).
+ * @retval -1 \p source is empty (if its size is equal zero) and \p target is non-empty.
+ * @retval -1 \p target is not found in `source`.
```
ahsonkhan, line 256:
```suggestion
 * @retval -1 \p target is not found in `source` OR \p source is empty (if its size is equal zero) and \p target is non-empty.
```

sdk/inc/azure/core/az_span.h
```diff
@@ -485,7 +514,7 @@ typedef struct
  * This method must never return an empty #az_span, unless the requested buffer size is not
  * available. In which case, it must return an error #az_result.
  *
- * @remarks The caller must check the return value using #az_succeeded() before continuing to use
+ * @remarks The caller must check the return value using #az_succeeded before continuing to use
```
ahsonkhan, line 517: Revert, `az_succeeded` is a method.
```suggestion
 * @remarks The caller must check the return value using #az_succeeded() before continuing to use
```

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -68,29 +66,30 @@ typedef struct
 /**
  * @brief Initialize a client with default options.
  *
- * @param client The blob client instance to initialize.
- * @param endpoint A url to a blob storage account.
- * @param credential The object used for authentication.
- *         #AZ_CREDENTIAL_ANONYMOUS should be used for SAS.
- * @param options A reference to an #az_storage_blobs_blob_client_options structure which defines
- * custom behavior of the client.
+ * @param[out] out_client The blob client instance to initialize.
```
ahsonkhan, line 69: I think we should keep this as `client`.

sdk/src/azure/core/az_json_private.h
```diff
@@ -86,15 +86,15 @@ AZ_INLINE _az_json_stack_item _az_json_stack_pop(_az_json_bit_stack* json_stack)
                                                         : _az_JSON_STACK_ARRAY;
 }
 
-AZ_INLINE void _az_json_stack_push(_az_json_bit_stack* json_stack, _az_json_stack_item item)
+AZ_INLINE void _az_json_stack_push(_az_json_bit_stack* ref_json_stack, _az_json_stack_item item)
```
ahsonkhan, line 89: Change pop a well, above.

sdk/inc/azure/storage/az_storage_blobs.h
```diff
@@ -108,39 +108,39 @@ typedef struct
...
+ * @retval #AZ_OK Success.
+ * @retval other Failure.
  */
 AZ_NODISCARD az_result az_storage_blobs_blob_upload(
     az_storage_blobs_blob_client* client,
     az_span content, /* Buffer of content*/
-    az_storage_blobs_blob_upload_options* options,
-    az_http_response* response);
+    az_storage_blobs_blob_upload_options const* options,
+    az_http_response* ref_response);
```
ahsonkhan, line 143: Shouldn't this be `out_response`?

sdk/inc/azure/core/_az_cfg.h
```diff
@@ -2,9 +2,10 @@
 // SPDX-License-Identifier: MIT
 
 /**
- * @file _az_cfg.h
+ * @file
```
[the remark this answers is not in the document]
antkmsft: I think it is better to have it, because otherwise, if the first entity in the header is not documented, it is for sure will be associated with that entity. When we have a `@file`, we clearly mark that the comment section should be applied to current file description.

sdk/inc/azure/core/az_config.h
```diff
@@ -21,10 +21,14 @@
 
 enum
 {
-  AZ_HTTP_REQUEST_URL_BUFFER_SIZE = 2 * 1024, ///< The maximum buffer size for a URL.
-  AZ_HTTP_REQUEST_BODY_BUFFER_SIZE = 1024, ///< The maximum buffer size for an HTTP request body.
+  /// The maximum buffer size for a URL.
```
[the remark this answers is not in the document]
antkmsft: Yes, they are, however if the line gets split into two `///<` (which is eay to happen when auto-formatting), first line goes to description, and second goes to details.

sdk/inc/azure/core/az_context.h
```diff
@@ -88,26 +90,26 @@ az_context_create_with_value(az_context const* parent, void const* key, void con
 /**
  * @brief Cancels the specified #az_context node; this cancels all the child nodes as well.
  *
- * @param[in] context A pointer to the #az_context node to be canceled; passing `NULL` cancels the
- * root #az_context.
+ * @param[in,out] ref_context __[nullable]__ A pointer to the #az_context node to be canceled;
```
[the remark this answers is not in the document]
antkmsft: No, I think clear specification is better. If anything, I think we should be rather making `[in]` as default - as the most common case, and everything else annotated specifically. But I think we should all 4 - `[in]`, `[out]`, `[in,out]`, and unannotated.
Annotated is the rare case when the parameter is technically immutable, but logically it is. Example: `append_int_to_span(az_span s, int i)`, where I leave `s` as non-annotated.

sdk/inc/azure/core/az_credentials.h
```diff
@@ -27,39 +27,31 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
-/**
- * @brief AZ_CREDENTIAL_ANONYMOUS is equivalent to no credential (NULL).
- *
- */
+/// @brief Equivalent to no credential (NULL).
```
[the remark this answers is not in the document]
antkmsft: Rationale: less lines and less visual clutter. If a comment can be a single-line, it should be a single line, IMO.

sdk/inc/azure/core/az_credentials.h
```diff
@@ -27,39 +27,31 @@
...
 enum
 {
   _az_TOKEN_BUFFER_SIZE = 2 * 1024,
 };
 
-/**
- * @brief Definition of a auth token. It is used by the auth policy from http pipeline as part of a
- * provided credential. User should not access _internal field.
- *
- */
```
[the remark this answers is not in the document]
antkmsft: I would prefer to have them as `//`, but keep the doxygen style, this way it won't leak. It is easy to transform regular comment into doxygen-styled one.

sdk/inc/azure/core/az_http.h
```diff
@@ -26,100 +26,111 @@
...
+  AZ_HTTP_STATUS_CODE_BAD_GATEWAY = 502, ///< HTTP 502 Bad Gateway.
+  AZ_HTTP_STATUS_CODE_SERVICE_UNAVAILABLE = 503, ///< HTTP 503 Unavailable.
+  AZ_HTTP_STATUS_CODE_GATEWAY_TIMEOUT = 504, ///< HTTP 504 Gateway Timeout.
+  AZ_HTTP_STATUS_CODE_HTTP_VERSION_NOT_SUPPORTED = 505, ///< HTTP 505 HTTP Version Not Supported.
+  AZ_HTTP_STATUS_CODE_VARIANT_ALSO_NEGOTIATES = 506, ///< HTTP 506 Variant Also Negotiates.
+  AZ_HTTP_STATUS_CODE_INSUFFICIENT_STORAGE = 507, ///< HTTP 507 Insufficient Storage.
+  AZ_HTTP_STATUS_CODE_LOOP_DETECTED = 508, ///< HTTP 508 Loop Detected.
+  AZ_HTTP_STATUS_CODE_NOT_EXTENDED = 510, ///< HTTP 510 Not Extended.
+  AZ_HTTP_STATUS_CODE_NETWORK_AUTHENTICATION_REQUIRED
+  = 511, ///< HTTP 511 Network Authentication Required.
```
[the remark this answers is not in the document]
antkmsft: Now they are documented, previously they were not. I find HTTP specific number to be specifically useful, because, maybe it's just me, but I feel more confident when I write the HTTP code and know it matches the integral number that I am expecting.

What do you mean by capitalization?

sdk/inc/azure/core/az_http.h
```diff
@@ -156,16 +166,21 @@ typedef struct
...
- * @param response The pointer to an az_http_response instance which is to be initialized.
- * @param buffer A span over the byte buffer that is to be filled with the HTTP response data. This
- * buffer must be large enough to hold the entire response.
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK Success.
+ * @retval other Initialization failed.
  */
-AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* response, az_span buffer)
+AZ_NODISCARD AZ_INLINE az_result
+az_http_response_init(az_http_response* out_response, az_span buffer)
```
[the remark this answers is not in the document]
antkmsft: I don't think we should be making exceptions for init. What makes init so different from `parse_int(span, int* out_result)`? Even more than that, I find it is also useful in init functions. When you see `out_`, you know you can pass an uninitialized garbage and it's ok.
`ref_` and `out_` is the current decision, and a work item to update guidelines with it is https://github.com/Azure/azure-sdk-for-c/issues/496

sdk/inc/azure/core/az_http.h
```diff
@@ -181,62 +196,64 @@ AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* respons
...
 typedef struct
 {
-  az_span reason_phrase;
-  az_http_status_code status_code;
-  uint8_t major_version;
-  uint8_t minor_version;
+  az_span reason_phrase; ///< Reason Phrase.
+  az_http_status_code status_code; ///< Status Code.
+  uint8_t major_version; ///< HTTP Major Version.
+  uint8_t minor_version; ///< HTTP Minor Version.
```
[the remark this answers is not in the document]
antkmsft: This doesn't look like a placeholder comment to me. Reason phrase is a standard domain-level terminology from RFCs, such as RFC7230 (specific line https://tools.ietf.org/html/rfc7230#section-3.1.2): `status-line = HTTP-version SP status-code SP reason-phrase CRLF`

sdk/inc/azure/core/az_http.h
```diff
@@ -181,62 +196,64 @@ AZ_NODISCARD AZ_INLINE az_result az_http_response_init(az_http_response* respons
...
- * @brief az_http_response_get_status_line returns the az_http_response_status_line
- * information within an HTTP response.
+ * @brief Returns the #az_http_response_status_line information within an HTTP response.
  *
- * @param response The az_http_response with an http response.
- * @param out_status_line The pointer to an az_http_response_status_line structure to be filled in
- * by this function.
- * @return AZ_OK = inline code is parsed and returned.<br>
- * Other value =  http response was not parsed
+ * @param[in,out] ref_response The #az_http_response with an HTTP response.
```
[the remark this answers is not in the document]
antkmsft: I disagree. If anything, `[in]` is more common, it should be the default. But I think we need all three + not specified in some cases.

sdk/inc/azure/core/az_config.h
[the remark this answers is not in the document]
ahsonkhan: I think we should always use the multi-line comment format. This way, if we write more comments, we don't need to switch from single line to multiple format on top of the  content change.

sdk/inc/azure/core/az_context.h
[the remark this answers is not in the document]
ahsonkhan: I like the idea of picking the most common one, and have it be the default (where we don't specify). Maybe that's `[in]`.

sdk/inc/azure/core/az_credentials.h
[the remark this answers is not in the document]
ahsonkhan: > Rationale: less lines and less visual clutter.

Gotcha.

> If a comment can be a single-line, it should be a single line, IMO.

I disagree. I think we should converge on the most flexible style and stick to it.

sdk/inc/azure/core/az_http.h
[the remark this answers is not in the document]
ahsonkhan: > What do you mean by capitalization?

For example, maybe change this to be lower-cased:
> HTTP 511 Network Authentication Required.

HTTP 511 Network authentication required.

sdk/inc/azure/core/az_json.h
antkmsft: It is an out parameter. Otherwise, there is no output from the function?
It is clearly an out paramter, because in the body, we do not read from it, and only write:
```
_az_PRECONDITION_NOT_NULL(out_json_writer);

  *out_json_writer = (az_json_writer){
    ._internal = {
      .destination_buffer = destination_buffer,
      .allocator_callback = NULL,
      .user_context = NULL,
      .bytes_written = 0,
      .total_bytes_written = 0,
      .need_comma = false,
      .token_kind = AZ_JSON_TOKEN_NONE,
      .bit_stack = { 0 },
      .options = options == NULL ? az_json_writer_options_default() : *options,
    },
  };
  return AZ_OK;
```

If this is not an output parameter, nothing is.

And I know that with out parameters, I can do
```
az_json_writer writer; // no initialization -gets initialized with random bytes from stack
if (az_succeeded(az_json_writer_init(&writer))
{
    writer-> ... // I can be sure that at this point, *writer initialized, and is safe to use, and it didn't matter that I dodn't initialize it with "writer = { 0 };" in the beginning.
}
```

sdk/inc/azure/core/az_http.h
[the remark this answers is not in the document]
ahsonkhan: I don't think I was part of the discussion where we decided to go through with these renames. I think we should bring this up with the group. It seems like unnecessary complexity that we are exposing to the user. 

I am not sure how valuable/helpful being super explicit is, in cases like `_init`.

sdk/inc/azure/core/az_http.h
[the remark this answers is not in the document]
ahsonkhan: Is it optional? Can it be a null or an empty az_span?

sdk/inc/azure/core/az_result.h
antkmsft: I pulled the comment from IoT's doxygen descriptions, which are very consistent when they are talking about this error code.
https://github.com/Azure/azure-sdk-for-c/search?q=%22+not+matching+the+expected+format%22&unscoped_q=%22+not+matching+the+expected+format%22

sdk/inc/azure/core/az_span.h
antkmsft: I'd leave it. If we are commenting, let's stick to some set of things that we always put there. Maybe you are right, maybe it is subjective. If I was just writing docs from myself, I'd also have some things writtern differently, or some left not documented. But what's obvious for me, may not be obvious for others.

sdk/inc/azure/core/az_credentials.h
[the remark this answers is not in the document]
ahsonkhan: > I would prefer to have them as `//`, but keep the doxygen style, this way it won't leak. 

I don't understand what you mean here.

> It is easy to transform regular comment into doxygen-styled one.

That may be true, but why incur the cost to change it? I don't see a benefit.

sdk/inc/azure/core/_az_cfg.h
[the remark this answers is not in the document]
ahsonkhan: > because otherwise, if the first entity in the header is not documented, it is for sure will be associated with that entity.

Can you clarify what happens if the first entity in the header is **not** documented and we **don't** have `@file`? How do we benefit from the association or how does the user/generated docs benefit?

Also, once we turn on docs validation in CI, we wouldn't have missing docs anyway (or else CI would fail).

sdk/inc/azure/core/az_span.h
antkmsft: Maybe it's my English, but I've seen "total number" everywhere but haven't seen "number of total". With previous wording, it sounds like "total" is attribute of a byte ("total byte"), and not an attribute of "number" (i.e. not "total number", but `number_of(total_bytes)`). I.e. previous is `count(sum(bytes))`, new is `sum(count(bytes))`.

Book ngrams: https://books.google.com/ngrams/graph?content=total+number+of%2Cnumber+of+total&year_start=1800&year_end=2019&corpus=26&smoothing=0

Search trends: https://trends.google.com/trends/explore?date=all&q=%22total%20number%20of%22,%22number%20of%20total%22

ahsonkhan: Sorry, let me clarify. Yo are right that it should be `total number of bytes` and **not** `number of total bytes`

I meant the following, for both ptr and size (keep the `the`):
```C
* @param[in] ptr The memory address of the first byte in the byte buffer.
 * @param[in] size The total number of bytes in the byte buffer.
```

ahsonkhan, line 63:
```suggestion
 * @param[in] ptr The memory address of the first byte in the byte buffer.
 * @param[in] size The total number of bytes in the byte buffer.
```

sdk/inc/azure/storage/az_storage_blobs.h
ahsonkhan: From offline discussion with @antkmsft, no, it should remain ref. We should document this in storage.

sdk/inc/azure/storage/az_storage_blobs.h
ahsonkhan: >  but factually it does.

What mutations are we doing to the client on a call to upload (even internal state)? That seems surprising to me.

sdk/src/azure/storage/az_storage_blobs_blob_client.c
```diff
@@ -54,19 +54,19 @@ AZ_NODISCARD az_storage_blobs_blob_client_options az_storage_blobs_blob_client_o
 }
 
 AZ_NODISCARD az_result az_storage_blobs_blob_client_init(
-    az_storage_blobs_blob_client* client,
-    az_span uri,
+    az_storage_blobs_blob_client* out_client,
+    az_span endpoint,
```
ahsonkhan, line 58: Good catch! I wish we had a linter tool that detected header params and source params matched. I ran one locally a while back.

PR #1074, opened 2020-08-19
update IoT unit test names to unify

danewalton:
Some names of unit tests were inconsistent. Patching those up.

### DIFF
--- a/sdk/tests/iot/hub/main.c
+++ b/sdk/tests/iot/hub/main.c
@@ -15,12 +15,12 @@ int main()
 {
   int result = 0;
 
-  result += test_iot_hub_client();
-  result += test_iot_hub_client_c2d();
-  result += test_iot_hub_client_methods();
-  result += test_iot_hub_client_sas_token();
-  result += test_iot_hub_client_telemetry();
-  result += test_iot_hub_client_twin();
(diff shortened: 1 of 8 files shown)


sdk/tests/iot/hub/main.c
```diff
@@ -15,12 +15,12 @@ int main()
 {
   int result = 0;
 
-  result += test_iot_hub_c2d();
   result += test_iot_hub_client();
-  result += test_iot_sas_token();
-  result += test_iot_hub_telemetry();
-  result += test_az_iot_hub_client_twin();
-  result += test_iot_hub_methods();
+  result += test_iot_hub_client_c2d();
```
momuno, line 19: Why not use test_az_iot_hub_client_x for these?
I notices that naming structure is used in these files, but it is for the static functions.  This non-static is the exception.

danewalton: Yea my original logic was api's shouldn't be prefix'd with `az_` unless they were public but these are prefix'd with `test_`. I blame my lack of coffee this morning. Will change

momuno: are you thinking the static functions lose the az, and the public one gains it?  Or something else?

PR #1100, opened 2020-08-22
Test PR for access

jbobotek:


jbobotek:
> @jbobotek - what are you trying to test?

I'm trying to set up a pipeline for the Azure IoT C SDK team so that we can have some continuous integration running. However, I have not been able to get it to show as an option to run a pipeline on. I'm assuming it's an access issue, which is why I was seeing if I had contributor or merge permissions for this repo. These may be required to initialize a pipeline. Keeping @ericwol-msft @danewalton posted too.

ahsonkhan:
As an aside, @danieljurek, @weshaggard, interestingly this PR causes CI legs to fail.

Even though we have `/build` in a our .gitignore and generally this shouldn't happen, I am wondering if we should add some robustness to this check and either clean/delete any existing `/build` folder on the CI machine before creating a new one, or continue using the existing one rather than failing:
https://github.com/Azure/azure-sdk-for-c/blob/bc577de6d8a601f845e3b695dca26e5f88ee1f8c/.gitignore#L300-L301

https://github.com/Azure/azure-sdk-for-c/blob/c6d381ef51168434a0a45db66548580d668b85b8/eng/pipelines/templates/steps/cmake-build.yml#L5-L6

https://dev.azure.com/azure-sdk/public/_build/results?buildId=506083&view=logs&j=65a3cc1a-1ab3-5128-7575-715a1aa65eff&t=1c18bbb5-6350-5432-f724-1a6db161bc91
```text
/bin/bash --noprofile --norc /home/vsts/work/_temp/bf2f2eb1-98a6-4528-b2f9-9387331e46c3.sh
mkdir: cannot create directory ‘build’: File exists

##[error]Bash exited with code '1'.
```

Or do you think that would be an undesired behavior?

Use `mkdir -p instead`?
https://stackoverflow.com/a/22738025

weshaggard:
We should probably make the build pipelines more resilient if the build directory already exists.

@jbobotek what sort of pipeline are you trying to setup? If in Azure DevOps which org are you using as this is already hooked up to the azure-sdk org and you cannot setup pipelines with automatic triggers in multiple orgs.

ahsonkhan:
> We should probably make the build pipelines more resilient if the build directory already exists.

https://github.com/Azure/azure-sdk-for-c/issues/1122

RickWinter:
@jbobotek Can this PR be closed?

jbobotek:
Yes.

PR #1126, opened 2020-08-25
Updates: iot_sample_common, c2d sample, methods sample, telemetry sample

momuno:
This commit focuses on:
- iot_sample_common.h and .c files.  The changes affecting all other sample files are also included.
- c2d, methods and telemetry samples.  

Items covered:
const, out_, ref_, correct commenting for functions, function naming, parameter naming, logging.


sdk/samples/iot/paho_iot_hub_c2d_sample.c
```diff
@@ -44,9 +44,9 @@ static void disconnect_mqtt_client_from_iot_hub(void);
 
 static void parse_c2d_message(
     char* topic,
-    int topic_len,
-    const MQTTClient_message* message,
-    az_iot_hub_client_c2d_request* c2d_request);
+    int const topic_len,
```
ahsonkhan, line 47: Primitive parameters like `int` that are passed by value (rather than ref/by pointer) don't need `const`.
```suggestion
    int topic_len,
```

sdk/samples/iot/paho_iot_hub_c2d_sample.c
```diff
@@ -244,23 +245,24 @@ static void disconnect_mqtt_client_from_iot_hub(void)
 
 static void parse_c2d_message(
     char* topic,
-    int topic_len,
-    const MQTTClient_message* message,
-    az_iot_hub_client_c2d_request* c2d_request)
+    int const topic_len,
```
ahsonkhan, line 248: Same here/elsewhere. Will stop mentioning.

```suggestion
    int topic_len,
```

sdk/samples/iot/iot_sample_common.h
```diff
@@ -0,0 +1,190 @@
...
+
+/*
+ * @brief Reads in environment variables set by user for purposes of running sample.
+ *
+ * @param[in] type The enumerated type of the sample.
+ * @param[in] name The enumerated name of the sample.
+ * @param[out] out_env_vars A pointer to the struct containing all read-in environment variables.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK All required environment variables successfuly read-in.
```
ahsonkhan, line 133:
```suggestion
 * @retval #AZ_OK All required environment variables successfully read-in.

```

sdk/samples/iot/iot_sample_common.h
```diff
@@ -0,0 +1,190 @@
...
+/*
+ * @brief Return total seconds passed including supplied minutes.
+ *
+ * @param[in] minutes Number of minutes to include in total seconds returned.
+ * @return Total time in seconds.
+ */
+uint32_t iot_sample_get_epoch_expiration_time_from_minutes(uint32_t minutes);
+
+/*
+ * @brief Generate the b64 encoded and signed signature using HMAC-SHA256 signing.
```
ahsonkhan, line 177: Is b64 commonly known shorthand? I think we should be explicit. Same with parameter names.

```suggestion
 * @brief Generate the base64 encoded and signed signature using HMAC-SHA256 signing.

```

ahsonkhan:
@antkmsft - can you please do a quick review of this as well.

sdk/samples/iot/iot_sample_common.h
```diff
@@ -0,0 +1,190 @@
...
+az_result iot_sample_create_mqtt_endpoint(
+    iot_sample_type type,
+    iot_sample_environment_variables const* env_vars,
+    char* out_endpoint,
+    size_t endpoint_size);
+
+/*
+ * @brief Sleep for given seconds.
+ *
+ * @param[in] seconds Number of seconds to sleep.
```
ahsonkhan, line 164: @antkmsft - I don't know how helpful `[in]` is for primitive value types being passed by value (rather than as a reference using `*`). Isn't it implicit/given that such parameters are `[in]`. We _could_ just leave these unspecified. What do you think?

sdk/samples/iot/iot_sample_common.c
```diff
@@ -46,27 +47,27 @@
 //
 //#define USE_WEB_SOCKET // Comment to use MQTT without WebSockets.
 #ifdef USE_WEB_SOCKET
-const az_span mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("wss://");
+az_span const mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("wss://");
```
danewalton, line 50: ew I don't like this but if our sdk guideline overlords say so I'll get used to it.

ps: actually a lot of the guidelines seem to have it on the left for declaration. 
https://azure.github.io/azure-sdk/clang_design.html#globals
@ahsonkhan are we to go through with this mismatch?

sdk/samples/iot/iot_sample_common.c
```diff
@@ -294,18 +311,21 @@ void sleep_for_seconds(uint32_t seconds)
...
+uint32_t iot_sample_get_epoch_expiration_time_from_minutes(uint32_t minutes)
 {
   return (uint32_t)(time(NULL) + minutes * 60);
 }
 
-static az_result base64_decode(az_span base64_encoded, az_span destination, az_span* out_span)
+static az_result base64_decode(
+    az_span base64_encoded,
+    az_span decoded_bytes,
+    az_span* out_decoded_bytes)
```
danewalton, line 322: 👍

sdk/samples/iot/iot_sample_common.h
danewalton: in the context it would but yea expanding it out here is fine

sdk/samples/iot/paho_iot_hub_c2d_sample.c
momuno: lol.  i must have been really excited about const, because I'm not sure why I did that :D

sdk/samples/iot/paho_iot_hub_c2d_sample.c
momuno: yeah, that was the definition signature for the declaration above

sdk/samples/iot/iot_sample_common.c
ahsonkhan: I don't think we ever discussed the ordering of const for local variables, or even that you should use them.

I have no opinion one way or the other. For local variables, I am fine with const on the left, or not even using it, unless you really care/benefit from knowing it won't change as you write the code.

To me, its an implementation detail/preference that may not necessarily belong in the guidelines since it shouldn't impact the end user.

ahsonkhan: @barcharcraz - where are const for local variables supposed to go? And do we need to update the guidelines to be consistent?

ahsonkhan: I just realized @danewalton that you are referring to public globals, not locals.

RHS const is more consistent.

As an aside, we should follow the global naming guidelines (capitalized and `AZ_...`).

sdk/samples/iot/iot_sample_common.h
momuno: I am renaming function names and parameters to be more specific.  Can you please review my next commit? Thanks.

sdk/samples/iot/iot_sample_common.h
momuno: While it is likely obvious to the developer that it is an `in`, I do like being consistent with the [in] on `in`s.  Then there is no question about what we mean when there is nothing mentioned (like in the case of an az_span that will be changed but not used, e.g. destination).

I do see [in] it used for primitive types in a few functions here: https://github.com/Azure/azure-sdk-for-c/blob/f36af1eb3c47b96ecf1f5c4010efb3d9e6ea261f/sdk/inc/azure/core/az_json.h#L453

sdk/samples/iot/iot_sample_common.c
ericwol-msft: const should be on the left side. Putting it after the makes pointer look weird. 

`Object const * const var = NULL;`

vs

`const Object* const var = NULL;`

momuno: lots here.  sounds like we are doing const RHS always, regardless of local, global, ptr, non ptr, etc.  si?

I'm seeing internal vs. private terminology in the link dane shared. Does this sound accurate:
Public - means we support the function/global variable as part of the public API.
Internal - means we don't support the function/global variable as part of the public API.

If a global is static, then it must ALSO be internal. It should use _az_BLAH_BLAH_BLAH.
If a global is internal, that does NOT mean it is has to be static.  It should still use _az_BLAH_BLAH_BLAH
If a global is public, then it is not internal, meaning it is also not static.  It should use AZ_BLAH_BLAH_BLAH.

sdk/samples/iot/iot_sample_common.h
ahsonkhan: > While it is likely obvious to the developer that it is an `in`, I do like being consistent with the [in] on `in`s.

Sure, makes sense. The contribution to "noise" in the docs would be minimal anyway.

> I do see [in] used for primitive types in a few functions here:

Yes, I saw those too, which is why my question isn't about your particular PR or code change (what you have is good). It just reminded me about it. :)

We have some rare instances in the repo that don't use `[in]`:
https://github.com/Azure/azure-sdk-for-c/blob/730e2c3ed7a2fb8952b5e8b6ba5b0abbc2ff5019/sdk/src/azure/core/az_span_private.h#L27-L35

sdk/samples/iot/iot_sample_common.c
momuno: side: I would prefer if we were consistent with RHS everywhere for const, otherwise it gets challenging to remember when the rule applies and when it doesn't.

ahsonkhan: > const should be on the left side. Putting it after the makes pointers look weird.

@ericwol-msft - we discussed this in our last Friday meeting. We use `const* const` in all such places in core.

ahsonkhan: > I'm seeing internal vs. private terminology in the link dane shared. Does this sound accurate:
Public - means we support the function/global variable as part of the public API.
Internal - means we don't support the function/global variable as part of the public API.

> If a global is static, then it must ALSO be internal. It should use _az_BLAH_BLAH_BLAH.
If a global is internal, that does NOT mean it is has to be static. It should still use _az_BLAH_BLAH_BLAH
If a global is public, then it is not internal, meaning it is also not static. It should use AZ_BLAH_BLAH_BLAH.

Yes. That's right.

> ... and what do we do about the `g`?  is it g_az_BLAH, and G_AZ_BLAH?  or g_AZ_BLAH?

Nothing, we should update the guidelines to remove that. We use such prefixes in C++ (`g_`, `s_`, `c_`, `m_`).

momuno: On second thought, we are not in the AZ API at all with these samples.  Note, iot_smple_common functions and variables do not begin with _az_iot_sample.  They begin with iot_sample.  

With that format, it would seem the globals should be named g_iot_sample_MQTT_URL_PREFIX

momuno: > Yes. That's right. Although...
> 
> > If a global is static, then it must ALSO be internal.
> 
> I don't think that's true (and does it matter?). Where did you read that?

I didn't read that, I was inferring it.  So it could be static (private to the file), but public with respect to the API?  I guess, why would be do that?  No developer should be using it if it is private to the file...?

sdk/samples/iot/iot_sample_common.h
momuno: ah interesting.  This would be a good item for us to agree upon in the next C99 meeting.  I am finding some other issues like this.  What if I start an email listing the issues (code vs. guideline differences, or code vs no guideline spec).  We can gather what we are finding before shipping it off to a C99 meeting?

sdk/samples/iot/iot_sample_common.c
ahsonkhan: The question is meaningless when it comes to public static globals in header files:
https://stackoverflow.com/q/11967502

sdk/samples/iot/iot_sample_common.h
ahsonkhan: Great idea. Go for it.

sdk/samples/iot/iot_sample_common.c
momuno: that's a great point, since a header is the public API.  I agree.  Adding to list topics I emailed out...

momuno:
I have made the following updates per comments and looking at C guidelines.  Can it please be re-reviewed?  Curious about opinions on style. 

-MACRO to inline function where a function can be used, per C guidelines. (only in source file.  Need to still apply to header)
-sas function and variable renaming to be more specific.
-renaming for global (non static) variables to IOT_SAMPLE_BLAH.  I'm trying this out, since the sample functions and global variables are not part of the public AZ API.  So they shouldn't use _az_ or AZ_ prefix.  They are global, so then should be all caps.

ahsonkhan:
> -MACRO to inline function where a function can be used, per C guidelines.

We have cases where we use macros in the SDK so I am not sure about that particular guideline.

Moreover, `IOT_SAMPLE_PRECONDITION_NOT_NULL` in particular should be a macro since it is based off the existing `_az_PRECONDITION_NOT_NULL` macro we already have in core. We should be consistent here.

momuno:
> > -MACRO to inline function where a function can be used, per C guidelines.
> 
> We have cases where we use macros in the SDK so I am not sure about that particular guideline.
> 
> Moreover, `IOT_SAMPLE_PRECONDITION_NOT_NULL` in particular should be a macro since it is based off the existing `_az_PRECONDITION_NOT_NULL` macro we already have in core. We should be consistent here.
> 
> > -sas function and variable renaming to be more specific.
> > -renaming for global (non static) variables to IOT_SAMPLE_BLAH.  I'm trying this out, since the sample functions and global variables are not part of the public AZ API.  So they shouldn't use _az_ or AZ_ prefix.  They are global, so then should be all caps.
> 
> All these changes look good.

Thanks.  Will keep as MACRO, and add that to the list :)  Will change IOT_SAMPLE_PRECONDITION_NOT_NULL.

sdk/samples/iot/iot_sample_common.c
momuno: > As an aside, we should follow the global naming guidelines (capitalized and `AZ_...`).

Realized these actually should be static.  Scoped to the file, but not true globals.  The guidelines actually don't say anything about static global variables that are scoped to file.  They do mention static functions though, which these are like.  In that case, no prefix is needed at all.

Adding this topic to the ongoing list of discrepancies.

ahsonkhan:
FYI, we are investigating the unrelated CI failures to get the PRs unblocked. The underlying cause is due to an MSVC compiler update on CI machines. It is flagging certain coding patterns in the SDK as a warning now.

sdk/samples/iot/iot_sample_common.c
```diff
@@ -41,32 +41,22 @@ do {                                          \
...
-  }
-}
 
 //
 // MQTT endpoints
 //
 //#define USE_WEB_SOCKET // Comment to use MQTT without WebSockets.
 #ifdef USE_WEB_SOCKET
-az_span const IOT_SAMPLE_MQTT_URL_PREFIX = AZ_SPAN_LITERAL_FROM_STR("wss://");
+static az_span const mqtt_url_prefix = AZ_SPAN_LITERAL_FROM_STR("wss://");
```
ahsonkhan, line 50: Why did we rename these back?

momuno: because theses are static globals and the globals the C guidelines speak of appear to be true globals.  However, it leaves one to assume that `extern` must have been used in some other file not shown.   The C guidelines do, however, speak to static functions as opposed to public or private/internal API functions, so I am following that standard (static being the common denominator) that there does not need to be a prefix.  

I have added this question on static globals to the list of C guideline topics to address. https://gist.github.com/momuno/70fabd6609b6459cc3599f1ba037af42

PR #1131, opened 2020-08-26
Remove Client Secret Credential

### LINKED ISSUE #1091: Remove anything relate to client secret credential from core, since it is not used by anything.
We don't use `az_credential_client_secret` or the `_az_credential_client_secret_apply_policy` policy in the storage SDK at the moment.

If we ever need it for other client libraries, we can certainly bring it back, but for now, remove them from `az_core`.

https://github.com/Azure/azure-sdk-for-c/blob/fcc7218ddb86a1c4e97a8c66e876b291cbefcfcb/sdk/src/azure/core/az_credential_client_secret.c#L91
https://github.com/Azure/azure-sdk-for-c/blob/fcc7218ddb86a1c4e97a8c66e876b291cbefcfcb/sdk/src/azure/core/az_credential_client_secret.c#L19

antkmsft:
Closes #1091.


ahsonkhan:
Once ready, also add a changelog entry, especially for breaking changes.

PR #1134, opened 2020-08-26
Add ARM compile to matrix

danieljurek:


cmake-modules/gcc-arm-toolchain.cmake
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_PROCESSOR arm)
+set(CMAKE_C_COMPILER arm-none-eabi-gcc)
+set(CMAKE_CXX_COMPILER arm-none-eabi-g++)
+string(APPEND CMAKE_C_FLAGS_INIT " -mcpu=cortex-m4 -ffreestanding -nostdlib -mthumb")
```
RickWinter, line 5:
```suggestion
string(APPEND CMAKE_C_FLAGS_INIT " -mcpu=cortex-m4 -ffreestanding -nostdlib -mthumb")

```

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -6,6 +6,21 @@ jobs:
   condition: and(succeededOrFailed(), ne(variables['Skip.Test'], 'true'))
   strategy:
     matrix:
+      Linux_ARM:
+        OSVmImage: 'ubuntu-18.04'
+        vcpkg.deps: ''
+        VCPKG_DEFAULT_TRIPLET: 'x64-linux'
```
RickWinter, line 12: I would expect arm64-linux as the triplet

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -6,6 +6,21 @@ jobs:
...
+        AptDependencies: 'gcc-arm-none-eabi'
+        build.args: '-DCMAKE_TOOLCHAIN_FILE=cmake-modules/gcc-arm-toolchain.cmake'
+        AZ_SDK_C_NO_SAMPLES: 'true'
+      Linux_ARM_no_preconditions:
+        OSVmImage: 'ubuntu-18.04'
+        vcpkg.deps: ''
+        VCPKG_DEFAULT_TRIPLET: 'x64-linux'
```
RickWinter, line 19: ditto, arm64-linux

eng/pipelines/templates/jobs/archetype-sdk-client.yml
danieljurek: I've removed the triplet because it's not necessary (we aren't using vcpkg)

eng/pipelines/templates/jobs/archetype-sdk-client.yml
danieljurek: I've removed the triplet because it's not necessary (we aren't using vcpkg)

danieljurek:
@barcharcraz  -- This is the work we're doing to build the Azure SDK for C using ARM GCC. It follows the suggestions you made last week. 


@RickWinter -- PTAL

cmake-modules/gcc-arm-toolchain.cmake
ahsonkhan, line 5: Can we add some comments to what all these flags being passed in mean? Like `-mthumb`, `-ffreestanding`?

cmake-modules/gcc-arm-toolchain.cmake
```diff
@@ -0,0 +1,5 @@
+set(CMAKE_SYSTEM_NAME Generic)
```
ahsonkhan, line 1: Why `Generic`? What does that mean?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -6,6 +6,19 @@ jobs:
   condition: and(succeededOrFailed(), ne(variables['Skip.Test'], 'true'))
   strategy:
     matrix:
+      Linux_ARM:
+        OSVmImage: 'ubuntu-18.04'
+        vcpkg.deps: ''
+        AptDependencies: 'gcc-arm-none-eabi'
+        build.args: '-DCMAKE_TOOLCHAIN_FILE=cmake-modules/gcc-arm-toolchain.cmake'
+        AZ_SDK_C_NO_SAMPLES: 'true'
```
ahsonkhan, line 14: Why is `AZ_SDK_C_NO_SAMPLES` set to true?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -18,26 +31,22 @@ jobs:
       Linux_x64:
         OSVmImage: 'ubuntu-18.04'
         vcpkg.deps: ''
-        VCPKG_DEFAULT_TRIPLET: 'x64-linux'
         AZ_SDK_C_NO_SAMPLES: 'true'
       Win_x86:
         OSVmImage: 'windows-2019'
         vcpkg.deps: ''
-        VCPKG_DEFAULT_TRIPLET: 'x86-windows-static'
```
ahsonkhan, line 26: We aren't using the triplet, even on windows? What do we do instead?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -6,6 +6,19 @@ jobs:
   condition: and(succeededOrFailed(), ne(variables['Skip.Test'], 'true'))
   strategy:
     matrix:
+      Linux_ARM:
```
ahsonkhan, line 9: Can we build samples as well on ARM? I think the customer can benefit from us having samples that build fine on ARM.

Maybe even tests (but I don't think that is absolutely necessary, nice to have).

I am surprised how little time these legs take:
image

eng/pipelines/templates/jobs/archetype-sdk-client.yml
danieljurek: We don't need triplets for vcpkg when there are no vcpkg dependencies. These triplets are redundant.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
danieljurek: These legs have no vcpkg dependencies and compile without samples. If we had to build dependencies in vcpkg they would run longer. There is no support in vcpkg for building ARM on Linux in vcpkg at this time: https://github.com/microsoft/vcpkg/tree/master/triplets 

Samples would need to be built for platforms on which they are running and, in the case of ARM we are building without standard libraries (e.g. printf, etc.) so samples do not build in this state.

cmake-modules/gcc-arm-toolchain.cmake
danieljurek: @barcharcraz helped create this module and can more accurately describe what's going on here. I know that this is setting us up for cross-compilation and for a target system that is different from the host system https://cmake.org/cmake/help/v3.10/variable/CMAKE_SYSTEM_NAME.html

eng/pipelines/templates/jobs/archetype-sdk-client.yml
ahsonkhan: > Samples would need to be built for platforms on which they are running and, in the case of ARM we are building without standard libraries (e.g. printf, etc.) so samples do not build in this state.

That's exactly the type of reasoning I was looking for! That makes sense. Thanks for sharing, Daniel.

ahsonkhan:
@danieljurek - ready to merge?

Issue #1142, opened 2020-08-27
CI failure in the Install Dependencies (vcpkg) step possibly due to out-dated artifacts on the machine


ahsonkhan:
CI failure in the `Install Dependencies (vcpkg)` step from https://github.com/Azure/azure-sdk-for-c/pull/1139

https://dev.azure.com/azure-sdk/public/_build/results?buildId=510971&view=logs&j=bc7ff4a5-0370-5b6c-2e51-844007afb52d&t=0a35b456-51b9-5a0b-b96e-7c034a4efd47​​​​​​​
```text
Starting: Install Dependencies (vcpkg) 
============================================================================== 
Task : Command line 
Description : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows 
Version : 2.164.2 
Author : Microsoft Corporation 
Help : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line 
============================================================================== 
Generating script. 
Script contents: 
vcpkg install cmocka 
========================== Starting Command Output =========================== 
"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\88094ebc-7c44-4e00-af4b-813aeffa1f9f.cmd"" 
Computing installation plan... 
Found both manifest and CONTROL file in port C:\vcpkg\ports\cmocka; please rename one or the other 
##[error]Cmd.exe exited with code '1'. 
Finishing: Install Dependencies (vcpkg)
```

cc @danieljurek, @weshaggard

kurtzeborn:
Swapping the "EngSys" label for "Client" since these are used to track/indicate _who_ is investigating or making the fix and not necessarily what area the issue is in.  "EngSys" and "Client" are intended to be mutually exclusive (hence the coloring).

ahsonkhan:
With the workaround in place, this is no longer blocking the release. That said, we should definitely fix and bring back the caching mechanism since that substantially reduced our CI build times.

RickWinter:
@ahsonkhan please verify and close

ahsonkhan:
This has been fixed by the updated vcpkg caching work by @danieljurek: https://github.com/Azure/azure-sdk-for-c/pull/1217

PR #1143, opened 2020-08-27
[Unblock clean CI] Remove the CI step to restore vcpkg from the cache and remove the analyze flag for MSVC

ahsonkhan:
Testing out some stopgap solution to https://github.com/Azure/azure-sdk-for-c/issues/1142 by removing the restore from cache step.

Also, remove the analyze flag for MSVC to get clean CI to test out a stopgap which unblocks existing PRs while investigation is pending for https://github.com/Azure/azure-sdk-for-c/issues/1137

### DIFF
--- a/eng/pipelines/templates/steps/vcpkg.yml
+++ b/eng/pipelines/templates/steps/vcpkg.yml
@@ -61,24 +61,6 @@ steps:
         contains(variables['OSVmImage'], 'ubuntu')
       )
 
-  # Attempt to restore vcpkg from the cache
-  - task: Cache@2
-    inputs:
-      key: >-
-        $(Agent.JobName)
-        | "${{ parameters.VcpkgVersion }}"
-        | $(Agent.Os)
-        | $(${{ parameters.DependenciesVariableName }})
-      path: $(VCPKG_INSTALLATION_ROOT)
-      cacheHitVar: VcpkgRestoredFromCache
-    displayName: Vcpkg Cache
-    condition: >-
-      and(
-        succeeded(),
-        not(eq(variables['${{ parameters.DependenciesVariableName }}'], '')),
-        not(eq(variables['Skip.VcpkgCache'], 'true'))
-      )
-
   # Install vcpkg on MacOS
   - task: Bash@3
     inputs:
--- a/eng/pipelines/templates/steps/vcpkg.yml
+++ b/eng/pipelines/templates/steps/vcpkg.yml
@@ -111,13 +111,28 @@ steps:
         git reset --hard HEAD
         echo git pull origin master
         git pull origin master
+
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/vcpkg.yml
```diff
@@ -88,6 +70,10 @@ steps:
         git clone https://github.com/Microsoft/vcpkg.git
         cd vcpkg
         git rev-parse --verify HEAD
+
+        # Make sure to fetch all the tags to avoid failures on checkout.
+        git fetch --all --tags
```
ahsonkhan, line 75: @danieljurek Turns out we were failing to checkout the vcpkg tag causing changes from latest master to come in, causing out-of-sync state on the machine.

`git checkout tags/2020.06` was failing with:
> pathspec '…' did not match any file(s) known to git

Fixed by fetching all the tags first.

momuno:
Why did the analyze flag create the problem?  Was it newly added, or did something underneath change?

ahsonkhan:
> Why did the analyze flag create a problem? Was it newly added, or did something underneath change?

The MSVC compiler on the build machines upgraded which brought in new (possibly false positive) analysis capabilities.

See https://github.com/Azure/azure-sdk-for-c/issues/1137#issuecomment-681078133

Nothing was changed recently from the SDK code perspective that caused these new warnings.

eng/pipelines/templates/steps/vcpkg.yml
weshaggard: Yes this is an interesting problem when ever the tag isn't rooted by the default branch you are forced to also fetch the tags.

weshaggard: FYI: "--all" isn't needed for our case as it just says to fetch from all remotes which we only have one of.

ahsonkhan: Btw, is there a way for us to capture failures from running the various git commands and fail the leg? I am not  sure for how long this `git checkout` command was failing for, silently.

weshaggard: Yes but we have to look at the exit codes explicitly.  See https://github.com/Azure/azure-sdk-tools/blob/master/eng/common/scripts/git-branch-push.ps1 as examples. If you don't explicitly look at the exit codes the script will simply use the exit code for the last command you ran. That is why it is become a little tricky to have a bunch of commands that might fail in an inline script.

ahsonkhan: That's a pretty good self-documenting script with good error messages being logged. Makes it quite easy to figure out what step failed and why just by looking at the logs.

PR #1145, opened 2020-08-27
Remove the CI step to restore vcpkg from the cache to show git checkout <tag> failure.

ahsonkhan:


ahsonkhan:
Was meant for illustrative purposes only to showcase the silent failure when doing `git checkout`:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=511453&view=logs&j=5f43df56-f144-584b-8cac-2916b6897a0a&t=51726468-f971-5cff-0e43-4d276a24f866

`error: pathspec 'tags/2020.06' did not match any file(s) known to git`

PR #1158, opened 2020-08-28
Context cleanups

RickWinter:
@JeffreyRichter We previously forced every context to belong to the application.  Cancelling the application, cancelled everything.  These updates, as agreed upon in the meeting, removes that behavior.

This was our previous statement
` @details The root node in the tree (ultimate parent) is #az_context_application which is a context for the entire application. Each new node is a child of some parent.`

### DIFF
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -56,8 +56,8 @@ extern az_context az_context_application;
 /**
  * @brief Creates a new expiring #az_context node that is a child of the specified parent.
  *
- * @param[in] parent The #az_context node that the new node is to be a child of.
- * @param[in] expiration The time when this new child node should be canceled.
+ * @param[in] parent The #az_context node that is the parent to the new node.
+ * @param[in] expiration The time when this new node should be canceled.
  *
  * @return The new child #az_context node.
  */
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -47,9 +47,9 @@ struct az_context
 #define _az_CONTEXT_MAX_EXPIRATION 0x7FFFFFFFFFFFFFFF
 
 /**
- * @brief The application root #az_context instances. The #az_context_application never expires but
- * you can explicitly cancel it by passing its address to #az_context_cancel which effectively
- * cancels all the #az_context child nodes.
+ * @brief The application root #az_context instances.
+ * @details The #az_context_application never expires but you can explicitly cancel it by passing its address to 
+ * #az_context_cancel() which effectively cancels all its #az_context child nodes.
  */
 extern az_context az_context_application;
 
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -16,6 +16,7 @@
 #define _az_CONTEXT_H
 
 #include <azure/core/az_result.h>
+#include <azure/core/internal/az_precondition_internal.h>
 
 #include <stddef.h>
 #include <stdint.h>
@@ -61,8 +62,7 @@ extern az_context az_context_application;
 /**
  * @brief Creates a new expiring #az_context node that is a child of the specified parent.
  *
- * @param[in] parent The #az_context node that the new node is to be a child of; passing `NULL` sets
- * the parent to #az_context_application.
+ * @param[in] parent The #az_context node that the new node is to be a child of.
  * @param[in] expiration The time when this new child node should be canceled.
  *
  * @return The new child #az_context node.
@@ -70,16 +70,16 @@ extern az_context az_context_application;
 AZ_NODISCARD AZ_INLINE az_context
 az_context_create_with_expiration(az_context const* parent, int64_t expiration)
 {
-  return (az_context){ ._internal
-                       = { .parent = ((parent != NULL) ? parent : &az_context_application),
-                           .expiration = expiration } };
+  _az_PRECONDITION_NOT_NULL(parent);
+
+  return (az_context){ ._internal = { .parent = parent,
+                                      .expiration = expiration } };
 }
 
 /**
  * @brief Creates a new key/value az_context node that is a child of the specified parent.
  *
- * @param[in] parent __[nullable]__ The #az_context node that the new node is to be a child of;
- * passing `NULL` sets the parent to #az_context_application.
+ * @param[in] parent The #az_context node that is the parent to the new node.
  * @param[in] key A pointer to the key of this new #az_context node.
  * @param[in] value A pointer to the value of this new #az_context node.
  *
@@ -88,7 +88,8 @@ az_context_create_with_expiration(az_context const* parent, int64_t expiration)
 AZ_NODISCARD AZ_INLINE az_context
 az_context_create_with_value(az_context const* parent, void const* key, void const* value)
 {
-  return (az_context){ ._internal = { .parent = (parent != NULL) ? parent : &az_context_application,
+  _az_PRECONDITION_NOT_NULL(parent);
+  return (az_context){ ._internal = { .parent = parent,
                                       .expiration = _az_CONTEXT_MAX_EXPIRATION,
                                       .key = key,
                                       .value = value } };
@@ -97,12 +98,12 @@ az_context_create_with_value(az_context const* parent, void const* key, void con
 /**
  * @brief Cancels the specified #az_context node; this cancels all the child nodes as well.
  *
- * @param[in,out] ref_context __[nullable]__ A pointer to the #az_context node to be canceled;
- * passing `NULL` cancels the #az_context_application.
+ * @param[in,out] ref_context A pointer to the #az_context node to be canceled.
  */
 AZ_INLINE void az_context_cancel(az_context* ref_context)
 {
-  ref_context = ((ref_context != NULL) ? ref_context : &az_context_application);
+  _az_PRECONDITION_NOT_NULL(ref_context);
+
   ref_context->_internal.expiration = 0; // The beginning of time
 }
 
@@ -118,12 +119,13 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context);
  * @brief Returns `true` if this #az_context node or any of its parent nodes' expiration is before
  * the \p current_time.
  *
- * @param[in] context A pointer to the #az_context node to check; passing `NULL` checks the root
- * #az_context_application.
+ * @param[in] context A pointer to the #az_context node to check.
  * @param[in] current_time The current time.
  */
 AZ_NODISCARD AZ_INLINE bool az_context_has_expired(az_context const* context, int64_t current_time)
 {
+  _az_PRECONDITION_NOT_NULL(context);
+
   return az_context_get_expiration(context) < current_time;
 }
 
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -24,8 +24,6 @@
 
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
- *
- * @details The root node in the tree (ultimate parent).
  */
 // Definition is below. Defining the typedef first is necessary here since there is a cycle.
 typedef struct az_context az_context;
@@ -49,10 +47,9 @@ struct az_context
 #define _az_CONTEXT_MAX_EXPIRATION 0x7FFFFFFFFFFFFFFF
 
 /**
- * @brief The ultimate root of all #az_context instances. It allows you to cancel your entire
- * application. The #az_context_application never expires but you can explicitly cancel it by
- * passing its address to #az_context_cancel which effectively cancels all the #az_context child
- * nodes.
+ * @brief The application root #az_context instances. The #az_context_application never expires but
+ * you can explicitly cancel it by passing its address to #az_context_cancel which effectively
+ * cancels all the #az_context child nodes.
  */
 extern az_context az_context_application;
 
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -26,8 +26,7 @@
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
  *
- * @details The root node in the tree (ultimate parent) is #az_context_application which is a
- * context for the entire application. Each new node is a child of some parent.
+ * @details The root node in the tree (ultimate parent)..
  */
 // Definition is below. Defining the typedef first is necessary here since there is a cycle.
 typedef struct az_context az_context;
@@ -35,8 +34,7 @@ typedef struct az_context az_context;
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
  *
- * @details The root node in the tree (ultimate parent) is #az_context_application which is a
- * context for the entire application. Each new node is a child of some parent.
+ * @details The root node in the tree (ultimate parent).
  */
 struct az_context
 {
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -16,7 +16,6 @@
 #define _az_CONTEXT_H
 
 #include <azure/core/az_result.h>
-#include <azure/core/internal/az_precondition_internal.h>
 
 #include <stddef.h>
 #include <stdint.h>
@@ -65,14 +64,8 @@ extern az_context az_context_application;
  *
  * @return The new child #az_context node.
  */
-AZ_NODISCARD AZ_INLINE az_context
-az_context_create_with_expiration(az_context const* parent, int64_t expiration)
-{
-  _az_PRECONDITION_NOT_NULL(parent);
-
-  return (az_context){ ._internal = { .parent = parent,
-                                      .expiration = expiration } };
-}
+AZ_NODISCARD az_context
+az_context_create_with_expiration(az_context const* parent, int64_t expiration);
 
 /**
  * @brief Creates a new key/value az_context node that is a child of the specified parent.
@@ -83,27 +76,15 @@ az_context_create_with_expiration(az_context const* parent, int64_t expiration)
  *
  * @return The new child #az_context node.
  */
-AZ_NODISCARD AZ_INLINE az_context
-az_context_create_with_value(az_context const* parent, void const* key, void const* value)
-{
-  _az_PRECONDITION_NOT_NULL(parent);
-  return (az_context){ ._internal = { .parent = parent,
-                                      .expiration = _az_CONTEXT_MAX_EXPIRATION,
-                                      .key = key,
-                                      .value = value } };
-}
+AZ_NODISCARD az_context
+az_context_create_with_value(az_context const* parent, void const* key, void const* value);
 
 /**
  * @brief Cancels the specified #az_context node; this cancels all the child nodes as well.
  *
  * @param[in,out] ref_context A pointer to the #az_context node to be canceled.
  */
-AZ_INLINE void az_context_cancel(az_context* ref_context)
-{
-  _az_PRECONDITION_NOT_NULL(ref_context);
-
-  ref_context->_internal.expiration = 0; // The beginning of time
-}
+void az_context_cancel(az_context* ref_context);
 
 /**
  * @brief Returns the soonest expiration time of this #az_context node or any of its parent nodes.
@@ -120,12 +101,7 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context);
  * @param[in] context A pointer to the #az_context node to check.
  * @param[in] current_time The current time.
  */
-AZ_NODISCARD AZ_INLINE bool az_context_has_expired(az_context const* context, int64_t current_time)
-{
-  _az_PRECONDITION_NOT_NULL(context);
-
-  return az_context_get_expiration(context) < current_time;
-}
+AZ_NODISCARD bool az_context_has_expired(az_context const* context, int64_t current_time);
 
 /**
  * @brief Walks up this #az_context node's parents until it find a node whose key matches the
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -25,7 +25,7 @@
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
  *
- * @details The root node in the tree (ultimate parent)..
+ * @details The root node in the tree (ultimate parent).
  */
 // Definition is below. Defining the typedef first is necessary here since there is a cycle.
 typedef struct az_context az_context;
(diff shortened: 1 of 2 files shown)


sdk/inc/azure/core/az_context.h
```diff
@@ -16,6 +16,7 @@
 #define _az_CONTEXT_H
 
 #include <azure/core/az_result.h>
+#include <azure/core/internal/az_precondition_internal.h>
```
ahsonkhan, line 19: A public header must not include an internal header. Make the methods that need preconditions non-inlined and move them to the `.c` source file.

sdk/inc/azure/core/az_context.h
```diff
@@ -25,17 +26,15 @@
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
  *
- * @details The root node in the tree (ultimate parent) is #az_context_application which is a
- * context for the entire application. Each new node is a child of some parent.
+ * @details The root node in the tree (ultimate parent)..
```
ahsonkhan, line 29:
```suggestion
 * @details The root node in the tree (ultimate parent).
```

sdk/inc/azure/core/az_context.h
```diff
@@ -61,25 +60,24 @@ extern az_context az_context_application;
...
  *
  * @return The new child #az_context node.
  */
 AZ_NODISCARD AZ_INLINE az_context
 az_context_create_with_expiration(az_context const* parent, int64_t expiration)
 {
-  return (az_context){ ._internal
-                       = { .parent = ((parent != NULL) ? parent : &az_context_application),
-                           .expiration = expiration } };
+  _az_PRECONDITION_NOT_NULL(parent);
```
ahsonkhan, line 71: Add another precondition that `expiration` is `>= 0`

sdk/inc/azure/core/az_context.h
```diff
@@ -61,25 +60,24 @@ extern az_context az_context_application;
 /**
  * @brief Creates a new expiring #az_context node that is a child of the specified parent.
  *
- * @param[in] parent The #az_context node that the new node is to be a child of; passing `NULL` sets
- * the parent to #az_context_application.
+ * @param[in] parent The #az_context node that the new node is to be a child of.
  * @param[in] expiration The time when this new child node should be canceled.
```
ahsonkhan, line 64:
```suggestion
 * @param[in] expiration The time when this new child node should be canceled. The expiration must be a positive number.
```

sdk/inc/azure/core/az_context.h
ahsonkhan: On second though, should we allow a creation of an already expired context, or should the precondition be `expiration > 0`?

sdk/inc/azure/core/az_context.h
```diff
@@ -118,12 +117,13 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context);
...
  * the \p current_time.
  *
- * @param[in] context A pointer to the #az_context node to check; passing `NULL` checks the root
- * #az_context_application.
+ * @param[in] context A pointer to the #az_context node to check.
  * @param[in] current_time The current time.
  */
 AZ_NODISCARD AZ_INLINE bool az_context_has_expired(az_context const* context, int64_t current_time)
 {
+  _az_PRECONDITION_NOT_NULL(context);
```
ahsonkhan, line 125: Similarly here, add a precondition that `current_time >= 0`

sdk/src/azure/core/az_context.c
```diff
@@ -33,6 +35,8 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context)
 AZ_NODISCARD az_result
 az_context_get_value(az_context const* context, void const* key, void const** out_value)
 {
+  _az_PRECONDITION_NOT_NULL(context);
```
ahsonkhan, line 38: Key can't be null either. Add that precondition.

Also, add a runtime check, if key is null, return `AZ_ERROR_ITEM_NOT_FOUND`, for when preconditions are off.

Also, tests :)

ahsonkhan:
> We previously forced every context to belong to the application. Cancelling the application, cancelled everything. These updates, as agreed upon in the meeting, removes that behavior.

How so? This PR (at least currently) is disallowing multiple roots, and we only have just one `az_context_application`. That is the only context with a null parent, and all other contexts must be created as children of it. There is no other way to create another context.

If we ever want to allow multiple roots in the  future, we can allow passing in `NULL` as a parent to the `create` methods, and  that's when we would get the behavior you are concerned with.

sdk/inc/azure/core/az_context.h
```diff
@@ -25,17 +26,15 @@
 /**
  * @brief A context is a node within a tree that represents expiration times and key/value pairs.
  *
- * @details The root node in the tree (ultimate parent) is #az_context_application which is a
```
ahsonkhan, line 28: This comment still holds. I would keep it and just remove `Each new node is a child of some parent.`

sdk/inc/azure/core/az_context.h
```diff
@@ -88,7 +86,8 @@ az_context_create_with_expiration(az_context const* parent, int64_t expiration)
 AZ_NODISCARD AZ_INLINE az_context
 az_context_create_with_value(az_context const* parent, void const* key, void const* value)
 {
-  return (az_context){ ._internal = { .parent = (parent != NULL) ? parent : &az_context_application,
+  _az_PRECONDITION_NOT_NULL(parent);
```
ahsonkhan, line 89: Null key isn't allowed.
```suggestion
  _az_PRECONDITION_NOT_NULL(parent);
  _az_PRECONDITION_NOT_NULL(key);
```

sdk/inc/azure/core/az_context.h
RickWinter: That seems an arbitrary requirement.  We should move the preconditions so they can be used in headers

sdk/src/azure/core/az_context.c
ahsonkhan, line 38: Also, out_value pointer can't be null since we dereference it to return the found value to the caller. Please add a test.
```C
az_PRECONDITION_NOT_NULL(*out_value);
```

sdk/inc/azure/core/az_context.h
RickWinter: Setting it to 0 is the equivallent of cancelling it.  I see no reason to not allow 0, its really no different than 1.  They'll both be expired before they could ever be used.  I'd say the same for values less than zero, though there is an argument that could be made that these values should be positive.

sdk/src/azure/core/az_context.c
```diff
@@ -33,6 +35,8 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context)
...
+
```
ahsonkhan, line 39:
```suggestion
if (key == NULL)
{
  *out_value = NULL;
  return AZ_ERROR_ITEM_NOT_FOUND;
}

```

sdk/src/azure/core/az_context.c
```diff
@@ -33,6 +35,8 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context)
 AZ_NODISCARD az_result
```
ahsonkhan, line 35: nit: Remove the duplicated comment above, in source. This comment belongs in the header file (and already exists).

It might be helpful to the caller to know that the search is happening by comparing the pointer references, and not by their value. Consider adding that `@remark` to the doc comments and add that to tests.

```C
void const* const key = "k";
void const* const value = "v";
az_context ctx1 = az_context_create_with_expiration(&az_context_application, 100);
az_context ctx2 = az_context_create_with_value(&ctx1, key, value);
az_context ctx3 = az_context_create_with_expiration(&ctx2, 250);

int64_t expiration = az_context_get_expiration(&ctx3);
void const* value2 = NULL;

// This returns AZ_ERROR_ITEM_NOT_FOUND, not AZ_OK, which is the intended behavior.
void const* const a_different_key = "k";
az_result r = az_context_get_value(&ctx3, a_different_key, &value2);

assert_true(r == AZ_OK); // Fails
assert_true(value2 == NULL);
```

sdk/inc/azure/core/az_context.h
ahsonkhan: > That seems an arbitrary requirement.

I don't follow. We don't want to make internal APIs/types accidentally visible to the caller.
https://azure.github.io/azure-sdk/clang_implementation.html#coding-style
⛔️ DO NOT include internal or private headers in public headers.

> We should move the preconditions so they can be used in headers

All these precondition macros are internal. Are you suggesting we make them public?

ahsonkhan:
Another design question, @JeffreyRichter, should `az_context_get_value(...)` include expired contexts as part of its search?

I think when looking for a value using a key of an expired context, we should return `AZ_ERROR_ITEM_NOT_FOUND`, or return the value with an error, `AZ_CONTEXT_EXPIRED`. This would mean that `az_context_get_value(...)` needs a `int64_t current_time` parameter.

sdk/inc/azure/core/az_context.h
```diff
@@ -97,12 +96,12 @@ az_context_create_with_value(az_context const* parent, void const* key, void con
...
  * @brief Cancels the specified #az_context node; this cancels all the child nodes as well.
  *
- * @param[in,out] ref_context __[nullable]__ A pointer to the #az_context node to be canceled;
- * passing `NULL` cancels the #az_context_application.
+ * @param[in,out] ref_context A pointer to the #az_context node to be canceled.
  */
 AZ_INLINE void az_context_cancel(az_context* ref_context)
 {
-  ref_context = ((ref_context != NULL) ? ref_context : &az_context_application);
+  _az_PRECONDITION_NOT_NULL(ref_context);
```
ahsonkhan, line 103: Maybe we should add a precondition that this  context hasn't already been cancelled.

```suggestion
  _az_PRECONDITION_NOT_NULL(ref_context);
  _az_PRECONDITION(ref_context->_internal.expiration != 0);
```
The caller should not try to cancel the same context multiple times, as that hints to a bug in their code.

sdk/inc/azure/core/az_context.h
RickWinter: Preventing the types from being used in the headers means we can't inline the functions.  Also these are _az functions  which by definition are internal.  They might be exposed to the user but they are documented as such.

RickWinter:
To support the precondition behavior we can no longer inline functions for the context.  

cc @JeffreyRichter

sdk/inc/azure/core/az_context.h
ahsonkhan: There's a few things to unpack here.

> Preventing the types from being used in the headers means we can't inline the functions

1) If it is an important perf-sensitive code path, especially used within tight loops, and wouldn't otherwise be inlined by the compiler automatically, without the hint, it is possible to inline the function by following what we do in `az_span_create`. Feel free to follow that approach.
https://github.com/Azure/azure-sdk-for-c/blob/9137165c0309e379aa4a3ecdb032d99e38380df9/sdk/inc/azure/core/az_span.h#L67-L76
https://github.com/Azure/azure-sdk-for-c/blob/9137165c0309e379aa4a3ecdb032d99e38380df9/sdk/src/azure/core/az_span.c#L23-L36

> Also these are _az functions which by definition are internal. They might be exposed to the user but they are documented as such.

2) We do not expose any such internal methods to the public via public header files. Any non-public `_az` function is either static/within the source .c file, OR in an internal/private header file if they are used across various files and component. Those header files are not exposed from any of the public header files either and are not used in samples. Why would we want to make an exception to that rule here?

If a user includes `az_span.h` and then sees/gets access to some internal method that started with `_az`, which they accidentally call (without realizing it was internal), it makes changing or breaking those methods, which are implementation detail, more difficult. Now, if the user really wanted to, they could explicitly include some internal header file but that is a more explicit, intentional action.

Also, although we should fix this, keep in mind, that not all of our internal and private methods start with `_az`. Without some static linting tool/rules, as part of code maintainability and code reviews, it is a lot easier to see what includes are in a header file and catch bugs than making sure no internal API is accidentally exposed.
https://github.com/Azure/azure-sdk-for-c/blob/9137165c0309e379aa4a3ecdb032d99e38380df9/sdk/inc/azure/core/internal/az_http_internal.h#L98
https://github.com/Azure/azure-sdk-for-c/blob/9137165c0309e379aa4a3ecdb032d99e38380df9/sdk/inc/azure/core/internal/az_http_internal.h#L180

3) In this particular instance, I don't think it's necessary to mark these methods inlined anyway. They are not perf-sensitive, won't be called often (certainly not in loops, and they might evolve overtime to become more complicated, making inlining not profitable. Generally, we should only inline APIs that we know are relatively small and customers would benefit from them being inlined. Keep in mind, inlining is just a hint to the compiler, and it is free to ignore it anyway. Even if the precondition macros were public, I would recommend not inlining. To get the precondition benefit, we had some APIs that are generally more perf sensitive and called in potentially hot code paths, that we still didn't inline. For example, the JSON init methods, because they are still only called once and not in tight loops, and bulk of the work/time is spent in the actual processing.
https://github.com/Azure/azure-sdk-for-c/blob/9137165c0309e379aa4a3ecdb032d99e38380df9/sdk/src/azure/core/az_json_reader.c#L15-L20

**All that said, if you feel strongly about this, please bring this up with the rest of the team, lets discuss this, and change the guidelines, or make the precondition macros public.**

RickWinter: I did look at what span did b/c I couldn't believe I was the first to hit this.  I'm against the span solution because of the code duplication.  There is a risk the fix in the C file may not make it into the .h inline version or vice-versa.  Is there performance data that shows the duplicated code is worth the overhead of maintaining two functions?

RickWinter:
> > We previously forced every context to belong to the application. Cancelling the application, cancelled everything. These updates, as agreed upon in the meeting, removes that behavior.
> 
> How does it remove that behavior? This PR (at least currently) is disallowing multiple roots, and we only have just one `az_context_application`. That is the only context with a null parent, and all other contexts must be created as children of it. There is no other way to create another context. We are still forcing every context to belong to the application, which means cancelling the application, cancels everything.
> 
> If we ever want to allow multiple roots in the future, we can allow passing in `NULL` as a parent to the `create` methods, and that's when we would get the behavior you are concerned with.

Keep in mind that the previous design didn't ensure everything was rooted to the application.  The statement was never true, but it was the statement the original author made.  I want to be clear that we are dropping that statement and not fixing it.  

> There is no other way to create another context.
```
az_context my_root_context = { 0 };
```

Here is the code to create a context that is not rooted as the application
``` 
   az_context my_root_context = { 0 };
   az_context context_with_null_parent = az_context_create_with_expiration(&my_root_context, 12345);
   az_context context_with_different_null_parent = az_context_create_with_expiration(&az_context_application, 12345);
```
If you look at the code you'll see that the parents are different, yet both are roots.

ahsonkhan:
> Here is the code to create a context that is not rooted as the application

Lol, not sure if that was sarcastic, but you got me there 😊

The customer can also do this too, doesn't mean it is what we expect or support them doing for defined behavior.
```C
az_context i_am_the_application_now = { ._internal.parent = NULL, ._internal.expiration = 10 };
az_context_application._internal.expiration = -1;
az_context_application._internal.parent = &i_am_the_application_now;
```

sdk/inc/azure/core/az_context.h
```diff
@@ -61,50 +59,32 @@ extern az_context az_context_application;
...
-}
+AZ_NODISCARD az_context
+az_context_create_with_expiration(az_context const* parent, int64_t expiration);
 
 /**
  * @brief Creates a new key/value az_context node that is a child of the specified parent.
  *
- * @param[in] parent __[nullable]__ The #az_context node that the new node is to be a child of;
- * passing `NULL` sets the parent to #az_context_application.
+ * @param[in] parent The #az_context node that is the parent to the new node.
```
ahsonkhan, line 73: The docs for the two create method should use similar wording.

> * @param[in] parent The #az_context node that is the parent to the new node.

>  * @param[in] parent The #az_context node that the new node is to be a child of.

sdk/src/azure/core/az_context.c
```diff
@@ -44,3 +49,36 @@ az_context_get_value(az_context const* context, void const* key, void const** ou
...
+  ref_context->_internal.expiration = 0; // The beginning of time
+}
+
+AZ_NODISCARD bool az_context_has_expired(az_context const* context, int64_t current_time)
+{
+  _az_PRECONDITION_NOT_NULL(context);
+
+  return az_context_get_expiration(context) < current_time;
+}
+
```
ahsonkhan, line 84: nit: remove extra line

JeffreyRichter:
> Another design question, @JeffreyRichter, should `az_context_get_value(...)` include expired contexts as part of its search?
> 
> I think when looking for a value using a key of an expired context, **maybe** we should return `AZ_ERROR_ITEM_NOT_FOUND`, or return the value with an error, `AZ_CONTEXT_EXPIRED`. This would mean that `az_context_get_value(...)` needs a `int64_t current_time` parameter.

Yes

sdk/src/azure/core/az_context.c
RickWinter: If its a precondition it shouldn't be a runtime check.  The precondition is the equivalent of an `if(NULL)` check.

sdk/inc/azure/core/az_context.h
RickWinter: The comment doesn't make sense on the type.  How is every az_context the root node in the tree?

sdk/inc/azure/core/az_context.h
RickWinter: Fixed

sdk/src/azure/core/az_context.c
RickWinter: Agreed.  There is a much more extensive C++ suite

sdk/src/azure/core/az_context.c
RickWinter: Key being a precondition should not all be a runtime check.  The precondition implies the application should never ever do this.  Where as the runtime check says its ok for the value to be something and we will take another path in that case.  We want the former behavior that the preconditions provide.

RickWinter:
> Another design question, @JeffreyRichter, should `az_context_get_value(...)` include expired contexts as part of its search?
> 
> I think when looking for a value using a key of an expired context, **maybe** we should return `AZ_ERROR_ITEM_NOT_FOUND`, or return the value with an error, `AZ_CONTEXT_EXPIRED`. This would mean that `az_context_get_value(...)` needs a `int64_t current_time` parameter.

The fact that expiration and key/value is on the context with the expiration is an implementation detail.  The are unrelated.  We merely keep them together as a convenient way to smuggle the data on the context.

sdk/inc/azure/core/az_context.h
RickWinter: Canceling a context multiple times is valid

RickWinter:
@ahsonkhan I've addressed your feedback please remove your block so this PR can be commited.

sdk/inc/azure/core/az_context.h
ahsonkhan: The comment isn't talking about any/every context, just specifically `#az_context_application`

It's letting the caller know that if you want the ultimate root az_context, go grab that one.

sdk/inc/azure/core/az_context.h
```diff
@@ -51,60 +47,41 @@ struct az_context
 #define _az_CONTEXT_MAX_EXPIRATION 0x7FFFFFFFFFFFFFFF
 
 /**
- * @brief The ultimate root of all #az_context instances. It allows you to cancel your entire
- * application. The #az_context_application never expires but you can explicitly cancel it by
- * passing its address to #az_context_cancel which effectively cancels all the #az_context child
- * nodes.
+ * @brief The application root #az_context instances. The #az_context_application never expires but
+ * you can explicitly cancel it by passing its address to #az_context_cancel which effectively
+ * cancels all the #az_context child nodes.
```
ahsonkhan, line 52: nit: Move non-brief details to details. Add `()` for the method, and the` -> `its`

```suggestion
 * @brief The application root #az_context instances.
 * @details The #az_context_application never expires but you can explicitly cancel it by passing its address to 
 * #az_context_cancel() which effectively cancels all its #az_context child nodes.
```

sdk/src/azure/core/az_context.c
```diff
@@ -44,3 +50,39 @@ az_context_get_value(az_context const* context, void const* key, void const** ou
...
+void az_context_cancel(az_context* ref_context)
+{
+  _az_PRECONDITION_NOT_NULL(ref_context);
+
+  ref_context->_internal.expiration = 0; // The beginning of time
+}
+
+AZ_NODISCARD bool az_context_has_expired(az_context const* context, int64_t current_time)
+{
+  _az_PRECONDITION_NOT_NULL(context);
```
ahsonkhan, line 84:
```suggestion
  _az_PRECONDITION_NOT_NULL(context);
  _az_PRECONDITION(current_time >= 0);
```

sdk/src/azure/core/az_context.c
```diff
@@ -33,6 +36,9 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context)
 AZ_NODISCARD az_result
 az_context_get_value(az_context const* context, void const* key, void const** out_value)
 {
+  _az_PRECONDITION_NOT_NULL(context);
+  _az_PRECONDITION_NOT_NULL(key);
```
ahsonkhan, line 40:
```suggestion
  _az_PRECONDITION_NOT_NULL(key);
  _az_PRECONDITION_NOT_NULL(*out_value);
```

sdk/src/azure/core/az_context.c
ahsonkhan: This wasn't fixed.

sdk/src/azure/core/az_context.c
```diff
@@ -33,6 +36,9 @@ AZ_NODISCARD int64_t az_context_get_expiration(az_context const* context)
 AZ_NODISCARD az_result
 az_context_get_value(az_context const* context, void const* key, void const** out_value)
 {
+  _az_PRECONDITION_NOT_NULL(context);
```
RickWinter, line 39:
```suggestion
  _az_PRECONDITION_NOT_NULL(context);
  _az_PRECONDITION_NOT_NULL(out_value);
  _az_PRECONDITION_NOT_NULL(*out_value);
```

sdk/src/azure/core/az_context.c
RickWinter: Added along with a check that out_value itself is not null

sdk/src/azure/core/az_context.c
```diff
@@ -38,7 +38,6 @@ az_context_get_value(az_context const* context, void const* key, void const** ou
 {
   _az_PRECONDITION_NOT_NULL(context);
   _az_PRECONDITION_NOT_NULL(out_value);
-  _az_PRECONDITION_NOT_NULL(*out_value);
```
ahsonkhan: Good catch.

PR #1160, opened 2020-08-28
Make AZ_RETURN_IF... internal.

### LINKED ISSUE #1103: Make AZ_RETURN_IF_FAILED and AZ_RETURN_IF_NOT_ENOUGH_SIZE macros internal
Though these macros are quite useful and used frequently within the SDK, they are currently not intended for consumers of the SDK to use.

Rename them using `_` to indicate they are internal only:
`AZ_RETURN_IF_FAILED` -> `_az_RETURN_IF_FAILED`
`AZ_RETURN_IF_NOT_ENOUGH_SIZE` -> `_az_RETURN_IF_NOT_ENOUGH_SIZE `

https://github.com/Azure/azure-sdk-for-c/blob/9ce11465d167e512e1abdaa912bc0e5c5dec3346/sdk/inc/azure/core/az_result.h#L44-L61

antkmsft:
Closes #1103.

### DIFF
--- a/sdk/samples/iot/iot_sample_common.c
+++ b/sdk/samples/iot/iot_sample_common.c
@@ -82,7 +82,7 @@ static az_result read_configuration_entry(
     (void)printf("%s = %s\n", env_name, hide_value ? "***" : env_value);
     az_span env_span = az_span_create_from_str(env_value);
 
-    AZ_RETURN_IF_NOT_ENOUGH_SIZE(destination, az_span_size(env_span));
+    IOT_SAMPLE_RETURN_IF_NOT_ENOUGH_SIZE(destination, az_span_size(env_span));
     az_span_copy(destination, env_span);
     *out_env_value = az_span_slice(destination, 0, az_span_size(env_span));
   }
@@ -105,7 +105,7 @@ az_result iot_sample_read_environment_variables(
   if (type == PAHO_IOT_HUB)
   {
     out_env_vars->hub_hostname = AZ_SPAN_FROM_BUFFER(iot_sample_hub_hostname_buffer);
-    AZ_RETURN_IF_FAILED(read_configuration_entry(
+    IOT_SAMPLE_RETURN_IF_FAILED(read_configuration_entry(
         IOT_SAMPLE_ENV_HUB_HOSTNAME,
         NULL,
         false,
@@ -121,7 +121,7 @@ az_result iot_sample_read_environment_variables(
       case PAHO_IOT_HUB_TELEMETRY_SAMPLE:
       case PAHO_IOT_HUB_TWIN_SAMPLE:
         out_env_vars->hub_device_id = AZ_SPAN_FROM_BUFFER(iot_sample_hub_device_id_buffer);
-        AZ_RETURN_IF_FAILED(read_configuration_entry(
+        IOT_SAMPLE_RETURN_IF_FAILED(read_configuration_entry(
             IOT_SAMPLE_ENV_HUB_DEVICE_ID,
             NULL,
             false,
@@ -130,7 +130,7 @@ az_result iot_sample_read_environment_variables(
 
         out_env_vars->x509_cert_pem_file_path
             = AZ_SPAN_FROM_BUFFER(iot_sample_x509_cert_pem_file_path_buffer);
-        AZ_RETURN_IF_FAILED(read_configuration_entry(
+        IOT_SAMPLE_RETURN_IF_FAILED(read_configuration_entry(
             IOT_SAMPLE_ENV_DEVICE_X509_CERT_PEM_FILE_PATH,
             NULL,
             false,
@@ -140,7 +140,7 @@ az_result iot_sample_read_environment_variables(
 
       case PAHO_IOT_HUB_SAS_TELEMETRY_SAMPLE:
         out_env_vars->hub_device_id = AZ_SPAN_FROM_BUFFER(iot_sample_hub_device_id_buffer);
-        AZ_RETURN_IF_FAILED(read_configuration_entry(
+        IOT_SAMPLE_RETURN_IF_FAILED(read_configuration_entry(
             IOT_SAMPLE_ENV_HUB_SAS_DEVICE_ID,
             NULL,
             false,
@@ -148,7 +148,7 @@ az_result iot_sample_read_environment_variables(
             &(out_env_vars->hub_device_id)));
 
         out_env_vars->hub_sas_key = AZ_SPAN_FROM_BUFFER(iot_sample_hub_sas_key_buffer);
-        AZ_RETURN_IF_FAILED(read_configuration_entry(
+        IOT_SAMPLE_RETURN_IF_FAILED(read_configuration_entry(
             IOT_SAMPLE_ENV_HUB_SAS_KEY,
             NULL,
             true,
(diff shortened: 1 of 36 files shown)


sdk/samples/iot/iot_sample_common.c
```diff
@@ -82,7 +82,7 @@ static az_result read_configuration_entry(
     (void)printf("%s = %s\n", env_name, hide_value ? "***" : env_value);
     az_span env_span = az_span_create_from_str(env_value);
 
-    AZ_RETURN_IF_NOT_ENOUGH_SIZE(destination, az_span_size(env_span));
+    IOT_SAMPLE_RETURN_IF_NOT_ENOUGH_SIZE(destination, az_span_size(env_span));
```
RickWinter, line 85: Make the names friendlier.  Remove the `IOT_SAMPLE` portion of the macros `RETURN_IF_NOT_ENOUGH_SIZE`.  If customers copy the code and use it they may also want the MACROS in the sample.

sdk/samples/iot/iot_sample_common.h
```diff
@@ -18,42 +18,65 @@
 //
 // Logging
 //
-#define IOT_SAMPLE_LOG_ERROR(...)                                                  \
-  do                                                                               \
-  {                                                                                \
+#define IOT_SAMPLE_LOG_ERROR(...) \
```
RickWinter, line 21: See my previous comment on the naming of these macros

sdk/src/azure/platform/az_curl.c
```diff
@@ -58,7 +59,8 @@ static AZ_NODISCARD az_result _az_http_client_curl_code_to_result(CURLcode code)
 }
 
 // returning AZ error on CURL Error
-#define AZ_RETURN_IF_CURL_FAILED(exp) AZ_RETURN_IF_FAILED(_az_http_client_curl_code_to_result(exp))
+#define _az_RETURN_IF_CURL_FAILED(exp) \
```
ahsonkhan, line 62: Question: Should this be in a do/while loop?

sdk/samples/iot/iot_sample_common.h
```diff
@@ -137,7 +160,8 @@ typedef enum
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK All required environment variables successfully read-in.
  * @retval #AZ_ERROR_ARG Sample type or name is undefined, or environment variable is not set.
- * @retval #AZ_RETURN_IF_NOT_ENOUGH_SIZE Not enough space set aside to store environment variable.
+ * @retval #AZ_ERROR_INSUFFICIENT_SPAN_SIZE Not enough space set aside to store environment
```
ahsonkhan, line 163: Good catch!

sdk/samples/iot/sample_pnp_device_info_component.c
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
+#include "iot_sample_common.h"
```
ahsonkhan, line 4: Does the ordering of this include matter compared to the others? That is, does it have to be first? Is that why you have a separate group for it?

sdk/samples/iot/sample_pnp_thermostat_component.h
```diff
@@ -9,6 +9,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include <azure/core/az_json.h>
```
ahsonkhan, line 12: Lol, why does this keep getting added? I am fine with it, but just curious :)

sdk/samples/iot/iot_sample_common.c
antkmsft: @momuno - you came up with this naming prefix recently. Should I keep `IOT_SAMPLE`, or make it something else?

sdk/samples/iot/iot_sample_common.h
antkmsft: @momuno - you came up with this naming prefix recently. Should I keep `IOT_SAMPLE`, or make it something else?

sdk/src/azure/platform/az_curl.c
antkmsft: No. It is the multi-statement macros "returning void" that should be. `AZ_RETURN_IF_FAILED()` would wrap it in do-while.

sdk/samples/iot/sample_pnp_device_info_component.c
antkmsft: I've seen in many other samples, this include file going first as a group, so this is matching. Plus, in my case, there seems to be a bug in VS formatter - it puts this file non-alphabetically when inside a group (between `s` and `s` - impossible).

sdk/samples/iot/sample_pnp_thermostat_component.h
antkmsft: It's really, the compilation fails without it. I don't know why id doesn't always. Include file rearrangement has something to do with it? Include file rearrangement should not normally break compilation, as the inclusion should take care of including everything it needs. And if you look in this file, it uses json types, but nowhere in the chain of include in this file does it include json.h. So it must be that the files using this header just always happened to have json.h included before. So, this is an improvement.

sdk/samples/iot/iot_sample_common.c
antkmsft: I contacted Mollie, we will likely follow up outside of this PR.

sdk/samples/iot/iot_sample_common.h
antkmsft: I contacted Mollie, we will likely follow up outside of this PR.

sdk/samples/iot/sample_pnp_thermostat_component.h
ahsonkhan: Agreed.

And I think you are right. The include reordering is probably breaking the build in some cases.

sdk/samples/iot/iot_sample_common.c
RickWinter: I suggest the naming as when code is copied it makes sense for the user w/o needing to rename/replace the MACRO name. 
 IoT can make the final decision on that.

momuno: I chose to go with IOT_SAMPLE on macros to follow the iot_sample prefix used on everything else in the iot_sample_commong.c and .h files.  There are many places that AZ_ prefixed functions are called in the sample, such as with json parsing.  Why does this MACRO name need to change?

sdk/samples/iot/iot_sample_common.h
momuno: See my earlier note.  These are used across samples based on the iot_sample_common files and meant to be distinguished from static functions belonging to each sample.  Additionally, it is meant to show that these are custom functions found in io_sample_common and not part of the AZ API.

sdk/samples/iot/sample_pnp_thermostat_component.h
momuno: I will look at the re-ording when I get to this sample for finalizing.  Thanks for finding this!

sdk/samples/iot/iot_sample_common.c
ahsonkhan: > Why does this MACRO name need to change?

We made the `AZ_RETURN_IF_NOT_ENOUGH_SIZE` macro internal. So, similar to `IOT_SAMPLE_PRECONDITION_NOT_NULL`, the samples need to define their own macro.

PR #1162, opened 2020-08-28
Macro cleanup

antkmsft:


sdk/src/azure/core/az_json_reader.c
```diff
@@ -700,6 +698,8 @@ AZ_NODISCARD static az_result _az_json_reader_process_number(az_json_reader* ref
   return AZ_OK;
 }
 
+AZ_INLINE int32_t _az_min(int32_t a, int32_t b) { return a < b ? a : b; }
```
ahsonkhan, line 701: When do we decide to go with one over the other?

sdk/tests/core/test_az_json.c
```diff
@@ -1405,187 +1405,187 @@ az_result write_str(az_span span, az_span s, az_span* out, int32_t* written)
...
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\":  "), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\": 123"), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\": 123  "), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\": 123,"), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\": 123,  "), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\": 123 ,  "), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\":\"value}"), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(AZ_SPAN_FROM_STR("{\"name\":\"value\""), AZ_ERROR_UNEXPECTED_END);
-  test_json_reader_invalid_helper(
+  TEST_JSON_READER_INVALID_HELPER(AZ_SPAN_FROM_STR("{"), AZ_ERROR_UNEXPECTED_END);
```
ahsonkhan, line 1552: Nice renames.

.clang-format
```diff
@@ -7,7 +7,7 @@ IncludeBlocks: Preserve
 NamespaceIndentation: None
 PointerAlignment: Left
 AlignAfterOpenBracket: AlwaysBreak
-AlignEscapedNewlines: DontAlign
+AlignEscapedNewlines: Left
```
ahsonkhan, line 10: There you go @momuno. Fixed :)

sdk/inc/azure/core/az_span.h
```diff
@@ -80,9 +80,9 @@ AZ_NODISCARD az_span az_span_create(uint8_t* ptr, int32_t size);
  */
 // When updating this macro, also update AZ_PAIR_NULL, which should be using this macro, but can't
 // due to warnings, and so it is using an expansion of this macro instead.
-#define AZ_SPAN_NULL \
-  (az_span) \
-  { \
+#define AZ_SPAN_NULL                       \
```
ahsonkhan, line 83: Did you run this across all the files in the repo?

sdk/samples/iot/iot_sample_common.h
```diff
@@ -44,16 +44,10 @@
...
-    char* buffer = (char*)az_span_ptr(span);                                 \
-    for (int32_t az_span_i = 0; az_span_i < az_span_size(span); az_span_i++) \
-    {                                                                        \
-      putchar(*buffer++);                                                    \
-    }                                                                        \
-    (void)printf("\n");                                                      \
+#define IOT_SAMPLE_LOG_AZ_SPAN(span_description, span)                                        \
+  do                                                                                          \
+  {                                                                                           \
+    (void)printf("\t\t%s %.*s\n", (span_description), az_span_size(span), az_span_ptr(span)); \
```
ahsonkhan, line 50: @momuno - is this an acceptable change? I believe the output will be identical. Please verify.

sdk/src/azure/core/az_json_reader.c
momuno: do you mean an inline function vs a macro?  I have the same question.

sdk/samples/iot/iot_sample_common.h
momuno: In the guidelines, I see:  https://azure.github.io/azure-sdk/clang_design.html#clang-design-naming-macros3

` do { `

on the same line.  Which do we want to go with?

.clang-format
momuno: yay!!

sdk/samples/iot/iot_sample_common.h
momuno: Oh, I see the bigger change now.
I recall asking the same question to @hihigupt , and there is a reason for using putchar instead of printf.  I believe it has to do with printing non ASCII characters.  Since we don't know what could come in the AZ_SPAN, we don't know if it is all ASCII.  I understand there to be unique scenarios where printf does not print non-ASCII  characters correctly.  @hihigupt can you please clarify?  Thanks.

ahsonkhan: Yes, I remember now:
https://github.com/Azure/azure-sdk-for-c/pull/911#discussion_r452443598

If the `span` contained a `0` byte, then this `printf` won't print all the bytes and stop there.
The `putchar` in a loop approach makes sure all the bytes of the span are printed.

momuno: That's it.  Thanks for remembering!  So, no please do not make that change :)

sdk/src/azure/core/az_json_reader.c
antkmsft: We should avoid macros, unless it is absolutely necessary, so it is impossible what we are trying to do using the normal functionality provided by the language. If it is not stated in the guidelines, we should put it there. Macros are harder to debug (try to step in), they would calculate argument twice (for instance, if `a` is a function call, it will be called twice when the macro decides to return `a` as result).

sdk/inc/azure/core/az_span.h
antkmsft: All that match `#define.*\\` regex :)

sdk/samples/iot/iot_sample_common.h
antkmsft: Oh, I see. I replaced it with `fwrite()` that doesn't stop at 0 terminator.

antkmsft: @momuno , @hihigupt , I replaced with `fwrite()` which doesn't stop when it reads 0, looks good?

.clang-format
RickWinter: Nit, the unaligned line continuations are less likely to require future updates.  As it is now the next time you change a line that increases or decreases the maximum width you will need to "unnecessarily" touch each line in the sequence to adjust.

antkmsft: @RickWinter, your point is valid, let me know if you think we need to vote. Either way is fine with me.
I did this because our guidelines actually say we should use aligned lines. And I've seen there's a discrepancy between us and the IoT team.
The specific point, about single line update, I do not see as a big problem, because if you change the single line of a macro - you update the entire macro, so I don't see it as a big problem for git history, especially given that normally there aren't that many macros are most of them are up to 5 lines long. On the other hand, single line history is even better, plus it is easier to format by hand, if it comes to that.

sdk/src/azure/core/az_json_reader.c
antkmsft: An advantage of a macro would be if it was used heavily, and used with different types - longs, ints, doubles, etc. C doesn't have templates, so it would not be possible to write a function that accepts all types universally while having the same name (assuming that would be a benefit to outweigh all the cons of macros). But here, macro is only used in a single place.

.clang-format
antkmsft: BTW, all the credit for driving this change and finding this setting goes to @ahsonkhan  - I didn't know this setting existed.

ahsonkhan: @RickWinter - if that becomes too much of a concern, we can use `Right` instead of `Left` for the value of `AlignEscapedNewlines` which moves all the `\` all the way to the right hand side (at around column 100), but still keeping them aligned.

I prefer `Left` though since the macros have relatively small line width, but either work.

sdk/samples/iot/iot_sample_common.h
```diff
@@ -18,65 +18,61 @@
...
-    for (int32_t az_span_i = 0; az_span_i < az_span_size(span); az_span_i++) \
-    { \
-      putchar(*buffer++); \
-    } \
-    (void)printf("\n"); \
+#define IOT_SAMPLE_LOG_AZ_SPAN(span_description, span)                                           \
+  do                                                                                             \
+  {                                                                                              \
+    (void)printf("\t\t%s ", span_description);                                                   \
+    (void)fwrite((char*)az_span_ptr(span), sizeof(uint8_t), (size_t)az_span_size(span), stdout); \
```
ahsonkhan, line 51: Seems reasonable. The only nit I have is that some IoT devices don't have a filesystem, so using `fwrite` in samples (which is primarily for files, even though it is writing to stdout), might give the wrong impression to customers.

sdk/samples/iot/iot_sample_common.h
ahsonkhan: Seems reasonable. The only nit I have is that some IoT devices don't have a filesystem, so using `fwrite` in samples (which is primarily for files, even though it is writing to stdout), might give the wrong impression to customers.

sdk/src/azure/core/az_json_reader.c
ahsonkhan: Maybe make the method static?

antkmsft: `AZ_INLINE` is already static - it is define as `static inline`. We absolutely should've used static if it wasn't already.

antkmsft: Ha! Guidelines even have `MAX()` macro as example, I didn't know/remember that :)

.clang-format
antkmsft: +1 for Left vs Right alignment for macros (and neutral on DontAlign).

sdk/samples/iot/iot_sample_common.h
antkmsft: It is hard for to tell whether it will be missing, or just support `stdout` when it is, but I won't be surprised it `fwrite(..., stdout)` does work there, and then we are doing it effectively, by not outputting bytes one-by-one. IoT team, I leave this to you - do you have info whether IoT compilers support fwrite to stdout?

ahsonkhan: Let's unblock and merge this PR in. If necessary, folks can follow up and revert.

momuno: Hi.  I am adding @hihigupt to this.

sdk/samples/iot/iot_sample_common.h
hihigupt, line 51: >fwrite [](start = 10, length = 6)

write?
[https://linux.die.net/man/2/write](https://linux.die.net/man/2/write)
on fd =0

hihigupt: ignore
having fwrite we will only include stdio.h

---
In reply to: [480293541](https://github.com/Azure/azure-sdk-for-c/pull/1162#discussion_r480293541) [](ancestors = 480293541)

PR #1163, opened 2020-08-28
Move pnp sample code into own folder

danewalton:
As per @ericwol-msft wishes. The main folder was getting pretty clogged.


sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include "iot_sample_common.h"
+#include "../iot_sample_common.h"
```
momuno, line 4: I'm curious if we can update cmake to recognize a path like "samples/iot/iot_sample_common.h".  While "../" works, I don't normally see that used in an include.

sdk/samples/iot/pnp/pnp_thermostat_component.c
```diff
@@ -7,11 +7,11 @@
...
 
-#include "iot_sample_common.h"
-#include "sample_pnp.h"
-#include "sample_pnp_mqtt_component.h"
+#include "../iot_sample_common.h"
+#include "pnp_protocol.h"
+#include "pnp_mqtt_message.h"
 
-#include "sample_pnp_thermostat_component.h"
+#include "pnp_thermostat_component.h"
```
momuno, line 15: Loving the directory and name updates.  Concise and organized!

sdk/samples/iot/pnp/pnp_protocol.c
danewalton: good point. updated feel free to check

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: MIT
 
-#include "iot_sample_common.h"
+#include <iot_sample_common.h>
```
momuno, line 4: Good eye.
I was reading up on the difference, and my understanding is that "" means to check current directory (or sub directories), while <> means to check via a directory list determined by the compiler. Does that sound accurate?

danewalton: yea IIRC it has to do with precedence of lookup. `""` is local directory first while I believe `<>` doesn't look up locally.

PR #1166, opened 2020-08-28
Add the doxygen warning log to gitignore and remove --all when fetching.

ahsonkhan:
Fixing based on @weshaggard recommmendation: https://github.com/Azure/azure-sdk-for-c/pull/1143#discussion_r477977213

@danieljurek - I can't move the warning log to `docs/` because that folder may not exists at the time of running doxygen, since it seems to be created near the end of the generation step. Because of that the file doesn't get written in that folder. So, just adding it to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -303,6 +303,7 @@ build/
 
 #Doxygen
 /[Dd]ocs/
+DoxygenWarningLog.txt
 
 # VS Code Files
 .vscode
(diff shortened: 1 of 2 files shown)


ahsonkhan:
Unless... we want to move the docs folder from `/sdk/docs/` to the root `/docs/`, and not have that be gitignored.

Issue #1082, opened 2020-08-19
Do not differentiate between an empty span and a null span


ahsonkhan:
- [x] Change AZ_SPAN_NULL macro to AZ_SPAN_EMPTY
- [x] Validate that we don't have any usages of `az_span_ptr(...) == NULL` in the code base, and replace those with `az_span_size(...) == 0`
- [ ] In the future, consider adding an `AZ_NODISCARD bool az_span_is_empty(az_span span);` convenience method.

As a side effect, we don't have a sentinel span. The user should not try to differentiate between an empty span and span with a null pointer. We haven't seen a common enough scenario that warrants it, and it adds confusion and burden to auditing `az_span_ptr` call sites, in code review.

Ideally, if we had indexer support in the language, we would have made `az_span_ptr` internal, and then byte access would be safe, by definition.

This follows the .NET precedent, where the BCL does not differentiate between those two either.
https://docs.microsoft.com/en-us/dotnet/api/system.span-1.empty?view=netcore-3.1
https://docs.microsoft.com/en-us/dotnet/api/system.span-1.isempty?view=netcore-3.1

cc @JeffreyRichter

PR #1168, opened 2020-08-28
Do not differentiate between empty and null span - rename macro.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1082

Also fixes https://github.com/Azure/azure-sdk-for-c/issues/1171

### DIFF
--- a/sdk/samples/iot/paho_iot_hub_pnp_sample.c
+++ b/sdk/samples/iot/paho_iot_hub_pnp_sample.c
@@ -852,17 +852,18 @@ static az_result invoke_getMaxMinReport(
       command_start_time_value_buffer,
       sizeof(command_start_time_value_buffer),
       &incoming_since_value_len));
-  az_span start_time_span
-      = az_span_create((uint8_t*)command_start_time_value_buffer, incoming_since_value_len);
-
-  IOT_SAMPLE_LOG_AZ_SPAN("start time:", start_time_span);
 
   // Set the response payload to error if the "since" value was empty.
-  if (az_span_ptr(start_time_span) == NULL)
+  if (incoming_since_value_len == 0)
   {
     *out_response = command_empty_response_payload;
     return AZ_ERROR_ITEM_NOT_FOUND;
   }
+
+  az_span start_time_span
+      = az_span_create((uint8_t*)command_start_time_value_buffer, incoming_since_value_len);
+
+  IOT_SAMPLE_LOG_AZ_SPAN("start time:", start_time_span);
 
   // Get the current time as a string.
   time_t rawtime;
--- a/sdk/samples/iot/paho_iot_hub_pnp_sample.c
+++ b/sdk/samples/iot/paho_iot_hub_pnp_sample.c
@@ -400,7 +400,7 @@ static void request_device_twin_document(void)
   }
 
   // Publish the twin document request.
-  mqtt_publish_message(twin_document_topic_buffer, AZ_SPAN_NULL, SAMPLE_PUBLISH_QOS);
+  mqtt_publish_message(twin_document_topic_buffer, AZ_SPAN_EMPTY, SAMPLE_PUBLISH_QOS);
 }
 
 static void receive_messages(void)
--- a/sdk/samples/iot/pnp/pnp_protocol.c
+++ b/sdk/samples/iot/pnp/pnp_protocol.c
@@ -124,7 +124,7 @@ static az_result is_component_in_model(
 {
   int32_t index = 0;
 
-  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
+  if (az_span_size(component_name) == 0)
   {
     return AZ_ERROR_UNEXPECTED_CHAR;
   }
@@ -154,7 +154,7 @@ az_result pnp_get_telemetry_topic(
 {
   az_iot_message_properties pnp_properties;
 
-  if (az_span_ptr(component_name) != NULL)
+  if (az_span_size(component_name) != 0)
   {
     if (properties == NULL)
     {
@@ -170,7 +170,7 @@ az_result pnp_get_telemetry_topic(
 
   AZ_RETURN_IF_FAILED(az_iot_hub_client_telemetry_get_publish_topic(
       client,
-      az_span_ptr(component_name) != NULL ? properties : NULL,
+      az_span_size(component_name) != 0 ? properties : NULL,
       mqtt_topic,
       mqtt_topic_size,
       out_mqtt_topic_length));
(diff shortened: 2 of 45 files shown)


ahsonkhan:
This is the rename and can be largely ignored for review:
[Do not differentiate between empty and null span - rename macro.](https://github.com/Azure/azure-sdk-for-c/pull/1168/commits/e10cd875cb8b2bb70239a3bf108400d3859f426c) - https://github.com/Azure/azure-sdk-for-c/pull/1168/commits/e10cd875cb8b2bb70239a3bf108400d3859f426c

**This is the important commit worth reviewing and folks should take a look at it:**
[Update all code paths that relied on az_span_ptr == NULL to differentiate between a null span and an empty span to use az_span_size == 0](https://github.com/Azure/azure-sdk-for-c/pull/1168/commits/5bdc9d323897684fe2bc4ca50384734e0961a088) - https://github.com/Azure/azure-sdk-for-c/pull/1168/commits/5bdc9d323897684fe2bc4ca50384734e0961a088

sdk/samples/iot/paho_iot_hub_pnp_sample.c
```diff
@@ -852,18 +852,19 @@ static az_result invoke_getMaxMinReport(
       command_start_time_value_buffer,
       sizeof(command_start_time_value_buffer),
       &incoming_since_value_len));
-  az_span start_time_span
-      = az_span_create((uint8_t*)command_start_time_value_buffer, incoming_since_value_len);
-
-  IOT_SAMPLE_LOG_AZ_SPAN("start time:", start_time_span);
```
ahsonkhan, line 858: @momuno - The span creation and logging should happen after the `start_time_span` size check, correct?

sdk/samples/iot/paho_iot_hub_pnp_sample.c
```diff
@@ -852,18 +852,19 @@ static az_result invoke_getMaxMinReport(
...
 
   // Set the response payload to error if the "since" value was empty.
-  if (az_span_ptr(start_time_span) == NULL)
```
ahsonkhan, line 861: @momuno - This would have never been true previously because `command_start_time_value_buffer` is not null. FYI.

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -124,7 +124,7 @@ static az_result is_component_in_model(
 {
   int32_t index = 0;
 
-  if (az_span_ptr(component_name) == NULL || az_span_size(component_name) == 0)
```
ahsonkhan, line 127: I don't see a reason to check for both ptr == null and size == 0.

If ptr == NULL, size must already be 0 anyway.
If ptr != NULL, size may or may not be 0.

Therefore, checking if size == 0, covers all the possibilities.

sdk/src/azure/iot/az_iot_hub_client_twin.c
```diff
@@ -131,20 +131,27 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
...
       // Get status and convert to enum
       uint32_t status_int;
       _az_RETURN_IF_FAILED(az_span_atou32(status_str, &status_int));
       out_twin_response->status = (az_iot_status)status_int;
 
+      if (index == -1)
+      {
+        return AZ_ERROR_UNEXPECTED_END;
+      }
+
```
ahsonkhan, line 154: FYI @danewalton, this is the fix for the bug https://github.com/Azure/azure-sdk-for-c/issues/1171 to prevent invalid slice later on line 157. It required updating the `_az_span_token` logic to return the index value.

sdk/tests/core/test_az_span.c
```diff
@@ -1751,17 +1755,11 @@ static void test_az_span_token_success(void** state)
...
+  assert_int_equal(index, -1);
   assert_true(az_span_ptr(token) == az_span_ptr(span));
   assert_int_equal(az_span_size(token), 4);
-  assert_true(az_span_ptr(out_span) == NULL);
   assert_true(az_span_size(out_span) == 0);
-
-  // Out_span is empty.
-  span = out_span;
-
-  token = _az_span_token(span, delim, &out_span);
```
ahsonkhan, line 1763: As previously documented, this is a precondition failure now.

sdk/samples/iot/paho_iot_hub_pnp_sample.c
momuno: oh of course.  good catch.

sdk/samples/iot/paho_iot_hub_pnp_sample.c
momuno: This looks correct.  @danewalton thoughts?

sdk/samples/iot/paho_iot_hub_pnp_sample.c
ahsonkhan: Also, as a recommendation, avoid using `az_span_ptr(...)` where possible, particularly when making just size-based decisions. Only use it when you need to access the contents of the span buffer either for reading or writing.

sdk/src/azure/iot/az_iot_hub_client_twin.c
danewalton: looks fine to me 👍

sdk/samples/iot/pnp/pnp_protocol.c
danewalton: 👍

sdk/samples/iot/paho_iot_hub_pnp_sample.c
danewalton: 👍

ahsonkhan:
> Should we update precondition macro (so that 0 can have null ptr, >0 can't), or review the existing preconditions we have in place?

I believe it already behaves this way.

I thought about removing the reliance on the null ptr check even in the preconditions, but given the preconditions also enforce that the span was correctly formed, I decided to leave it as is. Even though folks shouldn't rely on `null` ptr, we do create one with `null` (as an implementation detail) and are just using that fact within the precondition to validate well-formedness.

vhvb1989:
As a final user, can it happen that I want to know when az_span is not yet init/used  vs when it has already been?

```c
// Creating az_span like this will let ptr = NULL and size = 0
az_span my_span;

// at this point I know my_span is empty

// say I call some functions or do stuff here
my_span = foo(...);
// OR
az_result = foo(&out_my_span);  // Do we ask/request to never use az_span* for out params? 

// At this point my_span can still be empty. But ptr could be relevant for me.  

```

I wonder if by letting AZ_SPAN_NULL and empty az_span be different we support more than one programing style for using az_span as a customer.

ahsonkhan:
> I wonder if by letting AZ_SPAN_NULL and empty az_span be different we support more than one programing style for using az_span as a customer.


I thought about this for a while. I don't see a concrete scenario where the user would want to rely on the `NULL` ptr which cannot be substituted for a size check that would be worth the trade-off for increased complexity and security scrutiny.

There is pretty big precedence with .NET and even Golang which have the semantic behavior proposed in this PR and issue, which increases my confidence on the design.

ahsonkhan:
> As a final user, can it happen that I want to know when az_span is not yet init/used vs when it has already been?


@vhvb1989 - let me know if you have some other thoughts and scenarios on this issue and I will work with you to resolve them, or at least understand it. For now, merging this PR since it impacts API surface.

ahsonkhan:
Test reply from email.

Issue #1172, opened 2020-08-29
Providing an incomplete log classifications array results in AV


ahsonkhan:
This is a reliability bug. The following user code, which may be buggy, should not cause the SDK to AV:
```C
static void test_az_log_incomplete_list_fails_gracefully()
{
  az_log_classification const classifications[] = { AZ_LOG_HTTP_REQUEST };
  az_log_set_classifications(classifications);
  az_log_set_callback(_log_listener);

  assert_int_equal(0, _log_retry);
  _log_retry = 0;
  assert_int_equal(2229, az_iot_retry_calc_delay(5, 1, 500, 100000, 1234));
  assert_int_equal(0, _log_retry);

  az_log_set_callback(NULL);
  az_log_set_classifications(NULL);
}
```

```text
2: [  ERROR   ] --- EXCEPTION_ACCESS_VIOLATION occurred at 00007FF604D9635C.
2:
2: To debug in Visual Studio...
2: 1. Select menu item File->Open Project
2: 2. Change 'Files of type' to 'Executable Files'
2: 3. Open this executable.
2: 4. Select menu item Debug->Start
2:
2: Alternatively, set the environment variable
2: UNIT_TESTING_DEBUG to 1 and rebuild this executable,
2: then click 'Debug' in the popup dialog box.
2:
2:
2: [  FAILED  ] test_az_iot_provisioning_client_logging_succeed
2: [ RUN      ] test_az_log_incomplete_list_fails_gracefully
2: [  ERROR   ] --- EXCEPTION_ACCESS_VIOLATION occurred at 00007FF604D9635C.
2:
2: [  FAILED  ] test_az_log_incomplete_list_fails_gracefully
```

**This partially signals that we shouldn't be using an "end" sentinel to denote the end of list.**

![image](https://user-images.githubusercontent.com/6527137/91631590-8dfd3c00-e98f-11ea-9900-610e76317c06.png)

cc @antkmsft, @JeffreyRichter

JeffreyRichter:
[...]

PR #1174, opened 2020-08-29
Fix AV in logging when classification array isn't valid and add tests.

ahsonkhan:
Partially fixes https://github.com/Azure/azure-sdk-for-c/issues/1172 without removing the `AZ_LOG_END_OF_LIST` and keeping the design as is to use this sentinel to mark the end of the array.

Also added a bunch of interesting edge test cases.

I think we should consider removing this sentinel and having the user pass in the length of the array instead. That would be the most clean fix.

If we are sticking to preconditions, add tests: https://github.com/Azure/azure-sdk-for-c/pull/1174#discussion_r480851632


sdk/src/azure/core/az_log.c
```diff
@@ -57,8 +79,14 @@ static bool _az_log_write_engine(bool log_it, az_log_classification classificati
                            // is going to succeed and return from the function.
   }
 
-  for (az_log_classification const* cls = classifications; *cls != AZ_LOG_END_OF_LIST; ++cls)
+  az_log_classification const* cls = classifications;
+  while (*cls != AZ_LOG_END_OF_LIST)
```
antkmsft, line 83: Why did you remove the fol loop? If=t was line instead of 3, all iterations and exit logic in 1 place.

ahsonkhan: Good point, will revert.

sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -259,6 +258,20 @@ static void test_az_iot_provisioning_client_logging_succeed()
   az_log_set_classifications(NULL);
 }
 
+static void test_az_log_incomplete_list_fails_gracefully()
+{
+  az_log_classification const classifications[] = { AZ_LOG_HTTP_REQUEST };
+  az_log_set_classifications(classifications);
```
antkmsft, line 264: I don't see in this PR, how we would detect the end of the list consistently, without going to read the rest of the memory until we see a value that coincides with AZ_LOG_END_OF_LIST.

antkmsft: I see, so you're checking the classifications in the list, and not the classification bein logged. But in that case we are verifying first half the list (on average) on every log attempt. And still it doesn't prevent you from the AV - the code will still shoot in the memory, given the values are happen to be within the range of defined classifications.

sdk/src/azure/core/az_log.c
```diff
@@ -29,6 +29,27 @@ void az_log_set_callback(az_log_message_fn az_log_message_callback)
   _az_log_message_callback = az_log_message_callback;
 }
 
+// Verifies that  the classification that the user provided is one of the valid possibilties
+// and guards against looping past the end of the classification array.
+// Make sure to update the switch statement whenever new classifications are added.
+static bool _az_is_valid_log_classification(az_log_classification const* classification)
```
antkmsft, line 35: I think it is better to be AZ_INLINE to eliminate the invocation cost.

ahsonkhan: Making it a precondition for now, so leaving it as is.

ahsonkhan:
@antkmsft - how does it look now?

sdk/tests/core/test_az_logging.c
```diff
@@ -413,7 +366,7 @@ static void test_az_log_everything_on_null(void** state)
     _az_LOG_WRITE((az_log_classification)12345, AZ_SPAN_EMPTY);
     _az_LOG_WRITE(AZ_LOG_MQTT_RECEIVED_PAYLOAD, AZ_SPAN_EMPTY);
 
-    assert_int_equal(_az_BUILT_WITH_LOGGING(4, 0), _log_retry);
+    assert_int_equal(_az_BUILT_WITH_LOGGING(5, 0), _log_retry);
```
antkmsft, line 369: I was reading this, and `_log_retry` was confusing, especially given that there are log classifications having `RETRY` in their name. I had to expand the code above to make sure it what it really counts what I think it should. Rename to (`_number_of``) `_log_attempts`/`_log_invocations`?

sdk/src/azure/core/az_log.c
```diff
@@ -19,8 +19,38 @@
 static az_log_classification const* volatile _az_log_classifications = NULL;
 static az_log_message_fn volatile _az_log_message_callback = NULL;
 
+// Verifies that the classification that the user provided is one of the valid possibilties and
+// guards against looping past the end of the classification array.
+// Make sure to update the switch statement whenever new classifications are added.
+static bool _az_is_valid_log_classification(az_log_classification const* classifications)
```
antkmsft, line 25: The name should be plural, because this is what it does. I.e. `_az_is_valid_log_classifications`, or, better, I think, - `_az_log_classifications_are_valid`

sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -258,20 +258,6 @@ static void test_az_iot_provisioning_client_logging_succeed()
   az_log_set_classifications(NULL);
 }
 
-static void test_az_log_incomplete_list_fails_gracefully()
-{
-  az_log_classification const classifications[] = { AZ_LOG_HTTP_REQUEST };
```
antkmsft: Please write a test case covering that the precondition gets triggered when you pass an array of `{AZ_LOG_HTTP_REQUEST, (az_log_classification)12345}` (run it only when built with preconditions on).
There's a way to verify preconditions triggered, see some other tests. If you don't want to invest in writing that test until we held a broader discussion on preventing incorrect behavior on incorrect input data - that's fine, just please don't forget about it, create a work item to track if neccessary.

sdk/src/azure/core/az_log.c
antkmsft: Precondition should not interfere with inline. I see no reason to not make it such.

antkmsft:
@ahsonkhan rephrase the "fixes `#####`" word from the description - it looks like it is going to close the issue for you.

ahsonkhan:
> rephrase the "fixes #####" word from the description - it looks like it is going to close the issue for you.

I do want to close that issue. We already have a follow up #1174, so I don't want duplicate issues.

sdk/tests/iot/common/test_az_iot_common.c
ahsonkhan: I'll punt this for now, but will keep track of it and follow up as part of the design issue.

PR #1179, opened 2020-08-31
moving Doxyfile to /doc

### LINKED ISSUE #1146: Consider moving the Doxyfile from the root of the repo to under /eng/
See https://github.com/Azure/azure-sdk-for-c/pull/1133#discussion_r477501713

Mrlinkwii:
moves Doxyfile /eng

closes #1146


ahsonkhan:
Thanks for the contribution @Mrlinkwii 

It isn't yet clear whether we want this file to move to /eng/ or if we want to create a /docs/ folder and move it there. 

There is something nice about developers being able to run `doxygen Doxyfile` at the root of the repo, which creates a (currently) .gitignored `/docs/` folder with the generated docs. I have seen a few C/C++ repos where they have a docs folder which contains the doxyfile and other readme and docs.

Putting it in the `/eng/` folder is a bit odd, imo .I'll discuss the desired user and contributor experience with more folks and get back to you.

**What do you think the structure and behavior should be?**

ahsonkhan:
@weshaggard, @danieljurek, @RickWinter, @danewalton (picked as a stand-in for IoT, but other folks are definitely welcome o chime in) - what do we think about having a `/doc/` folder and moving the `Doxyfile` there? That seems to be a common pattern for other language SDKs and C/C++ GitHub repos in the wild (though they also contain conceptual docs there). We'll need to update some `.gitignore` and a couple of file paths.

**Several of our SDKs have a `/doc/` folder:**
https://github.com/Azure/azure-sdk-for-net/tree/master/doc
https://github.com/Azure/azure-sdk-for-python/tree/master/doc
https://github.com/Azure/azure-sdk-for-js/tree/master/documentation
https://github.com/Azure/azure-sdk-for-go/tree/master/documentation

**Repos in the wild that have a Doxyfile under a `/doc/` folder:**
https://github.com/eigenteam/eigen-git-mirror/blob/master/doc/Doxyfile.in
https://github.com/apache/xerces-c/blob/master/doc/Doxyfile.in
https://github.com/CGAL/cgal/blob/master/BGL/doc/BGL/Doxyfile.in
https://github.com/KDE/cocoon/blob/master/doc/Doxyfile

**This one has the Doxyfile under utilities:**
https://github.com/Kitware/VTK/tree/master/Documentation/Doxygen
https://github.com/Kitware/VTK/blob/master/Utilities/Doxygen/doxyfile.in

**This repo has it on the root, but also has a `/doc/` folder:**
https://github.com/freedesktop/dbus

weshaggard:
@ahsonkhan I'd be fine with putting this in a `/doc` folder as well.  My original comment was mostly trying to clean-up the root directory.

Mrlinkwii:
@weshaggard  ill create a `/doc` folder and move it in their :+1:  , it has been mentioned that the `.gitignore`  need to updated   ,dose  the /`doc` need to be included into it

ahsonkhan:
> it has been mentioned that the `.gitignore`  need to updated   ,dose  the /`doc` need to be included into it 

Nope, we don't need to put that in the .gitignore.

You'll have to update this path to `doc\Doxyfile`:
https://github.com/Azure/azure-sdk-for-c/blob/356e4ddf9f4e2ac2a74f736b6c9e4de7e3cd19fb/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L367

And update this readme step for generating the docs locally:
https://github.com/Azure/azure-sdk-for-c/blob/caff51dbd42d9d8d640bbdea7f3b29593521b2cc/CONTRIBUTING.md#build-docs

Try it out and make sure running `doxygen doc\Doxyfile` works for you locally.

Also, if there is a warning because of a missing doc comment in an API (you can simulate it by removing a comment from a method/field), then verify where the `DoxygenWarningLog.txt` gets generated. I think it will get generated in `doc\DoxygenWarningLog.txt`, but please verify that. What we might have to do, is update the .gitignore for the `DoxygenWarningLog.txt` log file (so that it doesn't get accidentally checked-in: https://github.com/Azure/azure-sdk-for-c/blob/98860972e427978c054544b263dcc660d66e234b/.gitignore#L304-L306

For example, remove this:
https://github.com/Azure/azure-sdk-for-c/blob/356e4ddf9f4e2ac2a74f736b6c9e4de7e3cd19fb/sdk/inc/azure/core/az_span.h#L43-L47

Let me know if you have any more questions or need some help.

CONTRIBUTING.md
```diff
@@ -113,7 +113,7 @@ cmake -DUNIT_TESTING=ON -DUNIT_TESTING_MOCKS=ON ..
 Running below command from root folder will create a new folder `docs` containing html file with documentation about CORE headers. Make sure you have `doxygen` installed on the system.
 
 ```bash
-doxygen Doxyfile
+doxygen doc\Doxyfile
```
ahsonkhan, line 116: You had it right before. Paths in bash/linux/osx use forward slash, `/`. Only Windows use baskslash, `\`.

```suggestion
doxygen doc/Doxyfile
```

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -340,7 +340,7 @@ jobs:
       # Assumes AZURE_SDK_VERSION is set in the environment
       # If doxygen produces any warnings, fail the step.
     - pwsh: |
-        Invoke-Expression "& `"C:\Program Files\doxygen\bin\doxygen.exe`" Doxyfile"
+        Invoke-Expression "& `"C:\Program Files\doxygen\bin\doxygen.exe`" doc\Doxyfile"
```
ahsonkhan, line 343: The path is right here, since it runs on Windows.

PR #1180, opened 2020-08-31
add error if powershell core isn't used

danewalton:


sdk/samples/iot/generate_certificate.ps1
```diff
@@ -4,6 +4,11 @@
 Set-StrictMode -Version 3.0
 $errorActionPreference="stop"
 
+if($PSVersionTable.PSEdition -ne "Core")
+{
+  Write-Error "Please install Powershell Core at aka.ms/pscore6."
```
ericwol-msft, line 9: better description
 ```suggestion
  Write-Error "This script requires Powershell Core. Please install Powershell Core at aka.ms/pscore6."
```

File: sdk/samples/iot/generate_certificate.ps1 [hunk 2]
```diff
@@ -4,6 +4,11 @@
 Set-StrictMode -Version 3.0
 $errorActionPreference="stop"
 
+if($PSVersionTable.PSEdition -ne "Core")
+{
+ Write-Error "This script requires Powershell Core. Please install Powershell Core at aka.ms/pscore6."
```
jspaith, line 9: I don't think `Write-Error` is enough - there's potential for this to get lost with other stuff script puts out.  We also should have a `throw` to bring the whole thing to a halt.

PR #1183, opened 2020-08-31
use throw instead of Writer-Error

danewalton:
as per @jspaith

### DIFF
--- a/sdk/samples/iot/generate_certificate.ps1
+++ b/sdk/samples/iot/generate_certificate.ps1
@@ -4,9 +4,8 @@
 Set-StrictMode -Version 3.0
 $errorActionPreference="stop"
 
-if($PSVersionTable.PSEdition -ne "Core")
-{
- Write-Error "This script requires Powershell Core. Please install Powershell Core at aka.ms/pscore6."
+if ($PSVersionTable.PSEdition -ne "Core") {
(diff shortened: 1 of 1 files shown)


sdk/samples/iot/generate_certificate.ps1
```diff
@@ -4,9 +4,8 @@
 Set-StrictMode -Version 3.0
 $errorActionPreference="stop"
 
-if($PSVersionTable.PSEdition -ne "Core")
-{
- Write-Error "This script requires Powershell Core. Please install Powershell Core at aka.ms/pscore6."
```
ahsonkhan, line 9: I used `Write-Error` in another script. Can you explain why `throw` is better? What's wrong with using `Write-Error`?

danewalton: The output stands out a little more. The "error" message with throw is a little more by itself so it easier to recognize from users.

jspaith: If another PowerShell script invoked Dane's script, then the throw would bubble up to the caller and it could then programmatically take action or just let the unhandled exception bubble up and stop its execution as well.  Not being on correct version of PS is serious enough to stop the show here.

https://stackoverflow.com/questions/9294949/when-should-i-use-write-error-vs-throw-terminating-vs-non-terminating-errors has some more on this.

PR #1189, opened 2020-09-01
update vscode config env var names

danewalton:


momuno:
> How can we make sure to keep these in sync?

There shouldn't be any more changes to these.  I did not know about the config file, so part of keeping things in sync is knowing all the files that need to be updated.  Still not sure what this config file does though. ?

momuno:
@danewalton what is the purpose of this config file?

danewalton:
Find and replace all using VSCode should in theory catch it.
These are used to use with VSCode and the CMake extension to help building and running tests and the samples

PR #1191, opened 2020-09-01
Remove validating precondition in 'az_log_set_classifications'

ahsonkhan:
Undoing most of https://github.com/Azure/azure-sdk-for-c/pull/1174 which was intended to put a band-aid over https://github.com/Azure/azure-sdk-for-c/issues/1172 but keeping the tests 

It is potentially useful to allow user-defined log classifications.

The downside is, developer bugs or misuse will result in undefined behavior, including AVs. We need to resolve this with better/more explicit documentation and code snippets - see https://github.com/Azure/azure-sdk-for-c/issues/1192

cc @JeffreyRichter

### DIFF
--- a/sdk/inc/azure/core/az_log.h
+++ b/sdk/inc/azure/core/az_log.h
@@ -33,8 +33,6 @@
 /**
  * @brief Identifies the classifications of log messages produced by the SDK.
  */
-// Make sure to update #_az_log_classifications_are_valid() whenever a new value is added to this
-// enum.
 typedef enum
 {
   AZ_LOG_END_OF_LIST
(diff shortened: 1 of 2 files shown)


antkmsft:
I think we still may end up with some form of a map, i.e. some solution that requires core to know about the valid/possible log classifications. If that's the case, then I think it is worth to have this precondition check, plus we may use it to verify the classification being logged.

ahsonkhan:
> Regarding keeping the unit tests. Isn't there a unit test that attempts malformed (without `END_OF_LIST`) list of classifications? If so, our tests could be AV'ing from time to time.

Nope, we don't have any such tests. I removed those. If we had such tests, they would AV deterministically :)

https://github.com/Azure/azure-sdk-for-c/blob/32a88e1b853c692b11f430670931934b097bbe5d/sdk/tests/core/test_az_logging.c#L284-L298

> I very strongly suggest for tests to stay within the defined behavior, especially if undefined behavior is non-deterministic and not 100% reproduceable.

Good point. Agreed.

vhvb1989 (approved):

PR #1193, opened 2020-09-02
Remove Client Secret Credential

### LINKED ISSUE #1091: Remove anything relate to client secret credential from core, since it is not used by anything.

antkmsft:
Closes #1091


sdk/inc/azure/core/az_credentials.h
```diff
@@ -92,53 +50,6 @@ typedef struct
   } _internal;
 } _az_credential;
```
ahsonkhan, line 51: Now this file only contains internal-only types and things. Move it to an internal header (`internal/az_credentials_internal.h`) or rename it to `_az_credentials.h` since the customer shouldn't be including it for anything.

sdk/tests/core/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ include(AddTestCMocka)
 
 # -ld link option is only available for gcc
 if(UNIT_TESTING_MOCKS)
-    set(WRAP_FUNCTIONS "-Wl,--wrap=az_platform_clock_msec -Wl,--wrap=az_http_client_send_request")
```
ahsonkhan, line 14: Why did we get rid of `az_http_client_send_request`?

sdk/inc/azure/core/az_credentials.h
ahsonkhan: Do we still need `_az_credential_set_scopes_fn`?

antkmsft: `AZ_CREDENTIAL_ANONYMOUS` is exactly what the users are going to be using, and they should include `az_credentials.h` for that. Later, as/if we implement more credential types (resurrect Client Secret Credential), they will go into this file. So I think it is useful to keep this file structure.

sdk/tests/core/CMakeLists.txt
antkmsft: No unit test mocks `az_http_client_send_request`. It was only the Client Secret credential that did.

sdk/inc/azure/core/az_credentials.h
ahsonkhan: At least for where we have landed now, it is certainly strange for the user to include an entirely separate header file just for one single line macro.

@barcharcraz - what do you think?

ahsonkhan: Also, @JeffreyRichter - thoughts on keeping `az_credential.h` for `AZ_CREDENTIAL_ANONYMOUS`?

ahsonkhan: Can we remove it for now?

PR #1195, opened 2020-09-02
Remove stdlib from SDK includes - curl.

ahsonkhan:
I don't think it was used for anything, so this was easy :)

Leaving uses of stdlib in samples and tests, for now.

cc @vhvb1989, @danieljurek


ahsonkhan:
Ah `malloc`. I am surprised only linux/mac CI legs that have samples failed, but not the windows ones.

PR #1198, opened 2020-09-02
Update all files in the repo to be ASCII only and save without a UTF-8 BOM

ahsonkhan:
As a follow up to this https://github.com/Azure/azure-sdk-for-c/pull/1196#discussion_r482486903, make sure all files in the SDK have ASCII-only characters.

Ran this script to detect issues:
```bash
#!/bin/bash

shopt -s globstar

for f in /sdk/* sdk/**/* ; do
  echo "Processing $f file..."
  grep --color='auto' -P -n "[^\x00-\x7F]" $f
done;
```

### DIFF
--- a/script.sh
+++ b/script.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+shopt -s globstar
+
+for f in /sdk/* sdk/**/* ; do
+  echo "Processing $f file..."
+  #hexdump -C $f > temp.txt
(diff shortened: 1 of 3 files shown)


ahsonkhan:
Looking for  a review here. Should be fairly quick.

RickWinter (approved):

vhvb1989 (approved):

PR #1194, opened 2020-09-02
Remove az pair

antkmsft:
Closes #1116.

PR #1203, opened 2020-09-03
PR #1194 follow up

antkmsft:


antkmsft:
@ahsonkhan, this is a response to the comments that you've left after #1194 was closed already.

ahsonkhan:
Hey @antkmsft - can you please apply the `Azure.Core` label to PRs that impact Core only. It helps me with parsing the PR list during my routine polling of outstanding PRs :)

sdk/inc/azure/core/internal/az_http_internal.h
```diff
@@ -225,15 +225,10 @@ AZ_NODISCARD az_result az_http_request_set_query_parameter(
...
- *   - *`AZ_ERROR_INSUFFICIENT_SPAN_SIZE`* there isn't enough space in the `p_request->buffer`
- * to add a header.
- *   - *`AZ_ERROR_ARG`*
- *     - `ref_request` is _NULL_.
- *     - `name` or `value` are invalid spans (see @ref _az_span_is_valid).
- *     - `name` or `value` are empty.
- *     - `name`'s or `value`'s buffer overlap resulting `url`'s buffer.
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK Success.
+ * @retval #AZ_ERROR_INSUFFICIENT_SPAN_SIZE there isn't enough space in the \p ref_request to add a
```
ahsonkhan, line 230:
```suggestion
 * @retval #AZ_ERROR_INSUFFICIENT_SPAN_SIZE There isn't enough space in the \p ref_request to add a
```

sdk/inc/azure/core/internal/az_http_internal.h
```diff
@@ -225,15 +225,10 @@ AZ_NODISCARD az_result az_http_request_set_query_parameter(
...
  *
- * @return
- *   - *`AZ_OK`* success.
- *   - *`AZ_ERROR_INSUFFICIENT_SPAN_SIZE`* there isn't enough space in the `p_request->buffer`
- * to add a header.
- *   - *`AZ_ERROR_ARG`*
- *     - `ref_request` is _NULL_.
- *     - `name` or `value` are invalid spans (see @ref _az_span_is_valid).
- *     - `name` or `value` are empty.
- *     - `name`'s or `value`'s buffer overlap resulting `url`'s buffer.
```
ahsonkhan, line 236: So, we never return this `az_result`? More reason to update the docs :)

ahsonkhan: What does actually happen if name/value are empty or overlapping?

ahsonkhan (approved):

vhvb1989 (approved):

sdk/inc/azure/core/internal/az_http_internal.h
antkmsft: Yes, we never return that here anymore, we have them as precondition now. For overlapping regions, that could've been a problem in the future, and since then the function was rewritten. Maybe it is still a problem, but it is not ovious from the implementation at the very least to a point that without going an checking all the function calls, it seems that the code would handle the overlapping correctly. Not necessarily the truth, but that's how it looks like. And the rest are preconditions. Btw, we do have a `_az_PRECONDITION_NO_OVERLAP_SPANS()` if we wanted to check for this, but again - from the implementation I it is not apparent that overlapping would create a problem.

PR #1206, opened 2020-09-03
Replace ==/!= AZ_OK with az_result_succeeded()/failed()

antkmsft:
I noticed this during the API review meeting.
I am updating our SDK code, samples, and docs. I am not updating tests, so they are more specifically ckeck for `AZ_OK` so any possible future update would be more intentional.

### DIFF
--- a/sdk/docs/iot/README.md
+++ b/sdk/docs/iot/README.md
@@ -298,15 +298,15 @@ void handle_iot_message(mqtt_client_message* msg)
   az_iot_hub_client_method_request method_request;
   az_iot_hub_client_c2d_request c2d_request;
   az_iot_hub_client_twin_response twin_response;
-  if (az_iot_hub_client_methods_parse_received_topic(&client, incoming_topic, &method_request) == AZ_OK)
+  if (az_result_succeeded(az_iot_hub_client_methods_parse_received_topic(&client, incoming_topic, &method_request)))
   {
     //Handle the method request
   }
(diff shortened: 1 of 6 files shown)


antkmsft:
@ahsonkhan, If only we could be sure that we will never need any other success code in the future other than AZ_OK, then I would be replacing these function calls with `==`/`!= AZ_OK`, and the part that you'd like in that would be that we would get rid of succeeded()/failed() functions, one being inverse of the other 😄

ahsonkhan:
Right! I just want to get rid of `az_result_succeeded()`, and we don't ever add any other success other than AZ_OK :)

Let's chat separately, and try some stuff, outside this PR. The intent of this PR is good as is (and not contentious).

ahsonkhan (approved):

RickWinter (approved):

vhvb1989 (approved):

momuno:
> Right! I just want to get rid of `az_result_succeeded()`, and we don't ever add any other success other than AZ_OK :)
> 
> Let's chat separately, and try some stuff, outside this PR. The intent of this PR is good as is (and not contentious).
> 
> I like using `az_result_failed()` instead of direct error code comparison though.

Ohhhhh both are being used in samples!  There will be some cleanup if it is removed, but I do find checking for failure and subsequently running "failure" code is a more common logic pattern than checking for success.

momuno (approved):

PR #1207, opened 2020-09-03
AZ_ERROR_DEPENDENCY_NOT_PROVIDED

antkmsft:
This came up during API review meeting. Replace some instances of AZ_ERROR_NOT_IMPLEMENTED with AZ_ERROR_DEPENDENCY_NOT_PROVIDED, and update the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ By default, when building the project with no options, the following static libr
   - az_storage_blobs
     - Storage SDK blobs client.
   - az_noplatform
-    - A platform abstraction which will compile but returns `AZ_ERROR_NOT_IMPLEMENTED` for all platform calls. This ensures the project can be compiled without the need to provide any specific platform implementation. This is useful if you want to use az_core without platform specific functions like `mutex` or `time`.
+    - A platform abstraction which will compile but returns 0 or does nothing for all platform calls. This ensures the project can be compiled without the need to provide any specific platform implementation. This is useful if you want to use az_core without platform specific functions like `time` or `sleep`.
   - az_nohttp
     - Library that provides a basic returning error when calling HTTP stack. Similar to az_noplatform, this library ensures the project can be compiled without requiring any HTTP stack implementation. This is useful if you want to use `az_core` without `az_http` functionality.
 
--- a/sdk/inc/azure/core/az_result.h
+++ b/sdk/inc/azure/core/az_result.h
@@ -80,7 +80,8 @@ typedef enum
   /// Not supported.
   AZ_ERROR_NOT_SUPPORTED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 7),
 
-  /// An external dependency required to perform operation was is not provided.
+  /// An external dependency required to perform the operation was not provided. The operation needs
+  /// an implementation of the platform layer or an HTTP transport adapter.
   AZ_ERROR_DEPENDENCY_NOT_PROVIDED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 8),
 
   // === Platform ===
--- a/sdk/inc/azure/core/az_result.h
+++ b/sdk/inc/azure/core/az_result.h
@@ -80,6 +80,9 @@ typedef enum
   /// Not supported.
   AZ_ERROR_NOT_SUPPORTED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 7),
 
+  /// An external dependency required to perform operation was is not provided.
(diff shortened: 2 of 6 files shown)


antkmsft:
image
😄

sdk/inc/azure/core/az_result.h
```diff
@@ -80,6 +80,9 @@ typedef enum
   /// Not supported.
   AZ_ERROR_NOT_SUPPORTED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 7),
 
+  /// An external dependency required to perform operation was is not provided.
+  AZ_ERROR_DEPENDENCY_NOT_PROVIDED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 8),
```
ahsonkhan, line 84: Should this go under the platform facility?

ahsonkhan: Also, consider adding a change log entry for this addition/change.

sdk/src/azure/platform/az_nohttp.c
```diff
@@ -19,5 +19,5 @@ az_http_client_send_request(az_http_request const* request, az_http_response* re
 {
   (void)request;
   (void)ref_response;
-  return AZ_ERROR_NOT_IMPLEMENTED;
+  return AZ_ERROR_DEPENDENCY_NOT_PROVIDED;
```
ahsonkhan, line 22: Update this too:
https://github.com/Azure/azure-sdk-for-c#cmake-options

> A platform abstraction which will compile but returns AZ_ERROR_NOT_IMPLEMENTED for all platform calls. ...

sdk/inc/azure/core/az_result.h
```diff
@@ -80,6 +80,9 @@ typedef enum
   /// Not supported.
   AZ_ERROR_NOT_SUPPORTED = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE, 7),
 
+  /// An external dependency required to perform operation was is not provided.
```
ahsonkhan, line 83:
```suggestion
  /// An external dependency required to perform the operation was not provided. The operations needs an implementation of the platform layer or an HTTP transport adapter.
```

ahsonkhan (approved):
Otherwise, looks good.

vhvb1989 (approved):

sdk/inc/azure/core/az_result.h
antkmsft: You could argue either way, but the error is general in nature. The same way, other core codes, if you look at them are general in nature, but other modules may return them. I'd say the module is not about the point where the error was originated, but about its being specific to a particular scope.

ahsonkhan: Makes sense. This one is sort of like the out-of-memory one though (which is under platform). Since core doesn't have any dependencies, it is odd to have one with that facility.

The facility is internal, if you feel strongly, we can defer changing it to later.

README.md
```diff
@@ -149,7 +149,7 @@ By default, when building the project with no options, the following static libr
   - az_storage_blobs
     - Storage SDK blobs client.
   - az_noplatform
-    - A platform abstraction which will compile but returns `AZ_ERROR_NOT_IMPLEMENTED` for all platform calls. This ensures the project can be compiled without the need to provide any specific platform implementation. This is useful if you want to use az_core without platform specific functions like `mutex` or `time`.
+    - A platform abstraction which will compile but returns 0 or does nothing for all platform calls. This ensures the project can be compiled without the need to provide any specific platform implementation. This is useful if you want to use az_core without platform specific functions like `time` or `sleep`.
   - az_nohttp
```
ahsonkhan, line 153: Mention that this will return `AZ_ERROR_DEPENDENCY_NOT_PROVIDED` at runtime.

antkmsft: And maybe it should, let's get what @JeffreyRichter thinks, and complete that in a separate PR.
Jeffrey, do you think we should change PAL functions (we only have two - `sleep()` and `clock()`) to return AZ_RESULT, and the `clock()` would return result as an out parameter? This way, if you ever seriously rely on a no-op PAL, the error will get to you, the stuff will start failing earlier, with better error code.

ahsonkhan: I am talking about `az_nohttp`, no the PAL.

antkmsft: But this section talks about platform, not HTTP. It used to say it would return `AZ_ERROR_NOT_IMPLEMENTED` but reality is that it won't. So I touched the comment, but instead of renaming, I updated it to say what it actually does, which is to return 0 or do nothing.

antkmsft: README.md line 25 talks about `az_nohttp`, and I updated it with DEPENDENCY_NOT_PROVIDED.

antkmsft: https://github.com/Azure/azure-sdk-for-c/pull/1219

PR #1210, opened 2020-09-04
PnP Component Sample Update

momuno:
- out_, ref_, const
- refactor to align with other samples
- refactor to align thermostat and temp controller component functions
- finalize logging output


ahsonkhan:
> out_, ref_, const

@antkmsft - can you please take a look.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -723,345 +857,440 @@ static void handle_device_twin_message(
...
+    az_span payload,
+    az_span* out_payload)
 {
   az_result rc;
+  az_span empty_span = AZ_SPAN_EMPTY; // For compilation.
 
-  // Get the Telemetry topic to publish the telemetry message and build the telemetry message.
-  if (az_result_failed(
-          rc = pnp_thermostat_get_telemetry_message(&hub_client, &thermostat_1, &mqtt_message)))
+  if (az_result_failed(rc = pnp_build_reported_property(payload, empty_span, twin_reported_serial_number_property_name, append_string_callback, (void*)&twin_reported_serial_number_property_value, out_payload)))
```
danewalton, line 1001: pedantic but here assign result and then check as discussed.

sdk/samples/iot/pnp/pnp_mqtt_message.h
```diff
@@ -9,19 +9,36 @@
...
   size_t* out_topic_length;
-  az_span payload_span;
-  az_span out_payload_span;
+  size_t topic_length;
+  az_span out_payload;
+  az_span payload;
 } pnp_mqtt_message;
 
-void pnp_mqtt_message_init(pnp_mqtt_message* mqtt_message);
+/*
```
danewalton, line 25: Thanks for doing all this work but a couple comments.

1. Note that this won't be rendered in doxygen at all since it isn't part of public API. Still useful to have since it shows what the function does.
2. IF it were intended to be processed by doxygen, note that the opening comment must be of the form 
```c
/**
```
instead of just one star. Doxygen processes double star and not single.

sdk/samples/iot/pnp/pnp_mqtt_message.h
```diff
@@ -9,19 +9,36 @@
...
+az_result pnp_mqtt_message_init(pnp_mqtt_message* out_mqtt_message);
 
-// Create request id span which increments request id integer each call. Capable of holding 8 digit
-// number.
-az_span get_request_id(void);
+/*
+ * @brief Creates a request id #az_span for use in sending twin messages. Value increments on each
+ * call.  Capable of holding a 10 digit number (base 10).
+ *
+ * @return An #az_span containing the requst id.
```
danewalton, line 40:
```suggestion
 * @return An #az_span containing the request id.
```

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -171,47 +170,55 @@ az_result pnp_get_telemetry_topic(
...
     az_span property_name,
     pnp_append_property_callback append_callback,
     void* context,
     az_span* out_span)
 {
+    (void)context;
+  (void)append_callback;
+  (void)property_name;
+  (void)component_name;
+    (void)json_buffer;
```
danewalton, line 211: formatting?

sdk/samples/iot/pnp/pnp_protocol.h
```diff
@@ -13,14 +13,22 @@
...
- * API.
+ * @brief Callback which is invoked for each property found by the
+ * #pnp_process_device_twin_message() API.
  */
 typedef void (*pnp_property_callback)(
     az_span component_name,
-    const az_json_token* property_name,
-    az_json_reader property_value_as_json,
+    az_json_token const* property_name,
+    az_json_reader* property_value_as_json,
```
danewalton, line 31: this should stay as a value not pointer. Reason being we want to process the document on our end in a predictable manner and know the state. If the user somehow shifts the token to a place we don't expect, it will cause issues inside the function which invoked this callback.

sdk/samples/iot/pnp/pnp_protocol.h
```diff
@@ -13,14 +13,22 @@
 #include <azure/core/az_span.h>
 #include <azure/iot/az_iot_hub_client.h>
 
+typedef enum
+{
+  PNP_METHODS_PUBLISH_TOPIC,
+  PNP_TELEMETRY_PUBLISH_TOPIC,
+  PNP_TWIN_DOCUMENT_PUBLISH_TOPIC,
+  PNP_TWIN_PATCH_PUBLISH_TOPIC
+} pnp_publish_topic;
```
danewalton, line 22: What is this used for?

sdk/samples/iot/pnp/pnp_protocol.h
```diff
@@ -85,19 +100,23 @@ az_result pnp_create_reported_property(
     az_span* out_span);
 
 /**
- * @brief Build a reported property with the ack status
+ * @brief Build a reported property with the ack status.
  *
- * @param[in] json_buffer The span into which the json payload will be placed.
+ * @param[out] json_buffer An #az_span with sufficient capacity to hold the json payload.
```
danewalton, line 105:
```suggestion
 * @param[in] json_buffer An #az_span with sufficient capacity to hold the json payload.
```

sdk/samples/iot/pnp/pnp_protocol.h
```diff
@@ -109,21 +128,44 @@ az_result pnp_create_reported_property_with_status(
     az_span* out_span);
 
 /**
- * @brief Iteratively get the next desired property.
+ * @brief Build a simple telemetry message using one property name and one value.
+ *
+ * @param[out] json_buffer An #az_span with sufficient capacity to hold the json payload.
```
danewalton, line 133:
```suggestion
 * @param[in] json_buffer An #az_span with sufficient capacity to hold the json payload.
```

sdk/samples/iot/pnp/pnp_thermostat_component.c
```diff
@@ -24,363 +24,326 @@
...
+az_result pnp_thermostat_process_command_request(
+    pnp_thermostat_component const* thermostat_component,
     az_span command_name,
-    az_span command_payload,
-    pnp_mqtt_message* mqtt_message,
-    az_iot_status* status)
+    az_span command_received_payload,
+    az_iot_status* out_status,
+    az_span payload,
+    az_span* out_payload)
```
danewalton, line 318:
```suggestion
    az_span payload,
    az_iot_status* out_status,
    az_span* out_payload)
```

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
-    pnp_mqtt_message* mqtt_message);
+    double initial_temperature);
 
-bool pnp_thermostat_get_max_temp_report(
-    const az_iot_hub_client* client,
+/**
+ * @brief Build the thermostat's temperature telemetry message.
+ *
+ * @param[in] thermostat_component A pointer to the themostat component to get data.
+ * @param[out] payload An #az_span with sufficient capacity to hold the json payload.
```
danewalton, line 50:
```suggestion
 * @param[in] payload An #az_span with sufficient capacity to hold the json payload.
```

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+ * @param[in] thermostat_component A pointer to the themostat component to get data.
+ * @param[out] out_property_name The name of the reported property to be sent.
+ * @param[out] payload An #az_span with sufficient capacity to hold the json payload.
+ * @param[out] out_payload A pointer to the #az_span containing the output json payload.
+ */
+void pnp_thermostat_build_maximum_temperature_reported_property(
     pnp_thermostat_component* thermostat_component,
+    az_span* out_property_name,
+    az_span payload,
+    az_span* out_payload);
```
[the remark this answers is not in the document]
danewalton: I believe with the exception of initialization functions

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
-    az_span component_name,
+/**
+ * @brief Update the thermostat's member variables and prepare reported property message.
+ *
+ * @param[in,out] ref_thermostat_component A pointer to the themostat component to update data.
+ * @param[in] property_name The name of the property to be updated.
+ * @param[in] property_value The value used for the property update.
+ * @param[in] version The version parsed from the received message, and used to prepare the returned
+ *                    reported property message.
+ * @param[out] payload An #az_span with sufficient capacity to hold the prepared reported property
```
danewalton, line 100:
```suggestion
 * @param[in] payload An #az_span with sufficient capacity to hold the prepared reported property
```

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+
+/**
+ * @brief Update the thermostat's member variables and prepare reported property message.
+ *
+ * @param[in] thermostat_component A pointer to the themostat on which to invoke the command.
+ * @param[in] command_name The name of the command to be invoked.
+ * @param[in] command_received_payload The received payload to be used by the invoked command.
+ * @param[out] out_status The status resulting from the invoked command, to be used in the command
+ *                        response.
+ * @param[out] payload An #az_span with sufficient capacity to hold the command response json
```
danewalton, line 121:
```suggestion
 * @param[in] payload An #az_span with sufficient capacity to hold the command response json
```

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+az_result pnp_thermostat_process_command_request(
+    pnp_thermostat_component const* thermostat_component,
     az_span command_name,
-    az_span command_payload,
-    pnp_mqtt_message* mqtt_message,
-    az_iot_status* status);
+    az_span command_received_payload,
+    az_iot_status* out_status,
+    az_span payload,
+    az_span* out_payload);
```
danewalton, line 132:
```suggestion
    az_span payload,
    az_iot_status* out_status,
    az_span* out_payload);
```

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -143,12 +143,11 @@ static az_result is_component_in_model(
...
 
-// Get the telemetry topic for PnP
-az_result pnp_get_telemetry_topic(
-    const az_iot_hub_client* client,
+az_result pnp_get_telemetry_publish_topic(
+    az_iot_hub_client const* client,
     az_iot_message_properties* properties,
     az_span component_name,
-    char* mqtt_topic,
+    char* out_mqtt_topic,
```
danewalton, line 150: I think this should stay `mqtt_topic` to match the API's

sdk/samples/iot/pnp/pnp_mqtt_message.h
danewalton: So no changes needed

danewalton (changes requested):

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+
+/**
+ * @brief Build the thermostat's error message with status.
+ *
+ * @param[in] component_name The name of the component for the reported property.
+ * @param[in] property_name The name of the property for which to send an update.
+ * @param[in] property_value The property value to be appended.
+ * @param[in] status The return status for the error message ack.
+ * @param[in] version The version for the reported property ack.
+ * @param[out] payload An #az_span with sufficient capacity to hold the json payload.
```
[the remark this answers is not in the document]
momuno: thanks I forgot to change these back.

sdk/samples/iot/iot_sample_common.h
```diff
@@ -203,12 +207,12 @@ uint32_t iot_sample_get_epoch_expiration_time_from_minutes(uint32_t minutes);
  * @brief Generate the base64 encoded and signed signature using HMAC-SHA256 signing.
  *
  * @param[in] sas_base64_encoded_key An #az_span containing the SAS key that will be used for
- * signing.
+ *                                   signing.
```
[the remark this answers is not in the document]
momuno: I copied the style I see in the iot API headers.  I can put it back, but then we should probably update those too later.
https://github.com/Azure/azure-sdk-for-c/blob/caa99de4006233eae37c0ac82a77f50689a9d0e6/sdk/inc/azure/iot/az_iot_hub_client.h#L113-L119

sdk/samples/iot/pnp/pnp_thermostat_component.h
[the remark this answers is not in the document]
momuno: oh i see. thanks

sdk/samples/iot/pnp/pnp_mqtt_message.h
momuno: oh. lol.  thanks!

sdk/samples/iot/pnp/pnp_thermostat_component.h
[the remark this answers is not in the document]
momuno: wait, nm above comment.  I confused [out[ with out_ prefix.  

@antkmsft what is your thought on this?  I have [out] set in another sample file like so:
```json
/*
 * @brief Generate the base64 encoded and signed signature using HMAC-SHA256 signing.
 *
 * @param[in] sas_base64_encoded_key An #az_span containing the SAS key that will be used for
 *                                   signing.
 * @param[in] sas_signature An #az_span containing the signature.
 * @param[out] sas_base64_encoded_signed_signature An #az_span with sufficient capacity to hold the
 *                                                 encoded signed signature.
 * @param[out] out_sas_base64_encoded_signed_signature A pointer to the #az_span containing the
 *                                                     encoded signed signature.
 */
void iot_sample_generate_sas_base64_encoded_signed_signature(
    az_span sas_base64_encoded_key,
    az_span sas_signature,
    az_span sas_base64_encoded_signed_signature,
    az_span* out_sas_base64_encoded_signed_signature);
```

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+    az_span* out_payload);
 
-az_result pnp_thermostat_process_property_update(
-    const az_iot_hub_client* client,
+/**
+ * @brief Build the thermostat's maximum temperature reported property message.
+ *
+ * @param[in] thermostat_component A pointer to the themostat component to get data.
+ * @param[out] out_property_name The name of the reported property to be sent.
+ * @param[out] payload An #az_span with sufficient capacity to hold the json payload.
```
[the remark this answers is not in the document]
momuno: confirming with anton.

sdk/samples/iot/pnp/pnp_thermostat_component.h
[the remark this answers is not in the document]
danewalton: The `[out]` should be changed to `[in]` because that span (without the `out_` prefix) tells the API what pointer and how much space it has to use. It isn't referenced again by the user after the function call for any information.

sdk/samples/iot/pnp/pnp_protocol.h
momuno: Do you mean shift the location of the token, or shift the value of the token?  We can keep the location in check by using az_json_reader *const, but def can't do az_json_reader const* to protect the token value because it will eventually be cast to void* in pnp_build_reported_property_with_status().

sdk/samples/iot/pnp/pnp_protocol.h
momuno: I consolidated all the publish topics into a single set_publish_topic function to eliminate some of the duplicated code everywhere for getting the publish topic.  Also an attempt to follow along with the publish_mqtt_message and not have az_result returns to check in so many places.  

With talk today on removing az_result from returning in general, this may no longer be as helpful.

sdk/samples/iot/pnp/pnp_thermostat_component.h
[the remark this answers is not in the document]
antkmsft: In this case, it is `[in]`, by the spirit of the parameter I'd say. You could argue that we do place output in that buffer, but we have an out parameter designated to mark the subspan in that buffer. So you pass in the space, you get out a region.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -92,51 +85,80 @@ static void connect_mqtt_client_to_iot_hub(void);
...
 
 // Callbacks
-static az_result append_string_callback(az_json_writer* jw, void* value);
-static az_result append_json_token_callback(az_json_writer* jw, void* value);
 static void property_callback(
     az_span component_name,
-    const az_json_token* property_name,
-    az_json_reader property_value,
+    az_json_token const* property_name,
+    az_json_reader* const property_value,
```
danewalton, line 150: From what I remember, we have tabled whether we add `const` on pointers. I would prefer not adding this because you might have to individually change it back.

sdk/samples/iot/pnp/pnp_thermostat_component.h
```diff
@@ -17,47 +17,118 @@
...
+ * reported property message.
+ * @param[in] payload An #az_span with sufficient capacity to hold the prepared reported property
+ * json payload.
+ * @param[out] out_payload A pointer to the #az_span containing the output reported property json
+ * payload.
+ */
+az_result pnp_thermostat_process_property_update(
+    pnp_thermostat_component* ref_thermostat_component,
+    az_json_token const* property_name,
+    az_json_reader const* const property_value,
```
danewalton, line 108: I don't think this is right. If we pass the `property_value` as an `az_json_reader`, this API will iterate on the reader which will change the thing it is pointed to. I think this should just be `az_json_reader property_value`

sdk/samples/iot/pnp/pnp_protocol.h
danewalton: I'm not so sure about this one. I can't picture what this would be used for and it might be trying to abstract too much from the user (sort of creating another API layer).

sdk/samples/iot/pnp/pnp_thermostat_component.h
momuno: i have changed back.  I did this change while I waiting to hear back from you on an earlier comment.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
momuno: I think that convo was more general in nature.  You can still have instances where you specifically want an address to be const.  This, however, was from my misunderstanding of an earlier comment you wrote.  I have changed it back.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -595,77 +658,142 @@ static void disconnect_mqtt_client_from_iot_hub(void)
   MQTTClient_destroy(&mqtt_client);
 }
 
-static void mqtt_publish_message(const char* topic, az_span payload, int qos)
+static void set_publish_topic(
```
danewalton, line 661: I'd still rather have these `_get_publish_topic` calls in whatever function is calling this to show the explicit API calls.

momuno: Then it seems we should also discard the publish_mqtt_message wrapper function which abstracts the explicit API call to MQTT.  Unless you are differentiating between MQTT and IoT Hub?  However, the call to pnp_get_telemetry_publish_topic is also an abstraction of the API.  So I'm not clear why we abstract certain IoT Hub calls and not others.

ahsonkhan:
Please wait to merge this until after we have a preview 5 release.

danewalton (approved):
lgtm other than Ahson's request to hold until the release

PR #1225, opened 2020-09-05
Small Dox correction: verbatim semver.org

antkmsft:
A proposal, but I think we should verbatim the site-name.
Plus I capitalized HTTP in one instance.

### DIFF
--- a/sdk/inc/azure/core/az_version.h
+++ b/sdk/inc/azure/core/az_version.h
@@ -15,7 +15,7 @@
 #ifndef _az_VERSION_H
 #define _az_VERSION_H
 
-/// The version in string format used for telemetry following the semver.org standard
+/// The version in string format used for telemetry following the `semver.org` standard
 /// (https://semver.org).
 #define AZ_SDK_VERSION_STRING "1.0.0-preview.5"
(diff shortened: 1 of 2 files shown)


sdk/inc/azure/core/az_version.h
```diff
@@ -15,7 +15,7 @@
 #ifndef _az_VERSION_H
 #define _az_VERSION_H
 
-/// The version in string format used for telemetry following the semver.org standard
+/// The version in string format used for telemetry following the `semver.org` standard
```
ahsonkhan, line 18: >  but I think we should verbatim the site-name.

Can you explain this to me. How was it not `verbatim` previously, or do you mean just add the back-ticks?

ahsonkhan (approved):

ahsonkhan:
Please wait to merge this until after preview5 is released.

sdk/inc/azure/core/az_version.h
antkmsft: Just adding the backticks, because "semver.org" is not a name itself, it is a place where the info can be found.

PR #1232, opened 2020-09-08
add further info for device id encoding

### LINKED ISSUE #1231: IoT hub client init
Add documentation for the DeviceID format

danewalton:
Closes #1231

### DIFF
--- a/sdk/inc/azure/iot/az_iot_hub_client.h
+++ b/sdk/inc/azure/iot/az_iot_hub_client.h
@@ -75,7 +75,12 @@ AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default();
  *
  * @param[out] client The #az_iot_hub_client to use for this call.
  * @param[in] iot_hub_hostname The IoT Hub Hostname.
- * @param[in] device_id The Device ID.
+ * @param[in] device_id The Device ID. If the ID contains any of the following characters, they must
+ * be percent-encoded as follows:
+ *         - `/` : `%2F`
+ *         - `%` : `%25`
+ *         - `#` : `%23`
+ *         - `&` : `%26`
  * @param[in] options A reference to an #az_iot_hub_client_options structure. If `NULL` is passed,
  * the hub client will use the default options. If using custom options, please initialize first by
  * calling az_iot_hub_client_options_default() and then populating relevant options with your own


ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -75,7 +75,12 @@ AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default();
  *
  * @param[out] client The #az_iot_hub_client to use for this call.
  * @param[in] iot_hub_hostname The IoT Hub Hostname.
- * @param[in] device_id The Device ID.
+ * @param[in] device_id The Device ID. If the ID contains any of the following characters, they must
```
ahsonkhan, line 78: What is the **data format** of the device id? Is it base64 encoded? Is it a Guid? Is it a number?

ahsonkhan:
Please wait until we have a preview5 release before merging.

sdk/inc/azure/iot/az_iot_hub_client.h
danewalton: I'm not sure if we need to go that far into detail. I think the minute specifics of the documentation are delegated to the service side docs which detail how IoT works in general. The "device id" is a veryyy fundamental and well known construct in IoT. If we are to go to that level of documentation here, tbh it's going to required a bunchhh of work and would be a headache.

ahsonkhan: Aside from whether we want to document that or not, for my edification, can you explain what the data format is?

danewalton: Sure it's a name given to the device by the user, typically containing mostly letters but could be numbers as well. Standard "C string" encoded as UTF-8 here.

ahsonkhan: How long can it be? Can it contain any non-ASCII characters? What about other special characters like `#%/:=?[\]^`{|}`

ericwol-msft: https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry

A case-sensitive string (up to 128 characters long) of ASCII 7-bit alphanumeric characters plus certain special characters: - . + % _ # * ? ! ( ) , : = @ $ '.

ahsonkhan: Is there a link/resource where I can read up on which of these characters need to be percent encoding?

It kind-of looks like the device ID needs to be URL encoded.

PR #1240, opened 2020-09-09
Do not validate that the version contains beta, since C uses preview for the first release.

ahsonkhan:
cc @RickWinter


RickWinter (approved):

weshaggard:
Thanks for fixing this workaround should be fine until you guys GA.

PR #1242, opened 2020-09-09
Remove trailing white space and insert final new lines.

ahsonkhan:
Ran `eclint fix * */**/*` using the specified `.editorconfig` but excluded changes from `/eng/common/`.


LICENSE
```diff
@@ -18,4 +18,4 @@
     AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
     LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
     OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-    SOFTWARE.
+    SOFTWARE.
```
ahsonkhan, line 21: This is changing the final new line from CRLF to LF. Same elsewhere.

.editorconfig
```diff
@@ -0,0 +1,12 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+[*]
```
RickWinter, line 9:
```suggestion
[*]
indent_style = space
indent_size = 2
```

ahsonkhan: I was exploring this and unfortunately this setting doesn't play well with our indentation within comment blocks. If I can find a way to make that work, I will add it. Otherwise, this is already defined in the .clang-format file, so not too concerning.

ahsonkhan: These settings don't work well and have unintended side effects.
Even outside of comments, it changes spacing on ternary operators and other if-blocks, causing misalignment, which we don't want:

image
image

.vscode-config/launch.json
```diff
@@ -52,4 +52,4 @@
             "preLaunchTask": "build",
         }
     ]
-}
\ No newline at end of file
+}
```
vhvb1989, line 55: think we decided to remove this file from the repo in the past. Surprised to see it!

.vscode-config/tasks.json
```diff
@@ -7,4 +7,4 @@
             "command": "cd Debug && rm -rf * && cmake -DCMAKE_BUILD_TYPE=Debug -DUNIT_TESTING=ON -DUNIT_TESTING_MOCK_ENABLED=ON .. && make"
         }
     ]
-}
\ No newline at end of file
+}
```
vhvb1989, line 10: lauch and tasks json for vscode could be removed.
otherwise, we need to update them to correct values

vhvb1989 (approved):

.vscode-config/launch.json
ahsonkhan: These are not the default `.vscode` files.

They were added by @danewalton to help customers copy/paste them, if needed, here: https://github.com/Azure/azure-sdk-for-c/pull/968

See https://github.com/Azure/azure-sdk-for-c#vscode

.vscode-config/tasks.json
ahsonkhan: > otherwise, we need to update them to correct values

What do you mean by correct values? File content or file name/path? These are intentionally written as they are, by Dane.

.vscode-config/launch.json
danewalton: yea these are fine. they are stored in the `.vscode-config` directory and are meant to be copied over to the `.vscode` directory by the user.

PR #1243, opened 2020-09-09
Clean-up the .clang-format file and run clang-format on all .c and .h files in the repo.

ahsonkhan:
Ran the following at the root of the repo:
```bash
find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;
```


.clang-format
```diff
@@ -1,23 +1,18 @@
...
-PenaltyReturnTypeOnItsOwnLine: 300
 BinPackArguments: false
 BinPackParameters: false
-IndentCaseLabels: true
-BreakBeforeTernaryOperators: true
 BreakBeforeBinaryOperators: All
 BreakBeforeBraces: Allman
+ColumnLimit: 100
 Cpp11BracedListStyle: false
-AlignConsecutiveAssignments: false
```
RickWinter, line 20: You removed these two.  Can you elaborate on why

ahsonkhan: These values are identical to the LLVM defaults. The only values we need to specify are those where we differ.

ahsonkhan: Good question. Keen eye :)

RickWinter (approved):

PR #1248, opened 2020-09-09
Update return code structure for all samples.

momuno:
Separates receiving the return code from a function call, and the actual check of the return code.
Commenting and typo fixes.


sdk/samples/iot/paho_iot_provisioning_sas_sample.c
```diff
@@ -249,33 +239,31 @@ static void receive_device_registration_status_message(void)
...
-         != MQTTCLIENT_SUCCESS)
-        && (MQTTCLIENT_TOPICNAME_TRUNCATED != rc))
+    rc = MQTTClient_receive(mqtt_client, &topic, &topic_len, &message, MQTT_TIMEOUT_RECEIVE_MS);
+    if ((rc != MQTTCLIENT_SUCCESS) && (rc != MQTTCLIENT_TOPICNAME_TRUNCATED))
     {
       IOT_SAMPLE_LOG_ERROR("Failed to receive message: MQTTClient return code %d.", rc);
       exit(rc);
     }
-    else if (NULL == message)
+    else if (message == NULL)
```
danewalton, line 255: 👍

sdk/samples/iot/paho_iot_provisioning_sas_sample.c
```diff
@@ -249,33 +239,31 @@ static void receive_device_registration_status_message(void)
...
     {
       IOT_SAMPLE_LOG_ERROR("Receive message timeout expired: MQTTClient return code %d.", rc);
       exit(rc);
     }
-    else if (MQTTCLIENT_TOPICNAME_TRUNCATED == rc)
+    else if (rc == MQTTCLIENT_TOPICNAME_TRUNCATED)
```
danewalton, line 260: 👍

danewalton (approved):

sdk/samples/iot/iot_sample_common.c
```diff
@@ -496,8 +496,9 @@ void iot_sample_generate_sas_base64_encoded_signed_signature(
   // Decode the sas base64 encoded key to use for HMAC signing.
   char sas_decoded_key_buffer[64];
   az_span sas_decoded_key = AZ_SPAN_FROM_BUFFER(sas_decoded_key_buffer);
-  if (az_result_failed(
-          rc = decode_base64_bytes(sas_base64_encoded_key, sas_decoded_key, &sas_decoded_key)))
+
+  rc = decode_base64_bytes(sas_base64_encoded_key, sas_decoded_key, &sas_decoded_key);
```
ahsonkhan, line 500: nit consider declaring it here?
```suggestion
  int rc = decode_base64_bytes(sas_base64_encoded_key, sas_decoded_key, &sas_decoded_key);
```

sdk/samples/iot/paho_iot_hub_c2d_sample.c
```diff
@@ -82,8 +82,8 @@ static void create_and_configure_mqtt_client(void)
   int rc;
 
   // Reads in environment variables set by user for purposes of running sample.
-  if (az_result_failed(
-          rc = iot_sample_read_environment_variables(SAMPLE_TYPE, SAMPLE_NAME, &env_vars)))
+  rc = iot_sample_read_environment_variables(SAMPLE_TYPE, SAMPLE_NAME, &env_vars);
```
ahsonkhan, line 85: Same nit here/elsewhere (will stop mentioning):
```suggestion
  int rc = iot_sample_read_environment_variables(SAMPLE_TYPE, SAMPLE_NAME, &env_vars);
```

sdk/samples/iot/paho_iot_hub_c2d_sample.c
```diff
@@ -176,11 +172,9 @@ static void connect_mqtt_client_to_iot_hub(void)
 
 static void subscribe_mqtt_client_to_iot_hub_topics(void)
 {
-  int rc;
-
   // Messages received on the C2D topic will be cloud-to-device messages.
-  if ((rc = MQTTClient_subscribe(mqtt_client, AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC, 1))
-      != MQTTCLIENT_SUCCESS)
+  int rc = MQTTClient_subscribe(mqtt_client, AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC, 1);
```
ahsonkhan, line 176: Nice :)

sdk/samples/iot/paho_iot_hub_c2d_sample.c
```diff
@@ -200,10 +194,8 @@ static void receive_c2d_messages(void)
     IOT_SAMPLE_LOG(" "); // Formatting
     IOT_SAMPLE_LOG("Waiting for C2D message.\n");
 
-    if (((rc
-          = MQTTClient_receive(mqtt_client, &topic, &topic_len, &message, MQTT_TIMEOUT_RECEIVE_MS))
-         != MQTTCLIENT_SUCCESS)
-        && (rc != MQTTCLIENT_TOPICNAME_TRUNCATED))
+    rc = MQTTClient_receive(mqtt_client, &topic, &topic_len, &message, MQTT_TIMEOUT_RECEIVE_MS);
+    if ((rc != MQTTCLIENT_SUCCESS) && (rc != MQTTCLIENT_TOPICNAME_TRUNCATED))
```
ahsonkhan, line 198: Significantly cleaner, imo.

nit: Remove unnecessary brackets

```suggestion
    if (rc != MQTTCLIENT_SUCCESS && rc != MQTTCLIENT_TOPICNAME_TRUNCATED)
```

sdk/samples/iot/paho_iot_hub_methods_sample.c
```diff
@@ -217,10 +211,8 @@ static void receive_method_messages(void)
     IOT_SAMPLE_LOG(" "); // Formatting
     IOT_SAMPLE_LOG("Waiting for method request.\n");
 
-    if (((rc
-          = MQTTClient_receive(mqtt_client, &topic, &topic_len, &message, MQTT_TIMEOUT_RECEIVE_MS))
-         != MQTTCLIENT_SUCCESS)
-        && (rc != MQTTCLIENT_TOPICNAME_TRUNCATED))
+    rc = MQTTClient_receive(mqtt_client, &topic, &topic_len, &message, MQTT_TIMEOUT_RECEIVE_MS);
+    if ((rc != MQTTCLIENT_SUCCESS) && (rc != MQTTCLIENT_TOPICNAME_TRUNCATED))
```
ahsonkhan, line 215: Same nit here/elsewhere (will stop mentioning)
```suggestion
    if (rc != MQTTCLIENT_SUCCESS && rc != MQTTCLIENT_TOPICNAME_TRUNCATED)
```

ahsonkhan, line 215: Btw, why do we check for these two particular enum values as the failure condition? Maybe add a comment?

https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/_m_q_t_t_client_8h.html

sdk/samples/iot/paho_iot_hub_methods_sample.c
```diff
@@ -255,9 +247,8 @@ static void receive_method_messages(void)
 
 static void disconnect_mqtt_client_from_iot_hub(void)
 {
-  int rc;
-
-  if ((rc = MQTTClient_disconnect(mqtt_client, MQTT_TIMEOUT_DISCONNECT_MS)) != MQTTCLIENT_SUCCESS)
+  int rc = MQTTClient_disconnect(mqtt_client, MQTT_TIMEOUT_DISCONNECT_MS);
```
ahsonkhan, line 250: 👍

sdk/samples/iot/paho_iot_provisioning_sas_sample.c
```diff
@@ -94,33 +93,32 @@ static void create_and_configure_mqtt_client(void)
...
-              NULL)))
+  // Initialize the provisioning client with the provisioning global endpoint and the default
+  // connection options.
+  rc = az_iot_provisioning_client_init(
+      &provisioning_client,
+      az_span_create_from_str(mqtt_endpoint_buffer),
+      env_vars.provisioning_id_scope,
+      env_vars.provisioning_registration_id,
+      NULL);
+  if (az_result_failed(rc))
```
ahsonkhan, line 121: nit: Consider adding a new line between rc returning operation and the error check (everywhere). Your call.

```suggestion
      NULL);

  if (az_result_failed(rc))
```

sdk/samples/iot/pnp/pnp_thermostat_component.c
```diff
@@ -343,7 +337,7 @@ az_result pnp_thermostat_process_command_request(
     *out_status = AZ_IOT_STATUS_NOT_FOUND;
 
     IOT_SAMPLE_LOG_AZ_SPAN("Command not supported on Thermostat Sensor component:", command_name);
-    rc = AZ_ERROR_ITEM_NOT_FOUND; // Unsupported command or not this component's command
+    rc = AZ_ERROR_ITEM_NOT_FOUND; // Unsupported command
```
ahsonkhan, line 340: Unrelated: Is this the correct `az_result` to set here?

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -375,7 +390,8 @@ az_result pnp_process_device_twin_message(
     }
     else if (jr.token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT)
     {
-      if (az_result_failed(az_json_reader_skip_children(&jr)))
+      rc = az_json_reader_skip_children(&jr);
+      if (az_result_failed(rc))
```
ahsonkhan, line 394: Unrelated: If the result is not getting logged here OR returned, do we need to set it?

Alternatively, should the `rc` be returned instead of a custom `az_result`?

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -351,18 +364,20 @@ az_result pnp_process_device_twin_message(
...
       {
         IOT_SAMPLE_LOG_ERROR("Failed to get next token.");
         return AZ_ERROR_UNEXPECTED_CHAR;
       }
 
-      if (jr.token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT && components_ptr != NULL
-          && (az_result_succeeded(
-              is_component_in_model(property_name.slice, components_ptr, components_num, &index))))
+      rc = is_component_in_model(property_name.slice, components_ptr, components_num, &index);
+      if ((jr.token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT) && (components_ptr != NULL)
```
ahsonkhan, line 375: Why is the `(jr.token.kind == AZ_JSON_TOKEN_BEGIN_OBJECT)` happening **after** `is_component_in_model`. That seems unrelated to JSON. Should it be before?

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -319,15 +329,17 @@ az_result pnp_process_device_twin_message(
...
     IOT_SAMPLE_LOG_ERROR("Failed to get desired property.");
     return AZ_ERROR_UNEXPECTED_CHAR;
   }
 
   copy_jr = jr;
-  if (az_result_failed(json_child_token_move(&copy_jr, iot_hub_twin_desired_version))
-      || az_result_failed(az_json_token_get_int32(&(copy_jr.token), (int32_t*)&version)))
+  az_result rc1 = json_child_token_move(&copy_jr, iot_hub_twin_desired_version);
+  az_result rc2 = az_json_token_get_int32(&(copy_jr.token), (int32_t*)&version);
+  if (az_result_failed(rc1) || az_result_failed(rc2))
```
ahsonkhan, line 342: I am guessing this was previously relying on short-circuiting to do the right thing. Without that, the behavior might be broken.

It would be OK to revert this since we aren't inlining the assignment (nor do we use those locals). Up to you no how to verify the correctness here and update if necessary.

sdk/samples/iot/iot_sample_common.c
momuno: good eye

sdk/samples/iot/paho_iot_hub_c2d_sample.c
momuno: I actually find it easier to read with the brackets here, even though they are not required.

sdk/samples/iot/pnp/pnp_thermostat_component.c
momuno: I am not certain what is most appropriate here.  What do you think?

sdk/samples/iot/pnp/pnp_protocol.c
momuno: Good question.  So I ran into the same boolean comparison bug we have in the vs compiler even if the function call was embedded inside an az_result_failed() check.

For that reason, I am getting the rc code first and then passing it to az_result_failed().  I think the call as it was (before my change here) reads fine, but didn't know how far to take this restructure due to the current bug.  What do you think?

sdk/samples/iot/pnp/pnp_protocol.c
momuno: @danewalton what are your thoughts on this?  I was mixed on pulling out the rc and then checking it in situations like this where we aren't using the rc for anything else.  See above about my note on the vs compiler bug.  If we believe the bug will get addressed, then I'm good with reverting my change in this type of scenario.

sdk/samples/iot/pnp/pnp_protocol.c
momuno: @danewalton thoughts?  I will take a look at as well.

sdk/samples/iot/pnp/pnp_thermostat_component.c
danewalton: Yup it serves its purpose here.

sdk/samples/iot/paho_iot_hub_c2d_sample.c
danewalton: fan of the parens too if it means anything

sdk/samples/iot/pnp/pnp_protocol.c
danewalton: No so the marker of a component is if the property name is equal to a registered component and if its value is a JSON object (`AZ_JSON_TOKEN_BEGIN_OBJECT`). So it first checks to see if the prop name is a component and if it is, it makes sure the next token is a `AZ_JSON_TOKEN_BEGIN_OBJECT`.

sdk/samples/iot/pnp/pnp_protocol.c
danewalton: I'd say revert to keep it simple

sdk/samples/iot/iot_sample_common.c
momuno: I want to keep it at the top unintialized because it is initialized/change multiple times through the function, rather than a one time use.

sdk/samples/iot/paho_iot_hub_c2d_sample.c
momuno: same as above.

sdk/samples/iot/iot_sample_common.c
```diff
@@ -491,13 +489,14 @@ void iot_sample_generate_sas_base64_encoded_signed_signature(
 {
   IOT_SAMPLE_PRECONDITION_NOT_NULL(out_sas_base64_encoded_signed_signature);
 
-  int rc;
+  az_result rc;
```
ericwol-msft, line 492: humm `rc` is return code...but we are not returning it. exit_code?

sdk/samples/iot/paho_iot_hub_twin_sample.c
```diff
@@ -79,7 +79,7 @@ static az_result build_reported_property(
  * A property named `device_count` is supported for this sample. To send a device twin desired
  * property message, select your device's Device Twin tab in the Azure Portal of your IoT Hub. Add
  * the property `device_count` along with a corresponding value to the `desired` section of the
- * JSON. Select Save to udate the twin document and send the twin message to the device.
+ * JSON. Select Save to update the twin document and send the twin message to the device.
```
ericwol-msft, line 82:
```suggestion
 * JSON. Select 'Save' to update the twin document and send the twin message to the device.
```

ericwol-msft (approved):

sdk/samples/iot/pnp/pnp_protocol.c
ahsonkhan: To confirm on **correctness**, if `json_child_token_move` fails (returns non-AZ_OK result), is it OK to call `az_json_token_get_int32`?

sdk/samples/iot/pnp/pnp_protocol.c
ahsonkhan: > Alternatively, should the `rc` be returned instead of a custom `az_result`?

I don't know why we are explicitly returning `AZ_ERROR_UNEXPECTED_CHAR` here rather than just returning the `az_result` from the `az_json_reader_skip_children` call. Another reason that could fail is `AZ_ERROR_UNEXPECTED_END`. How do you know which error to return?

momuno: this is all being cleared up in my next PR

sdk/samples/iot/iot_sample_common.c
momuno: this is all being removed in my next PR

sdk/samples/iot/pnp/pnp_protocol.c
ahsonkhan: > it doesn't make sense to call az_json_token_get_int32 if json_child_token_move fails.

> I'd say revert to keep it simple


OK, good. So reverting this back to what it originally was, wasn't just a simplicity/readability concern, but also correctness.

It's certainly more difficult to review the correctness of samples since they don't have tests :p

PR #1252, opened 2020-09-10
Remove az_result return type for custom sample functions

momuno:
- Sample functions only return void or bool (with a few exceptions: callbacks, and structure initializers).
- IOT_SAMPLE_EXIT_IF_CALLED macro is rewritten to log and then exit instead of return (previously IOT_SAMPLE_RETURN_IF_CALLED).  

***Still under test.  I want to get this out so others can start reviewing and we can discuss it.

***Please IGNORE plethora of commits listed... I forgot to make a new branch for this work, so this is a long history compiled from previous PRs and commits.


sdk/samples/iot/iot_sample_common.h
```diff
@@ -149,19 +154,26 @@ typedef enum
 
 extern bool is_device_operational;
 
+/**
+ * @brief Initialize an #iot_sample_error_log which holds error logging.
+ *
+ * @param[out] out_error_log A pointer to a #out_error_log instance to initialize.
```
ericwol-msft, line 160: should out params be at the end of the param list?

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -30,125 +30,111 @@ static az_span const iot_hub_twin_desired_version = AZ_SPAN_LITERAL_FROM_STR("$v
...
     az_span const** components_ptr,
     int32_t components_num,
     int32_t* out_index)
 {
-  int32_t index = 0;
-
   if (az_span_size(component_name) == 0)
   {
-    return AZ_ERROR_UNEXPECTED_CHAR;
+    return false;
```
ericwol-msft, line 109: `IOT_SAMPLE_EXIT_IF_FAILED` here?

sdk/samples/iot/iot_sample_common.h
```diff
@@ -149,19 +154,26 @@ typedef enum
...
 extern bool is_device_operational;
 
+/**
+ * @brief Initialize an #iot_sample_error_log which holds error logging.
+ *
+ * @param[out] out_error_log A pointer to a #out_error_log instance to initialize.
+ * @param[in] message The message to log.
+ * @param[in] parameter The #az_span parameter to be used by the message. May be AZ_SPAN_EMPTY.
+ */
+void iot_sample_error_log_init(
```
ericwol-msft, line 164: were is this implemented? Did you miss a file?

sdk/samples/iot/iot_sample_common.h
danewalton: init functions are the exception

ericwol-msft (approved):

sdk/samples/iot/pnp/pnp_protocol.c
momuno: No, because if the component is not in the model, there is separate logic that needs to be executed upon return.

antkmsft (approved):
Thank you Mollie! I like this change. I am surprised it is only "+701 -902" in terms of lines :) It is cool what you do there with variable arg macros, using their full potential!

momuno:
> Thank you Mollie! I like this change. I am surprised it is only "+701 -902" in terms of lines :) It is cool what you do there with variable arg macros, using their full potential!

Thanks Anton!  I'm glad you were ok with the az_span size and ptr wrapper functions.  Will commit this (omce I fix the buid errors ad test), and we can continue to improve if need be.

antkmsft:
> Thanks Anton! I'm glad you were ok with the az_span size and ptr wrapper functions.
I assume, if you did it, it was needed :) I am surprized that `az_span_size/ptr(##__VA_ARGS__)` didn't work.

ahsonkhan:
@momuno - this is still a draft PR. What are your plans for this samples update? Is the goal to try to get these changes in, this week?

momuno:
> @momuno - this is still a draft PR. What are your plans for this samples update? Is the goal to try to get these changes in, this week?

Thanks for noticing that. I will change to PR.  I would love to have this is for GA.

momuno:
@antkmsft Had to make some changes to the macros.  I can explain more tomorrow.  I do think this version is cleaner, so I'm happy I needed to make the changes.

sdk/samples/iot/pnp/pnp_protocol.c
```diff
@@ -194,43 +176,47 @@ void pnp_parse_command_name(
...
-az_result pnp_build_reported_property(
+void pnp_build_reported_property(
     az_span json_buffer,
     az_span component_name,
     az_span property_name,
     pnp_append_property_callback append_callback,
     void* context,
     az_span* out_span)
 {
+  char const* log = "Failed to build `%.*s` reported property";
```
antkmsft, line 187: I see you have this pattern in multiple places, so I would suggestthe following for all occurrences:
1. Rename `log` to something like `error_msg`. But if that would result in autoformatting giving us too many line breaks, leave it as is, or use something shorter. `log` is not that bad, after all, I just think that `error_msg` is more descriptive. Up to you.
2. I would declare it as `char const* const`, so that when I am reading the code, I know you don't reassign it, so I shouldn't expect to track it being changed, as  read the code. Any place within the function, all the same meaning. But: for the SDK code I would definitely ask you to do that; here, in the sample, I leave it up to you, whether you think `char const* const` is too much of a complication for the majority of the audience, or not. For me, it certainly would be helpful, even if I wasn't one of the SDK devs :)

sdk/samples/iot/pnp/pnp_thermostat_component.c
```diff
@@ -50,66 +50,80 @@ static char command_end_time_value_buffer[32];
 // IoT Hub Telemetry Values
 static az_span const telemetry_temperature_name = AZ_SPAN_LITERAL_FROM_STR("temperature");
 
-static az_result build_command_response_payload(
```
antkmsft, line 53: This file includes az_result.h. It looks like it is no longer needed. Please check other sample files, if it is the case with them too. Probably, only the file that defines the `IOT_SAMPLE_EXIT_IF_AZ_FAILED` needs to include it; also any file that uses az_result itself, any specific value, such as AZ_OK, or the az_result_failed()/succeeded(), if there are any left.

sdk/samples/iot/pnp/pnp_protocol.c
momuno: I agree `error_msg` is much more descriptive of what it actually is.  The reason I went with `log` is because it was shorter.  Some of the lines are right on the cusp of having to use a second line if I go with anything much longer than `log`.  Intent was to save lines, making it cleaner to read. I'm split on it.

* const is definitely more descriptive.  I'll update those. Thanks.

PR #1253, opened 2020-09-10
update iot test names

### LINKED ISSUE #1167: Consider updating IoT test names that take char* mqtt ptr and size_t length rather than span.
There are a few such instances, mainly when testing preconditions.

For example, in this case, a span cannot be null. It is referring to the ptr being passed in.
`test_az_iot_hub_client_get_user_name_NULL_input_span_fails`

https://github.com/Azure/azure-sdk-for-c/blob/730e2c3ed7a2fb8952b5e8b6ba5b0abbc2ff5019/sdk/tests/iot/hub/test_az_iot_hub_client.c#L89-L100

https://github.com/Azure/azure-sdk-for-c/blob/14ca09fca5bd7785044dcbd41070baf049a09ac9/sdk/tests/iot/hub/test_az_iot_hub_client_twin.c#L72-L83

Some others:
https://github.com/Azure/azure-sdk-for-c/blob/14ca09fca5bd7785044dcbd41070baf049a09ac9/sdk/tests/iot/hub/test_az_iot_hub_client_twin.c#L380-L392

cc @danewalton

danewalton:
closes #1167

### DIFF
--- a/sdk/tests/iot/hub/test_az_iot_hub_client.c
+++ b/sdk/tests/iot/hub/test_az_iot_hub_client.c
@@ -86,7 +86,7 @@ static void test_az_iot_hub_client_get_user_name_NULL_client_fails(void** state)
       az_iot_hub_client_get_user_name(NULL, test_buf, sizeof(test_buf), &test_length));
 }
 
-static void test_az_iot_hub_client_get_user_name_NULL_input_span_fails(void** state)
+static void test_az_iot_hub_client_get_user_name_NULL_char_buf_fails(void** state)
 {
   (void)state;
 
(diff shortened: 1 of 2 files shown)


sdk/tests/iot/hub/test_az_iot_hub_client.c
```diff
@@ -122,7 +122,7 @@ static void test_az_iot_hub_client_get_client_id_NULL_client_fails(void** state)
       az_iot_hub_client_get_client_id(NULL, test_buf, sizeof(test_buf), &test_length));
 }
 
-static void test_az_iot_hub_client_get_client_id_NULL_input_span_fails(void** state)
+static void test_az_iot_hub_client_get_client_id_NULL_char_buf_fails(void** state)
 {
```
momuno, line 126: The only reason I would expand buf to buffer in the naming (here and elsewhere) is because of this note in the C guidelines: https://azure.github.io/azure-sdk/clang_design.html#clang-design-naming-concise

But I'm not too concerned about that in the test file.  Your choice :)

momuno (approved):

ahsonkhan (approved):

PR #1257, opened 2020-09-11
Invoking functions from az_noplatform fails precondition.

### LINKED ISSUE #1226: sleep() and clock() should return az_result

antkmsft:
This is variant B of the proposed behavior. Variant A: https://github.com/Azure/azure-sdk-for-c/pull/1256

Closes #1226

### DIFF
--- a/sdk/src/azure/platform/az_noplatform.c
+++ b/sdk/src/azure/platform/az_noplatform.c
@@ -2,9 +2,24 @@
 // SPDX-License-Identifier: MIT
 
 #include <azure/core/az_platform.h>
+#include <azure/core/internal/az_precondition_internal.h>
 
 #include <azure/core/_az_cfg.h>
 
-AZ_NODISCARD int64_t az_platform_clock_msec() { return 0; }
+AZ_NODISCARD int64_t az_platform_clock_msec()
+{
+  // This function does not have meaningful implementation to be called and its return value being
+  // relied upon.
+  _az_PRECONDITION(false);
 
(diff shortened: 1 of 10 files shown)


ahsonkhan (approved):
I prefer this version :)

sdk/src/azure/platform/az_noplatform.c
```diff
@@ -2,9 +2,24 @@
...
+  // relied upon.
+  _az_PRECONDITION(false);
 
-void az_platform_sleep_msec(int32_t milliseconds) { (void)milliseconds; }
+  return 0;
+}
+
+void az_platform_sleep_msec(int32_t milliseconds)
+{
+  (void)milliseconds;
```
ahsonkhan, line 20: unrelated: precondition that `milliseconds` is positive.

antkmsft: So far we say in the docs, `@remarks The behavior is undefined when \p milliseconds is a non-positive value (0 or less than 0).` That's because on some platform, it would yield to other thread, but will return control back if no other thread needs it. We don't use this feature though.

Issue #1190, opened 2020-09-01
Consider decoupling log classification and az_result enum values between Core and services like IoT, if there is no usability impact


ahsonkhan:
There is a desire to be able to decouple the enum values, so that core doesn't have dependency to non-core values, since enums are just plain integers.

One approach is to make the enum anonymous. One of the downsides is discovery of the enum values in doxygen docs. Making this change depends on us resolving the user experience concerns.

If we are going to do it, it would be best to do it before GA.

https://github.com/Azure/azure-sdk-for-c/blob/bb1b4532ea06872c7eb1194d89913deb3787277a/sdk/inc/azure/core/az_log.h#L33-L66
https://github.com/Azure/azure-sdk-for-c/blob/bb1b4532ea06872c7eb1194d89913deb3787277a/sdk/inc/azure/core/az_result.h#L49-L132

cc @JeffreyRichter, @RickWinter

PR #1275, opened 2020-09-16
Move IOT specific az_result and log classification enum values to separate headers.

ahsonkhan:
Trying some creative solutions to solve https://github.com/Azure/azure-sdk-for-c/issues/1190

### DIFF
--- a/sdk/inc/azure/iot/az_iot_common.h
+++ b/sdk/inc/azure/iot/az_iot_common.h
@@ -25,17 +25,17 @@
 #include <azure/core/_az_cfg_prefix.h>
 
 /**
- * @brief B The type represents the various success and error conditions.
+ * @brief The type represents the various success and error conditions.
  */
 typedef az_result az_iot_result;
 
 /**
- * @brief B Identifies the classifications of log messages produced by the SDK.
+ * @brief Identifies the classifications of log messages produced by the SDK.
  */
 typedef az_log_classification az_iot_log_classification;
 
 /**
- * @brief C The type represents the various success and error conditions.
+ * @brief The type represents the various success and error conditions.
  */
 enum az_iot_result
 {
@@ -48,7 +48,7 @@ enum az_iot_result
 };
 
 /**
- * @brief C Identifies the classifications of log messages produced by the SDK.
+ * @brief Identifies the classifications of log messages produced by the SDK.
  */
 enum az_iot_log_classification
 {
--- a/sdk/inc/azure/iot/az_iot_common.h
+++ b/sdk/inc/azure/iot/az_iot_common.h
@@ -15,6 +15,7 @@
 #ifndef _az_IOT_CORE_H
 #define _az_IOT_CORE_H
 
+#include <azure/core/az_log.h>
 #include <azure/core/az_result.h>
 #include <azure/core/az_span.h>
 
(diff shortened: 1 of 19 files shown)


ahsonkhan:
@JeffreyRichter - please take a look.

Also, IoT folks and upstream consumers, heads up. cc @hihigupt, @TiejunMS

danewalton:
mmmmmmm....I don't know. If someone looks up the docs for `az_result` in a given (IoT) doxygen document, it will pull up "core" `az_result` values and the IoT/any downstream libs won't be there. It will be a hunt to actually find any documentation.

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -23,6 +24,53 @@
...
+ * @brief Identifies the classifications of log messages produced specifically by the IoT clients
+ * within the SDK.
+ */
+typedef az_log_classification az_iot_log_classification;
+
+/**
+ * @brief The type represents the various success and error conditions specific to the IoT clients
+ * within the SDK.
+ */
+enum az_iot_result
```
JeffreyRichter, line 43: The type name here is meaningless; so I think I'd underscore it. Alterntively, you could define const az_result AZ_ERROR_* = ... for each value so that these are all of the az_result type. Or, remove the "az_iot_result" between enum and {

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -23,6 +24,53 @@
...
+
+  /// While iterating, there are no more properties to return.
+  AZ_ERROR_IOT_END_OF_PROPERTIES = _az_RESULT_MAKE_ERROR(_az_FACILITY_IOT, 2),
+};
+
+/**
+ * @brief Identifies the classifications of log messages produced specifically by the IoT clients
+ * within the SDK.
+ */
+enum az_iot_log_classification
```
JeffreyRichter, line 57: Same thing here

sdk/src/azure/iot/az_iot_common.c
```diff
@@ -119,7 +119,7 @@ AZ_NODISCARD az_result az_iot_message_properties_next(
   {
     *out_name = AZ_SPAN_EMPTY;
     *out_value = AZ_SPAN_EMPTY;
-    return AZ_ERROR_IOT_END_OF_PROPERTIES;
+    return (az_result)AZ_ERROR_IOT_END_OF_PROPERTIES;
```
JeffreyRichter, line 122: Is this cast required? If so, that is very unfortunate and requires discussion.

sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -233,20 +233,21 @@ static void test_az_iot_calculate_retry_delay_overflow_time_success()
 static int _log_retry = 0;
 static void _log_listener(az_log_classification classification, az_span message)
 {
-  switch (classification)
+  if (classification == (az_log_classification)AZ_LOG_IOT_RETRY)
```
JeffreyRichter, line 236: Is this cast required?

JeffreyRichter:
> mmmmmmm....I don't know. If someone looks up the docs for `az_result` in a given (IoT) doxygen document, it will pull up "core" `az_result` values and the IoT/any downstream libs won't be there. It will be a hunt to actually find any documentation.
> 
> Why is decoupling really necessary here? Is it mainly visually? Because core doesn't need downstream libs to compile itself?

I agree with you Dane. However, the thinking here is that this is the right architecture and architectures have to last for years and (we hope) that we can fix the documentation to get what we want. Also, I don't think customers will scan the docs for these and will instead read the docs for the function which will indicate what results come back.

ericwol-msft:
Are these changes really necessary? Considering we are TWO days away from GA, this is necessary code refactoring. We already did our bug bash yesterday. I don't want regressions at this stage of the game. What is the bug bar here???

ericwol-msft (changes requested):
unneeded code changes before GA

sdk/src/azure/iot/az_iot_common.c
ahsonkhan: Yes, with the current solution, the casts are required. I'll follow up and discuss.

ahsonkhan:
Thanks for the feedback @ericwol-msft and @danewalton. Though the goal/intent of the change is valuable, I am going to re-evaluate the approach and get back to you.

ahsonkhan:
Closing in favor of other approaches:
https://github.com/Azure/azure-sdk-for-c/pull/1295
https://github.com/Azure/azure-sdk-for-c/pull/1297

PR #1277, opened 2020-09-17
Proposal: assert(false) in the default precondition failed handler

antkmsft:
Currently, if precondition fails, your program ends up spinning in the endless loop. Maybe it is the only thing that can be done on an IoT device, I think that if we also do `assert(false)`, nothing will change for the IoT device, but on the desktop systems at least, you will break into debugger/program will abort with console message immediately - no need to wait until you figure out it is not doing anything, plus if you break into debugger, you get callstack. And if our unit tests fail, you get results immediately instead of being terminated by timeout at an unknown place, especially useful in CI/CD, I think.

Note that I am not even doing `assert(!"Precondition failed!")`, or try to include file and line information: so that it does not contribute as much to the binary size (all these would be strings need to be stored somewhere in the binary).

Update: to make would be message on assert failure even smaller, I changed `assert(false)` to `assert(0)`, so if the implementation choses to store "false" or "0" as a message, "0" is shorter, but carries the same information (little).

### DIFF
--- a/sdk/src/azure/core/az_precondition.c
+++ b/sdk/src/azure/core/az_precondition.c
@@ -10,7 +10,7 @@
 
 static void az_precondition_failed_default()
 {
-  assert(false);
+  assert(0);
   /* By default, when a precondition fails the calling thread spins forever */
   while (1)
   {
--- a/sdk/src/azure/core/az_precondition.c
+++ b/sdk/src/azure/core/az_precondition.c
@@ -2,12 +2,15 @@
 // SPDX-License-Identifier: MIT
 
 #include <azure/core/internal/az_precondition_internal.h>
+
+#include <assert.h>
 #include <stdint.h>
 
 #include <azure/core/_az_cfg.h>
 
 static void az_precondition_failed_default()
 {
+  assert(false);
   /* By default, when a precondition fails the calling thread spins forever */
   while (1)
   {
--- a/sdk/src/azure/core/az_precondition.c
+++ b/sdk/src/azure/core/az_precondition.c
@@ -10,7 +10,9 @@
 
 static void az_precondition_failed_default()
 {
+  // When assert() fails, some platforms would break into debugger, print callstack info, etc.
   assert(0);
+
   /* By default, when a precondition fails the calling thread spins forever */
   while (1)
   {


antkmsft:
@RickWinter , you told earlier today that it was not very obvious when the precondition failed at some point. With this, I think, it will be more obvious when precondition fails, and easier to debug.

ahsonkhan:
I like this change! 

> Note that I am not even doing `assert(!"Precondition failed!")`, or try to include file and line information: so that it does not contribute as much to the binary size (all these would be strings need to be stored somewhere in the binary).

I think default diagnosability trumps code size here given it is a precondition, that can be turned off, and is the default callback, which can be changed by the user, if they care.

So, if we can put a meaningful message like "precondition failed", that would be better.

Even with assert(0), in VS, I get an unhandled exception when the test fails. Can we do anything to improve that experience?
image

sdk/src/azure/core/az_precondition.c
```diff
@@ -2,12 +2,15 @@
 // SPDX-License-Identifier: MIT
 
 #include <azure/core/internal/az_precondition_internal.h>
+
+#include <assert.h>
```
ahsonkhan, line 6: Is `assert.h` always available, even on IoT devices (a no-op is fine) and C99 compatible?

sdk/src/azure/core/az_precondition.c
```diff
@@ -2,12 +2,15 @@
...
 #include <stdint.h>
 
 #include <azure/core/_az_cfg.h>
 
 static void az_precondition_failed_default()
 {
+  assert(0);
```
ahsonkhan, line 13: nit: Consider adding a comment.

ahsonkhan (approved):

sdk/src/azure/core/az_precondition.c
antkmsft: It is part of the standard, and it is a macro. Nothing prevents the platform compiler to define it as no-op.

antkmsft:
> I like this change!
> 
> > Note that I am not even doing `assert(!"Precondition failed!")`, or try to include file and line information: so that it does not contribute as much to the binary size (all these would be strings need to be stored somewhere in the binary).
> 
> I think default diagnosability trumps code size here given it is a precondition, that can be turned off, and is the default callback, which can be changed by the user, if they care.
> 
 I would collect more feedback from the team, but including "Precondition failed!" literal is not as bad as including every file and line literal, for all the places where precondition is used. But I think `assert(0)` is a significant step forward from the plain infinite loop.

antkmsft:
> Even with assert(0), in VS, I get an unhandled exception when the test fails. Can we do anything to improve that experience?
image

I don't think it is related to a failed precondition. It might be a bug in the test macro, the way it jumps back. It uses setjmp an it looks like trying to jump to a non-aligned address.

antkmsft:
@ericwol-msft , @danewalton , @momuno , @ewertons , @hihigupt - what do you guys think about this change, especially in the context of building for IoT devices?

ahsonkhan (changes requested):
Based on offline discussions, please wait to merge this after post-GA.

momuno:
Thanks for digging into this @antkmsft.  I had noticed it earlier on and thought it was peculiar but must have gotten distracted with samples...  For the larger desktop systems, I like this change.  For embedded devices, I don't know enough about the space to understand why a forever while-loop would be beneficial over an assert(false) (and the other way around).  @hihigupt what do you think?

danewalton (changes requested):
The `assert` as is in the arm gcc toolchain could cause many programs to blow up in size. It requires the line number, filepath, function name, and expression being tested in the debug build as text. In some cases, it can cause the binary to increase in size by 10x purely by those values needing a place to live.

If we were to do this, it would require us to roll our own version of an `assert` which would likely add to the PAL. We would have to give an easy way for users to sub in their own version since often times, people will create their own version of an assert to have high debuggability with low code size.

I'm blocking mainly because IF we were to do this, the standard `assert` is not the way to go.

ahsonkhan:
Thank you @danewalton from bringing up the code size concern. If we can solve that concern, simply, let's try it.

However, let's make sure to keep the context in mind. This is about preconditions, which themselves can be turned off (particularly for code-size sensitive deployments), OR, the callback can be overridden, with a custom implementation without asserts or a cheaper assert.

All this change is doing is making the default implementation more user-friendly regarding debugging and diagnosing.

Let me do some quick measurements.

ahsonkhan:
Where's the 10x coming from?

Optimized, with preconditions, in IAR.

**Before:**
Total - 1'284

**After:**
Total - 1'312

Used this sample code just as a demo. We can try and measure, other, more representative code samples, if desired.
```C
static az_span request_url
    = AZ_SPAN_LITERAL_FROM_STR("https://antk-keyvault.vault.azure.net/secrets/Password");
int A(void)
{
  uint8_t buf[100];
  uint8_t header_buf[(2 * sizeof(az_pair))];
  memset(buf, 0, sizeof(buf));
  memset(header_buf, 0, sizeof(header_buf));
  
  az_span url_span = AZ_SPAN_FROM_BUFFER(buf);
  az_span remainder = az_span_copy(url_span, request_url);
  az_span header_span = AZ_SPAN_FROM_BUFFER(header_buf);

  az_http_request request;
  az_result result = az_http_request_init(
      &request,
      &az_context_application,
      az_http_method_get(),
      url_span,
      az_span_size(request_url),
      header_span,
      AZ_SPAN_FROM_STR("body"));
    
  result = az_http_request_append_path(&request, AZ_SPAN_FROM_STR("path"), false);
  printf("%d\n", result);
  return 0;
}
```

ahsonkhan (approved):

danewalton:
> Where's the 10x coming from? 

https://robitzki.de/blog/ for example.

ahsonkhan:
> https://robitzki.de/blog/ for example. I meant that `assert()` in practice can cause an inflation in code size. Not necessarily that *ours* would.

I wonder if C99 has any differences compared to C++ on this topic.

Regardless, if you use the same compiler optimization flags (like -Og), for an apples-to-apples comparison, there is **not** a lot of difference:
image

antkmsft:
`assert()` does insert information about file and line - that is true, but the only file it is used in is `az_precondition.c, line 14` ,regardless of how often AZ_PRECONDITION is used. So, the comparison is not correct, and the size will not grow to the extant shown above.

antkmsft:
This is not getting enough traction, although I don't think that the size concern applies to this specific assert. I am closing this for now, until there's a request to improve this experience.

ahsonkhan:
cc @CIPop

ahsonkhan:
> until there's a request to improve this experience.

There is such a request. See https://github.com/Azure/azure-sdk-for-c/issues/1421#issuecomment-715623397

@antkmsft - can we revive this PR?

CIPop:
I personally like assert: they are part of the C99 standard and are properly handled on most platforms.

I propose we solve all concerns by:
- adding a cmake option to enable/disable the assert function.
- allowing preconditions to either loop or assert (or anything else like save the state then reboot + report to cloud).

danewalton:
Yea I'm okay with that. Options are always a good thing :)

PR #1278, opened 2020-09-17
Add script that validates all *.c and *.h files in the repo have been properly formatted according to .clang-format

ahsonkhan:
If this CI validation fails, please run clang-format to fix the issue.
> find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -359,6 +359,25 @@ jobs:
...
+        echo Some files were not formatted correctly according to the .clang-format file.
+        echo Please run clang-format to fix the issue.
+        exit 1
+      fi
+      
+      echo Success, all files are formatted correctly according to the .clang-format file.
+      exit 0
+      
+    displayName: Validate Clang Format
+    condition: eq(variables['AZ_SDK_CODE_COV'], 1)
```
RickWinter, line 379: Nit, Its a little odd we condition this and the utf bom check to run only if a code coverage build.

ahsonkhan: I agree. It's a simple way to run the checks on linux, but only on one leg/once. That leg takes < 5 minutes to run, and is one of the quickest, which makes it a good candidate. It's a stop gap.

antkmsft (approved):

antkmsft (commented):
Ideally, there should be a way to override this while the PR is not merged, but if there's no such an option, I'd take this one as is.

ahsonkhan:
> there should be a way to override this while the PR is not merged

Can you explain in what scenario that would be helpful? Please feel free to open a backlog, engsys, issue.

PR #1280, opened 2020-09-17
Proposal: az_platform_sleep_msec(): _az_PRECONDITION(milliseconds > 0)

antkmsft:


ahsonkhan (approved):

sdk/src/azure/platform/az_noplatform.c
```diff
@@ -15,6 +15,7 @@ AZ_NODISCARD az_result az_platform_clock_msec(int64_t* out_clock_msec)
 
 AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
 {
+  _az_PRECONDITION(milliseconds > 0);
```
JeffreyRichter, line 18: If you want this, then you want _az_PRECONDITION(milliseconds >= 0);
0 is valid here; it has a valid side-affect. 
See https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep and search for "zero"

antkmsft: But on POSIX it does not have the same effect. So, we should either use the subset of values behaving the same on both platforms, or not to go with this change - i.e. no precondition, and documentation stating that the behavior for values<= 0 is not defined, each implementation may treat it as it likes.
I am leaning towards using a subset, i.e. values that are >0, but there is nothing in the middle for me, so if you think we should allow 0, we probably should just not commit this change, and it is fine, this PR is a proposal.

sdk/inc/azure/core/az_platform.h
```diff
@@ -43,8 +43,7 @@ AZ_NODISCARD az_result az_platform_clock_msec(int64_t* out_clock_msec);
  *
  * @param[in] milliseconds Number of milliseconds to sleep.
  *
- * @remarks The behavior is undefined when \p milliseconds is a non-positive value (0 or less than
- * 0).
+ * @remarks Precondition: \p milliseconds should be greater than 0.
```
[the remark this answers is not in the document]
antkmsft: I'll revert it to previous wording.

sdk/src/azure/platform/az_noplatform.c
ahsonkhan: I am leaning towards using a subset, i.e. values that are >0

antkmsft: https://pubs.opengroup.org/onlinepubs/7908799/xsh/usleep.html : "If the value of useconds is 0, then the call has no effect."
On the other hand, it also says that "The useconds argument must be less than 1,000,000. " But it is the original documentation, and I know that on Ubuntu you can sleep for more than 1 million microseconds.

So, after all, maybe it is not possible to make a guaranteed subset that would guaranteed work across all implementations.

I would not go with this PR then.

ahsonkhan: Implementation specific preconditions are fine. The user can create their own implementation with any custom behavior too. 

If we want to strengthen the contract in some way that applies to all implementation, we should provide that guidance in the docs.

Issue #1286, opened 2020-09-17
The user experience with consuming and using our headers in includes should be simpler.


ahsonkhan:
Currently, the user has to include several granular headers to get specific functionality, and there isn't an easy way for them to get all of core or all of IoT.

Those, that level of granularity is helpful in some cases, in others, folks just want to include `az_core.h`, `az_iot.h`.

This is similar to consuming a single Core or Service package in other languages.

PR #1289, opened 2020-09-17
Add high-level az_core.h and az_iot.h files for simpler include experience for customers.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1286


sdk/samples/storage/blobs/src/blobs_client_example.c
```diff
@@ -17,13 +17,12 @@
  * 4) Get payload from response and parse it
  */
 
-#include <azure/core/az_context.h>
+#include <azure/az_core.h>
```
ahsonkhan, line 20: Open question: Should these headers be in `/inc/azure/core/az_core.h` or is it better for them to be one-level higher, in `/inc/azure/az_core.h`.

**Either:**
/inc/azure/core/az_core.h
/inc/azure/iot/az_iot.h

**OR:**
/inc/azure/az_core.h
/inc/azure/az_iot.h

I prefer the latter.

sdk/inc/azure/az_iot.h
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ahsonkhan, line 1: @momuno, once this is merged, can you please see if our IoT samples could benefit from this?

danewalton (changes requested):
Similar to the result one, I don't think this should really be merged before GA. It changes the way we tell people to consume our libraries and docs and samples would have to be updated to reflect that.

sdk/inc/azure/az_iot.h
momuno: These merges may break the sample functionality, so it would be best for the samples to get updated and tested before anything is merged.  I do not have time to do this currently, so can it please wait until after GA? Thanks.

ahsonkhan: Updates to IoT samples can be made post-GA.

This PR itself does not break any samples though and only provides customers a better experience, if they choose it.

ahsonkhan:
>  docs and samples would have to be updated to reflect that.

I agree that the docs/samples updates can be made post-GA. 

However, I think it is really valuable for the end customer experience to have this particular change in now. It is purely additive.

danewalton:
> However, I think it is really valuable for the end customer experience to have this particular change in now. It is purely additive.

I will counter with the fact that if it is additive, then it does no harm to add it later (it doesn't break anything). I agree there may be a good use for it, but I would rather wait until we support it and document it for people to use. Customers have already used the headers as is and haven't had issues, so not having this for GA I don't believe is a loss. 

I do thank you for doing the work regardless.

JeffreyRichter:
I agree that these are nice to have and I also agree with Dane that they can waituntil after GA.

antkmsft:
I would recommend to update IoT and storage samples to use these headers in this PR - to follow the principle that if we add something, it is used somewhere, add ref count from the get go, so to speak :) . Up to you, we don't have to, just a recommendation.

sdk/samples/storage/blobs/src/blobs_client_example.c
danewalton: vote for latter

danewalton (approved):

Issue #1276, opened 2020-09-17
az_json_writer: IAR warning


hihigupt:
Hi,

When compiling az_json_writer.c using IAR. I am getting following warning, can we please fix this warning for IAR.

az_json_writer.c  
iccarm.exe <....>\azure_iot\azure-sdk-for-c\sdk\src\azure\core\az_json_writer.c -D TX_ENABLE_EXECUTION_CHANGE_NOTIFY -D AZ_NO_PRECONDITION_CHECKING -D COLLECTOR_HEARTBEAT_ENABLED -D COLLECTOR_SYSTEM_INFORMATION_ENABLED -D  
COLLECTOR_NETWORK_ACTIVITY_ENABLED -o C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\Debug\Obj --no_cse --no_unroll --no_inline --no_code_motion --no_tbaa --no_clustering --no_scheduling --debug --endian=little --cpu=Cortex-M7 -e --fpu=VFPv5_sp  
--dlib_config C:\Program Files (x86)\IAR Systems\Embedded Workbench 8.4\arm\inc\c\DLib_Config_Full.h -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\tx\ -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\..\..\..\..\..\..\ 
..\externals\threadx\common\inc\ -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\..\..\..\..\..\..\..\externals\threadx\ports\cortex_m7\iar\inc\ -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\nxd\ -I C:\gitroot\iot_8\netx\Test\ 
Regression\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\..\..\..\..\..\..\..\netxduo\common\inc\ -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\azure-iot-sdk-c\..\..\..\..\..\..\..\..\addons\cloud\ -I C<....>\azure_iot\project\st\stm32f746g-disco\iar\ 
[...]

PR #1304, opened 2020-09-18
Reorder fields within az_json_reader for better packing and to avoid IAR warning for int assignment to an enum.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1276

This is the line in question causing a warning in IAR:
https://github.com/Azure/azure-sdk-for-c/blob/183b7932ff74fbc7083aa602434f6700ddd65979/sdk/src/azure/core/az_json_writer.c#L724

We use this pattern elsewhere too, but for the `az_json_reader` all other instances are in tests.

Discussion about the IAR issue: https://stackoverflow.com/a/27138813

cc @hihigupt, @TiejunMS

### DIFF
--- a/sdk/inc/azure/core/az_json.h
+++ b/sdk/inc/azure/core/az_json.h
@@ -67,10 +67,6 @@ typedef struct
  */
 typedef struct
 {
-  /// This read-only field gives access to the type of the token returned by the #az_json_reader,
-  /// and it shouldn't be modified by the caller.
-  az_json_token_kind kind;
-
   /// This read-only field gives access to the slice of the JSON text that represents the token
   /// value, and it shouldn't be modified by the caller.
   /// If the token straddles non-contiguous buffers, this is set to the partial token value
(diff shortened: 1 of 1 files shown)


antkmsft (approved):

sdk/inc/azure/core/az_json.h
```diff
@@ -80,6 +76,13 @@ typedef struct
   /// In the case of JSON strings, the slice does not include the surrounding quotes.
   az_span slice;
 
+  // Avoid using enum as the first field within structs, to allow for { 0 } initialization.
+  // This is a workaround for IAR compiler warning [Pe188]: enumerated type mixed with another type.
+
+  /// This read-only field gives access to the type of the token returned by the #az_json_reader,
+  /// and it shouldn't be modified by the caller.
+  az_json_token_kind kind;
+
```
momuno, line 85: I want to learn more about packing and padding. I definitely see why it's better where you moved the enum (1 byte?) than where it was, however, why not move it below the int32_t size?  The next data type is a struct that begins with two bools (2 bytes total?).  Would the compiler see the enum next to the two bools and only pad 1 more byte? or is there a cut-off at the struct, so it doesn't matter?  This is probably easier to discuss in a call.

ahsonkhan: Sure, we can chat. I believe enums in C are `int`-backed, so 4 bytes.

To clarify, I didn't re-order them for maximum struct packing, but we could.

hihigupt: When initializing a struct, the first initializer in the list initializes the first declared member and omitting fields in initialization  implicitly initialized the same as for objects that have static storage duration.

This above principle is getting used here to initialize the struct with zero, as you can see in line azure-sdk-for-c/sdk/src/azure/core/az_json_writer.c:774. 

Now, in this (az_json_writer.c:774) initialization we are initializing enum with 0, which is causing issue in compilers (like IAR) that do not want INT to be mixed with enum. Now with shuffling the fields, we are avoiding the warning.

PR #1308, opened 2020-09-18
Docs: More details about az_log_set_classifications()

### LINKED ISSUE #1192: Make sure to clearly document the behavior and supported scenario for log classifications
We need explicit docs that state:
- The user cannot and should not modify the array after passing it to `az_log_set_classifications`
- The user must make sure to end the array with a terminator.
- What does passing `NULL` or an empty array mean, and why it is distinct.
- ~User-defined classifications are supported.~

cc @JeffreyRichter, @RickWinter

### LINKED ISSUE #1227: It is not possible to put az_log_classifications on a stack and expect reliable behavior in multithreaded app.
```
void customer_func()
{
    az_log_classification cls[] = { ... }
    az_log_set_classifications(cls);
    ...
    az_log_set_classifications(NULL); // it STILL may not be safe to return from customer_func(), because another thread may still be logging => iterating through cls allocated in this stack.
}
```

A fix would be to either mutex/spinlock classification array pointer inside logging.c, or to have a "copy" of that array - which doesn't have to be the exact copy, but can have essentially a map from logging classification -> to bool whether it is enabled. The downside in the second case is that core needs to be updated accordingly to contain all possible classifications, but a small additional benefit is that on every log attempt we try to find out whether a classification is enabled, currently it is O(N/2) on average, it probably will be O(logN) on average.

antkmsft:
Closes #1227.
Closes #1192.

I don't think there is much more we can do for #1227.  We still can - maybe we can have a local copy of an array of some finite size, and copy elements there, but besides that I don't see much options. az_log_set_classifications() would need to return az_result in this case, if the limit is exceeded. Then, if you want more than the predefined limit, one of the options is to pass NULL to receive all classifications, and filter them out in log callback, which does receive classification in addition to message.

I had two solutions in mind:
1. Use mutex, and lock it when setting classifications, and while iterating over them. I do not think it is possible, given that az_log is used by IoT, and mutex is in PAL, and we don't want the dependency on PAL for the IoT scenarios.

2. If we know all the possible classifications, make some sort of a map of `classification -> bool is_enabled`, and use that internal copy; this would also have some perf benefits as side effect. But given that we now extracted IoT classifications from core into a file in the iot directory, I do not see a realistic good and simple way of doing this anymore.

### DIFF
--- a/sdk/docs/core/README.md
+++ b/sdk/docs/core/README.md
@@ -106,10 +106,11 @@ Log classifications allow your application to select which specific log messages
       printf("%.*s\n", az_span_size(message), az_span_ptr(message));
    }
 
+   az_log_classification const log_classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
+
    int main()
    {
-      az_log_classification const classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
-      az_log_set_classifications(classifications);
+      az_log_set_classifications(log_classifications);
       az_log_set_callback(test_log_func);
 
       // More code goes here...
--- a/sdk/docs/core/README.md
+++ b/sdk/docs/core/README.md
@@ -101,12 +101,12 @@ Now, whenever our SDK wants to send a log message, it will invoke your callback 
 Log classifications allow your application to select which specific log messages it wants to receive. Here is a complete example that logs HTTP request and response messages to standard output:
 
    ```C
-   void test_log_func(az_log_classification classification, az_span message)
+   static void test_log_func(az_log_classification classification, az_span message)
    {
       printf("%.*s\n", az_span_size(message), az_span_ptr(message));
    }
 
-   az_log_classification const log_classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
+   static az_log_classification const log_classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
 
    int main()
    {
--- a/sdk/inc/azure/core/az_log.h
+++ b/sdk/inc/azure/core/az_log.h
@@ -83,11 +83,11 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span 
  * receive log messages for all #az_log_classification values.
  * @details If \p classifications is not `NULL`, it must point to an array of
  * #az_log_classification, terminated by #AZ_LOG_END_OF_LIST.
- * @details In constrast to \p classifications being `NULL`, \p classifications pointing to an empty
- * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an aplication is not
- * interested in receiving any message, regardless of classification.
+ * @details In contrast to \p classifications being `NULL`, \p classifications pointing to an empty
+ * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an application is
+ * not interested in receiving any log messages.
  *
- * @warning Users should not change the \p classifications array elements, once it is passed to this
+ * @warning Users must not change the \p classifications array elements, once it is passed to this
  * function. If \p classifications array is allocated on a stack, program behavior in multi-threaded
  * environment is undefined.
  */
--- a/sdk/inc/azure/core/az_log.h
+++ b/sdk/inc/azure/core/az_log.h
@@ -79,11 +79,11 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span 
  * @param[in] classifications __[nullable]__ An array of #az_log_classification values, terminated
  * by #AZ_LOG_END_OF_LIST.
  *
- * @detail If no classifications are set (\p classifications is `NULL`), the application will
+ * @details If no classifications are set (\p classifications is `NULL`), the application will
  * receive log messages for all #az_log_classification values.
- * @detail If \p classifications is not `NULL`, it must point to an array of #az_log_classification,
- * terminated by #AZ_LOG_END_OF_LIST.
- * @detail In constrast to \p classifications being `NULL`, \p classifications pointing to an empty
+ * @details If \p classifications is not `NULL`, it must point to an array of
+ * #az_log_classification, terminated by #AZ_LOG_END_OF_LIST.
+ * @details In constrast to \p classifications being `NULL`, \p classifications pointing to an empty
  * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an aplication is not
  * interested in receiving any message, regardless of classification.
  *
--- a/sdk/inc/azure/core/az_log.h
+++ b/sdk/inc/azure/core/az_log.h
@@ -78,7 +78,7 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span 
  *
  * @param[in] classifications __[nullable]__ An array of #az_log_classification values, terminated
  * by #AZ_LOG_END_OF_LIST.
- * 
+ *
  * @detail If no classifications are set (\p classifications is `NULL`), the application will
  * receive log messages for all #az_log_classification values.
  * @detail If \p classifications is not `NULL`, it must point to an array of #az_log_classification,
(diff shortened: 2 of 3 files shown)


sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
- * @details If no classifications are set (`NULL`), the application will receive log messages for
- * all #az_log_classification values.
+ * @param[in] classifications __[nullable]__ An array of #az_log_classification values, terminated
+ * by #AZ_LOG_END_OF_LIST.
  *
- * @param[in] classifications __[nullable]__ An array of az_log_classification values, terminated by
- * #AZ_LOG_END_OF_LIST.
+ * @detail If no classifications are set (\p classifications is `NULL`), the application will
+ * receive log messages for all #az_log_classification values.
+ * @detail If \p classifications is not `NULL`, it must point to an array of #az_log_classification,
```
ahsonkhan, line 84: This is what CI is complaining about:
```suggestion
 * @details If \p classifications is not `NULL`, it must point to an array of #az_log_classification,
```

vhvb1989 (approved):

sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
+ * @param[in] classifications __[nullable]__ An array of #az_log_classification values, terminated
+ * by #AZ_LOG_END_OF_LIST.
  *
- * @param[in] classifications __[nullable]__ An array of az_log_classification values, terminated by
- * #AZ_LOG_END_OF_LIST.
+ * @details If no classifications are set (\p classifications is `NULL`), the application will
+ * receive log messages for all #az_log_classification values.
+ * @details If \p classifications is not `NULL`, it must point to an array of
+ * #az_log_classification, terminated by #AZ_LOG_END_OF_LIST.
+ * @details In constrast to \p classifications being `NULL`, \p classifications pointing to an empty
```
RickWinter, line 86:
```suggestion
 * @details In contrast to \p classifications being `NULL`, \p classifications pointing to an empty
```

sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
+ * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an aplication is not
+ * interested in receiving any message, regardless of classification.
```
RickWinter, line 88:
```suggestion
 * interested in receiving any log messages.
```

sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
+ *
+ * @warning Users should not change the \p classifications array elements, once it is passed to this
```
RickWinter, line 90:
```suggestion
 * @warning Users must not change the \p classifications array elements, once it is passed to this
```

RickWinter (approved):
Few minor tweaks

ahsonkhan (approved):

sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
+ * function. If \p classifications array is allocated on a stack, program behavior in multi-threaded
+ * environment is undefined.
```
ahsonkhan, line 92: We should add code snippets, and maybe even some tests for such edge cases, but not critical.

sdk/inc/azure/core/az_log.h
```diff
@@ -76,11 +76,20 @@ typedef void (*az_log_message_fn)(az_log_classification classification, az_span
...
+ * by #AZ_LOG_END_OF_LIST.
  *
- * @param[in] classifications __[nullable]__ An array of az_log_classification values, terminated by
- * #AZ_LOG_END_OF_LIST.
+ * @details If no classifications are set (\p classifications is `NULL`), the application will
+ * receive log messages for all #az_log_classification values.
+ * @details If \p classifications is not `NULL`, it must point to an array of
+ * #az_log_classification, terminated by #AZ_LOG_END_OF_LIST.
+ * @details In constrast to \p classifications being `NULL`, \p classifications pointing to an empty
+ * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an aplication is not
```
ahsonkhan, line 87:
```suggestion
 * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an application is not
```

sdk/inc/azure/core/az_log.h
antkmsft: A C programmer can't spell "cont" without "s" :)

ahsonkhan: What a constroversial statement. It depends on the contsext.

sdk/inc/azure/core/az_log.h
antkmsft: Thank you Ahson, I just added update to samples to use safer pattern of declaring classifications for samples and for readme.

I so far don't know if it is possible to update the test to verify the conditions that we are documenting against, given that when that happens, the timing needs to be just precisely right; plus they are going to verify that the behavior is going to be undefined as we are telling, I am not sure what an error or success for such test case would mean.

As of the multi-threaded testing, I think eventually we should get some sort of a mechanism for testing that allows us to run multiple threads, and do so in such a controlled precision (with stopping and going), that they are guaranteed to reproduce a race condition; maybe we don't really need multiple threads for that, and single thread simulating two threads, by jumping from one location to another at precisely pre-programmed moments is better, but this is a bigger item that won't be fixed within this PR, let's revisit the whole idea as part of a bigger picture, there could be some other places (the ones that we wrote with multi-threading in mind) where we would use such mechanism to verify that the multi threaded behavior is exactly as we intended it to be, with 100% accuracy.

sdk/docs/core/README.md
```diff
@@ -106,10 +106,11 @@ Log classifications allow your application to select which specific log messages
       printf("%.*s\n", az_span_size(message), az_span_ptr(message));
    }
 
+   az_log_classification const log_classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
```
ahsonkhan, line 109: Is this allowed in C, will it compile? Should it be static?

ahsonkhan: Add a comment above this line why we are declaring it here and that it can't be a local.

sdk/samples/storage/blobs/src/blobs_client_example.c
```diff
@@ -37,6 +37,9 @@ static az_span content_to_upload = AZ_SPAN_LITERAL_FROM_STR("Some test content")
 #endif
 
 // Enable logging
+az_log_classification const log_classifications[]
```
ahsonkhan, line 40: Good catch. Add comments so someone copying code knows what they  should and shouldn't do.

ahsonkhan (approved):

sdk/docs/core/README.md
antkmsft: This is allowed in C - it is going to be a global variable, but static won't hurt and is even better (local to a translation unit). Let me fix this.

sdk/samples/storage/blobs/src/blobs_client_example.c
antkmsft: I am hesitant to draw so much of the reader's attention in that instance. We follow a good practice, and I think it is sufficient. The problem is only when we run in a multi-threaded app, but the sample does not have multiple threads. I.e. you came to see the storage sample, and right away you get a couple lines of detailed information on the multi-threaded aspect of logging.

antkmsft:
I think I might give this whole idea another shot in the next sprint, I put more details in the PR description (local array copy of a finite size + returning az_result when limit is exceeded); so maybe after all we will be able to safely handle stack-allocated arrays.

sdk/docs/core/README.md
ahsonkhan, line 109:
```suggestion
   static az_log_classification const log_classifications[] = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
```

antkmsft: Thank you Ahson, good catch, I had this change, but didn't save the changes to file before git commit. I am also adding static for the log callback function.

PR #1313, opened 2020-09-18
update API version

danewalton:
Updates the API version for the iot client to the latest.

I tested this with a current West hub and the newer canary Hub. Telemetry sample and PnP samples connect in both cases.

### DIFF
--- a/sdk/tests/iot/hub/test_az_iot_hub_client.c
+++ b/sdk/tests/iot/hub/test_az_iot_hub_client.c
@@ -34,11 +34,11 @@ static const az_span test_hub_hostname = AZ_SPAN_LITERAL_FROM_STR(TEST_HUB_HOSTN
 static const char test_correct_user_name[]
     = "myiothub.azure-devices.net/my_device/?api-version=2018-06-30&" PLATFORM_USER_AGENT;
 static const char test_correct_user_name_with_model_id[]
-    = "myiothub.azure-devices.net/my_device/?api-version=2020-05-31-preview&" PLATFORM_USER_AGENT
+    = "myiothub.azure-devices.net/my_device/?api-version=2020-09-30&" PLATFORM_USER_AGENT
       "&model-id=dtmi%3AYOUR_COMPANY_NAME_HERE%3Asample_device%3B1";
 static const char test_correct_user_name_with_model_id_with_module_id[]
     = "myiothub.azure-devices.net/my_device/my_module_id/"
-      "?api-version=2020-05-31-preview&os=azrtos&model-id=dtmi%3AYOUR_COMPANY_NAME_HERE%"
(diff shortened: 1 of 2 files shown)


jspaith (changes requested):
We cannot get rid of `hub_service_preview_api_version` yet because 2020-09-30 has not yet deployed to the world.  This round we need to update `hub_service_preview_api_version` to be the 2020-09-30 but still leave the if(modelId) stuff around.

When 2020-09-30 goes to the world then we can make the remove preview change.

sdk/tests/iot/hub/test_az_iot_hub_client.c
```diff
@@ -335,7 +335,8 @@ static void test_az_iot_hub_client_get_user_name_with_model_id_small_buffer_firs
   assert_int_equal(
       az_iot_hub_client_init(&client, test_hub_hostname, test_device_id, &options), AZ_OK);
 
-  char mqtt_topic_buf[sizeof(test_correct_user_name) + 2];
+  char mqtt_topic_buf[sizeof(test_correct_user_name) + 53]; // Plus 53 is to test the percent
```
CIPop, line 338: Is there a test that can be checked in to account for the comments?

danewalton:
> 
> 
> We cannot get rid of `hub_service_preview_api_version` yet because 2020-09-30 has not yet deployed to the world. This round we need to update `hub_service_preview_api_version` to be the 2020-09-30 but still leave the if(modelId) stuff around.
> 
> When 2020-09-30 goes to the world then we can make the remove preview change.

Yup just talked with Eric. Making the change to only modify the preview version. Misunderstood the change.

sdk/tests/iot/hub/test_az_iot_hub_client.c
```diff
@@ -335,7 +335,7 @@ static void test_az_iot_hub_client_get_user_name_with_model_id_small_buffer_firs
   assert_int_equal(
       az_iot_hub_client_init(&client, test_hub_hostname, test_device_id, &options), AZ_OK);
 
-  char mqtt_topic_buf[sizeof(test_correct_user_name) + 2];
+  char mqtt_topic_buf[sizeof(test_correct_user_name) - 1];
```
danewalton, line 338: Note: this change is due to the fact that the preview name is now shorter. This compares the user name without a model id (`test_correct_user_name`) vs with a model id (`test_correct_user_name_with_model_id`).

ericwol-msft (approved):

jspaith (approved):

sdk/tests/iot/hub/test_az_iot_hub_client.c
ahsonkhan: I don't understand. How does that result in a change of 3 fewer chars?

PR #1314, opened 2020-09-18
Add precondition check to validate clients are initialized before passed in to public APIs.

ahsonkhan:
~**TODO:**~
- ~Verify other places are good~
- ~Add doc comments.~

### DIFF
--- a/sdk/src/azure/iot/az_iot_hub_client_c2d.c
+++ b/sdk/src/azure/iot/az_iot_hub_client_c2d.c
@@ -22,6 +22,7 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
     az_iot_hub_client_c2d_request* out_request)
 {
   _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_VALID_SPAN(client->_internal.iot_hub_hostname, 1, false);
   _az_PRECONDITION_VALID_SPAN(received_topic, 1, false);
   _az_PRECONDITION_NOT_NULL(out_request);
   (void)client;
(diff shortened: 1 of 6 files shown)


JeffreyRichter (approved):

sdk/src/azure/iot/az_iot_hub_client_c2d.c
```diff
@@ -22,6 +22,7 @@ AZ_NODISCARD az_result az_iot_hub_client_c2d_parse_received_topic(
     az_iot_hub_client_c2d_request* out_request)
 {
   _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_VALID_SPAN(client->_internal.iot_hub_hostname, 1, false);
```
danewalton, line 25: just to be clear this doesn't ensure that it was initialized. but this covers the 90% case.

danewalton (approved):

sdk/src/azure/iot/az_iot_hub_client_c2d.c
ahsonkhan: Do you mean something like this?
```C
az_iot_hub_client cl;
cl._internal.iot_hub_hostname = AZ_SPAN_FROM_STR("foo");
```

The goal here is to give users a hint that they forgot a call to `az_iot_hub_client_init`. Using a single sentinel field is good enough to detect that. I don't know if it is feasible to cover more than that in a simple way.

danewalton: No I mean if the hub client is initialized on the stack or malloc'd, the `iot_hub_hostname` field could be in a state which passes this precondition.

ahsonkhan: Interesting. Can you show me a code example on what that would look like? I want to understand how likely that will be and potential side effects.

Issue #1184, opened 2020-08-31
Re-consider the design of using sentinels to mark end of arrays, such as log classification and HTTP status.


ahsonkhan:
See related issue https://github.com/Azure/azure-sdk-for-c/issues/1172

cc @JeffreyRichter, @RickWinter

ahsonkhan:
Also FYI, @antkmsft

PR #1317, opened 2020-09-19
Use a should_write callback method to let the caller decide what should and shouldn't get logged.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1172 and also addresses a big part of https://github.com/Azure/azure-sdk-for-c/issues/1184.

Follow up to https://github.com/Azure/azure-sdk-for-c/issues/1227 and https://github.com/Azure/azure-sdk-for-c/issues/1192

**Here are some benefits of this design:**
1) No lifetimes concerns with user-provided array (especially if they declared it on the stack, or want to re-use it after).
2) Reduced potential for AVs due to accidental misuse of the the API.
3) No loop-lookups during logging which will improve runtime performance and code size (120 vs 92, ~30%).
4) Minimize global state just to function pointers.
5) Gives the user the flexibility to use an allow list or a deny list.

This is a more complete and cleaner alternative to https://github.com/Azure/azure-sdk-for-c/pull/1319, https://github.com/Azure/azure-sdk-for-c/pull/1321, with fewer downsides.

~TODO:~
- ~Changelog update~
- ~README update~


sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
...
+  _az_PRECONDITION_VALID_SPAN(message, 0, true);
+
+  // Copy the volatile fields to local variables so that they don't change within this function.
+  az_log_message_fn const message_callback = _az_log_message_callback;
+
+  // If the user hasn't registered a should_write_callback, then we log everything, as long as a
+  // message_callback metho was provided.
+  if (_az_log_should_write(classification))
+  {
+    message_callback(classification, message);
```
antkmsft, line 62: Make _az_log_should_write return the az_log_message_fn, instead of bool. This is the advantage we had with the "engine" function: we check and write, and when we do that, message_callback pointer has the same value in all points.
Here, you get the callback, it is NULL, other thread sets it to non-NULL, _az_log_should_write() returns you true, and you invoke a NULL callback.

sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
...
+  return (should_write_callback == NULL || should_write_callback(classification))
+      && message_callback != NULL;
 }
 
 // This function attempts to log the passed-in message.
 void _az_log_write(az_log_classification classification, az_span message)
 {
-  (void)_az_log_write_engine(true, classification, message);
+  _az_PRECONDITION(classification > 0);
+  _az_PRECONDITION_VALID_SPAN(message, 0, true);
```
antkmsft, line 53: I still don't like the `_az_PRECONDITION(classification > 0)`, but I agree that `_az_PRECONDITION_VALID_SPAN(message, 0, true)` is rather useful, only that if we doing that check, I'd say don't allow empty and null spans, because it is likely a bug, there is no point, at this time at least, in such message.

sdk/tests/core/inc/az_test_log.h
```diff
@@ -6,10 +6,10 @@
 
 // This macro helps with the expected values for tests when verifying logging.
 
-#ifndef AZ_NO_LOGGING
-#define _az_BUILT_WITH_LOGGING(value_if_yes, value_if_no) (value_if_yes)
-#else
+#ifdef AZ_NO_LOGGING
```
antkmsft, line 9: I strongly recommend to revert it back, as I said above.

sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -244,32 +244,47 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_log_retry(az_log_classification classification)
```
antkmsft, line 247:
```suggestion
static bool _should_write_iot_retry_only(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
```diff
@@ -211,12 +211,34 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_MQTT(az_log_classification classification)
```
antkmsft, line 214:
```suggestion
static bool _should_write_everything(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
```diff
@@ -211,12 +211,34 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  {
+    case AZ_LOG_MQTT_RECEIVED_TOPIC:
+    case AZ_LOG_MQTT_RECEIVED_PAYLOAD:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_MQTT_received_payload_only(
```
antkmsft, line 226:
```suggestion
static bool _should_write_mqtt_received_payload_only(
```

sdk/tests/core/test_az_logging.c
```diff
@@ -281,19 +308,41 @@ static void _log_listener_no_op(az_log_classification classification, az_span me
   (void)message;
 }
 
+static bool _log_listener_should_write_http_retry(az_log_classification classification)
```
antkmsft, line 311:
```suggestion
static bool _should_write_http_retry_only(az_log_classification classification)
```

sdk/tests/core/test_az_logging.c
```diff
@@ -66,6 +66,30 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_everything_valid(az_log_classification classification)
```
antkmsft, line 69:
```suggestion
static bool _should_write_everything(az_log_classification classification)
```

sdk/tests/core/test_az_logging.c
```diff
@@ -66,6 +66,30 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+    case AZ_LOG_HTTP_RETRY:
+    case AZ_LOG_HTTP_RESPONSE:
+    case AZ_LOG_HTTP_REQUEST:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_request(az_log_classification classification)
```
antkmsft, line 82:
```suggestion
static bool _should_write_http_request_only(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_methods.c
```diff
@@ -381,12 +381,34 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_MQTT(az_log_classification classification)
```
antkmsft, line 384:
```suggestion
static bool _should_write_everything(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_methods.c
```diff
@@ -381,12 +381,34 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  {
+    case AZ_LOG_MQTT_RECEIVED_TOPIC:
+    case AZ_LOG_MQTT_RECEIVED_PAYLOAD:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_MQTT_received_payload_only(
```
antkmsft, line 396:
```suggestion
static bool _should_write_mqtt_received_payload_only(
```

sdk/tests/iot/hub/test_az_iot_hub_client_sas.c
```diff
@@ -438,11 +438,27 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_sas(az_log_classification classification)
```
antkmsft, line 441:
```suggestion
static bool _should_write_iot_sas_token_only(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_sas.c
```diff
@@ -438,11 +438,27 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  switch (classification)
+  {
+    case AZ_LOG_IOT_SAS_TOKEN:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_nothing(az_log_classification classification)
```
antkmsft, line 452:
```suggestion
static bool _should_write_nothing(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -355,12 +355,34 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_MQTT(az_log_classification classification)
```
antkmsft, line 358:
```suggestion
static bool _should_write_everything(az_log_classification classification)
```

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -355,12 +355,34 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  {
+    case AZ_LOG_MQTT_RECEIVED_TOPIC:
+    case AZ_LOG_MQTT_RECEIVED_PAYLOAD:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_MQTT_received_payload_only(
```
antkmsft, line 370:
```suggestion
static bool _should_write_mqtt_received_payload_only(
```

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
```diff
@@ -503,12 +503,28 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_MQTT(az_log_classification classification)
```
antkmsft, line 506:
```suggestion
static bool _should_write_everything(az_log_classification classification)
```

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
```diff
@@ -503,12 +503,28 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  {
+    case AZ_LOG_MQTT_RECEIVED_TOPIC:
+    case AZ_LOG_MQTT_RECEIVED_PAYLOAD:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_nothing(az_log_classification classification)
```
antkmsft, line 518:
```suggestion
static bool _should_write_nothing(az_log_classification classification)
```

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_sas.c
```diff
@@ -282,11 +282,27 @@ static void _log_listener(az_log_classification classification, az_span message)
...
+  switch (classification)
+  {
+    case AZ_LOG_IOT_SAS_TOKEN:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_nothing(az_log_classification classification)
```
antkmsft, line 296:
```suggestion
static bool _should_write_nothing(az_log_classification classification)
```

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_sas.c
```diff
@@ -282,11 +282,27 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _log_listener_should_write_sas(az_log_classification classification)
```
antkmsft, line 285:
```suggestion
static bool _should_write_iot_sas_token_only(az_log_classification classification)
```

sdk/tests/core/test_az_logging.c
```diff
@@ -281,19 +308,41 @@ static void _log_listener_no_op(az_log_classification classification, az_span me
...
+  switch (classification)
+  {
+    case AZ_LOG_HTTP_RETRY:
+      return true;
+    default:
+      return false;
+  }
+}
+
+static bool _log_listener_should_write_nothing(az_log_classification classification)
```
antkmsft, line 322:
```suggestion
static bool _should_write_nothing(az_log_classification classification)
```

antkmsft (commented):
Looks good, please fix the feedback, I think this can go to GA, but let's collect more feedback.
AND, also huge thing, we need to do the same callback-filter mechanism for retry: instead of status codes being terminated by `AZ_HTTP_STATUS_CODE_END_OF_LIST`, az_http_policy_retry_options should have a callback pointer. If null, we use our default list to retry. A quickest fix for GA would be to remove `az_http_status_code const* status_codes` from the `az_http_policy_retry_options` struct, then we use just the default list, and it is not customizable, until we later introduce a callback pointer as a new field in the az_http_policy_retry_options struct.

antkmsft (changes requested):
Actually, blocking the PR, just in case, you can always override with one positive vote (if you need this to go on the weekend and I am not around), but there are definitely things that should be fixed, such as possible invocation of a NULL callback fn.

ahsonkhan:
> we need to do the same callback-filter mechanism for retry

I understand. I will do that in a separate PR.

sdk/inc/azure/core/az_log.h
```diff
@@ -64,56 +61,63 @@ enum az_log_classification_core
...
- * @warning Users must not change the \p classifications array elements, once it is passed to this
- * function. If \p classifications array is allocated on a stack, program behavior in multi-threaded
- * environment is undefined.
+ * @remarks By default, this is `NULL`, which means, no function is invoked.
  */
-#ifndef AZ_NO_LOGGING
-void az_log_set_classifications(az_log_classification const classifications[]);
-#else
-AZ_INLINE void az_log_set_classifications(az_log_classification const classifications[])
+#ifdef AZ_NO_LOGGING
```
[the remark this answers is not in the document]
ahsonkhan: The double negative of `ifndef AZ_NO_LOGGING` is confusing. Every time someone reads this they have to parse what this exactly means. I saw this first hand when discussing something unrelated with Daniel.

sdk/inc/azure/core/az_log.h
```diff
@@ -64,56 +61,63 @@ enum az_log_classification_core
...
+ * @brief Defines the signature of the callback function that application developers must implement
+ * which will be used to check whether a particular log classification should be logged.
+ *
+ * @param[in] classification The log message's #az_log_classification.
  *
- * @param[in] classifications __[nullable]__ An array of #az_log_classification values, terminated
- * by #AZ_LOG_END_OF_LIST.
+ * @return Whether or not a log message with the provided classification should be logged.
+ */
+typedef bool (*az_log_should_write_fn)(az_log_classification classification);
```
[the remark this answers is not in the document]
ahsonkhan: > I think, "message_filter" is probably better than "classification_filter", fron the user's persective. Not _what_ it does, but _why_ it does what it does.

But the user doesn't filter based on the `az_span message`, they filter on the `az_log_classification`.

sdk/inc/azure/core/az_log.h
```diff
@@ -64,56 +61,63 @@ enum az_log_classification_core
...
 /**
- * @brief Defines the signature of the callback function that application developers must write in
- * order to receive Azure SDK log messages.
+ * @brief Defines the signature of the callback function that application developers must implement
+ * in order to receive Azure SDK log messages.
  *
  * @param[in] classification The log message's #az_log_classification.
  * @param[in] message The log message.
  */
 typedef void (*az_log_message_fn)(az_log_classification classification, az_span message);
```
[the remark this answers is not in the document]
ahsonkhan: I understand. Its listed in my TODO in the PR description.

sdk/samples/storage/blobs/src/blobs_client_example.c
```diff
@@ -43,8 +43,17 @@ static void test_log_func(az_log_classification classification, az_span message)
   printf("%.*s\n", az_span_size(message), az_span_ptr(message));
 }
 
-static az_log_classification const log_classifications[]
-    = { AZ_LOG_HTTP_REQUEST, AZ_LOG_HTTP_RESPONSE, AZ_LOG_END_OF_LIST };
+static bool test_log_should_write_func(az_log_classification classification)
```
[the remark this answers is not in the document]
ahsonkhan: > why the "test" got in the name, BTW? It should not be there in any case)

This was how it was before, will fix.

sdk/inc/azure/core/az_log.h
```diff
@@ -64,56 +61,63 @@ enum az_log_classification_core
...
  *
- * @param[in] az_log_message_callback __[nullable]__ A pointer to the function that will be invoked
- * when the SDK reports a log message matching one of the #az_log_classification passed to
- * #az_log_set_classifications(). If `NULL`, no function will be invoked.
+ * @param[in] az_log_should_write_callback __[nullable]__ A pointer to the function that will be
+ * invoked when the SDK checks whether a log message of a particular #az_log_classification should
+ * be logged. If `NULL`, log messages for all classifications will be logged, by passing them to the
+ * #az_log_message_fn provided to #az_log_set_message_callback().
+ *
+ * @remarks By default, this is `NULL`, in which case no function is invoked to check whether a
```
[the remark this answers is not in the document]
ahsonkhan: We didn't say that on `set_classifications` before. I don't think it is necessary, especially because that is implementation detail. Either the user cares about logging and want to filter to some particular messages, or they want all of them.

sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
...
-  // This message's classification is not in the customer-provided list; we should not log it.
-  return false;
+  _az_log_should_write_callback = az_log_should_write_callback;
 }
 
 // This function returns whether or not the passed-in message should be logged.
 bool _az_log_should_write(az_log_classification classification)
 {
-  return _az_log_write_engine(false, classification, AZ_SPAN_EMPTY);
+  _az_PRECONDITION(classification > 0);
```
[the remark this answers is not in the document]
ahsonkhan: >  First, classification can be anything. Today it can't be <0, tomorrow it can, there's no special meaning behind that, no grand plan.

Exactly, once that is true _in the future_, we can remove the precondition. Weakening the precondition is all non-breaking. Also, this is an internal method that our code calls and guards against accidental bugs in our code, like a `Debug.Assert(...)`.

> Plus, this is a precondition in logging engine, which is already better to have non-faulty, because logging can be used during the error handling

I don't understand this. In some public APIs, whenever for simplicity, we removed az_result, we substituted the failure case with a precondition and docs making it the caller's responsibility to check and fix the bug in their code. The same holds true here.

> Plus, it is us who call these functions, so user does nothing wrong, it is our "bug", and has to deal with precondition failure, and what are they supposed to do?

Again, exactly. There is no harm in keeping it, and it helps us detects bugs in our code. This is conveying intent to other callers to make sure they set the log classification enum value correctly, and they aren't passing in a value that _currently_ doesn't make sense.

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
ahsonkhan: Why `everything`? That's not what this is doing. It is only returning true for `AZ_LOG_MQTT_RECEIVED_PAYLOAD`

sdk/inc/azure/core/az_log.h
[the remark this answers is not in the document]
JeffreyRichter: I don't think the current names are bad. The function is called asking the callee if the classification should be logged. Maybe az_log_should_log_fn? It has 'log' twice in the name which is not ideal; but not terrible.

sdk/inc/azure/core/az_log.h
```diff
@@ -64,56 +61,63 @@ enum az_log_classification_core
...
-#else
-AZ_INLINE void az_log_set_callback(az_log_message_fn az_log_message_callback)
+#ifdef AZ_NO_LOGGING
+AZ_INLINE void az_log_set_filter_callback(az_log_should_write_fn az_log_should_write_callback)
 {
-  (void)az_log_message_callback;
+  (void)az_log_should_write_callback;
 }
+#else
+void az_log_set_filter_callback(az_log_should_write_fn az_log_should_write_callback);
```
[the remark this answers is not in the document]
JeffreyRichter: How about naming the function: az_log_set_should_write_callback

sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
 
 #ifndef AZ_NO_LOGGING
 
-static az_log_classification const* volatile _az_log_classifications = NULL;
+// Only using volatile here, not for thread safety, but so that the compiler does not optimize what
+// it falsely thinks are stale reads.
 static az_log_message_fn volatile _az_log_message_callback = NULL;
+static az_log_should_write_fn volatile _az_log_should_write_callback = NULL;
```
JeffreyRichter, line 22: I can't remember if we default to writing all classifications; but, if so, you could initialize this to point to a function (in core) that simply returns true.

sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
...
-        callback(classification, message);
-      }
-
-      return true;
-    }
-  }
-
-  // This message's classification is not in the customer-provided list; we should not log it.
-  return false;
+  _az_log_should_write_callback = az_log_should_write_callback;
```
JeffreyRichter, line 31: @antkmsft : Didn't we have a compare exchange in the PAL? We should call it here to guarantee atomicity.

sdk/src/azure/core/az_log.c
```diff
@@ -16,75 +16,51 @@
...
+  _az_PRECONDITION(classification > 0);
+
+  // Copy the volatile fields to local variables so that they don't change within this function.
+  az_log_message_fn const message_callback = _az_log_message_callback;
+  az_log_should_write_fn const should_write_callback = _az_log_should_write_callback;
+
+  // If the user hasn't registered a should_write_callback, then we log everything, as long as a
+  // message_callback metho was provided.
+  return (should_write_callback == NULL || should_write_callback(classification))
+      && message_callback != NULL;
```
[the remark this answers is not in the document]
JeffreyRichter: You should capture the function pointer in a local variable first & then compare to NULL and invoke it. As written, the code suffers from a potential race condition where this thread see not NULL, another changes the pointer to NULL and then you invoke NULL causing an access violation.

You could also remove the NULL check if you guarantee that the global variable always refers to a valid function. For example, we could have a function n core that always returns true (or false).

sdk/src/azure/core/az_log.c
ahsonkhan: Yes, we default to write all if the callback is `NULL`.

sdk/inc/azure/core/az_log.h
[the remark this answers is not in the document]
antkmsft: I understand, double negation is not perfect, but the order of branches that follow is more important, in my opinion.
You know, yesterday I was more categorical about this, was it because it was a middle of the night, I dunno - but, now I look at this and I don't think it has that much importance. If you strongly think we should prioritize getting rid of double negations, and also you can think of it as "error first, happy path last", do it.

sdk/inc/azure/core/az_log.h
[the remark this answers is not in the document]
antkmsft: So, if we go with a verb, shouldn't we do the same for the writer callback?
We pick them in pairs: `(log)+should_log`,  `write+should_write`, `message+classification_filter`, or `message+message_filter`:
```
void az_log_set_*_callback(az_log_*_fn);
```

Which ever pair we pick, I the naming pattern should be consistent.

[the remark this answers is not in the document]
antkmsft: > But the user doesn't filter based on the `az_span message`, they filter on the `az_log_classification`.

Yes, but they filter on classifications, in order to see less messages. The same way, users receive messages, not spans - first is type, second is meaning, we don't call the first "az_log_span_callback", but call it "az_log_message_callback"

sdk/samples/storage/blobs/src/blobs_client_example.c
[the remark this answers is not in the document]
antkmsft: I know, this is not you who put that "test" there, thank you for looking to fix it!

sdk/inc/azure/core/az_log.h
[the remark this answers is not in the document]
antkmsft: Jeffrey, I listed your suggestion as one of the possible name pairs for log()+should_log() functions, let's decide on any one :)
But here, in this comment, I am talking about the argument name.

sdk/src/azure/core/az_log.c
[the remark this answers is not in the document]
antkmsft: I'll let you decide, but it is much harder to get an invalid classification, than to get an invalid span. But we will be checking this up to three times during each logging attempt (should_log, then log), and if we change the bit representation to allow more values, nothing tells us to update this place - there's nothing like _az_PRECONDITION_IS_VALID_CLASSIFICATION(), so the only way to find out would be to trip on these preconditions.

sdk/src/azure/core/az_log.c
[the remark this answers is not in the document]
antkmsft: I think checking for NULL is better than empty callback: less cycles spent on each place that logs when user is not listening, plus, we don't pointer-jump unless there's a callback, which is also good because they may invalidate CPU cache.

sdk/tests/core/inc/az_test_log.h
antkmsft: Now I don't feel that strong, but I still like the previous order and optimizing for happy path first, it is better from the reader's perspective, IMO.

sdk/src/azure/core/az_log.c
antkmsft: Ah, if only we could! Logging is used by IoT, and they don't want PAL dependency, because their customers will have to provide az_noplatform.lib to linker at all times.

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
antkmsft: I think github may have misplaced the comment, because the function I am looking at would return true for any `mqtt`: for `AZ_LOG_MQTT_RECEIVED_TOPIC`, and for `AZ_LOG_MQTT_RECEIVED_PAYLOAD`.

sdk/src/azure/core/az_log.c
JeffreyRichter: Well, we need something to do this assignment atomically; volatile by itself doesn't guarantee that.

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
ahsonkhan: I wouldn't name that as `everything` either.

sdk/src/azure/core/az_log.c
antkmsft: Don't we "assume it is atomic"? You can do volatile and be reasonably good, and would cause no AVs in many environments, in this specific case, especially given that some compilers/platforms may have flags/guarantees to do that atomically.

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
antkmsft: `_should_write_any_mqtt`? But on the other hand, you call it "everything" in other places, such as in the context of core tests you only include AZ_LOG_HTTP ones, but call it everything.

antkmsft (commented):
I like this approach more and more. Previously, with classifications list, it was a whitelist only. Now it can be a whitelist or a blacklist, depending how you wish to write the filter function.

sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c
ahsonkhan: I call it everything there because it returns true for all `core` log classifications.

JeffreyRichter (approved):

CHANGELOG.md
```diff
@@ -23,11 +23,14 @@
   - `AZ_LOG_IOT_RETRY`
   - `AZ_LOG_IOT_SAS_TOKEN`
   - `AZ_LOG_IOT_AZURERTOS`
+- Removed `AZ_LOG_END_OF_LIST` log classification and `az_log_set_classifications()` from `az_log.h`.
```
antkmsft, line 26: I would remove the "from az_log.h" part.

CHANGELOG.md
```diff
@@ -23,11 +23,14 @@
...
+- Renamed `az_log_set_callback()` to `az_log_set_message_callback()` and renamed the parameter from `az_log_message_callback` to `log_message_callback`.
```
antkmsft, line 27: Parameter names are not breaking changes.

CHANGELOG.md
```diff
@@ -23,11 +23,14 @@
...
 
 ### Bug Fixes
 
 ### Other Changes and Improvements
 
+- Added a `az_log_filter_fn` callback function along with a setter method `void az_log_set_classification_filter_callback(az_log_filter_fn message_filter_callback);`, allowing the caller to filter log messages.
```
antkmsft, line 33:
```suggestion
- Added a `az_log_filter_fn` callback function type along with a setter `az_log_set_classification_filter_callback()`, allowing the caller to filter log messages.
```

sdk/inc/azure/core/az_log.h
```diff
@@ -64,55 +61,62 @@ enum az_log_classification_core
...
+ * function will be invoked.
  *
- * @warning Users must not change the \p classifications array elements, once it is passed to this
- * function. If \p classifications array is allocated on a stack, program behavior in multi-threaded
- * environment is undefined.
+ * @remarks By default, this is `NULL`, which means, no function is invoked.
  */
 #ifndef AZ_NO_LOGGING
-void az_log_set_classifications(az_log_classification const classifications[]);
+void az_log_set_message_callback(az_log_message_fn log_message_callback);
```
antkmsft, line 93: I would leave it as `az_log_set_callback()`, reason being that this is the main line of thought when logging, so you think of it as "log callback", while "classification filter callback" is a specialization which you may, or may not provide.

sdk/src/azure/core/az_log.c
```diff
@@ -39,39 +43,24 @@ void az_log_set_callback(az_log_message_fn az_log_message_callback)
 // false indicating whether it was logged.
 static bool _az_log_write_engine(bool log_it, az_log_classification classification, az_span message)
```
antkmsft, line 44: Having this method is good, we don't get any race conditions, but I would simplify to having an inline function returning a callback. So, in _az_log_should_write() you do "return _az_log_get_message_callback() != NULL", and in _az_log_write, you have "_az_log_get_callback()(classification, callback)";

If you want to keep the "engine" function, I would mark it as AZ_INLINE now - previously it was a longer method with iterations, which we didn't neccessarily want to be inlined, and now it is very short. But IMO,
```
AZ_INLINE az_log_message_fn _az_log_get_message_callback()
{
    az_log_message_fn const message_callback = _az_log_message_callback;
    az_log_filter_fn const message_filter_callback = _az_message_filter_callback;

    return _az_log_message_callback != NULL
        && (_az_message_filter_callback != NULL || _az_message_filter_callback(classification))
            ? _az_log_message_callback
            : NULL;
}
```
is better.

sdk/tests/core/test_az_logging.c
```diff
@@ -66,6 +66,30 @@ static void _log_listener(az_log_classification classification, az_span message)
   }
 }
 
+static bool _should_write_everything_valid(az_log_classification classification)
```
antkmsft, line 69: To match the `_should_write_any_mqtt()`, shouldn't this one be called `_should_write_any_http()`?

antkmsft (approved):

sdk/src/azure/core/az_log.c
antkmsft: An example of us doing this is here: https://github.com/Azure/azure-sdk-for-cpp/blob/c5438d623bd95d9bae3256aa1a4c11ceed6c097d/sdk/core/azure-core/src/logging/logging.cpp#L32

sdk/inc/azure/core/az_log.h
```diff
@@ -64,55 +61,62 @@ enum az_log_classification_core
...
- * @details If no classifications are set (\p classifications is `NULL`), the application will
- * receive log messages for all #az_log_classification values.
- * @details If \p classifications is not `NULL`, it must point to an array of
- * #az_log_classification, terminated by #AZ_LOG_END_OF_LIST.
- * @details In contrast to \p classifications being `NULL`, \p classifications pointing to an empty
- * array (which still should be terminated by #AZ_LOG_END_OF_LIST), states that an application is
- * not interested in receiving any log messages.
+ * @return Whether or not a log message with the provided classification should be logged.
+ */
+typedef bool (*az_log_filter_fn)(az_log_classification classification);
```
RickWinter, line 80:
```suggestion
typedef bool (*az_log_classification_filter_fn)(az_log_classification classification);
```

sdk/tests/core/test_az_logging.c
ahsonkhan: I would like to leave this as is, especially since this is just naming in tests.

sdk/src/azure/core/az_log.c
ahsonkhan: I am going to defer such changes to a subsequent PR.

CHANGELOG.md
ahsonkhan: Why? I prefer keeping it as it helps users know how they will be impacted and which header is being changed.

sdk/docs/core/README.md
```diff
@@ -85,13 +85,25 @@ As our SDK performs operations, it can send log messages to a customer-defined c
 To enable logging, you must first write a callback function that our logging mechanism will invoke periodically with messages. The function signature must match this type definition (defined in the [az_log.h](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/inc/azure/core/az_log.h) file):
 
    ```C
-   typedef void (*az_log_fn)(az_log_classification classification, az_span message);
+   typedef void (*az_log_message_fn)(az_log_classification classification, az_span message);
```
antkmsft, line 88: This rename should go to changelog.

ahsonkhan: This rename didn't occur in any recent release. It was done quite a while back, well before preview 4. It doesn't make sense to add an entry to the changelog for a change that wasn't made in that release. All this is doing is caching the docs up to what already changed.

sdk/src/azure/core/az_log.c
ahsonkhan: With an optimizing compiler (-Os) both gcc and clang produce basically identical assembly for both what you have, and what is currently in the PR using `_az_log_write_engine`, when calling `_az_log_write`.

`_az_log_should_write` actually performs slightly better on the original version, compared to yours, when optimized.

But, the current PR performs poorly when not-optimized.

PR #1336, opened 2020-09-22
Rev version

RickWinter:


ahsonkhan (approved):

antkmsft (approved):
I think that maybe we should undef AZ_SDK_VERSION_PRERELEASE, instead of making it empty.

antkmsft (commented):
Because you can't write a condition `#ifdef AZ_SDK_VERSION_PRERELEASE == "anything"`, including `#ifdef AZ_SDK_VERSION_PRERELEASE == ""`.
But you can do `#ifdef AZ_SDK_VERSION_PRERELEASE` and then your code knows that you are compiling with Beta.
If you go with undefing it, which I strongly  suggest, update the comment above it.


Update: the comment is good, and thank you for undefing. I like the way you did it, with having both "" and then the undef.

ahsonkhan:
Please hold off on merging this until we are ready to release.

sdk/inc/azure/core/az_version.h
```diff
@@ -29,6 +29,6 @@
 #define AZ_SDK_VERSION_PATCH 0
 
 /// Optional pre-release identifier. SDK is in a pre-release state when present.
-#define AZ_SDK_VERSION_PRERELEASE "preview.6"
+#undef AZ_SDK_VERSION_PRERELEASE
```
antkmsft, line 32: For Doxygen:
```suggestion
#define AZ_SDK_VERSION_PRERELEASE
#undef AZ_SDK_VERSION_PRERELEASE
```

antkmsft: This is still not ideal. The docs will look as if it is defined. But if we go and undef it without ever defining, the info on AZ_SDK_VERSION_PRERELEASE will be missing from the docs.

ahsonkhan: Where do you see that?
I don't see `AZ_SDK_VERSION_PRERELEASE` anymore in the docs from the generated artifacts:
image

antkmsft: With this suggestion (define then undef next line), you will be able to:
image
image

ahsonkhan: It might be better that it is hidden and not visible in the docs, which we get by jus keeping undef. But your call Rick.

antkmsft: If we don't want in documentation, I suggest to wrap the whole thing, including the doxygen comment, as 
```
/*
/// Optional pre-release identifier. SDK is in a pre-release state when present.
#define AZ_SDK_VERSION_PRERELEASE "beta.1"
*/
``` 
- this way, note that it sets the format, and there is not Doxygen comment
If we have a doxygen comment, and then an undef entry but without a #define, the doxygen comment woud stickto the next documentable entity, if there's any in the future. But if we comment (note that we don't use the doxygen-style multi-line comment), then it is excluded alltogether.

ahsonkhan: I see. In that case, define/undef, like you suggested makes sense.

PR #1337, opened 2020-09-22
Update cgmanifest to include vcpkg dependencies and linux tools we use in CI.

ahsonkhan:
Documenting all `AptDependencies` and `vcpkg.deps`

**These, currently, include:**
- curl (which transitively depends on openssl, and zlib)
- paho-mqtt
- cmocka
- gcc-arm-none-eabi
- gcovr
- lcov
- clang-format
- doxygen

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -57,6 +57,28 @@
       "Component": {
         "Type": "other",
         "Other": {
+          "Name": "openssl",
+          "Version": "1.1.1d",
+          "DownloadUrl": "https://www.openssl.org"
+        }
+      },
+      "DevelopmentDependency": false
+    },
+    {
+      "Component": {
+        "Type": "other",
+        "Other": {
+          "Name": "zlib",
+          "Version": "1.2.11-6",
+          "DownloadUrl": "https://www.zlib.net/"
+        }
+      },
+      "DevelopmentDependency": false
+    },
+    {
+      "Component": {
+        "Type": "other",
+        "Other": {
           "Name": "gcc-arm-none-eabi",
           "Version": "6.3.1+svn253039-1build1",
           "DownloadUrl": "https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/gcc-arm-none-eabi_6.3.1+svn253039-1build1_amd64.deb.html"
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -343,6 +343,22 @@ jobs:
       # Run clang-format recursively on each source and header file within the repo.
       find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;
       
+      echo CLANG FORMAT INFO
```
ahsonkhan, line 346: Unrelated change which I will revert. Just using it to verify version numbers.

antkmsft (approved):
I see we are freezing the version numbers which is the best approach for build stability. However, what is our plan to upgrade these dependencies, do we have a process, cadence, or is there an event that would trigger us to go and grab more latest versions?

cgmanifest.json
```diff
@@ -19,6 +19,116 @@
...
+    {
+      "Component": {
+        "Type": "other",
+        "Other": {
+          "Name": "cmocka",
+          "Version": "1.1.5-1",
+          "DownloadUrl": "https://gitlab.com/cmocka/cmocka"
+        }
+      },
+      "DevelopmentDependency": false
```
RickWinter, line 32: Are we including parts of CMocka outside of the tests?  Otherwise I'd expect this one is true that its a dev only dependency

RickWinter (approved):

ahsonkhan:
> I see we are freezing the version numbers which is the best approach for build stability.

This file and PR in itself doesn't freeze anything (that isn't the purpose of cgmanifest, all we are doing is declaring our dependencies). We rely on vcpkg to handle the versioning for us. The other versions are defaults that we get on our build images based on OSes like ubuntu-18.04, etc. when installed.

We pinned the vcpkg version to this:
https://github.com/Azure/azure-sdk-for-c/blob/fad6ccb651a438ec6214d3d4944c051f72f71630/eng/vcpkg-commit.txt#L1

vhvb1989 (approved):
Great work Sr.! Thank you

PR #1338, opened 2020-09-22
Fix libcmt/libcmtd.lib link warning

### LINKED ISSUE #1020: Don't suppress linker warning about libcmtd.lib
https://github.com/Azure/azure-sdk-for-c/blob/1215f7283eafd7cb16d0bd65b5da0f5cde157915/cmake-modules/AddTestCMocka.cmake#L108

This warning is important, it indicates you've mixed debug and release CRTs, or shared and static CRTs. You need to fix the mixing not just suppress the warning. If you mix libs like this many CRT calls are undefined behavior.

antkmsft:
Closes #1020

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -81,7 +81,7 @@ jobs:
       WindowsX86_Release_MapFiles_UnitTests:
         OSVmImage: 'windows-2019'
         vcpkg.deps: 'cmocka'
-        VCPKG_DEFAULT_TRIPLET: 'x86-windows-static'
+        VCPKG_DEFAULT_TRIPLET: 'x86-windows-static-md'
         build.args: '-DPRECONDITIONS=OFF -DUNIT_TESTING=ON -DLOGGING=OFF'
         CMAKE_GENERATOR: 'Visual Studio 16 2019'
         CMAKE_GENERATOR_PLATFORM: Win32
@@ -169,7 +169,7 @@ jobs:
       Windows_Preconditions_UnitTests_Samples:
         OSVmImage: 'windows-2019'
         vcpkg.deps: 'curl[winssl] paho-mqtt cmocka'
-        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static'
+        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static-md'
         build.args: '-DTRANSPORT_CURL=ON -DTRANSPORT_PAHO=ON -DAZ_PLATFORM_IMPL=WIN32 -DUNIT_TESTING=ON -DLOGGING=OFF'
         CMAKE_GENERATOR: 'Visual Studio 16 2019'
         CMAKE_GENERATOR_PLATFORM: x64
@@ -215,7 +215,7 @@ jobs:
       Windows_Release_MapFiles:
         OSVmImage: 'windows-2019'
         vcpkg.deps: ''
-        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static'
+        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static-md'
         build.args: '-DPRECONDITIONS=OFF -DLOGGING=OFF'
         CMAKE_GENERATOR: 'Visual Studio 16 2019'
         CMAKE_GENERATOR_PLATFORM: x64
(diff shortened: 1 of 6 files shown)


sdk/tests/iot/common/CMakeLists.txt
```diff
@@ -18,5 +18,4 @@ add_cmocka_test(az_iot_common_test SOURCES
                 LINK_TARGETS
                     az_iot_common
                     az_core
-                    ${PAL}
```
vhvb1989, line 21: I believed iot were using clock, at some point..  calling wait(), or something like that. Linker was failing to find the function definition if not linking a PAL.  Did that change?

antkmsft: IoT does not have a dependency on PAL and we're not/should not be using using clock there. Precondition no longer invokes sleep for exactly this reason.
It all builds, links, runs, and passes, so no problem.

vhvb1989 (approved):

RickWinter (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -215,7 +215,7 @@ jobs:
       Windows_Release_MapFiles:
         OSVmImage: 'windows-2019'
         vcpkg.deps: ''
-        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static'
+        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static-md'
```
ahsonkhan, line 218: Does it make sense to use `-md` even for the release leg?

sdk/tests/iot/common/CMakeLists.txt
ahsonkhan: nit: In the future, please do such cleanup changes in a separate PR/commit.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -169,7 +169,7 @@ jobs:
       Windows_Preconditions_UnitTests_Samples:
         OSVmImage: 'windows-2019'
         vcpkg.deps: 'curl[winssl] paho-mqtt cmocka'
-        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static'
+        VCPKG_DEFAULT_TRIPLET: 'x64-windows-static-md'
```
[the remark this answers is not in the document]
ahsonkhan: I haven't looked too closely, but this might be interesting to follow up to understand why:
https://github.com/microsoft/vcpkg/issues/1131#issuecomment-303060932

> It is discouraged to use `static-md` triplet...

I wonder if the issue is with the `static` part or the `-md` part.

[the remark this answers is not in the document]
ahsonkhan: @barcharcraz - can you please take a look at this as well and share your thoughts.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
antkmsft: It is not. `-md` stands for "multithreaded dynamic".

eng/pipelines/templates/jobs/archetype-sdk-client.yml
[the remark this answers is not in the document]
antkmsft: That thread is from 2017. Probably there were issues back then. And also I can see that these triplets were added to the officiale vcpkg repo on February, so maybe there were other issues at a time. But now it is there. https://github.com/microsoft/vcpkg/commit/71a9def3df1cd418a8c00e69413f19507ab25fa8.  Right now, it seems permanent and I don't see why we can't use it.
The reason why we were getting the warning was that _we_ built our SDK with dynamic runtime, but cmocka was built with a reference to static runtime, and we were inking them together, and that is what the warning was telling us. Now we build cmocka with the same runtime as our SDK, so now they are consistent, and there is no warning anymore.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
ahsonkhan: Consider updating references `x64-windows-static` in docs to include `-md`. We have a couple of places if you search for `triplet`:
https://github.com/Azure/azure-sdk-for-c/blob/ef22522414e5b56acba056eae84dd3facda63865/sdk/samples/iot/README.md#set-environment-variables

https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/docs/how_to_iot_hub_samples_windows.md#setup-instructions

PR #1333, opened 2020-09-21
Make az_log_set_classifications internal.

ahsonkhan:
Intentionally kept the changes here to a minimum.

Fixes https://github.com/Azure/azure-sdk-for-c/issues/1172 and also addresses a big part of https://github.com/Azure/azure-sdk-for-c/issues/1184.

Follow up to https://github.com/Azure/azure-sdk-for-c/issues/1227 and https://github.com/Azure/azure-sdk-for-c/issues/1192

- Removed `AZ_LOG_END_OF_LIST` log classification and `az_log_set_classifications()` from `az_log.h`.
- Renamed `az_log_set_callback()` to `az_log_set_message_callback()`.

The subsequent PR - https://github.com/Azure/azure-sdk-for-c/pull/1317, will add a `az_log_set_classification_filter_callback` mechanism to let the users filter messages.

PR #1340, opened 2020-09-22
Remove end of array sentinel for HTTP status codes.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1184

Minimal change following https://github.com/Azure/azure-sdk-for-c/pull/1333 to avoid having to make a breaking change post-GA. With the storage client pushed out to 1.1.0, there are no callers within the SDK impacted by this.


antkmsft (approved):
Looks good to me, thank you for doing this. A few things could me written simpler, such as using `status_codes[i]` instead of incrementing it, but all that can be done later.
The only thing we would need is changelog update ("Removed `status_codes` from `az_http_policy_retry_options`, and `AZ_HTTP_STATUS_CODE_END_OF_LIST` HTTP status code."?)

vhvb1989 (approved):

RickWinter (approved):

PR #1344, opened 2020-09-22
Use return instead of retval for boolean values in az_result docs.

ahsonkhan:
Fixes the doc rendering issue with `retval` which doesn't handle back-ticks well:
https://github.com/Azure/azure-sdk-for-c/issues/1104#issuecomment-678612494

**Before:**
image

**After:**
image

### DIFF
--- a/sdk/inc/azure/core/az_result.h
+++ b/sdk/inc/azure/core/az_result.h
@@ -143,8 +143,7 @@ enum az_result_core
  *
  * @param[in] result Result value to check for failure.
  *
- * @retval true The operation that returned this \p result failed.
- * @retval false The operation that returned this \p result was successful.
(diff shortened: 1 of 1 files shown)


vhvb1989 (approved):

momuno:
Do you know why this is an issue for booleans?  Do we not use @retval for specific return explanation of each value that could be returned?

ahsonkhan:
Since boolean has a closed, known, two-state return value, @return makes a lot more sense here. We should reserve retval for things like az_result (or ints), or more open, not-always known, and multiple return value cases.

RickWinter (approved):

antkmsft (approved):

RickWinter:
@ahsonkhan Please commit

PR #1349, opened 2020-09-22
Remove mentions of Storage from Readme

antkmsft:


vhvb1989 (approved):

README.md
```diff
@@ -248,17 +248,6 @@ The following compilation, preprocessor options will add or remove functionality
 See [cmake options](#cmake-options) to learn about how to build an HTTP transport adapter, how to build IoT samples, and to turn logging on.
 
 
-### Storage Sample
```
ahsonkhan, line 251: There is more under the table in https://github.com/Azure/azure-sdk-for-c/blob/cdd2c4a2a3e9631c497b12a61091ca29b0441eb7/README.md#cmake-options

ahsonkhan:
I don't think this change is necessary post-GA, since the storage could isn't going to come back shortly, and such a change will make reverting difficult because it might introduce a merge conflict.

PR #1353, opened 2020-09-22
Update header includes in samples

momuno:
Per https://github.com/Azure/azure-sdk-for-c/pull/1289, I checked sample to see if any should use the consolidated headers.
None can, since none use all the headers in either consolidation.

Discovered a few samples should be including `az_iot_common.h`, due to use of `az_iot_status` and `az_iot_message_properties`

This will not be merged until after branch reopens Friday 9/25.


ahsonkhan:
> None can, since none use all the headers in either consolidation.

I think it is valuable to present samples with simplified headers, even if not all will be used directly. Any motivated customer, if desired, can pick individual/granular headers, but having samples that showcase simple "include core, include iot", and you are good to go, is quite valuable, imo.

Other package-based languages/distribution have single acquisition points (and generally one root namespace, with potential sub-namespaces) for core and per service. This helps customers with the "Getting Started" scenario.

sdk/samples/iot/pnp/pnp_thermostat_component.c
```diff
@@ -22,6 +21,7 @@
 #include <azure/core/az_json.h>
 #include <azure/core/az_result.h>
 #include <azure/core/az_span.h>
+#include <azure/iot/az_iot_common.h>
```
ahsonkhan, line 24: What made you decide to explicitly include this header, when it isn't _technically_ necessary (because other headers are transitively including it anyway)?

I think it is a good idea, but I wonder what other such headers are missing. How would we detect that vs. figure out when a header is unused and need to remove it?

momuno:
> > None can, since none use all the headers in either consolidation.
> 
> I think it is valuable to present samples with simplified headers, even if not **all** will be used directly. Any motivated customer, if desired, can pick individual/granular headers, but having samples that showcase simple "include core, include iot", and you are good to go, is quite valuable, imo. Providing that level of detail/breakdown of our individual headers is not necessary for the audience of samples.
> 
> Other package-based languages/distribution have single acquisition points (and generally one root namespace, with potential sub-namespaces) for core and per service. This helps customers with the "Getting Started" scenario.

After discussion with @danewalton on the style of includes, I disagree on this one.  Originally I thought fewer is better as you say, and at first glance it does appear cleaner.  But I think there is a lot of value in presenting the exact files that are being referenced in these samples so that there is no guess work for the user.  It can be very painful to try to track down the chain of includes in large systems. (Note: We do follow this approach currently, with `NULL` being an example.  Although it is included in `stdlib.h`, it is specifically defined in `stddef.h`.  So we include `stddef.h` for that, and `stdlib.h` for things specifically defined in that file e.g. `exit()`)

I do think we should have your simplification header option though for those developers who prefer that approach.

sdk/samples/iot/pnp/pnp_thermostat_component.c
momuno: Hi.  Added comment below.  Yes, there is more onus on the developer to stay up-to-date on what headers to remove and add since the compiler will not automatically detect it.  It is not critical, but I think a good practice to keep headers up to date when one is going to add standard functionality that would require a header (even if it is already included transitively).

momuno:
We had a convo discussing this with @antkmsft at somepoint I think, but can't recall where it is...  it's certainly a good topic.

ahsonkhan:
> But I think there is a lot of value in presenting the exact files that are being referenced in these samples so that there is no guess work for the user.

Help me understand what you mean by guess work? What types of questions, do you think, would the customer be asking and why?

>  It can be very painful to try to track down the chain of includes in large systems.

Exactly, if there is no chain because we give examples on how to use simplified headers, we circumvent the problem all together.

> I do think we should have your simplification header option though for those developers who prefer that approach.

Generally speaking, hiding complexity from the user, an defaulting to simplicity is a favorable approach (follows progressive disclosure). I am still not convinced we want samples to show a block of SDK includes. Your call though :)

ahsonkhan:
I like the phrasing here: https://softwareengineering.stackexchange.com/a/185775
`the art to balance contradicting goals`

Compare the depth of what `windows.h` gives (and how often people use that header **only**) to something like `az_span.h`.
https://en.wikipedia.org/wiki/Windows.h

sdk/samples/iot/pnp/pnp_thermostat_component.c
ahsonkhan: > It is not critical, but I think a good practice to keep headers up to date when one is going to add standard functionality that would require a header (even if it is already included transitively).

Let's make sure we are discussing the same thing to avoid conflating audiences.

One topic is on what we want **_customers of the SDK_** to do, which is the primarily purpose of the samples. Your comment below is talking about that (and using single headers vs. many).

My comment here, is about the **author of samples and contributors to this repo and SDK itself**. For such developers, I understand and agree that we want to use granular header files, only include the ones we need, and include them explicitly.

My question is, how do we make sure that happens within our repo? This is not something we can detect by code reviews without investing a lot of time. So, is there a compiler tool/static analysis that can help make sure all transitive header files are explicitly included, and also one which highlights unused headers or dead code paths? Even if we thoroughly go through the SDK and fix all such instances today, how do we plan to stay green? We need some sort of automation here. Got any ideas worth exploring?

momuno:
> > But I think there is a lot of value in presenting the exact files that are being referenced in these samples so that there is no guess work for the user.
> 
> Help me understand what you mean by guess work? What types of questions, do you think, would the customer be asking and why?
> 
> > It can be very painful to try to track down the chain of includes in large systems.
> 
> Exactly, if there is no chain because we give examples on how to use simplified headers, we circumvent the problem all together.
> 
> > I do think we should have your simplification header option though for those developers who prefer that approach.
> 
> Generally speaking, hiding complexity from the user, and **defaulting to simplicity** is a favorable approach (follows [progressive disclosure](https://en.wikipedia.org/wiki/Progressive_disclosure)). I am still not convinced we want samples to show a block of SDK includes. Your call though :)

By guess work I mean tracking down what headers to be looking via the chain of includes.  Speaking from personal experience when learning a large repository, I have found it to be very frustrating.  While abstraction has its benefits (which is what we are doing through the simplification of all headers into one) for a user, there are also the downsides.  In this specific scenario of the samples, because they are so small and require only a few very specific headers, I prefer to showcase that specificity and do not think it overcomplicates the sample.  Instead I think it will assist those wanting to learn the repository.  For full production code that will reference far more headers, I do like the header simplification you have provided.

ahsonkhan:
>Perhaps it would be beneficial to include a comment at the headers for each sample referencing the option of the all-encompassing header for production code? That way they know it is available. Our goal is for the samples to be strong learning tools for the user to learn the API.

That would be great!

momuno:
> > Perhaps it would be beneficial to include a comment at the headers for each sample referencing the option of the all-encompassing header for production code? That way they know it is available. Our goal is for the samples to be strong learning tools for the user to learn the API.
> 
> That would be great!

I will add that to this PR.   wait for it .... lol

ahsonkhan:
Would you agree, though, that the README docs (unlike samples), should only have the simplified headers?

momuno:
> Would you agree, though, that the README docs (unlike samples), should only have the simplified headers?

What README docs are you referring to?

ahsonkhan:
> What README docs are you referring to?

Any and all of them, where relevant/where we mention headers to include. I didn't have a specific on in mind.

**For example, mentioning them in these:**
https://github.com/Azure/azure-sdk-for-c/blob/master/README.md#source-files-ide-command-line-etc
https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/docs/iot/README.md#iot-hub-client-with-mqtt-stack
https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/docs/core/README.md

danewalton (approved):

ahsonkhan (approved):

sdk/samples/iot/pnp/pnp_thermostat_component.c
ahsonkhan: https://github.com/Azure/azure-sdk-for-c/issues/1371

PR #1354, opened 2020-09-23
For vcpkg dependencies, update the component type to git, within cgmanifest.

ahsonkhan:


cgmanifest.json
```diff
@@ -22,55 +22,50 @@
...
     {
       "Component": {
-        "Type": "other",
-        "Other": {
-          "Name": "paho-mqtt",
-          "Version": "1.3.1",
-          "DownloadUrl": "https://github.com/eclipse/paho.mqtt.c"
+        "Type": "git",
+        "git": {
+          "RepositoryUrl": "https://github.com/eclipse/paho.mqtt.c",
```
CIPop, line 37: How are we planning to keep the hashes in sync with VCPKG?

ahsonkhan: We have pinned the vcpkg version that we use. These hashes won't change until we do a manual update. At that point, we will update these too.
https://github.com/Azure/azure-sdk-for-c/blob/6395acce257d21df607d92efdffea6d4729e4655/eng/vcpkg-commit.txt#L1

These hashes correspond to the particular version we use.

RickWinter (approved):

cgmanifest.json
weshaggard: It might be worth adding some validation or automation to keep these in sync as I do think this will likely be forgotten. However I will also say that this is mostly for pulling the license information which rarely if ever changes so it might not be a big deal.

ahsonkhan: Hopefully this issue captures your concern (I am assuming you are referring to keeping vcpkg-commit file and the corresponding commit hashes in cgmanifest):
https://github.com/Azure/azure-sdk-for-c/issues/1364

PR #1360, opened 2020-09-24
Add a .gitattributes file and normalize line endings.

ahsonkhan:
Borrowing from other language SDKs:
https://github.com/Azure/azure-sdk-for-net/blob/0b93d0d19c677f7378e19b9e34681841d5f9131e/.gitattributes#L1
https://github.com/Azure/azure-sdk-for-java/blob/69fdb8508c19800d23a7c2cd54190e1b88378486/.gitattributes#L1-L3
https://github.com/Azure/azure-sdk-for-js/blob/213a83f0871a46834f288b776bba67eddcd387b3/.gitattributes#L17-L18

https://git-scm.com/docs/gitattributes#_end_of_line_conversion
```bash
$ echo "* text=auto" >.gitattributes
$ git add --renormalize .
$ git status        # Show files that will be normalized
$ git commit -m "Introduce end-of-line normalization"
```

If you `Hide whitespace changes`, you will see there are no code changes:
https://github.com/Azure/azure-sdk-for-c/pull/1360/files?diff=split&w=1


ahsonkhan:
If you `Hide whitespace changes`, you will see there are no code changes:
https://github.com/Azure/azure-sdk-for-c/pull/1360/files?diff=split&w=1

vhvb1989 (approved):

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Use unix line endings everywhere, even on Windows
+* text=auto eol=lf
```
danewalton, line 2: does this disallow people from the git setting "check out with windows/commit with Unix"

ahsonkhan: No, I don't believe it does. Git should continue to do its magic for devs working on Windows, as far as I can tell.

ahsonkhan: Let me know if you find otherwise. I have been playing around with the `git config --global core.autocrlf true` settings locally, so my experience and dev box might be in a non-default state.

PR #1365, opened 2020-09-25
update az_iot_pnp_client_twin_get_next_component_property() API

danewalton:
Updated the `az_iot_pnp_client_twin_get_next_component_property()` API to pass back only the `az_json_reader` pointing to the property name. It saves us 60 bytes on the stack by not having another `az_json_token` and the only overhead for the user is that they have to call `az_json_reader_next_token()` to get to the value.

### DIFF
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -695,7 +695,8 @@ AZ_NODISCARD az_result az_iot_pnp_client_twin_get_property_version(
  * instance.
  *
  * @return An #az_result value indicating the result of the operation.
- * @retval #AZ_OK If the function returned a valid #az_json_reader pointing to the property name.
+ * @retval #AZ_OK If the function returned a valid #az_json_reader pointing to the property name and
+ * the #az_span with a component name.
  * @retval #AZ_ERROR_IOT_END_OF_PROPERTIES If there are no more properties left for the component.
  */
 AZ_NODISCARD az_result az_iot_pnp_client_twin_get_next_component_property(
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -676,22 +676,26 @@ AZ_NODISCARD az_result az_iot_pnp_client_twin_get_property_version(
  * only checked and compared. Internally, the #az_span is only changed if the component name changes
  * in the JSON document and is not necessarily set every invocation of the function.
  *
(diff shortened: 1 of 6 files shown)


sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -676,31 +676,35 @@ AZ_NODISCARD az_result az_iot_pnp_client_twin_get_property_version(
...
  * @retval #AZ_ERROR_IOT_END_OF_PROPERTIES If there are no more properties left for the component.
  */
 AZ_NODISCARD az_result az_iot_pnp_client_twin_get_next_component_property(
     az_iot_pnp_client const* client,
     az_json_reader* ref_json_reader,
     az_iot_pnp_client_twin_response_type response_type,
     az_span* out_component_name,
-    az_json_token* out_property_name,
-    az_json_reader* out_property_value);
+    az_json_reader* out_property_name_and_value);
```
ahsonkhan, line 707: It _feels_ like we should be able to design this API differently, without having to use/return the JSON reader like this, but I don't have a great suggestion atm.

When I have some time, I will take a look at sample usage code to fully understand the design/scenario.

hihigupt: Basically this api, should move the cursor to next property and user should get the handle of that cursor. 
With this approach you are asking caller to allocate too much of memory.

hihigupt (approved):

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -882,7 +879,15 @@ static void process_twin_message(
...
-        rc = append_simple_json_token(&jw, &property_value.token);
+        rc = az_json_reader_next_token(&property_name_and_value);
+        if (az_result_failed(rc))
+        {
+          IOT_SAMPLE_LOG_ERROR(
+              "Could not advance to the property value: az_result return code 0x%08x.", rc);
+          exit(rc);
+        }
+
+        rc = append_simple_json_token(&jw, &property_name_and_value.token);
```
momuno, line 890: Same as my comment on the other PR, I would use the IOT_SAMPLE_EXIT_IF_AZ_FAILED in these spots since there are many repetitive log and exits:

```
rc = az_json_reader_next_token(&property_name_and_value);
IOT_SAMPLE_EXIT_IF_AZ_FAILED(rc, "Could not advance to the property value");
```

Please note, with these the colon and return code part of the comment are appended in the macro, so no need to include here.

danewalton: noted tackling in another PR since there are other places where this could be improved

PR #1369, opened 2020-09-25
update twin APIs to property

danewalton:


sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -373,66 +373,67 @@ AZ_NODISCARD az_result az_iot_pnp_client_commands_response_get_publish_topic(
...
   az_span
       request_id; /**< Request ID matches the ID specified when issuing a Get or Patch command. */
-  az_span version; /**< The Twin object version.
-                    * @note This is only returned when
-                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_DESIRED_PROPERTIES
-                    * or
-                    * response_type==AZ_IOT_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES. */
-} az_iot_pnp_client_twin_response;
+  az_span
+      version; /**< The property object version.
```
jspaith, line 419: Nit - it's always going to be returned because it's in a struct.  I think you mean it's only going to be specified (non az_span_null) in this case??

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -582,18 +583,18 @@ AZ_NODISCARD az_result az_iot_pnp_client_twin_property_builder_end_component(
  * @endcode
  *
  * To send a status for a property belonging to a component, first call the
- * az_iot_pnp_client_twin_property_builder_begin_component() API to prefix the payload with the
+ * az_iot_pnp_client_property_builder_begin_component() API to prefix the payload with the
  * necessary identification. The API call flow would look like the following with the listed JSON
  * payload being generated.
  *
  * **With Component**
```
jspaith, line 590: Nit - it's unclear `Component` should be in caps here.  It may be per being in a heading, but component isn't capitilized in https://docs.microsoft.com/en-us/azure/iot-pnp/concepts-components

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -38,7 +38,7 @@
 bool is_device_operational = true;
 
 // * PnP Values *
-// The model id is the JSON document (also called the Digital Twins Model Identifier or DTMI) which
+// The model id is the JSON document (also called the Digital s Model Identifier or DTMI) which
 // defines the capability of your device. The functionality of the device should match what is
```
jspaith, line 42: `Digital s Model` - not sure what the `s` means here?

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -57,11 +57,11 @@ static az_span pnp_device_components[] = { AZ_SPAN_LITERAL_FROM_STR("thermostat1
 static int32_t const pnp_components_length
     = sizeof(pnp_device_components) / sizeof(pnp_device_components[0]);
 
-// IoT Hub Device Twin Values
-static az_span const twin_reported_serial_number_property_name
+// IoT Hub Device Property Values
```
jspaith, line 60: Nit - i'm not sure this is really `IoT Hub Device property` since it's not the Hub twin anymore.  Could this be PnP property??

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -85,7 +85,7 @@ static void subscribe_mqtt_client_to_iot_hub_topics(void);
 static void initialize_components(void);
 static void send_device_info(void);
 static void send_serial_number(void);
-static void request_device_twin_document(void);
+static void request_device_property_document(void);
```
jspaith, line 88: `property_document` isn't really a term i've seen used other places in PnP.  Do you mean something like `request_all_properties_from_service` (with a better name of course, i don't like the `from_service` specifically - maybe quset `request_all_properties` would be OK if i read this right.)

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -164,7 +164,7 @@ static az_result append_simple_json_token(az_json_writer* jw, az_json_token* jso
  * temperature reached since boot.
  *
  * On initial bootup of the device, the sample will send the Temperature Controller reported
- * properties to the service. You will see the following in the device twin JSON.
+ * properties to the service. You will see the following in the device property JSON.
  *   {
```
jspaith, line 168: In general for documenting the JSON here (and even in API docs if it's still there?).  Is this really needed?  My gut is no.  If I'm a dev either I don't care and want the API to "just work" for me, or otherwise I can look at the .c implementing this or even better yet the PnP protocol spec someplace else.  This feels duplicative in this spot to me.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -391,23 +391,25 @@ static void subscribe_mqtt_client_to_iot_hub_topics(void)
...
   if (rc != MQTTCLIENT_SUCCESS)
   {
     IOT_SAMPLE_LOG_ERROR(
-        "Failed to subscribe to the Twin Patch topic: MQTTClient return code %d.", rc);
+        "Failed to subscribe to the property PATCH topic: MQTTClient return code %d.", rc);
     exit(rc);
   }
 
-  // Messages received on Twin Response topic will be response statuses from the server.
+  // Messages received on property Response topic will be response statuses from the server.
```
jspaith, line 404: Nit - should it be `property Response` or `property response`?  Not a bad idea to look at caps after the `Twin` => `property` in general.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -488,21 +492,21 @@ static void send_serial_number(void)
       publish_message.topic, publish_message.out_payload, IOT_SAMPLE_MQTT_PUBLISH_QOS);
   IOT_SAMPLE_LOG_SUCCESS(
       "Client sent `%.*s` reported property message.",
-      az_span_size(twin_reported_serial_number_property_name),
-      az_span_ptr(twin_reported_serial_number_property_name));
+      az_span_size(property_reported_serial_number_property_name),
```
jspaith, line 495: Nit - the `property_...property_name` is stuttering - do we need it twice here?  `property_reported_serial_number_name` instead.  ??

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -488,21 +492,21 @@ static void send_serial_number(void)
...
-static void request_device_twin_document(void)
+static void request_device_property_document(void)
 {
-  IOT_SAMPLE_LOG("Client requesting device twin document from service.");
+  IOT_SAMPLE_LOG("Client requesting device property document from service.");
 
-  // Get the Twin Document topic to publish the twin document request.
-  az_result rc = az_iot_pnp_client_twin_document_get_publish_topic(
+  // Get the property document topic to publish the property document request.
+  az_result rc = az_iot_pnp_client_property_document_get_publish_topic(
```
jspaith, line 509: I hadn't noticed `property_document` in api name but my comment above RE this being non-standard would apply here too.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
danewalton: ahh yea find/replace error. thanks

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
danewalton: Yea this is symmetrical with the Hub API. We should probably discuss whether we want to deviate from that.

sdk/inc/azure/iot/az_iot_pnp_client.h
danewalton: that's fair. originally had it there more as a "title" but I'll lowercase to unify

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
danewalton: I think I agree that the amount of doc's here is a little overwhelming. The one thing I will say is that because we are working at this low level (and the user as well), I'd argue it is useful to show the user what to expect since they will have to interact with this solution at a lower level than some other SDKs.

jspaith (approved):

PR #1374, opened 2020-09-29
Component sample update

danewalton:
- Updated exit strategy on the sample (uses a macro).
- Updates docs on the sample.

### DIFF
--- a/sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
+++ b/sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
@@ -349,7 +349,7 @@ static void connect_mqtt_client_to_iot_hub(void)
   mqtt_connect_options.ssl = &mqtt_ssl_options;
 
   // Connect MQTT client to the Azure IoT Hub.
-  int rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options);
+  rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options);
   if (rc != MQTTCLIENT_SUCCESS)
   {
     IOT_SAMPLE_LOG_ERROR(
--- a/sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
+++ b/sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
@@ -352,7 +352,7 @@ static void connect_mqtt_client_to_iot_hub(void)
   mqtt_connect_options.ssl = &mqtt_ssl_options;
 
   // Connect MQTT client to the Azure IoT Hub.
-  rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options);
+  int rc = MQTTClient_connect(mqtt_client, &mqtt_connect_options);
   if (rc != MQTTCLIENT_SUCCESS)
   {
     IOT_SAMPLE_LOG_ERROR(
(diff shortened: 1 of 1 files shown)


sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -310,23 +306,17 @@ static void create_and_configure_mqtt_client(void)
...
   options.component_names = pnp_device_components;
   options.component_names_length = pnp_components_length;
-  rc = az_iot_pnp_client_init(
-      &pnp_client, env_vars.hub_hostname, env_vars.hub_device_id, model_id, &options);
-  if (az_result_failed(rc))
-  {
-    IOT_SAMPLE_LOG_ERROR("Failed to initialize pnp client: az_result return code 0x%08x.", rc);
-    exit(rc);
-  }
+  IOT_SAMPLE_EXIT_IF_AZ_FAILED(
```
momuno, line 309: The only reason I would leave some of these as is, is because they match what is in the other samples.  But I can go either way.  Since the macro has been introduced for the pnp component sample, it also makes sense to update all the places that log and exit on an az_result fail.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -341,13 +331,10 @@ static void create_and_configure_mqtt_client(void)
...
 {
   // Get the MQTT client username.
-  int rc = az_iot_pnp_client_get_user_name(
-      &pnp_client, mqtt_client_username_buffer, sizeof(mqtt_client_username_buffer), NULL);
-  if (az_result_failed(rc))
-  {
-    IOT_SAMPLE_LOG_ERROR("Failed to get MQTT client username: az_result return code 0x%08x.", rc);
-    exit(rc);
-  }
+  IOT_SAMPLE_EXIT_IF_AZ_FAILED(
```
momuno, line 334: Similar to above, a reason for keeping as is here is because it matches the other samples PLUS it mimics the log and exit for the MQTTClient_connect call below.  Your call.

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
```diff
@@ -469,19 +442,14 @@ static void send_device_info(void)
...
-      NULL);
-
-  if (az_result_failed(rc))
-  {
-    IOT_SAMPLE_LOG_ERROR(
-        "Failed to get the property PATCH topic: az_result return code 0x%08x.", rc);
-    exit(rc);
-  }
+  IOT_SAMPLE_EXIT_IF_AZ_FAILED(
+      az_iot_pnp_client_property_patch_get_publish_topic(
```
momuno, line 446: On some of these longer function calls, would it be easier to read if the call was made first (as prior), and then the result passed in to the macro?

```
az_result rc = az_iot_pnp_client_property_patch_get_publish_topic(
      &pnp_client,
      pnp_mqtt_get_request_id(),
      publish_message.topic,
      publish_message.topic_length,
      NULL);
IOT_SAMPLE_EXIT_IF_AZ_FAILED(rc, "Failed to get the property PATCH topic");
```

momuno (approved):

sdk/samples/iot/paho_iot_hub_pnp_component_sample.c
danewalton: yes great call. Will update for cases that maybe don't have chained/dependent successive calls.

ericwol-msft (approved):

PR #1390, opened 2020-10-01
Add a notice for preview Plug and Play features

danewalton:


sdk/docs/iot/README.md
```diff
@@ -51,6 +51,7 @@ The library targets made available via CMake are the following:
 
 - `az::iot::hub` - For Azure IoT Hub features ([API documentation here][azure_sdk_for_c_doxygen_hub_docs])
 - `az::iot::provisioning` - For Azure IoT Provisioning features ([API documentation here][azure_sdk_for_c_doxygen_provisioning_docs])
+- `az::iot::pnp` - **[PREVIEW]** For Azure IoT Plug and Play features ([feature branch here](https://github.com/Azure/azure-sdk-for-c/tree/feature/iot_pnp))
```
momuno, line 55: sometimes I see `tree` used and other times I see `blob/master`.  Do you know what the difference is?

momuno (approved):

ericwol-msft (approved):

sdk/docs/iot/README.md
danewalton: interesting question. This might be better than what I would explain.
https://stackoverflow.com/questions/39400848/in-github-urls-what-is-the-difference-between-a-tree-and-a-blob

momuno: blobs for files, trees for directories... cool.

PR #1391, opened 2020-10-01
Update live tests to use the same package triplet as CI/CD does

antkmsft:


ahsonkhan:
Given the  change in https://github.com/Azure/azure-sdk-for-c/pull/1394, we can close this, correct?

PR #1398, opened 2020-10-02
update api string to latest

danewalton:


sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -20,9 +20,7 @@ static const az_span hub_client_param_separator_span = AZ_SPAN_LITERAL_FROM_STR(
 static const az_span hub_client_param_equals_span = AZ_SPAN_LITERAL_FROM_STR("=");
 
 static const az_span hub_digital_twin_model_id = AZ_SPAN_LITERAL_FROM_STR("model-id");
-static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2018-06-30");
-static const az_span hub_service_preview_api_version
-    = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2020-09-30");
+static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-version=2020-09-30");
```
ahsonkhan, line 23: When do we typically update these?

How do we make sure we don't accidentally ship with a version that is too old?

danewalton: It's determined by the service and the rollout of new features/API. The hubs are backwards compatible so if we are shipping something that has an older API string, it will still work. Just won't be able to take advantage of the newer features.

jspaith (approved):
:) !!

Issue #1352, opened 2020-09-22
Fail gracefully on invalid/incomplete JSON payload processing containing \uXXXX characters


ahsonkhan:
Broken off from https://github.com/Azure/azure-sdk-for-c/issues/717

Currently, for incomplete/invalid JSON payloads, this will access memory we shouldn't have access to:
https://github.com/Azure/azure-sdk-for-c/blob/9f6d569ded33c626b98d4512a54851c90e3ec671/sdk/src/azure/core/az_json_reader.c#L294-L312

Also add tests, along with variations to improve test coverage.

ahsonkhan:
cc @hihigupt

hihigupt:
Following are some of json input that reader is causing the crash:

[
[json_reader_crashes.zip](https://github.com/Azure/azure-sdk-for-c/files/5295651/json_reader_crashes.zip)
](url)

One of the stack trace is following, but almost all of these crashes happen at az_json_reader.c:306.

[...]

PR #1404, opened 2020-10-03
Fix bounds check while processing JSON string to avoid out-of-range access.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1352


sdk/tests/core/test_az_json.c
```diff
@@ -1531,6 +1531,18 @@ static void test_json_reader_invalid(void** state)
       AZ_SPAN_FROM_STR("{\r\n\"isActive\":false \"\r\n}"), AZ_ERROR_UNEXPECTED_CHAR);
 }
 
+static void temp_fuzz_test(az_span json, az_result expected_result)
```
ahsonkhan, line 1534: Just using this to test an example fuzz-tested JSON input in CI, will remove shortly.

vhvb1989 (approved):

PR #1405, opened 2020-10-03
Simplify should-retry logic

antkmsft:


sdk/src/azure/core/az_http_policy_retry.c
```diff
@@ -84,67 +75,77 @@ AZ_INLINE AZ_NODISCARD int32_t _az_uint32_span_to_int32(az_span span)
...
-  for (; *status_codes != _az_HTTP_STATUS_CODE_END_OF_LIST; ++status_codes)
+  if (!_az_http_policy_retry_should_retry_http_response_code(status_line.status_code))
   {
-    if (*status_codes != response_code)
-    {
-      continue;
-    }
+    *should_retry = false;
+    *retry_after_msec = -1;
+    return AZ_OK;
```
ahsonkhan, line 107: Are you sure we should be returning `AZ_OK` in this case rather than bubbling up some error `az_result` to the caller?

ahsonkhan (approved):

sdk/src/azure/core/az_http_policy_retry.c
```diff
@@ -84,67 +75,77 @@ AZ_INLINE AZ_NODISCARD int32_t _az_uint32_span_to_int32(az_span span)
   return value < INT32_MAX ? (int32_t)value : INT32_MAX;
 }
 
+AZ_INLINE AZ_NODISCARD bool _az_http_policy_retry_should_retry_http_response_code(
```
ahsonkhan, line 78: Probably outside the scope of this PR, but looks like Storage, in .NET, modifies the retry check to be different than the base implementation:
https://github.com/Azure/azure-sdk-for-net/blob/3065b59a09795d9bf14064c921e80784d34bd63e/sdk/storage/Azure.Storage.Common/src/Shared/StorageResponseClassifier.cs#L23-L44

Also, credential adds 404 as another status  code that triggers a retry:
https://github.com/Azure/azure-sdk-for-net/blob/c8a4acbe0de9eb70f9cb95af35eed5c93484ea88/sdk/identity/Azure.Identity/src/CredentialPipeline.cs#L63-L66

cc @JeffreyRichter

sdk/src/azure/core/az_http_policy_retry.c
antkmsft: Yes, in this place, we only return failures if we failed to parse the response status line.
This is a local function. If the function that calls it has exhausted all the retry attempts - then the error code gets returned to the customer.

sdk/src/azure/core/az_http_policy_retry.c
ahsonkhan: Also, the base impl of the Java retry policy is different for some reason:
https://github.com/Azure/azure-sdk-for-java/blob/126f19539f0d232ff1d6ce57bb1b34329f069324/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryStrategy.java#L39-L46

sdk/src/azure/core/az_http_policy_retry.c
ahsonkhan: > If the function that calls it has exhausted all the retry attempts - only then we fail with error.

Is that what's happening? I looks like if `should_retry` is false, we just return `AZ_OK` here:
https://github.com/Azure/azure-sdk-for-c/blob/dbb8393c63c7c7278cd30d86e3a978b797753fd6/sdk/src/azure/core/az_http_policy_retry.c#L194-L197

sdk/src/azure/core/az_http_policy_retry.c
antkmsft: Yep - too many opinions on this even in different SDK. Java also does `>= HTTP_INTERNAL_ERROR`, that we can now also do. Go is also different, BTW.
We don't have a credential now, but should we re-add it again - we would need to add `struct { _az_should_retry_fn } _internal` to the az_retry_options; or the auth policy will need to have its own retry policy, which preferably reuses the existing one in some effective manner.

sdk/src/azure/core/az_http_policy_retry.c
antkmsft: And that is fine too - we return the latest response. We fail with != AZ_OK only when there is a critical error - we can't parse the HTTP response, we can't mark the retry headers, we can't initialize the response, we can't remove the retry headers.
If after 5 retry attempts the response you get from the server is still with `AZ_HTTP_STATUS_CODE_BAD_GATEWAY` code, that response ultimately gets returned to the client.

antkmsft: .NET, for example, also behaves this way: https://github.com/Azure/azure-sdk-for-net/blob/d879da95f4c642031089ea2208f25de55231a98a/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs#L110

antkmsft:
https://github.com/Azure/azure-sdk-for-c/issues/1406

PR #1419, opened 2020-10-08
Readme edits

momuno:
formatting and update vcpkg time for windows.

### DIFF
--- a/sdk/samples/iot/README.md
+++ b/sdk/samples/iot/README.md
@@ -82,7 +82,7 @@ To run the samples, ensure you have the following programs and tools installed o
     </details>
 
 - Have [Git](https://git-scm.com/download) installed.
-- Have Microsoft [vcpkg](https://github.com/microsoft/vcpkg) package manager and [Eclipse Paho MQTT C client](https://www.eclipse.org/paho/) installed. On Linux, this installation may take an extended amount of time (~15-20 minutes).
+- Have Microsoft [vcpkg](https://github.com/microsoft/vcpkg) package manager and [Eclipse Paho MQTT C client](https://www.eclipse.org/paho/) installed. This installation may take an extended amount of time (~15-20 minutes).
 
     <details><summary><i>Instructions:</i></summary>
     <p>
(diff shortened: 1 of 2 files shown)


RickWinter (approved):

ahsonkhan (approved):

ahsonkhan:
Please wait for **all** the CI legs to get triggered and run before merging PRs :)

This was just docs, so that was fine, but we have other legs like link verification/etc. that would help catch any issues that might cause master (or other PRs) to break.

momuno:
Gotcha.  With the reset and evaluate the check-enforcer remained stuck.  What do I do then to get it unstuck?

ahsonkhan:
@mitchdenny, @weshaggard - can you please take a look at why the `c - client - ci` pipeline didn't trigger and run?
https://dev.azure.com/azure-sdk/public/_build?definitionId=722

> Gotcha. With the reset and evaluate the check-enforcer remained stuck. What do I do then to get it unstuck?

@momuno - the issue is less about overriding the check-enforcer (which just validates all ~23 legs are passing). The other 23 legs (that run the unit tests on various platforms) were skipped. There is likely an outage or Azure DevOps infra issue.

momuno:
Thanks.  Would it be best for me to revert then, or do you think this is ok since it was just a readme edit?

ahsonkhan:
> Would it be best for me to revert then, or do you think this is ok since it was just a readme edit?

But if CI isn't running, how will you merge the revert PR? 😜

In all seriousness, it's all good. I just triggered the pipeline on what's in master manually and it passed all the other ~20 legs, including link verification (not surprising given it was just a docs change).
https://dev.azure.com/azure-sdk/public/_build/results?buildId=567820&view=results
image

Just wanted to give a heads-up going forward. **Sometimes, due to outages or other infra issues, pipelines may not run as expected, and we should just wait for the issue to get resolved** (and notify folks in case they aren't aware, or even try to help them fix it, if possible).

For example, see this PR in the C++ repo which is also waiting:
https://github.com/Azure/azure-sdk-for-cpp/pull/718
image

PR #1425, opened 2020-10-14
Update Plug and Play find version API

danewalton:
Changes the `az_iot_pnp_client_property_get_property_version()` to pass the json reader by reference instead of value. Few reasons:

1. This saves us another `az_json_reader` used on the stack.
2. Passing by value leads to stack corruption on the PIC sample (for reasons I'm still investigating). But this change rectifies the problem and with the memory savings, I believe it is worth it.

The user does have to call another `az_json_reader_init()` in between but it's a minor usage change.

cc @hihigupt @jspaith

### DIFF
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -664,9 +664,9 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
  * az_iot_pnp_client_property_get_next_component_property().
  *
  * @param[in] client The #az_iot_pnp_client to use for this call.
- * @param[in] ref_json_reader The #az_json_reader used to parse through the JSON payload. An
+ * @param[in] ref_json_reader The #az_json_reader* used to parse through the JSON payload. An
  * internal copy is made to maintain the index of \p ref_json_reader.
- * @param[in] response_type The #az_iot_pnp_client_property_response_type* representing the message
+ * @param[in] response_type The #az_iot_pnp_client_property_response_type representing the message
  * type associated with the payload.
  * @param[out] out_version The numeric version of the properties in the JSON payload.
  *
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -664,14 +664,14 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
  * az_iot_pnp_client_property_get_next_component_property().
  *
  * @param[in] client The #az_iot_pnp_client to use for this call.
- * @param[in] json_reader The #az_json_reader used to parse through the JSON payload. An internal
- * copy is made to maintain the index of \p json_reader.
+ * @param[in] ref_json_reader The #az_json_reader used to parse through the JSON payload. An
+ * internal copy is made to maintain the index of \p ref_json_reader.
  * @param[in] response_type The #az_iot_pnp_client_property_response_type* representing the message
  * type associated with the payload.
  * @param[out] out_version The numeric version of the properties in the JSON payload.
  *
  * @pre \p client must not be `NULL`.
- * @pre \p json_reader must not be `NULL`.
+ * @pre \p ref_json_reader must not be `NULL`.
  * @pre \p out_version must not be `NULL`.
  *
  * @return An #az_result value indicating the result of the operation.
@@ -679,7 +679,7 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
  */
 AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
     az_iot_pnp_client const* client,
-    az_json_reader* json_reader,
+    az_json_reader* ref_json_reader,
     az_iot_pnp_client_property_response_type response_type,
     int32_t* out_version);
 
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -658,6 +658,11 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
 /**
  * @brief Read the IoT Plug and Play property version.
  *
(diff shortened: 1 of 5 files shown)


sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -674,7 +679,7 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
  */
 AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
     az_iot_pnp_client const* client,
-    az_json_reader json_reader,
+    az_json_reader* json_reader,
```
momuno, line 682: Is it possible to make this `az_json_reader const*` since it is an [in]?  I recall having trouble with that at some point because it later was cast to a void*???  Not sure if that is relevant.  Also, I recollect something about not wanting the json structure to be altered, which is way it was originally pass by value.  Would const alleviate this (I haven't looked at the az_json_reader structure)?

momuno (approved):

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -658,10 +658,15 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
 /**
  * @brief Read the IoT Plug and Play property version.
  *
+ * @warning This modifies the state of the json reader. To then use the same json reader
+ * with az_iot_pnp_client_property_get_next_component_property(), you must call
+ * az_json_reader_init() again after this call and before the call to
```
ahsonkhan, line 663: How? What if I don't have the JSON data to pass to `_init()` anymore and all I have is the reader?

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -658,10 +658,15 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
...
+ * az_iot_pnp_client_property_get_next_component_property().
+ *
  * @param[in] client The #az_iot_pnp_client to use for this call.
  * @param[in] json_reader The #az_json_reader used to parse through the JSON payload. An internal
  * copy is made to maintain the index of \p json_reader.
- * @param[in] response_type The #az_iot_pnp_client_property_response_type representing the message
+ * @param[in] response_type The #az_iot_pnp_client_property_response_type* representing the message
```
ahsonkhan, line 669: Did you intend to make this doc change? API/docs don't match anymore.

Also, maybe word it as `The pointer to an #az_...`

sdk/inc/azure/iot/az_iot_pnp_client.h
danewalton: yes you are right it should be `const`

sdk/inc/azure/iot/az_iot_pnp_client.h
danewalton: woops yea edited the wrong parameter

ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_pnp_client.h
ahsonkhan: The question isn't about the overhead, it is about feasibility.  If I, as a the caller, only have a json reader (someone else gave  it to me, who also owns the data and promises to keep it live), how am I supposed to re-create the json reader by calling az_json_reader_init?

> If you don't have access to the data anymore you're kinda out of luck.

And we are OK with that behavior for the user as a trade-off? Seems like a decision to over-optimize for stack size given the potential negative impact for the caller.

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -658,23 +658,28 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_builder_end_reported_status(
...
  *
+ * @warning This modifies the state of the json reader. To then use the same json reader
+ * with az_iot_pnp_client_property_get_next_component_property(), you must call
+ * az_json_reader_init() again after this call and before the call to
+ * az_iot_pnp_client_property_get_next_component_property().
+ *
  * @param[in] client The #az_iot_pnp_client to use for this call.
- * @param[in] json_reader The #az_json_reader used to parse through the JSON payload. An internal
- * copy is made to maintain the index of \p json_reader.
+ * @param[in] ref_json_reader The #az_json_reader* used to parse through the JSON payload. An
```
ahsonkhan, line 667:
```suggestion
 * @param[in] ref_json_reader The pointer to an #az_json_reader used to parse through the JSON payload. An
```

sdk/inc/azure/iot/az_iot_pnp_client.h
momuno: By that do you mean you expect the az_json_reader to be altered?  In that case, would the header description be [in,out] ?
@antkmsft Can you please refresh me on this lol. I don't know if what I'm recalling is accurate.

sdk/inc/azure/iot/az_iot_pnp_client.h
momuno: [in,out] question above. should have put it here. @antkmsft ?

sdk/inc/azure/iot/az_iot_pnp_client.h
danewalton: Good question

If a caller has access to the json reader, in a state that is as required, then they themselves can make a copy via
```c
az_json_reader copy_json = *passed_json;
```
before they call the first API. At that point, they are taking the same hit on stack usage as before. They can use a separate json reader for each call at that point.

sdk/inc/azure/iot/az_iot_pnp_client.h
danewalton: ahh crap yea you're right. I'll send another

sdk/inc/azure/iot/az_iot_pnp_client.h
ahsonkhan: Gotcha. In that case, let's document that explicitly (make a copy of the reader) rather than asking the caller to call `init`. Or maybe have both, if you really want to call out the init use-case where they own/have access to the data.

PR #1461, opened 2020-10-27
Adjust the payload from aziot_esp8266.ino to work with Azure TSI

ewertons:
This PR modifies the payload sent by the esp8266 sample to send a json content.

### DIFF
--- a/sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
+++ b/sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
@@ -42,6 +42,8 @@ static unsigned char encrypted_signature[32];
 static char base64_decoded_device_key[32];
 static unsigned long next_telemetry_send_time_ms = 0;
 static char telemetry_topic[128];
+static uint8_t telemetry_payload[100];
+static uint32_t telemetry_send_count = 0;
 
 static void connectToWiFi()
 {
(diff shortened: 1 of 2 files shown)


sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
```diff
@@ -268,6 +268,19 @@ void setup()
   establishConnection();
 }
 
+static uint8_t m_payload[100];
+static uint32_t m_msgCount = 0;
+static char* getTelemetryPayload()
```
momuno, line 273: Curiosity, since m_msgCount is only used within the getTelemetryPayload(), why not declare it inside the function to keep its scope local?

momuno (approved):

sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
ewertons: Because we want to keep tracking the count throughout the application lifetime. 
If we define it inside the function, the counter will get lost every time the function is called.

momuno: not if it is static though.

sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
```diff
@@ -268,6 +268,19 @@ void setup()
   establishConnection();
 }
 
+static uint8_t m_payload[100];
```
ericwol-msft, line 271: What does m_ mean?

ericwol-msft (approved):

sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
ewertons: Good job getting the info.
For the sake of keeping the pattern we have used before, I'll keep the variable defined outside the function. I'll move it to the top though.

sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino
ewertons: Old syntax for "member". I was going to rename it anyway, just waiting for Wellington to verify the change against Time Series Insights.

wduraes (approved):

Issue #1471, opened 2020-11-03
az_iot_message_properties_next fails if properties buffer is larger than the properties string it contains


ewertons:
**Describe the bug**

Mark Radbourne reported having a precondition failure when using az_iot_message_properties_next:

```c
>	sample1.exe!az_precondition_failed_default(...) Line 12	C
 	sample1.exe!_az_span_diff(az_span sliced_span, az_span original_span) Line 42	C
 	sample1.exe!az_iot_message_properties_next(az_iot_message_properties * properties, az_span * out_name, az_span * out_value) Line 136	C
```
String used by Mark:
Byte array containing "Mark=wasHere&Mark=wasHere" at position 0; the variable is larger than its content.

**To Reproduce**

Code version: master@azure-sdk-for-c (7f45b34b88e92806be292ddf60e3a34a19287d87)

The following code was used for a repro:

```c
int main(void)
{
  az_span string = AZ_SPAN_LITERAL_FROM_STR("Mark=wasHere&Mark=wasHere");
  
  uint8_t a[64];
  az_span s = AZ_SPAN_FROM_BUFFER(a);
  
  az_span_copy(s, string);
  
  az_iot_message_properties p;
  (void)az_iot_message_properties_init(&p, s, 25);

  az_span property_name;
  az_span property_value;
  while (az_result_succeeded(az_iot_message_properties_next(&p, &property_name, &property_value)))
  {
    printf(
        "%.*s: %.*s\r\n",
        (int)az_span_size(property_name),
        (unsigned char*)az_span_ptr(property_name),
        (int)az_span_size(property_value),
        (unsigned char*)az_span_ptr(property_value));
  }


  return 0;
}
```

**Analysis and Root Cause**

In the code above, right before calling az_iot_message_properties_next, these are the values of the local variables:

```
		az_iot_message_properties_init returned	65536	int
+		a	0x00000079422ffc20 "Mark=wasHere&Mark=wasHereÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌÌ...	unsigned char[64]
[...]

PR #1472, opened 2020-11-03
Fix for az_iot_message_properties_next if properties buffer is larger than string it contains (#1471)

ewertons:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1471

### DIFF
--- a/sdk/tests/iot/common/test_az_iot_common.c
+++ b/sdk/tests/iot/common/test_az_iot_common.c
@@ -145,6 +145,47 @@ static void test_az_iot_message_properties_next_NULL_out_value_fail(void** state
   ASSERT_PRECONDITION_CHECKED(az_iot_message_properties_next(&props, &name, NULL));
 }
 
+static void test_az_iot_message_properties_next_written_less_than_size_succeed(void** state)
+{
+  (void)state;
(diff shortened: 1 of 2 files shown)


sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -145,6 +145,48 @@ static void test_az_iot_message_properties_next_NULL_out_value_fail(void** state
   ASSERT_PRECONDITION_CHECKED(az_iot_message_properties_next(&props, &name, NULL));
 }
 
+static void test_az_iot_message_properties_next_written_less_than_size_succeed(void** state)
+{
+  (void)state;
+  #define STRSIZE sizeof(TEST_KEY_VALUE_THREE)/sizeof(TEST_KEY_VALUE_THREE[0])
```
danewalton, line 151: nit: any reason to not have this just be a var

danewalton (approved):

sdk/tests/iot/common/test_az_iot_common.c
ewertons: No specific reason at all.

danewalton: just looks a little odd to have in a function

PR #1494, opened 2020-11-11
Add pnp feature to the version name to keep them unique

ahsonkhan:


danewalton (approved):
seems reasonable to me. should it be `1.1.0-pnp.beta.3` to stay ahead of the master version?

ahsonkhan:
I don't think so. I think master will move to beta.3 as needed, later.

Since we shipped beta.1 already, the next in line is beta.2 :)

PR #1500, opened 2020-11-11
Revert "Update semver validation regex to be more flexible and allow the prerelease label to contain a "feature." string"

ahsonkhan:
Now that we have published the docs for `1.1.0-pnp.beta.2`, reverts the temporary workaround from Azure/azure-sdk-for-c#1497

If we want to support `<feature>.<beta>.<rev>` versioning long term, we should follow the [eng/common workflow](https://github.com/Azure/azure-sdk-tools/blob/2c58ce43e5133a9dfa681652151e41d09e25f5c5/doc/common/common_engsys.md#workflow) and make such a fix again using that approach, starting in `azure-sdk-tools`.

### DIFF
--- a/eng/common/scripts/SemVer.ps1
+++ b/eng/common/scripts/SemVer.ps1
@@ -23,7 +23,7 @@ class AzureEngSemanticVersion {
   [bool] $IsSemVerFormat
   [string] $DefaultPrereleaseLabel
   # Regex inspired but simplified from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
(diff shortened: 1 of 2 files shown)


mitchdenny:
@weshaggard I think we are going to need to revise our versioning guidelines to accommodate scenarios where we have a 1:1 mapping between package and repository. The ```1.0.0-beta.1``` format doesn't leave room for teams to stabilize new features on a branch and ship previews to customers.

We wouldn't want to merge a feature onto master until we are confident that it wouldn't stop subsequent releases from the master branch (the alternative is long lived diverging servicing branches which is probably less desirable).

We are likely to have the same issue with the iOS repository.

weshaggard (approved):

weshaggard:
Thanks for reverting these as these changes are definitely incomplete. There is a teams conversation going on about potential solutions for this issue but until we come up with that plan we should avoid release any versions that don't match our versioning guidelines otherwise we will be fighting a lot of different things not just these 2 changes.

PR #1511, opened 2020-11-13
PnP API Update - az_iot_pnp_client_property_get_next_component_property

danewalton:
This updates the `az_iot_pnp_client_property_get_next_component_property()` API to only require one JSON reader. This cuts the stack usage down from ~200 bytes to ~100 bytes. 

The trade off is that the only JSON reader is shared between the API and the user to read the properties. That means that if the user doesn't iterate correctly through the JSON, the API won't be able to recover. There is therefore more responsibility on the user to correctly process the properties and deliver it back to the API in an expected location. However, in order to properly read the properties in the first place, they would have had to already correctly call the JSON APIs. That means that the overhead is one call to `az_json_reader_next_token()` in the case of properties they care about and the following in the case they don't (which is detailed in the docs):

```c
 az_json_reader_skip_children(&jr);
 az_json_reader_next_token(&jr)
```

### DIFF
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -742,6 +742,7 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK If the function returned a valid #az_json_reader pointing to the property name and
  * the #az_span with a component name.
+ * @retval #AZ_ERROR_JSON_INVALID_STATE If the json reader is passed in at an unexpected location.
  * @retval #AZ_ERROR_IOT_END_OF_PROPERTIES If there are no more properties left for the component.
  */
 AZ_NODISCARD az_result az_iot_pnp_client_property_get_next_component_property(
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -711,6 +711,7 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
  *     az_json_reader_next_token(&jr);
  *
  *     // Get the property value here
+ *     // Example: az_json_token_get_int32(&jr.token, &user_int);
  *
  *     // Skip to next property value
  *     az_json_reader_next_token(&jr);
--- a/sdk/inc/azure/iot/az_iot_pnp_client.h
+++ b/sdk/inc/azure/iot/az_iot_pnp_client.h
@@ -685,29 +685,59 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
     int32_t* out_version);
 
 /**
- * @brief Iteratively Read the IoT Plug and Play component properties.
+ * @brief Iteratively read the IoT Plug and Play component properties.
  *
  * Note that between calls, the #az_span pointed to by \p out_component_name shall not be modified,
  * only checked and compared. Internally, the #az_span is only changed if the component name changes
  * in the JSON document and is not necessarily set every invocation of the function.
  *
- * The `out_property_name_and_value` is passed back as an #az_json_reader. On success, the reader
- * will be pointing to the property name. After checking the property name, the reader can be
- * advanced to the property value by calling az_json_reader_next_token().
+ * On success, the `ref_json_reader` will be set on a valid property name. After checking the
+ * property name, the reader can be advanced to the property value by calling
+ * az_json_reader_next_token(). Note that on the subsequent call to this API, it is expected that
+ * the json reader will be placed AFTER the read property name and value. That means that after
+ * reading the property value (including single values or complex objects), the user must call
+ * az_json_reader_next_token().
+ *
+ * Below is a code snippet which you can use as a starting point:
+ *
+ * @code
+ *
+ * while (az_result_succeeded(az_iot_pnp_client_property_get_next_component_property(
+ *       &pnp_client, &jr, response_type, &component_name)))
+ * {
+ *   // Check if property is of interest (sub user_property for your own)
+ *   if (az_json_token_is_text_equal(
+ *           &jr.token, user_property))
+ *   {
+ *     az_json_reader_next_token(&jr);
+ * 
+ *     // Get the property value here
+ * 
+ *     // Skip to next property value
+ *     az_json_reader_next_token(&jr);
+ *   }
+ *   else
+ *   {
+ *     // The JSON reader must be advanced regardless of whether the property
+ *     // is of interest or not.
+ *     // Skip children in case the property value is an object
+ *     az_json_reader_skip_children(&jr);
+ *     az_json_reader_next_token(&jr);
+ *   }
+ * }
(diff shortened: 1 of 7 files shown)


sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -685,41 +685,71 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
...
- * @pre \p out_property_name_and_value must not be `NULL`. It must point to an #az_json_reader
- * instance.
  *
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK If the function returned a valid #az_json_reader pointing to the property name and
  * the #az_span with a component name.
+ * @retval #AZ_ERROR_JSON_INVALID_STATE If the json reader is passed in at an unexpected location.
  * @retval #AZ_ERROR_IOT_END_OF_PROPERTIES If there are no more properties left for the component.
  */
 AZ_NODISCARD az_result az_iot_pnp_client_property_get_next_component_property(
```
hihigupt, line 748: >next_ [](start = 54, length = 5)

Now that we expect user to do next, do you think next in the API is adding confusion?

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -685,41 +685,71 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
...
+ * Below is a code snippet which you can use as a starting point:
+ *
+ * @code
+ *
+ * while (az_result_succeeded(az_iot_pnp_client_property_get_next_component_property(
+ *       &pnp_client, &jr, response_type, &component_name)))
+ * {
+ *   // Check if property is of interest (sub user_property for your own)
+ *   if (az_json_token_is_text_equal(
+ *           &jr.token, user_property))
```
ericwol-msft, line 710: does this really need to wrap?

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -685,41 +685,71 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
...
+ * az_json_reader_next_token().
+ *
+ * Below is a code snippet which you can use as a starting point:
+ *
+ * @code
+ *
+ * while (az_result_succeeded(az_iot_pnp_client_property_get_next_component_property(
+ *       &pnp_client, &jr, response_type, &component_name)))
+ * {
+ *   // Check if property is of interest (sub user_property for your own)
```
ericwol-msft, line 708:
```suggestion
 *   // Check if property is of interest (substitute user_property for your own property name)
```

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -685,41 +685,71 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
...
+ *   if (az_json_token_is_text_equal(
+ *           &jr.token, user_property))
+ *   {
+ *     az_json_reader_next_token(&jr);
+ *
+ *     // Get the property value here
+ *
```
ericwol-msft, line 715: add an example here... string?

sdk/samples/iot/paho_iot_pnp_sample.c
```diff
@@ -591,6 +589,37 @@ static void process_device_property_message(
         send_reported_property(
             property_reported_maximum_temperature_name, device_maximum_temperature, -1, confirm);
       }
+
+      // Skip to next property value
+      rc = az_json_reader_next_token(&jr);
+      if (az_result_failed(rc))
+      {
+        IOT_SAMPLE_LOG_ERROR("Could not move to next property name");
```
ericwol-msft, line 597:
```suggestion
        IOT_SAMPLE_LOG_ERROR("Invalid JSON. Could not move to next property name");
```

sdk/samples/iot/paho_iot_pnp_sample.c
```diff
@@ -591,6 +589,37 @@ static void process_device_property_message(
...
+    }
+    else
+    {
+      IOT_SAMPLE_LOG_AZ_SPAN("Unknown Property Received:", jr.token.slice);
+      // The JSON reader must be advanced regardless of whether the property
+      // is of interest or not.
+      rc = az_json_reader_next_token(&jr);
+      if (az_result_failed(rc))
+      {
+        IOT_SAMPLE_LOG_ERROR("Could not move to next property value");
```
ericwol-msft, line 608:
```suggestion
        IOT_SAMPLE_LOG_ERROR("Invalid JSON. Could not move to next property value");
```

sdk/samples/iot/paho_iot_pnp_sample.c
```diff
@@ -591,6 +589,37 @@ static void process_device_property_message(
...
+      }
+
+      // Skip children in case the property value is an object
+      rc = az_json_reader_skip_children(&jr);
+      if (az_result_failed(rc))
+      {
+        IOT_SAMPLE_LOG_ERROR("Could not skip children");
```
ericwol-msft, line 615:
```suggestion
        IOT_SAMPLE_LOG_ERROR("Invalid JSON. Could not skip children");
```

sdk/samples/iot/paho_iot_pnp_sample.c
```diff
@@ -591,6 +589,37 @@ static void process_device_property_message(
...
+      }
+
+      rc = az_json_reader_next_token(&jr);
+      if (az_result_failed(rc))
+      {
+        IOT_SAMPLE_LOG_ERROR("Could not move to next property name");
```
ericwol-msft, line 621:
```suggestion
        IOT_SAMPLE_LOG_ERROR("Invalid JSON. Could not move to next property name");
```

ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_pnp_client.h
```diff
@@ -685,41 +685,71 @@ AZ_NODISCARD az_result az_iot_pnp_client_property_get_property_version(
...
+ *     // Skip to next property value
+ *     az_json_reader_next_token(&jr);
+ *   }
+ *   else
+ *   {
+ *     // The JSON reader must be advanced regardless of whether the property
+ *     // is of interest or not.
+ *     // Skip children in case the property value is an object
+ *     az_json_reader_skip_children(&jr);
+ *     az_json_reader_next_token(&jr);
```
hihigupt, line 725: this should be :
next_token;  // for property name
next_token/skip_children; // for value

danewalton: This stems from this:
https://github.com/danewalton/azure-sdk-for-c/pull/10#discussion_r517729633

I can add the skip here too to be explicit.

sdk/inc/azure/iot/az_iot_pnp_client.h
hihigupt: I think next is source of confusion... 

---
In reply to: [523178570](https://github.com/Azure/azure-sdk-for-c/pull/1511#discussion_r523178570) [](ancestors = 523178570)

PR #1520, opened 2020-11-16
Merge master into feature/iot_pnp

danewalton:
This needs a little patchwork on the iot sample readme which I have in another branch. Will send after this PR to keep this on topic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,19 @@
 # Release History
 
-## 1.1.0-beta.3 (Unreleased)
+## 1.1.0-beta.4 (Unreleased)
 
+### New Features
+
+
+### Bug Fixes
+
+- [[#1472]](https://github.com/Azure/azure-sdk-for-c/pull/1472) Fix `az_iot_message_properties_next()` when the buffer in which the properties were kept was bigger than the length of characters in the buffer.
+
+### Other Changes and Improvements
+
+- [[#1473]](https://github.com/Azure/azure-sdk-for-c/pull/1473) Add remove server certificate validation on paho and ESP8266 samples.
+- [[#1449]](https://github.com/Azure/azure-sdk-for-c/pull/1449) Add basic reconnection capability for the ESP8266 sample.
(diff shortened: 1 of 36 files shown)


ericwol-msft (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
 # Release History
 
-## 1.1.0-beta.3 (Unreleased)
+## 1.1.0-beta.4 (Unreleased)
```
ahsonkhan, line 3: We don't want to bump this version. `.3` in the feature branch was correct. @weshaggard - what do you suggest we do here?

The workflow in this feature branch includes frequent pull/merge from master. Should we limit those to only prior to a beta release?

CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
...
+
+### Bug Fixes
+
+- [[#1472]](https://github.com/Azure/azure-sdk-for-c/pull/1472) Fix `az_iot_message_properties_next()` when the buffer in which the properties were kept was bigger than the length of characters in the buffer.
+
+### Other Changes and Improvements
+
+- [[#1473]](https://github.com/Azure/azure-sdk-for-c/pull/1473) Add remove server certificate validation on paho and ESP8266 samples.
+- [[#1449]](https://github.com/Azure/azure-sdk-for-c/pull/1449) Add basic reconnection capability for the ESP8266 sample.
+- [[#1490]](https://github.com/Azure/azure-sdk-for-c/pull/1490) Fix static analyzer flagging of non-checked return value in `az_iot_hub_client_c2d_parse_received_topic()`.
```
ahsonkhan, line 16: @danewalton These Changelog changes are incorrect and duplicated from the beta.2 release below.

CHANGELOG.md
weshaggard: Unfortunately there isn't any magic solution here you will just have to carefully watch the versions of the components that you plan to potentially ship in the feature branch and ensure the version matches that versioning plan. 

I would suggest you take pretty frequent changes as we don't want a big surprise late in the process if the merge has any issues.

CHANGELOG.md
danewalton: #1523

PR #1564, opened 2021-01-12
Enable 1ES hosted pools.

mitchdenny:
This PR moves the Windows and Linux jobs over to the 1ES hosted pools (this has already been done for C#, Java, Python, and JS. The changes are roughly as follows:

1. Added WindowsPool and LinuxPool parameters to stage template and job template.
2. Added variables which contain the WindowsPool and LinuxPool values.
3. Used variables in the test matrix.
4. Other minor changes to support above.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -41,7 +41,8 @@ jobs:
   strategy:
     matrix:
       LinuxARM_Release_Logging:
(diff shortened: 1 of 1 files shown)


mitchdenny:
Before this can be merged I need to bring up a 1604 pool to cover those legs.

RickWinter (approved):

mitchdenny:
Not quite ready to merge yet @RickWinter, but my subsequent changes to this will have the same effect.

weshaggard (approved):

PR #1575, opened 2021-01-22
Adding a facility for uLib to az_result.

mamokarz:


sdk/inc/azure/core/az_result.h
```diff
@@ -28,6 +28,7 @@ enum
   _az_FACILITY_CORE_HTTP = 0x4,
   _az_FACILITY_IOT = 0x5,
   _az_FACILITY_IOT_MQTT = 0x6,
+  _az_FACILITY_ULIB = 0x7,
```
ahsonkhan, line 31: FWIW, these are meant to be used internal to the SDK only, since they can be renamed or changed since we don't consider anything starting with `_az` to be part of the public surface area (and hence don't guarantee breaking changes).

For context, how do you plan to consume/use this?

mamokarz: This is related to the discussion that we have a couple of months ago with @JeffreyRichter. Az_result was created to be the result messages in Azure, not only in the SDKs. ULIB ([https://github.com/Azure/azure-ulib-c](https://github.com/Azure/azure-ulib-c)) is one of the cases where the code is Azure, but not SDK.

---
In reply to: [562958058](https://github.com/Azure/azure-sdk-for-c/pull/1575#discussion_r562958058) [](ancestors = 562958058)

JeffreyRichter: Yes, Marcos & I discussed this a few months ago and him adding this facility code is the proper thing to do.

JeffreyRichter (approved):

ahsonkhan (approved):
From offline discussion, the consumption model is by source copy, which works for consuming internal-only APIs since it decouples the ship trains.

PR #1580, opened 2021-01-27
wrap CMP0091 in MSVC check

### LINKED ISSUE #1514: SDK no longer builds on Ubuntu 18.04
CMakeLists.txt file in repository root now has [the line](https://github.com/Azure/azure-sdk-for-c/blob/ae9548d71d2b39df1d52eb2c01fd52a92d171ea7/CMakeLists.txt#L49):
```cmake
cmake_policy(SET CMP0091 NEW) # https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
```
When attempting to build the SDK on Ubuntu 18.04 this causes the following failure:
```
markrad@markradedge:~/source/azure-sdk-for-c/build$ cmake ..
CMake Error at CMakeLists.txt:49 (cmake_policy):
  Policy "CMP0091" is not known to this version of CMake. 
-- The C compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Read version from /home/markrad/source/azure-sdk-for-c/sdk/inc/azure/core/az_version.h
-- AZ_SDK_VERSION_MAJOR 1
-- AZ_SDK_VERSION_MAJOR 1
-- AZ_SDK_VERSION_PATCH 0
-- AZ_SDK_VERSION_PRERELEASE beta.2
-- Configuring incomplete, errors occurred!
See also "/home/markrad/source/azure-sdk-for-c/build/CMakeFiles/CMakeOutput.log".
```
The problem here is twofold. First this particular policy is only supported in cmake 3.15 and yet the CMakeLists.txt file indicates that the minimum version required is __3.10__. From the file in question, [line 4](https://github.com/Azure/azure-sdk-for-c/blob/ae9548d71d2b39df1d52eb2c01fd52a92d171ea7/CMakeLists.txt#L4):
```cmake
```
The other problem, though not necessarily a bug is that the latest version of cmake available for Ubuntu 18:04 is __3.10.2__ as shown by:
```bash
markrad@markradedge:~/source/azure-embedded-sdk-sys/azure-sdk-for-c$ apt search cmake | grep ^cmake/

cmake/bionic-updates,now 3.10.2-1ubuntu2.18.04.1 amd64 [installed]
```
This is further frustrating because the policy only relevant to Windows builds so this could have been easily avoided.

**To Reproduce**
Clone repository and build as documented.

**Expected behavior**
Either the cmake script should accurately specify the version of cmake required or the cmake facilities that use higher versions should be removed. The latter would be my preferred solution since a lot of people are still on Ubuntu 18.04.

- OS: Ubuntu 18.04
- IDE: None - built from command line
- Commit: ae9548d71d2b39df1d52eb2c01fd52a92d171ea7

**Information Checklist**
Please make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report

- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

danewalton:
fixes #1514


vhvb1989 (approved):

antkmsft (approved):

CMakeLists.txt
```diff
@@ -44,11 +44,13 @@ if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
...
-cmake_policy(SET CMP0091 NEW) # https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
-if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
-  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+if(MSVC)
+  # Use the static runtime libraries when building statically for consistency with vcpkg's
+  # arch-windows-static triplets:
+  cmake_policy(SET CMP0091 NEW) # https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
+  if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+  endif()
```
ahsonkhan, line 53: I am not sure why the windows CI legs are now failing with that linker warning on Windows:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=705929&view=logs&j=1822ef0b-3ac4-586f-233b-6fdf166a140f&t=b129bef7-7ee0-56b0-68be-0867e926d36b&l=102
```text
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library [D:\vss\_work\1\s\build\sdk\tests\iot\provisioning\az_iot_provisioning_test.vcxproj]
LINK : error LNK1218: warning treated as error; no output file generated [D:\vss\_work\1\s\build\sdk\tests\iot\provisioning\az_iot_provisioning_test.vcxproj]
##[error]Cmd.exe exited with code '1'.
```

Presumable `if (MSVC)` condition is returning false, but I don't know why. I wonder if you need to do something like this instead: `if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")`. Try some tests @danewalton using `message(...)` to see if we actually enter those code blocks.

@BillyONeal maybe you have some ideas.

antkmsft: I know! `MSVC` us not defined before the line saying `project(az LANGUAGES C)`. Try moving the check after that line, it should work.

antkmsft:
MSVC is not defined before the `project(... C)`, but the `CMP0091` and `CMAKE_MSVC_RUNTIME_LIBRARY` things that we do has to be done before the `project(... C)`, or they have no effect. The real problem even on ubuntu is that we use a feature from cmake 3.15, while claiming that we need cmake 3.10, so when customer who has exactly cmake 3.10 tries to build it, even for Ubuntu, there are errors. The proper fix would be to either bump up the cmake version to 3.15, or to introduce an optional build parameter that only has effect during the MSVC builds and indicates which runtime the user wants to link with. I've seen other libraries using that approach, I think we should do it too. We have the same problem that needs to be solved for the C++ repo (https://github.com/Azure/azure-sdk-for-cpp/pull/1223).

danewalton:
@ahsonkhan @antkmsft thoughts here? Latest addition seems to be working

danewalton:
> @ahsonkhan @antkmsft thoughts here? Latest addition seems to be working

just checked with Ubuntu 18.04 and looks to be working

image

eng/cmake/global_compile_options.txt
```diff
@@ -5,6 +5,12 @@ if(MSVC)
   endif()
 
   add_compile_options(/W4 ${WARNINGS_AS_ERRORS_FLAG} /wd5031 /wd4668 /wd4820 /wd4255 /wd4710 /analyze)
+
+  add_compile_options(
+    $<$<CONFIG:>:/MT> #---------|
+    $<$<CONFIG:Debug>:/MTd> #---|-- Statically link the runtime libraries
+    $<$<CONFIG:Release>:/MT> #--|
+  )
```
ahsonkhan, line 13: @barcharcraz and @BillyONeal - can you please help review and sign-off on this approach.

CMakeLists.txt
```diff
@@ -44,16 +44,18 @@ if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
...
 enable_testing ()
 
+if(MSVC OR WIN32)
+  # Use the static runtime libraries when building statically for consistency with vcpkg's
+  # arch-windows-static triplets:
+  cmake_policy(SET CMP0091 NEW) # https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html
+  if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+  endif()
+endif()
```
ahsonkhan, line 57: If we are explicitly adding some compile options in `global_compile_options`, do we need to set `cmake_policy` anymore?

eng/cmake/global_compile_options.txt
danewalton: Just for reference I pulled it from here:
https://gitlab.kitware.com/cmake/cmake/-/issues/18390

CMakeLists.txt
danewalton: testing

RickWinter (approved):

danewalton:
Merging for now but let me know if this needs to be changed

CMakeLists.txt
```diff
@@ -44,13 +44,6 @@ if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
   set(VCPKG_TARGET_TRIPLET "$ENV{VCPKG_DEFAULT_TRIPLET}" CACHE STRING "")
 endif()
 
-# Use the static runtime libraries when building statically for consistency with vcpkg's
```
ahsonkhan, line 47: Consider including this fix in a  changelog entry.

ahsonkhan: done in https://github.com/Azure/azure-sdk-for-c/pull/1605

Issue #1263, opened 2020-09-14
Add preconditions to the doxygen docs


danewalton:
Discovered the `@pre` tag to list preconditions for functions. Would look something like:

![image](https://user-images.githubusercontent.com/36710865/93136030-1d913300-f690-11ea-9574-669ea384e3da.png)

![image](https://user-images.githubusercontent.com/36710865/93135976-08b49f80-f690-11ea-9100-fc3fc62d9d04.png)

Is this something we would want to add to all the API's?

ahsonkhan:
Nice!

> Is this something we would want to add to all the API's?

I think we should add them. Good idea.

ahsonkhan:
@RickWinter, @antkmsft, @vhvb1989, @JeffreyRichter - thoughts?

JeffreyRichter:
I really like this too!

danewalton:
Tackling IOT side of it with #1619

PR #1619, opened 2021-02-25
add precondition documentation to iot apis

danewalton:
tackles the iot side of #1263

### DIFF
--- a/sdk/inc/azure/iot/az_iot_common.h
+++ b/sdk/inc/azure/iot/az_iot_common.h
@@ -172,6 +172,9 @@ typedef struct
  * with the following format: {name}={value}&{name}={value}.
  * @param[in] written_length The length of the properly formatted properties already initialized
  * within the buffer. If the \p buffer is unfilled (uninitialized), this should be 0.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p buffer must be a valid span of size greater than 0.
+ * @pre \p written_length must be greater than or equal to 0.
  * @return An #az_result value indicating the result of the operation.
  */
 AZ_NODISCARD az_result az_iot_message_properties_init(
@@ -194,6 +197,9 @@ AZ_NODISCARD az_result az_iot_message_properties_init(
  * @param[in] properties The #az_iot_message_properties to use for this call.
  * @param[in] name The name of the property. Must be a valid, non-empty span.
  * @param[in] value The value of the property. Must be a valid, non-empty span.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p name must be a valid span of size greater than 0.
+ * @pre \p value must be a valid span of size greater than 0.
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK The operation was performed successfully.
  * @retval #AZ_ERROR_NOT_ENOUGH_SPACE There was not enough space to append the property.
@@ -211,6 +217,9 @@ AZ_NODISCARD az_result az_iot_message_properties_append(
  * @param[in] properties The #az_iot_message_properties to use for this call.
  * @param[in] name The name of the property to search for.
  * @param[out] out_value An #az_span containing the value of the found property.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p name must be a valid span of size greater than 0.
+ * @pre \p out_value must not be `NULL`.
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK The property was successfully found.
  * @retval #AZ_ERROR_ITEM_NOT_FOUND The property could not be found.
@@ -226,6 +235,9 @@ AZ_NODISCARD az_result az_iot_message_properties_find(
  * @param[in] properties The #az_iot_message_properties to use for this call.
  * @param[out] out_name A pointer to an #az_span containing the name of the next property.
  * @param[out] out_value A pointer to an #az_span containing the value of the next property.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p out_name must not be `NULL`.
+ * @pre \p out_value must not be `NULL`.
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK A property was retrieved successfully.
  * @retval #AZ_ERROR_IOT_END_OF_PROPERTIES The API reached the end of the properties to retrieve.
@@ -268,6 +280,11 @@ AZ_NODISCARD AZ_INLINE bool az_iot_status_retriable(az_iot_status status)
  * @param[in] max_retry_delay_msec The maximum time, in milliseconds, to wait before a retry.
  * @param[in] random_jitter_msec A random value between 0 and the maximum allowed jitter, in
  * milliseconds.
+ * @pre \p operation_msec must be between 0 and INT32_MAX - 1.
+ * @pre \p attempt must be between 0 and INT16_MAX - 1.
+ * @pre \p min_retry_delay_msec must be between 0 and INT32_MAX - 1.
+ * @pre \p max_retry_delay_msec must be between 0 and INT32_MAX - 1.
+ * @pre \p random_jitter_msec must be between 0 and INT32_MAX - 1.
  * @return The recommended delay in milliseconds.
  */
 AZ_NODISCARD int32_t az_iot_calculate_retry_delay(
--- a/sdk/inc/azure/iot/az_iot_hub_client.h
+++ b/sdk/inc/azure/iot/az_iot_hub_client.h
@@ -398,7 +398,7 @@ typedef struct
  * @param[in] received_topic An #az_span containing the received topic.
  * @param[out] out_request If the message is a method request, this will contain the
  * #az_iot_hub_client_method_request.
- * @pre \p client must not be `NULL`.
+ * @pre \p client must not be `NULL` and must already be initialized by first calling az_iot_hub_client_init().
  * @pre \p received_topic must be a valid span of size greater than 0.
  * @pre \p out_request must not be `NULL`.
  * @return An #az_result value indicating the result of the operation.
@@ -425,7 +425,7 @@ AZ_NODISCARD az_result az_iot_hub_client_methods_parse_received_topic(
  * @param[in] mqtt_topic_size The size, in bytes of \p mqtt_topic.
  * @param[out] out_mqtt_topic_length __[nullable]__ Contains the string length, in bytes, of \p
  * mqtt_topic. Can be `NULL`.
- * @pre \p client must not be `NULL`.
+ * @pre \p client must not be `NULL` and must already be initialized by first calling az_iot_hub_client_init().
  * @pre \p request_id must be a valid span of size greater than 0.
  * @pre \p mqtt_topic must not be `NULL`.
  * @pre \p mqtt_topic_size must be greater than 0.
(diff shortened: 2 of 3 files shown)


sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -174,6 +188,11 @@ AZ_NODISCARD az_result az_iot_provisioning_client_sas_get_signature(
  * @param[in] mqtt_password_size The size, in bytes of \p mqtt_password.
  * @param[out] out_mqtt_password_length __[nullable]__ Contains the string length, in bytes, of \p
  * mqtt_password. Can be `NULL`.
+ * @pre \p client must not be `NULL`.
+ * @pre \p base64_hmac_sha256_signature must be a valid span of size greater than 0.
```
danewalton, line 192: note here that we had switched `base64_hmac_sha256_signature` and `token_expiration_epoch_time` from recommendation of the SDK team so that the two SAS key APIs match (token expiration is second). This was never changed but we might just have to live with it.
https://github.com/Azure/azure-sdk-for-c/blob/66cd00042a12466d998811c9c504615f0d5afd29/sdk/src/azure/iot/az_iot_hub_client_sas.c#L85-L86

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -322,6 +341,10 @@ typedef struct
  * @param[in] received_payload An #az_span containing the received MQTT payload.
  * @param[out] out_response If the message is register-operation related, this will contain the
  * #az_iot_provisioning_client_register_response.
+ * @pre \p client must not be `NULL`.
+ * @pre \p received_topic must be a valid span of size greater than or equal to 0.
+ * @pre \p received_payload must be a valid span of size greater than or equal to 0.
```
danewalton, line 346: note on these two. These are currently written to the spec of the API here:
https://github.com/Azure/azure-sdk-for-c/blob/66cd00042a12466d998811c9c504615f0d5afd29/sdk/src/azure/iot/az_iot_provisioning_client.c#L521-L522
I don't think we have any other cases where we have received topics or really spans for that matter where 0 is accepted as valid. Are we able to change these to make them more strict?
@ahsonkhan @RickWinter

danewalton: I remember @JeffreyRichter saying we can only relax req's to not break people.

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: cc @ericwol-msft

sdk/inc/azure/iot/az_iot_provisioning_client.h
ericwol-msft: why would the topic be zero?

ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: that's what I'm saying. So I'd like to change these to minimum size 1 like the rest of the APIs but wanted to check with the SDK team if it's allowed

danewalton:
@ahsonkhan if you also want to give this a look that would be great. I'd hope to use a lot of this wording elsewhere so if we do find and replace they're all the same. So pick through the wording with a fine tooth comb.

ahsonkhan:
> I'd hope to use a lot of this wording elsewhere so if we do find and replace they're all the same. So pick through the wording with a fine tooth comb.

For sure! I can't context switch currently, but I will make sure to get through this before eod today. Hope that's ok.

danewalton:
> > I'd hope to use a lot of this wording elsewhere so if we do find and replace they're all the same. So pick through the wording with a fine tooth comb.
> 
> For sure! I can't context switch currently, but I will make sure to get through this before eod today. Hope that's ok.

Yup totally fine!

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: Oh interesting, oversight like this can some times happen. I think leaving it as is seems reasonable rather than forcing an API break. There isn't substantial customer value to justify it, but for future APIs let's try to be consistent with the SAS key ones.

Maybe add a document `@remark` to call attention to the parameter order, but we could also leave that alone too, since the compiler should error out if you try to pass an int to the signature parameter.

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: In practice, is the minimum size of a valid topic actually 1? What's an example of such a topic string?

I think it's reasonable to consider strengthening the precondition if we are confident no customer would reasonably be impacted by such a change because no one would want to pass an empty topic. We can be pragmatic about what contractual strengthening/changes are actual breaks vs just bug/correctness fixes.

That said, let's do that outside of this precondition documentation PR.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -322,7 +343,10 @@ typedef struct
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] received_topic An #az_span containing the received topic.
  * @param[out] out_request If the message is a C2D request, this will contain the
- * #az_iot_hub_client_c2d_request
+ * #az_iot_hub_client_c2d_request.
+ * @pre \p client must not be `NULL`.
+ * @pre \p received_topic must be a valid span of size greater than 0.
+ * @pre \p out_request must not be `NULL`.
```
ahsonkhan, line 349: Missing a precondition doc for the client internal host name. Was this skipped intentionally?
Not only must the client not be `NULL`, it needs to be initialized first by calling `az_iot_hub_client_init` (from https://github.com/Azure/azure-sdk-for-c/pull/1314)

https://github.com/Azure/azure-sdk-for-c/blob/4b2f787bc155d1cd67f64ad1eebfdbc1438f6ce5/sdk/src/azure/iot/az_iot_hub_client_c2d.c#L25

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -398,6 +425,10 @@ AZ_NODISCARD az_result az_iot_hub_client_methods_parse_received_topic(
  * @param[in] mqtt_topic_size The size, in bytes of \p mqtt_topic.
  * @param[out] out_mqtt_topic_length __[nullable]__ Contains the string length, in bytes, of \p
  * mqtt_topic. Can be `NULL`.
+ * @pre \p client must not be `NULL`.
+ * @pre \p request_id must be a valid span of size greater than 0.
+ * @pre \p mqtt_topic must not be `NULL`.
+ * @pre \p mqtt_topic_size must be greater than 0.
```
ahsonkhan, line 431: Is this the correct precondition in source, should it be `_az_PRECONDITION(mqtt_topic_size > 0);`?
https://github.com/Azure/azure-sdk-for-c/blob/4b2f787bc155d1cd67f64ad1eebfdbc1438f6ce5/sdk/src/azure/iot/az_iot_hub_client_methods.c#L91

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -322,7 +343,10 @@ typedef struct
  * @param[in] client The #az_iot_hub_client to use for this call.
  * @param[in] received_topic An #az_span containing the received topic.
  * @param[out] out_request If the message is a C2D request, this will contain the
- * #az_iot_hub_client_c2d_request
+ * #az_iot_hub_client_c2d_request.
+ * @pre \p client must not be `NULL`.
```
ahsonkhan, line 347:
```suggestion
 * @pre \p client must not be `NULL` and must already be initialized by first calling az_iot_hub_client_init().
```

ahsonkhan (commented):
Some questions and nit suggestions. If you take any of the suggestions, apply them across the board.

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: Looks like we need at least 23, if not more, looking at:
https://github.com/Azure/azure-sdk-for-c/blob/4b2f787bc155d1cd67f64ad1eebfdbc1438f6ce5/sdk/src/azure/iot/az_iot_provisioning_client.c#L21-L28

For example, this slice to end call would fail the precondition anyway, if `received_topic` is smaller than the expected size:
https://github.com/Azure/azure-sdk-for-c/blob/4b2f787bc155d1cd67f64ad1eebfdbc1438f6ce5/sdk/src/azure/iot/az_iot_provisioning_client.c#L536

https://github.com/Azure/azure-sdk-for-c/blob/222010d40d108e356896985277e14e91b5b9a5d8/sdk/src/azure/core/az_span.c#L60-L65

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -172,6 +172,9 @@ typedef struct
  * with the following format: {name}={value}&{name}={value}.
  * @param[in] written_length The length of the properly formatted properties already initialized
  * within the buffer. If the \p buffer is unfilled (uninitialized), this should be 0.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p buffer must be a valid span of size greater than 0.
+ * @pre \p written_length must be greater than or equal to 0.
```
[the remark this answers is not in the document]
danewalton: I'll probably keep as is (written in plain, positive, "this is what it needs to be"). Not a mixture of what it can't and can.

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -268,6 +280,11 @@ AZ_NODISCARD AZ_INLINE bool az_iot_status_retriable(az_iot_status status)
  * @param[in] max_retry_delay_msec The maximum time, in milliseconds, to wait before a retry.
  * @param[in] random_jitter_msec A random value between 0 and the maximum allowed jitter, in
  * milliseconds.
+ * @pre \p operation_msec must be between 0 and INT32_MAX - 1.
```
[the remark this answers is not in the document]
danewalton: you know that's a good question. Off the top of my head I'm not quite sure but I can find out.

sdk/inc/azure/iot/az_iot_hub_client.h
danewalton: good call

sdk/inc/azure/iot/az_iot_hub_client.h
danewalton: fixed https://github.com/Azure/azure-sdk-for-c/pull/1621

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: gonna leave this open until @JeffreyRichter gives his 2 cents.

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -172,6 +172,9 @@ typedef struct
  * with the following format: {name}={value}&{name}={value}.
  * @param[in] written_length The length of the properly formatted properties already initialized
  * within the buffer. If the \p buffer is unfilled (uninitialized), this should be 0.
+ * @pre \p properties must not be `NULL`.
+ * @pre \p buffer must be a valid span of size greater than 0.
```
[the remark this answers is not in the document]
danewalton: this I'd like to leave open for now as well to see if we can strengthen them.

sdk/inc/azure/iot/az_iot_common.h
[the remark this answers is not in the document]
ahsonkhan: It would be good to know what's special about the max integer number in certain cases. Let me know what you find :)

PR #1646, opened 2021-03-12
add error checking for iot live test scripts

danewalton:


danieljurek (approved):
Thanks for adding these. It should help in working out which steps failed in later runs of the live tests. Ideally if a pattern emerges we can add mitigations (retry, check for status change in a loop, etc.)

sdk/samples/iot/templates/iothub/test-resources-post.ps1
```diff
@@ -65,12 +65,30 @@ Add-AzIotHubDevice `
 -PrimaryThumbprint $fingerprint `
 -SecondaryThumbprint $fingerprint
 
+if ($LASTEXITCODE -ne 0)
+{
+  Write-Host "Adding cert device failed: LAST_ERROR_CODE=${LAST_ERROR_CODE}"
```
ahsonkhan, line 70: Would it be helpful to also log some other things like the `$deviceID` or are those considered private?

sdk/samples/iot/templates/iothub/test-resources-post.ps1
```diff
@@ -65,12 +65,30 @@ Add-AzIotHubDevice `
...
+  exit $LASTEXITCODE
+}
+
 # Download Baltimore Cert
```
ahsonkhan, line 74: Turn this comment into a `Write-Host` prior to running the command, so we get some log info about the steps even when things are running successfully.

sdk/samples/iot/templates/iothub/test-resources-post.ps1
```diff
@@ -65,12 +65,30 @@ Add-AzIotHubDevice `
...
+  Write-Host "Downloading root cert failed: LAST_ERROR_CODE=${LAST_ERROR_CODE}"
+  exit $LASTEXITCODE
+}
+
 # Link IoTHub to DPS service
 $hubConnectionString = Get-AzIotHubConnectionString -ResourceGroupName $ResourceGroupName -Name $iothubName -KeyName "iothubowner"
 
+if ($LASTEXITCODE -ne 0)
+{
+  Write-Host "Getting connection string for cert device failed: LAST_ERROR_CODE=${LAST_ERROR_CODE}"
```
ahsonkhan, line 88: Same here, any non-secret parameter values to the command above might be helpful to log.

ahsonkhan (approved):

sdk/samples/iot/templates/iothub/test-resources-post.ps1
danewalton: it wouldn't be that useful. All we really need to know is that it failed.

PR #1656, opened 2021-03-19
generate map files for iot samples

danewalton:


momuno:
What is a map file and we do we need them? :)

danewalton:
> What is a map file and we do we need them? :)

Good breakdown here on what they are and how/what to do with them:
https://interrupt.memfault.com/blog/get-the-most-out-of-the-linker-map-file

To quote it though here:
> The map file provides valuable information that can help you understand and optimize memory. I highly recommend keeping that file for any firmware running in production. This is one of the few artifacts I keep in my CD pipeline.

> The map file is a symbol table for the whole program. Let’s dive into it to see how simple it is and how you can effectively use it. I will try to illustrate with some examples, all described with GNU binutils.

It can help us keep track of sizes of libraries. In the future, we can do some work to automatically pull out the interesting bits (some kind of `grep` thing), but for now they are published as a whole here on devops:

image

danewalton:
@momuno I think I should change the title of the PR to "generate map files" instead of "add map files" :)

ericwol-msft (approved):

ericwol-msft (approved):

danewalton:
cc @danieljurek to double check before I merge. Want to make sure this is done properly.

PR #1640, opened 2021-03-09
update precondition on provisioning to minimum size 1

danewalton:

PR #1699, opened 2021-04-16
add test for too many written properties

danewalton:
cc @ahsonkhan @RickWinter since this is strengthening preconditions. But I think this is an obvious one we don't want to have happen.


sdk/src/azure/iot/az_iot_common.c
```diff
@@ -27,6 +27,7 @@ AZ_NODISCARD az_result az_iot_message_properties_init(
   _az_PRECONDITION_NOT_NULL(properties);
   _az_PRECONDITION_VALID_SPAN(buffer, 0, true);
   _az_PRECONDITION(written_length >= 0);
+  _az_PRECONDITION(written_length <= az_span_size(buffer));
```
ahsonkhan, line 30: Similar to https://github.com/Azure/azure-sdk-for-c/pull/1640, should this be considered a bug fix for the CL?

sdk/tests/iot/common/test_az_iot_common.c
```diff
@@ -59,6 +59,17 @@ static void test_az_iot_message_properties_init_NULL_props_fails(void** state)
...
 
+static void test_az_iot_message_properties_init_user_set_params_too_many_written_fail(void** state)
+{
+  (void)state;
+
+  az_span test_span = az_span_create_from_str(TEST_KEY_VALUE_ONE);
+  az_iot_message_properties props;
+
+  ASSERT_PRECONDITION_CHECKED(
+      az_iot_message_properties_init(&props, test_span, az_span_size(test_span) + 1));
```
ahsonkhan, line 70: While you are here, can you verify if a test exists for negative written length, and add one if it doesn't?

sdk/src/azure/iot/az_iot_common.c
```diff
@@ -27,6 +27,7 @@ AZ_NODISCARD az_result az_iot_message_properties_init(
   _az_PRECONDITION_NOT_NULL(properties);
   _az_PRECONDITION_VALID_SPAN(buffer, 0, true);
   _az_PRECONDITION(written_length >= 0);
```
ahsonkhan, line 29: nit, you could combine the condition like this:
```suggestion
  _az_PRECONDITION_RANGE(0, written_length, az_span_size(buffer));
```

ahsonkhan (approved):

sdk/src/azure/iot/az_iot_common.c
```diff
@@ -27,6 +27,7 @@ AZ_NODISCARD az_result az_iot_message_properties_init(
   _az_PRECONDITION_NOT_NULL(properties);
   _az_PRECONDITION_VALID_SPAN(buffer, 0, true);
```
ahsonkhan, line 28: Hmm, another way to think about the fix is to do this:
```suggestion
  _az_PRECONDITION_VALID_SPAN(buffer, written_length, true);
```

So, maybe first check the `_az_PRECONDITION(written_length >= 0);` and then ensure the minimum size of the span is  equal to written_length.

ahsonkhan: Actually, you wouldn't even need to check `_az_PRECONDITION(written_length >= 0);`

`_az_PRECONDITION_VALID_SPAN` does it for you.

sdk/src/azure/iot/az_iot_common.c
danewalton: Yea I like this one

sdk/src/azure/iot/az_iot_common.c
```diff
@@ -26,7 +26,7 @@ AZ_NODISCARD az_result az_iot_message_properties_init(
 {
   _az_PRECONDITION_NOT_NULL(properties);
   _az_PRECONDITION_VALID_SPAN(buffer, 0, true);
```
ahsonkhan, line 28: Not a fan of using this one to do both checks, eh? :p

ahsonkhan (approved):

sdk/src/azure/iot/az_iot_common.c
danewalton: haha no only because I like having separate, explicit purpose preconditions here. Makes it a little easier to read

ahsonkhan: I agree.

PR #1702, opened 2021-04-17
CBOR documentation

momuno:
- CELA OSS requirements
- README sample instructions


ahsonkhan:
@momuno did you expect or want this to be reviewed by others or were you just trying to test CI and have it merge into the feature branch?

NOTICE.txt
```diff
@@ -332,6 +337,33 @@ Except as contained in this notice, the name of a copyright holder shall not
 be used in advertising or otherwise to promote the sale, use or other dealings
 in this Software without prior written authorization of the copyright holder.
 
+
+License notice for intel/tinycbor
+------------------------------------------------------------------------------
+
+MIT License
```
ahsonkhan, line 344: Nice!

cgmanifest.json
```diff
@@ -50,6 +50,16 @@
...
     },
+    {
+      "Component": {
+        "Type": "git",
+        "git": {
+          "RepositoryUrl": "https://github.com/intel/tinycbor",
+          "CommitHash": "755f9ef932f9830a63a712fd2ac971d838b131f1"
+        }
+      },
+      "DevelopmentDependency": false
```
ahsonkhan, line 61: This needs to be true. It isn't a real customer facing dependency of the SDK (like curl), just a devdep.

sdk/samples/iot/README.md
```diff
@@ -176,13 +173,53 @@ To run the samples, ensure you have the following programs and tools installed o
...
+    git clone https://github.com/intel/tinycbor.git
+    cd tinycbor
+    git checkout v0.5.3
+    make
+    sudo make install
+    ```
+
+    - Please check your `/usr/local/lib` directory and update permissions on the installed libraries if needed.
+
+    Windows (Select the Command Prompt):
```
ahsonkhan, line 199: nit: Is there a way to add some line break/spacing here?
Looks squished in the markdown:
image

sdk/samples/iot/README.md
```diff
@@ -176,13 +173,53 @@ To run the samples, ensure you have the following programs and tools installed o
...
+    - x64 system: Open the x64 Native Tools Command Prompt for Visual Studio.
+
+    ```
+    git clone https://github.com/intel/tinycbor.git
+    cd tinycbor
+    git checkout v0.5.3
+    NMAKE /F Makefile.nmake
+    ```
+
+    - Update your Path environment variable to include the tinycbor directory.
```
ahsonkhan, line 211: Why is this required?

It might be helpful to add more details on what exactly I need to set my path environment var too, with an example.

sdk/samples/iot/README.md
```diff
@@ -484,6 +521,12 @@ This section provides an overview of the different samples available to run and
 
   This [sample](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/paho_iot_hub_sas_telemetry_sample.c) sends five telemetry messages to the Azure IoT Hub. SAS authentication is used.
 
+### IoT Hub Twin CBOR Sample
+
+- *Executable:* `paho_iot_hub_twin_cbor_sample`
+
+  This [sample](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/paho_iot_hub_twin_sample.c) is the same as the `paho_iot_hub_twin_sample` below, with the exception that the device uses CBOR encoding and decoding for the Twin Document instead of JSON. The service side still uses JSON. To run this sample, the [MIT licensed](https://github.com/intel/tinycbor/blob/master/LICENSE) intel/tinycbor [library](https://github.com/intel/tinycbor) must be installed. Please see the [prerequisites](#prerequisites) section for instructions.
```
ahsonkhan, line 528: nit: This section should come after `### IoT Hub Twin Sample`

Then you can say:
```suggestion
### IoT Hub Twin CBOR Sample

- *Executable:* `paho_iot_hub_twin_cbor_sample`

  This [sample](https://github.com/Azure/azure-sdk-for-c/blob/master/sdk/samples/iot/paho_iot_hub_twin_sample.c) is the same as the `paho_iot_hub_twin_sample` above, with the exception that the device uses CBOR encoding and decoding for the Twin Document instead of JSON. The service side still uses JSON. To run this sample, the [MIT licensed](https://github.com/intel/tinycbor/blob/master/LICENSE) intel/tinycbor [library](https://github.com/intel/tinycbor) must be installed. Please see the [prerequisites](#prerequisites) section for instructions.
```

NOTICE.txt
```diff
@@ -30,6 +30,7 @@ bring it to our attention. Post an issue or email us:
 
 The attached notices are provided for information only.
 
+
```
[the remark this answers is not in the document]
momuno: Because it wasn't clear at first glance where the delineation was between licensing info.

cgmanifest.json
momuno: oh i forgot to change that. Thanks.

sdk/samples/iot/README.md
momuno: good eye

sdk/samples/iot/README.md
momuno: The CMake needs to know where to look for the tinycbor library.  Since the we don't know where the user will install it, I can tell CMake to search in a specific path (e.g., Env Path).  The user just has to include the library in the Env Path.

momuno: The Linux install is nice, because `make install` puts the library in the /usr/local/lib.  That is one of the locations CMake automatically looks, so no extra instructions need to give to it.

sdk/samples/iot/README.md
momuno: you're right. I thought the listing was alphabetical, but it's not.

sdk/samples/iot/README.md
ahsonkhan: > I assumed if someone was running this on windows, they would know how to do it, or could look it up. Maybe that's a bad assumption.

On Windows, what is the default install location that I need to set my PATH env to once I call `NMAKE /F Makefile.nmake`? 
It would be helpful to use that as an example `PATH` so I, as the reader, don't have to spend time figuring this out :)

If I want to override that location to some other directory, how can I do that?

ahsonkhan: Agreed. That makes a lot of sense.

> Otherwise, I think a basic "open the system properties window to edit the PATH environment variable" should suffice.

Yep. A tweak like this might just be good enough:
```suggestion
    - Open the system properties window and update your Path environment variable to include the location of the tinycbor directory you cloned above.
```

PR #1703, opened 2021-04-17
CBOR Twin code implementation

momuno:
- Adds options for method/twin content type
- Adds logic to build username
- Adds unit testing for building of username.
- Refactors unit test static variable naming


sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
    * The model ID used to identify the capabilities of a device based on the Digital Twin document.
    */
   az_span model_id;
+
+  /**
+   * The data_format is sent in the username to inform IoT Hub what format the device expects Twin
```
ahsonkhan, line 60: What is this `data_format` referring to? Is this some object we define? I am asking because I am trying to understand why we have the underscore in it?

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+   * Document and Direct Method payloads to use. If this option is not set, the default is JSON. If
+   * data format provided is not supported by IoT Hub, service will close the connection with error
```
ahsonkhan, line 62:
```suggestion
   * the data format provided is not supported by IoT Hub, the service will close the connection with error
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+  az_span method_twin_content_type;
+
 } az_iot_hub_client_options;
 
+/**
+ * @brief Azure IoT Hub Client option values. Not all options will have pre-defined values.
+ *
+ */
+
+/**
```
ahsonkhan, line 74: How does this render in the generated doxygen docs? I wonder where this doc actually shows up compared to where we want it to.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+   * Document and Direct Method payloads to use. If this option is not set, the default is JSON. If
+   * data format provided is not supported by IoT Hub, service will close the connection with error
+   * 4 code indicating Bad Username/Password.
+   */
+  az_span method_twin_content_type;
+
 } az_iot_hub_client_options;
 
+/**
+ * @brief Azure IoT Hub Client option values. Not all options will have pre-defined values.
```
ahsonkhan, line 70: > Not all options will have pre-defined values.

What do you mean by this?

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+ *
+ */
+
+/**
+ * @brief CBOR content type. Sent in username to inform IoT Hub the device expects CBOR for Twin
+ * Document and Direct Method payloads.
+ * @note It can be used by wrapping the macro in a #AZ_SPAN_FROM_STR macro as a parameter.
```
ahsonkhan, line 77:
```suggestion
 * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR "application/cbor"
+
+/**
+ * @brief JSON content type. Sent in username to inform IoT Hub the device expects JSON for Twin
+ * Document and Direct Method payloads.
+ * @note It can be used by wrapping the macro in a #AZ_SPAN_FROM_STR macro as a parameter.
```
ahsonkhan, line 84:
```suggestion
 * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
```

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -61,27 +64,38 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
+
   if (az_span_size(*module_id) > 0)
   {
-    required_length += az_span_size(*module_id) + (int32_t)sizeof(hub_client_forward_slash);
+    required_length += (int32_t)sizeof(hub_client_forward_slash) + az_span_size(*module_id);
   }
   if (az_span_size(*user_agent) > 0)
   {
-    required_length += az_span_size(*user_agent) + az_span_size(hub_client_param_separator_span);
+    required_length += az_span_size(hub_client_param_separator_span) + az_span_size(*user_agent);
```
ahsonkhan, line 82: Why swap the order of operations here. This change seems unnecessary (unless I am missing some logical difference).

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -61,27 +64,38 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
   }
-  // Note we skip the length of the model id since we have to url encode it. Bound checking is done
-  // later.
   if (az_span_size(*model_id) > 0)
+  {
+    // Temporarily use mqtt_user_name_span. If not big enough for encoded_length, return error here.
+    int32_t encoded_length = 0;
+    _az_RETURN_IF_FAILED(_az_span_url_encode(mqtt_user_name_span, *model_id, &encoded_length));
```
ahsonkhan, line 88: I am just thinking out loud here, but doesn't overriding `mqtt_user_name_span` while doing url encoding result in the initial data being trampled over that the user provided, since url encoding typically increases the size of the data?

Why is that OK to do?

sdk/tests/iot/hub/test_az_iot_hub_client.c
```diff
@@ -25,25 +25,40 @@
 #define TEST_HUB_HOSTNAME_STR "myiothub.azure-devices.net"
 #define TEST_MODULE_ID "my_module_id"
 #define TEST_USER_AGENT "os=azrtos"
-#define TEST_MODEL_ID "dtmi:YOUR_COMPANY_NAME_HERE:sample_device;1"
 #define PLATFORM_USER_AGENT "DeviceClientType=c%2F" AZ_SDK_VERSION_STRING
+#define TEST_MODEL_ID "dtmi:YOUR_COMPANY_NAME_HERE:sample_device;1"
+#define MODEL_ID_PARAMETER "model-id=dtmi%3AYOUR_COMPANY_NAME_HERE%3Asample_device%3B1"
+#define TEST_CONTENT_TYPE_CBOR "application/cbor"
```
ahsonkhan, line 31: Why do we need to define this in tests, can we re-use `AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR` from the public header?

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: Left over from an earlier version! It should be method_twin_content_type.  Updating.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: This is the first time one of the options has macro defined options offered in the header file.  The other options don't have them.

sdk/src/azure/iot/az_iot_hub_client.c
momuno: It's not necessary technically, but the swapped order now matched the order the username will be built in.  It was harder to envision what was being measured when it was in the backwards order.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: Can you remind me how to test these comments against doxygen?

sdk/src/azure/iot/az_iot_hub_client.c
momuno: `mqtt_user_name_span` is created but not yet used, so it won't be trampling over anything.  I considered creating a different az_span of the same size just to get the encoded length, but thought that was redundant.  However, it could prevent against a future bug if someone didn't catch this nuance.

sdk/tests/iot/hub/test_az_iot_hub_client.c
momuno: good idea

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
    * The model ID used to identify the capabilities of a device based on the Digital Twin document.
    */
   az_span model_id;
+
+  /**
+   * The method_twin_content_type is sent in the username to inform IoT Hub what format the device
+   * expects Twin Document and Direct Method payloads to use. If this option is not set, the default
+   * is JSON. If the content type provided is not supported by IoT Hub, the service will close the
+   * connection with error 4 code indicating Bad Username/Password.
```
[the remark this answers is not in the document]
danewalton: this is great thorough docs though

danewalton (changes requested):

sdk/src/azure/iot/az_iot_hub_client.c
ahsonkhan: > `mqtt_user_name_span` is created but not yet used, so it won't be trampling over anything. 

The underlying buffer that we are writing to (that you created the span out of) is provided by the user. Do they have data in there that we need?

sdk/src/azure/iot/az_iot_hub_client.c
ahsonkhan: Gotcha, that makes sense. I just wanted to make sure it wasn't some compiler optimization/re-ordering issue.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: I have updated the commenting for doxygen.  Also moved the defines in front of the telemetry QOS enum and iot hub client options struct to follow the ordering in the file: defines, enums, structs, functions.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -55,8 +55,36 @@ typedef struct
...
+ * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR "application%2Fcbor"
+
+/**
+ * @brief JSON content type. Sent in username to inform IoT Hub the device expects JSON for Twin
+ * Document and Direct Method payloads.
+ * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_JSON "application%2Fjson"
```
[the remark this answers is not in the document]
momuno: I updated to the following.  What do you think?

```
/**
 * @brief Azure IoT Hub Client method_twin_content_type option value: CBOR.
 * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
 */
#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR "application%2Fcbor"

/**
 * @brief Azure IoT Hub Client method_twin_content_type option value: JSON.
 * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
 */
#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_JSON "application%2Fjson"
```

sdk/src/azure/iot/az_iot_hub_client.c
momuno: The problem is that if we url encode and check the required total length later (when model-id is added to the username), then we aren't getting an accurate url required length check since the content type has to go after the model id.

The first required length check is done sans url-encoded model id.
The username is constructed with url-encoded model id, and the content type is added after that.  
There is no accurate required length check done, unless we do it AGAIN.

momuno: I have rewritten the logic to do pre-url encode size bound checking and building of username, then logic for the model id url encoding, and then a post-url encode size bound checking and building of username.

momuno: A different option is to bound check size and build the username for each item as we go along.  
Or, build the complete username, and bound check at the end.

momuno: @jspaith, interested in what you think about a best approach here.

danewalton: > Unless, its ok for model_id to go at the end.

Yes that should be fine since they are really just url parameters. Hub will parse them out wherever they are. I would put the model-id at the end and check the size of everything before that url encode section.

eng/common/TestResources/build-test-resource-config.yml
```diff
@@ -0,0 +1,65 @@
...
+  - name: SubscriptionConfiguration
+    type: string
+    default: $(sub-config-azure-cloud-test-resources)
+  - name: SubscriptionConfigurations
+    type: object
+    default: null
+
+steps:
+  - ${{ if parameters.SubscriptionConfiguration }}:
+    - pwsh: |
```
danewalton, line 11: is this file supposed to be here?

momuno: Oh snap.  I bet this is because I merged in master.  Let me remove it and make that update a separate PR.  Thanks!

momuno: Thanks Dane.  Managing a feature branch is a bit of a dance in its own right.  Figuring it out....

danewalton: that it is haha. no worries

sdk/src/azure/iot/az_iot_hub_client.c
momuno: After thinking about it, for now I'm going to keep it with the order specified in the docs for consistency.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: comment removed in doxygen fix commit.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -26,6 +26,18 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
+/**
+ * @brief Azure IoT Hub Client method_twin_content_type option value: CBOR.
+ * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR "application%2Fcbor"
```
danewalton, line 33: Yup I would only suggest adding backticks to the method_twin_content_type so it renders in typewriter font in doxygen

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -26,6 +26,18 @@
...
+
+/**
+ * @brief Azure IoT Hub Client method_twin_content_type option value: JSON.
```
danewalton, line 36:
```suggestion
 * @brief Azure IoT Hub Client `method_twin_content_type` option value: JSON.
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -42,19 +54,28 @@ enum
 typedef struct
 {
   /**
-   * The module name (if a module identity is used).
+   * The module_id represents the module name (if a module identity is used).
```
danewalton, line 57: Can we backtick all of these?
```suggestion
   * The `module_id` represents the module name (if a module identity is used).
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -26,6 +26,18 @@
...
+ * @brief Azure IoT Hub Client method_twin_content_type option value: CBOR.
+ * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_CBOR "application%2Fcbor"
+
+/**
+ * @brief Azure IoT Hub Client method_twin_content_type option value: JSON.
+ * @note It can be used by wrapping the macro in an #AZ_SPAN_FROM_STR macro as a parameter.
+ */
+#define AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_JSON "application%2Fjson"
```
danewalton, line 39: might be a dumb question, but in what scenario would someone use this (if it defaults to JSON)? I'm guessing this is just for clarity/parity?

danewalton (approved):
unblocking on my end

sdk/src/azure/iot/az_iot_hub_client.c
momuno: Will try this out.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: Yeah, they should all be updated.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: It is a supported option, so I'd like to include it for completeness and documentation.  Specific use case, maybe a group of devices that some use cbor and some use json, or for some reason need to switch part way in their life. There is a nice consistency to the implementation to have both supported.  But right away, I doubt anyone will use this option for json.

sdk/src/azure/iot/az_iot_hub_client.c
momuno: Fixed, and tests updated.

PR #1707, opened 2021-04-20
Merge master into feature/iot_cbor

momuno:


danewalton (approved):

danewalton:
Oh oh oh. One thing. See if you are able to not squash and merge. Actually have all these commit's merged. It might take a setting and coordination with @RickWinter but that way the history is clean and as long as you merge master in consistently, then this will only show `x ahead of master` instead of some combo between ahead and behind.

momuno:
Hey @RickWinter .  I am trying to merge in master into this feature branch because I merged master into my dev branches that I'm committing into feature/iot_cbor.

Is there a clean way to make this commit so it doesn't make things look wonky in the history?

PR #1713, opened 2021-04-21
CBOR Twin Sample

momuno:
- Adds CMake logic to find intel/tinycbor and only build CBOR sample if this is found.
- Adds the CBOR Twin Sample -- NOTE: reported properties is still being debugged.
- Adds IOT_SAMPLE_LOG_HEX to print payload in bytes
- Adds username printout to all samples


sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
```diff
@@ -0,0 +1,533 @@
...
+  {
+    IOT_SAMPLE_LOG(
+        "Parsed desired `%s`: %" PRIi64,
+        desired_device_count_property_name,
+        *out_parsed_device_count);
+  }
+  else
+  {
+    IOT_SAMPLE_LOG(
+        "`%s` property was not found in desired property response.",
```
danewalton, line 493: nitnit: desired in this case is a message not a response. 
```suggestion
        "`%s` property was not found in desired property message.",
```

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
```diff
@@ -0,0 +1,533 @@
...
+        desired_device_count_property_name);
+    return false;
```
danewalton, line 495: maybe just return `property_found` at the end? single return place

sdk/samples/iot/CMakeLists.txt
```diff
@@ -137,6 +137,51 @@ target_link_libraries(paho_iot_hub_twin_sample
...
+    PATH_SUFFIXES tinycbor/lib lib
+  endif()
+)
+
+if (${tinycbor} STREQUAL "tinycbor-NOTFOUND")
+  message("-- Could NOT find tinycbor. Not compiling paho_iot_hub_twin_cbor_sample")
+else()
+  if(MSVC)
+    set(CMAKE_EXE_LINKER_FLAGS
+        "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:LIBCMT.lib")
```
danewalton, line 153: what is this doing?

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
```diff
@@ -0,0 +1,533 @@
...
+ *     }
+ *   }
+ * }
+ *
+ * No other property names sent in a desired property message are supported. If any are sent, the
+ * IOT_SAMPLE_LOG will report there is nothing to update.
+ */
+int main(void)
+{
+  create_and_configure_mqtt_client();
```
danewalton, line 105: maybe have a print statement here which calls out that preview hubs are needed? (I think they're needed). And to expect a disconnect if you use this without one (if that's the behavior). sorry some ignorance here on what exactly happens between device and hub for this feature.

sdk/samples/iot/paho_iot_hub_twin_sample.c
```diff
@@ -121,7 +121,14 @@ static void create_and_configure_mqtt_client(void)
...
   // Initialize the hub client with the default connection options.
-  rc = az_iot_hub_client_init(&hub_client, env_vars.hub_hostname, env_vars.hub_device_id, NULL);
+  az_iot_hub_client_options options = az_iot_hub_client_options_default();
+
+  // Set the content type to JSON for Direct Method payloads and Twin Document.
+  // This option is not required to be set for JSON, since the default content type is JSON.
+  // options.method_twin_content_type =
+  // AZ_SPAN_FROM_STR(AZ_IOT_HUB_CLIENT_OPTION_METHOD_TWIN_CONTENT_TYPE_JSON);
+
+  rc = az_iot_hub_client_init(&hub_client, env_vars.hub_hostname, env_vars.hub_device_id, &options);
```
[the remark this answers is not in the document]
momuno: Good point.  I'd like the user to be aware of it, so those needing the option for CBOR would be the only ones who might care to also use an option for JSON, since normal users of JSON wont care about this option.

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
```diff
@@ -0,0 +1,533 @@
...
+static void build_reported_property(
+    uint8_t* reported_property_payload,
+    size_t reported_property_payload_size,
+    size_t* out_reported_property_length)
+{
+  // WARNING: Check the return of all API calls when developing your solution. Many return checks
+  //          are ommited from this sample for simplification.
+
+  CborEncoder cbor_encoder_root;
+  CborEncoder cbor_encoder_root_container;
```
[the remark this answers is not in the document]
momuno: I want to be careful how much instruction we are giving on the tinycbor library.  It's not part of our SDK, so I think I'd like to leave reading up on how tinycbor works as an "exercise to the reader".

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
momuno: will change in reg twin sample too.

sdk/samples/iot/CMakeLists.txt
momuno: suppresses this warning: 
```
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs; use /NODEFAULTLIB:library [C:\Repos\momuno_azu
re-sdk-for-c\build\sdk\samples\iot\paho_iot_hub_twin_cbor_sample.vcxproj]
```
The cbor library uses by default a different default runtime library (the multithreaded LIMCMT.lib) than what the SDK is using (the multithreaded debug LIBCMTD.lib).
https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4098?view=msvc-160

@ahsonkhan , I have tried to use the add_compile_options you suggested, and even attempted these three simplified versions:
`add_compile_options( /MT )`, `add_compile_options( /MTd )`, `add_compile_options( /MT /MTd)`, but the issue is that only one runtime default library can be used, not two.  If I use `/MT` I get the LIBCMTD.lib complaint.  If I use `/MTd` I get the LIBCMT.lib complaint.  It looks like I have to choose one, so will need to use the `/NODEFAULT:<lib>` on one of the options.

momuno: @ahsonkhan side note, I'm not sure how effective this is to include multiple types of runtime libraries.  Only one can be used.
```
  add_compile_options(
    $<$<CONFIG:>:/MT> #---------|
    $<$<CONFIG:Debug>:/MTd> #---|-- Statically link the runtime libraries
    $<$<CONFIG:Release>:/MT> #--|
  )
  ```

File: sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c (hunk 4 above)
Comment (momuno) [reply sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c:105]: This only works on a Dogfood hub right now (not an official public preview hub).  There is no official public preview right now of this feature.  Also, I'm concerned that if commenting is tied too closely to a feature branch, it will get missed when the merge into master happens, and then be wrong.  I'd like to bring up feature branch management and public-preview branch management in the Friday meeting.

File: sdk/samples/iot/CMakeLists.txt (hunk 3 above)
Comment (momuno) [reply sdk/samples/iot/CMakeLists.txt:153]: There are multiple libraries being linked to the samples, such as paho and openssl, beyond just the SDK compiled code.  Even when the SDK appears to have no build type set, which means it should be using /MT, it is possible one of these other libraries was compiled with Debug, which could explain the persistent /MTd complaint when I compile the CBOR sample with /MT.  I think it is simplest for the CBOR sample to suppress the /MT library requirement.

File: sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c (hunk 6 above)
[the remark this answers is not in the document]
Comment (momuno) [reply sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c:519]: I'm going to change the naming to make it more clear what the purpose is.

File: sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c [hunk 7]
```diff
@@ -460,37 +461,56 @@ static bool parse_desired_device_count_property(
   bool property_found = false;
   *out_parsed_device_count = 0;
 
+  CborError rc; // CborNoError == 0
+
   // Parse message_span.
-  CborParser cbor_parser;
+  CborParser parser;
```
ericwol-msft, line 467: why get rid of cbor_?

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
```diff
@@ -460,37 +461,56 @@ static bool parse_desired_device_count_property(
...
+    else
+    {
+      IOT_SAMPLE_LOG("`%s` property was not an integer", desired_device_count_property_name);
+    }
   }
   else
   {
     IOT_SAMPLE_LOG(
-        "`%s` property was not found in desired property response.",
+        "`%s` property was not found in desired property message.",
```
ericwol-msft, line 513: sometimes there is a trailing period and something there isn't.

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
momuno: why have it?  It shortened lines which plays better with clang-format.  If it's preferred I can put it back.

sdk/samples/iot/paho_iot_hub_twin_cbor_sample.c
momuno: I found a sentence that missed a period and have added it.  Other places don't have periods because they aren't sentences.

sdk/samples/iot/CMakeLists.txt
danewalton: Might be beating a dead horse here but I had added these guys here a while back for that:
https://github.com/Azure/azure-sdk-for-c/blob/eb9b39275c973874f029a986fbf9010a6be8b486/eng/cmake/global_compile_options.txt#L9-L13

Have you tried these?

sdk/samples/iot/CMakeLists.txt
```diff
@@ -137,6 +137,51 @@ target_link_libraries(paho_iot_hub_twin_sample
...
+    PATHS
+      if(UNIX) /usr/local/include
+      endif()
+
+    PATH_SUFFIXES
+      if(WIN32) tinycbor/src src
+      elseif(UNIX) tinycbor
+      endif()
+  )
+  include_directories(${TINYCBOR_INCLUDE_DIR})
```
danewalton, line 180: should this be 
```suggestion
  target_include_directories(paho_iot_hub_twin_cbor_sample PRIVATE ${TINYCBOR_INCLUDE_DIR})
```

danewalton (approved):

sdk/samples/iot/CMakeLists.txt
momuno: Yes, this is exactly what I had tried and it was not working.  I'm not sure if its fully effective tbh.  WIth not setting a CMAKE_BUILD_TYPE, the empty CONFIG should be getting used, but all the projects show up as /MTd, which make me wonder if Visual Stuido uses that as a default?  Not sure, and I ran out of steam trying to debug it. :/

sdk/samples/iot/CMakeLists.txt
momuno: So i've fixed this and committed it: 976ce6d
But don't know why its not showing up in the PR

momuno: https://github.com/momuno/azure-sdk-for-c/commit/976ce6d3264009adb7094e08db61d9b9cf156560

PR #1732, opened 2021-05-04
Merge master into feature/iot_cbor

momuno:


danewalton:
I don't think this is needed. I think the only reason why there's a diff at all is because the previous merge of master squashed and this one has the history. I wouldn't go through with it because there's no files changed.

PR #1733, opened 2021-05-05
Update naming to twin and decouple twin-content-type option from characters in username.

momuno:
Previously, used a macro string literal as both an option to the user-application and as content for the username's twin content-type key.  Decoupling this option and username twin content-type key.  Option now uses an SDK defined enum.  Characters for username content-type key moved to source file.  

This aims to protect user application from a potential breaking change if username's twin content-type key is to change in future and application had used a string literal instead of the supplied macro for the option.


sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -47,6 +35,17 @@ enum
...
+/**
+ * @brief Azure IoT Hub Client option values.
+ *
+ */
+typedef enum
+{
+  AZ_IOT_HUB_CLIENT_OPTION_VALUE_UNDEFINED = 0,
+  AZ_IOT_HUB_CLIENT_OPTION_VALUE_TWIN_CONTENT_TYPE_CBOR = 1,
+  AZ_IOT_HUB_CLIENT_OPTION_VALUE_TWIN_CONTENT_TYPE_JSON = 2,
+} az_iot_hub_client_option_values;
```
danewalton, line 47: I think this should be named something more specific to this feature. Something like `az_iot_hub_client_content_type` or `az_iot_hub_client_content_type_value`. At the very least here the `s` should be dropped from the end.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -64,11 +63,10 @@ typedef struct
...
 
   /**
-   * The `method_twin_content_type` is sent in the username to inform IoT Hub what format the device
-   * expects Twin Document and Direct Method payloads to use. If this option is not set, the default
-   * is JSON.
+   * The `twin_content_type` is sent in the username to inform IoT Hub what format the device
+   * expects Twin Document to use. If this option is not set, the default is JSON.
    */
-  az_span method_twin_content_type;
+  az_iot_hub_client_option_values twin_content_type;
```
danewalton, line 69: can we put this at the end of the struct? That way new options are always added at the end.

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -63,7 +64,8 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
 
   const az_span* const module_id = &(client->_internal.options.module_id);
   const az_span* const user_agent = &(client->_internal.options.user_agent);
-  const az_span* const method_twin_ct = &(client->_internal.options.method_twin_content_type);
+  const az_iot_hub_client_option_values* const twin_ct
```
danewalton, line 67: just take the value instead of the pointer since it's just an enum?

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: i like it

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: The reason I put it here is to follow the logic order in the building of the username. model_id has to go at the end.  It keeps the order consistent for the reader.

danewalton: CC @ahsonkhan do we have guidance on where to put new options? I'm thinking from a compatibility standpoint leaving already released options in the same point in the struct would be beneficial and any new options should be tacked on to the end?

momuno: I'm curious to know more.  What kind of compatibility issues might arise here?  It's not an enum, so element number assignments aren't being changed.  The only think I can see being affected by inserting this new element instead of tacking on, is that because it is of a different type than az_span, possibly the size of the struct is not optimized?  But i'm not sure I see a compatibility change.

ewertons (approved):
What was the decision process to select using enums instead of passing az_span for twin_content_type?

momuno:
> What was the decision process to select using enums instead of passing az_span for twin_content_type?

Great question.
What prompted this is a potential future naming change of the content-type within the username.  Ultimately, I think any value in the username that is internally defined should not be defined by the user application.

Currently the user-application can define the exact values used in the username for module_id, user_agent, and model_id.  I believe these are all application dependent, so it makes sense that the app has this level of control.  However, the content-type value is internally defined -- it is agreed upon between the IoT Hub and SDK.  The application should have control over what content-type is selected, but not how the username value for the content-type is defined. 

Where the existing coupling can become an issue is if the IoT Hub and SDK agreement on the value used for the content type changes.  Example: "application%2Fcbor" later is shortened to "cbor".  If an application has not made use of the provided string-literal defined macro when setting the content type option, and instead has provided their own matching string-literal, then when the macro is updated to the new naming, their application will break. I think it is both a cleaner design and a better approach for backwards compatibility if we separate internal definitions from the application.  The app can still select which content-type it wants to use, but the SDK controls the details of the content-type definition, not the app.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -47,6 +35,17 @@ enum
   AZ_HUB_CLIENT_DEFAULT_MQTT_TELEMETRY_QOS = 0
 };
 
+/**
+ * @brief Azure IoT Hub Client twin_content_type option values.
+ *
+ */
+typedef enum
+{
+  AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_UNDEFINED = 0,
```
ericwol-msft, line 44: Why undefined? We know there only can be two types. Default to JSON. Also consider making JSON the first one.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -64,11 +63,11 @@ typedef struct
   az_span user_agent;
 
   /**
-   * The `method_twin_content_type` is sent in the username to inform IoT Hub what format the device
-   * expects Twin Document and Direct Method payloads to use. If this option is not set, the default
+   * The `twin_content_type` determines the format to send in the username, which will inform IoT
+   * Hub what format the device expects Twin Document to use. If this option is not set, the default
```
ericwol-msft, line 67:
```suggestion
   * Hub what format the device expects twin document to use. If this option is not set, the default
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -64,11 +63,11 @@ typedef struct
...
    * is JSON.
    */
-  az_span method_twin_content_type;
+  az_iot_hub_client_option_twin_content_type twin_content_type;
```
ericwol-msft, line 70: new fields should be at the end of the struct.

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -23,15 +23,17 @@ static const az_span hub_service_api_version = AZ_SPAN_LITERAL_FROM_STR("/?api-v
...
+static const az_span hub_twin_content_type_json = AZ_SPAN_LITERAL_FROM_STR("application%2Fjson");
 
 AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default()
 {
   return (az_iot_hub_client_options){ .module_id = AZ_SPAN_EMPTY,
                                       .user_agent = hub_client_sdk_version,
-                                      .model_id = AZ_SPAN_EMPTY,
-                                      .method_twin_content_type = AZ_SPAN_EMPTY };
+                                      .twin_content_type
+                                      = AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_UNDEFINED,
```
ericwol-msft, line 35: AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_JSON

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -63,7 +65,8 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
 
   const az_span* const module_id = &(client->_internal.options.module_id);
   const az_span* const user_agent = &(client->_internal.options.user_agent);
-  const az_span* const method_twin_ct = &(client->_internal.options.method_twin_content_type);
+  az_iot_hub_client_option_twin_content_type const twin_ct
```
ericwol-msft, line 68: move below line 70?

sdk/src/azure/iot/az_iot_hub_client.c
```diff
@@ -82,11 +85,19 @@ AZ_NODISCARD az_result az_iot_hub_client_get_user_name(
...
-        + az_span_size(*method_twin_ct);
+        + az_span_size(hub_twin_content_type) + az_span_size(hub_client_param_equals_span);
+
+    if (twin_ct == AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_CBOR)
+    {
+      required_length += az_span_size(hub_twin_content_type_cbor);
+    }
+    else if (twin_ct == AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_JSON)
+    {
+      required_length += az_span_size(hub_twin_content_type_json);
```
ericwol-msft, line 99: humm, should we do this if the CT is JSON. We know the default is JSON.

sdk/inc/azure/iot/az_iot_hub_client.h
ericwol-msft: or AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_UNDEFINED -> AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_DEFAULT

momuno: Hmm, I was thinking for backwards compat reasons we would need the option for the username to not include the content-type at all, but I think that's not needed now. With the release of this feature, it will be using a defined hub api that can handle content type in the username.  In this case, then yes we can always include JSON as the chosen content type in the username, and I can get rid of undefined.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: is direct methods, telemetry, and c2d also not capitalized?  What is the standard on all these?

sdk/src/azure/iot/az_iot_hub_client.c
momuno: If we can always include the content type in the username, then yes, and will update.  can you please double check my thought on backwards compat in comment above?

ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_hub_client.h
ericwol-msft: I was thinking not sending it to save bytes across the wire.
AZ_IOT_HUB_CLIENT_OPTION_TWIN_CONTENT_TYPE_DEFAULT_JSON?

momuno: That's a better reason not to.  Thanks for thinking of that.  
If we want to not be able to send anything, then there needs to be a generic undefined/default so the logic will recognize the option has not been set.  This assumes that if an app does set the option to use JSON, then we SHOULD send it across the wire.

The other approach is to always have the default option set to JSON, and even if an app decides to set it again, we will never send the JSON content type selection across the wire.  "default-content-type:application%2Fjson" is never used.

sdk/inc/azure/iot/az_iot_hub_client.h
ericwol-msft: We don't ship dynamic library, so not an issue. But it is still good practice to follow.

sdk/inc/azure/iot/az_iot_hub_client.h
momuno: why?  Dane had similar question. Not clear to me yet.

danewalton: Usually it's memory ordering. If you always add to the end, the position of fields never changes between versions (only amended). In theory there is a possibility of your code pulling in new headers but having old libs and the ordering would be off if you dot reference the fields.

danewalton: I guess I should say when I say "usually" I mean in my head. It could just be convention for other reasons

momuno:
Closing.  All changes will now be in forked repo.

PR #1751, opened 2021-05-13
Draft only: check tests work (no merge, do not review)

jspaith:
**REVIEWERS - PLEASE DON'T REVIEW THIS.** Marking draft only.

This is only for a quick gate run on some test changes.


jspaith:
For gate pre-run only.  Closing.

PR #1755, opened 2021-05-17
add clang format version

danewalton:


mitchdenny (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -365,6 +365,7 @@ jobs:
   # Validate all the files are formatted correctly according to the .clang-format file.
   - bash: |
       # Run clang-format recursively on each source and header file within the repo.
+      clang-format --version
```
ahsonkhan, line 368: Good idea. Out of curiosity, why did you notice you needed it? Did you observe the version mismatch between CI and what you had locally causing formatting difference?

ahsonkhan: Looks like CI currently has it as:
`clang-format version 9.0.0-2~ubuntu18.04.2 (tags/RELEASE_900/final)`

ahsonkhan: For C++, we stated we want v10 (and moved the test to a MacOS leg), so that our CI machines use the slightly newer version:
https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#third-party-dependencies

We seem to be OK with v9 here, which is totally fine, it might be worth mentioning the version to contributors though, here:
https://github.com/Azure/azure-sdk-for-c/blob/main/CONTRIBUTING.md#prerequisites

ahsonkhan: Found the issue that motivated us making the change in C++:
https://github.com/Azure/azure-sdk-for-cpp/issues/1207

@vhvb1989 what are your thoughts?

danewalton: iirc it was because different versions create differently rendered files. this was just to make sure people could easily see which version is being used at the gates

Issue #1784, opened 2021-06-17
Add static_assert style mechanism


jspaith:
There are scenarios where having a C++ style `static_assert` would be useful for this SDK as well.  Could we please add something like a `AZ_STATIC_ASSERT`?  

Thanks

PR #1786, opened 2021-06-18
Incorporate feedback from Azure SDK review board on PnP APIs.

jspaith:
Highligts of changes:
* Prefer the word "write" to "build" when writing JSON properties.  Samples also prefer "write" for JSON creation.
* Remove "_ack" from writable response names.  So status_code instead of ack_code, e.g.
* az_iot.h includes reference to properties header for the one-stop-shopper.
* Have the properties enums be hard-coded #'s, and not point at the underlying twin values they cast from.
  * Adding a UT for now to make sure they're equal.
  * Longer term https://github.com/Azure/azure-sdk-for-c/issues/1784 tracks adding a static_assert style mechanism for this.
* `az_iot_hub_client_properties_update_get_publish_topic` => `az_iot_hub_client_properties_get_reported_publish_topic`.  There's not an ideal way to represent this, but what's in this PR is an improvement over original.


jspaith:
Closing this PR.  The git branch setup was confused such that the base branch to merge into wasn't correct.  Re-creating.

PR #1791, opened 2021-06-21
only add compiler flags for tests

danewalton:
Having compiler flags on all the time limits compatibility with some platforms and forces customers to compile with flags they might not want. Having the flags on only for tests allows us to verify on every PR that our code is clean but allows integration to be more flexible.


ewertons (approved):

RickWinter (changes requested):
It seems reasonable to keep a lower level of warnings by default.  However I would put that into the global_compile_options.txt rather than a condition in the CMakeLists.  It not intended that global_compile_options only apply to tests.

danewalton:
@RickWinter how's that?

eng/cmake/global_compile_options.txt
```diff
@@ -1,28 +1,33 @@
...
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /WX")
-  endif()
-
-  add_compile_options(/W4 ${WARNINGS_AS_ERRORS_FLAG} /wd5031 /wd4668 /wd4820 /wd4255 /wd4710 /analyze)
-
   add_compile_options(
-    $<$<CONFIG:>:/MT> #---------|
-    $<$<CONFIG:Debug>:/MTd> #---|-- Statically link the runtime libraries
-    $<$<CONFIG:Release>:/MT> #--|
+      $<$<CONFIG:>:/MT> #---------|
```
hihigupt, line 3: are the compiler flags are must for sdk? or they are also optional ?

danewalton: these I'm almost positive are a must to build for windows. I'll let @ahsonkhan maybe confirm.

RickWinter (approved):

Issue #1744, opened 2021-05-07
Add public depth field for JSON reader


danewalton:
This will be needed for PnP property reader. Example usage here:
https://github.com/Azure/azure-sdk-for-c/blob/93fb3362a37753c67aca4b52b852441e097f4dbf/sdk/src/azure/iot/az_iot_pnp_client_property.c#L240

ahsonkhan:
New usage link:
https://github.com/Azure/azure-sdk-for-c/blob/75afd98a1efd41d0c1f4475f692da88547a61582/sdk/src/azure/iot/az_iot_hub_client_properties.c#L242

From:
https://github.com/Azure/azure-sdk-for-c/blob/feature/iot_pnp_stage/sdk/src/azure/iot/az_iot_hub_client_properties.c#L242

PR #1802, opened 2021-06-29
Add public current depth field to JSON reader.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1744

The following snippet can now be re-written using `public` APIs:
https://github.com/Azure/azure-sdk-for-c/blob/75afd98a1efd41d0c1f4475f692da88547a61582/sdk/src/azure/iot/az_iot_hub_client_properties.c#L242
As follows:
```C
if (jr->current_depth == 0)
```

Note when the depth is incremented:
```text
0 {
1, 1	"foo": 123,
1, 1	"bar": {
2, 2		"baz": 1
1	}
0 }
```

cc @danewalton


sdk/inc/azure/core/az_json.h
```diff
@@ -614,6 +614,11 @@ typedef struct
   /// and it shouldn't be modified by the caller.
   az_json_token token;
 
+  /// The depth of the current token. This read-only field tracks the recursive depth of the nested
+  /// objects or arrays within the JSON text processed so far, and it shouldn't be modified by the
+  /// caller.
+  int32_t current_depth;
```
ahsonkhan, line 620: @JeffreyRichter, do you think it makes more sense to expose this field from the `az_json_token` struct rather than from the `az_json_reader`? I have it from the reader because it is easier to access from the caller (the user doesn't have to fish it out from the nested token field), but the depth can be thought of as a property of a particular JSON token.

RickWinter: Can you share the customer code that uses this field.  The usage patterns would probably highlight if it should be on the reader on the token.

ahsonkhan: I did above in the PR description.

ahsonkhan: The reason why I didn't want to add it to the token is because right now, the token can be created as a standalone type, and used outside of the reader (at least as looking at the public fields). Adding depth to it turns it into something that has to be a part of a larger JSON payload.

ahsonkhan: All that said, this is still in beta and we can revisit it separately, if necessary.

danewalton: fwiw I'd prefer in the json reader rather than the token unless there's a good reason for it to be in the token.

sdk/src/azure/core/az_json_reader.c
```diff
@@ -104,6 +107,15 @@ static void _az_json_reader_update_state(
 {
   ref_json_reader->token.kind = token_kind;
   ref_json_reader->token.size = consumed;
+  ref_json_reader->current_depth = ref_json_reader->_internal.bit_stack._internal.current_depth;
+
+  // The depth of the start of the container will be one less than the bit stack managing the state.
+  // That is because we push on the stack when we see a start of the container (above in the call
+  // stack), but it's actual depth and "indentation" level is one lower.
```
danewalton, line 114:
```suggestion
  // stack), but its actual depth and "indentation" level is one lower.
```

danewalton (approved):

sdk/inc/azure/core/az_json.h
ahsonkhan: Will close on it later, and decide if a change is needed or we are happy where we are:
https://github.com/Azure/azure-sdk-for-c/issues/1808

ahsonkhan:
@RickWinter or @antkmsft any other feedback or thoughts before I merge this?

vhvb1989 (approved):

PR #1787, opened 2021-06-18
Incorporate feedback from Azure SDK review board on PnP APIs.

jspaith:
Highligts of changes:
* Prefer the word "write" to "build" when writing JSON properties.  Samples also prefer "write" for JSON creation.
* Remove "_ack" from writable response names.  So status_code instead of ack_code, e.g.
* az_iot.h includes reference to properties header for the one-stop-shopper.
* Have the properties enums be hard-coded #'s, and not point at the underlying twin values they cast from.
  * Adding a UT for now to make sure they're equal.
  * Longer term https://github.com/Azure/azure-sdk-for-c/issues/1784 tracks adding a static_assert style mechanism for this.
* `az_iot_hub_client_properties_update_get_publish_topic` => `az_iot_hub_client_properties_get_reported_publish_topic`.  There's not an ideal way to represent this, but what's in this PR is an improvement over original.

PR #1804, opened 2021-06-29
Incorporating PR feedback from #1787

jspaith:
Take additional PR feedback from #1787.

This will be last non-essential change we make before the beta.


ahsonkhan (approved):
Thanks for addressing the late feedback.

danewalton (approved):

PR #1815, opened 2021-07-07
increase wait period for deployment of hub

danewalton:


RickWinter (approved):

ericwol-msft (approved):

danewalton:
Merging since I can't get the failure to repro. Debugging is now on with the powershell script though so hopefully when/if it rears its head again it will be easier to diagnose.

Issue #1618, opened 2021-02-24
Add base64 encoding and decoding to az_span


danewalton:
Request: Add functions to encode and decode bytes to/from base64.

PR #1816, opened 2021-07-08
Add base64 encoding and decoding APIs that accept az_span.

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/1618

cc @danewalton

### DIFF
--- a/sdk/inc/azure/core/az_base64.h
+++ b/sdk/inc/azure/core/az_base64.h
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ *
+ * @brief Defines APIs to convert between binary data and UTF-8 encoded text that is represented in
+ * base 64.
+ *
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
+ */
+
+#ifndef _az_BASE64_H
+#define _az_BASE64_H
+
+#include <azure/core/az_result.h>
+#include <azure/core/az_span.h>
+
+#include <stdint.h>
+
+#include <azure/core/_az_cfg_prefix.h>
+
+/**
+ * @brief Encodes the span of binary data into UTF-8 encoded text represented as base 64.
+ *
+ * @param destination_base64_text The output #az_span where the encoded base 64 text should be
+ * copied to as a result of the operation.
+ * @param[in] source_bytes The input #az_span that contains binary data to be encoded.
+ * @param[out] out_written A pointer to an `int32_t` that receives the number of bytes written into
+ * the destination #az_span. This can be used to slice the output for subsequent calls, if
+ * necessary.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK Success.
+ * @retval #AZ_ERROR_NOT_ENOUGH_SPACE The \p destination_base64_text is not large enough to contain
+ * the encoded bytes.
+ */
+AZ_NODISCARD az_result
+az_base64_encode(az_span destination_base64_text, az_span source_bytes, int32_t* out_written);
+
+/**
+ * @brief Returns the maximum length of the result if you were to encode an #az_span of the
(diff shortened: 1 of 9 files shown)


sdk/src/azure/core/az_base64.c
```diff
@@ -0,0 +1,521 @@
...
+// The maximum integer length of binary data that can be encoded into base 64 text and still fit
+// into an az_span which has an int32_t length, i.e. (INT32_MAX / 4) * 3;
+#define _az_MAX_SAFE_ENCODED_LENGTH 1610612733
+
+#define _az_ENCODING_PAD '='
+
+static const char _az_base64_encode_array[65]
+    = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+static const int8_t _az_base64_decode_array[256] = {
```
antkmsft, line 18: I would try save some of that 256 bytes and try to implement it with ifs/switch-cases.

sdk/src/azure/core/az_base64.c
```diff
@@ -0,0 +1,521 @@
...
+#include <azure/core/az_base64.h>
+#include <azure/core/internal/az_precondition_internal.h>
+
+#include <azure/core/_az_cfg.h>
+
+// The maximum integer length of binary data that can be encoded into base 64 text and still fit
+// into an az_span which has an int32_t length, i.e. (INT32_MAX / 4) * 3;
+#define _az_MAX_SAFE_ENCODED_LENGTH 1610612733
+
+#define _az_ENCODING_PAD '='
```
antkmsft, line 13: Replace `#define` with unnamed enum?

sdk/src/azure/core/az_base64.c
```diff
@@ -0,0 +1,521 @@
...
+
+static const char _az_base64_encode_array[65]
```
antkmsft, line 15: I think we use right-side const?

antkmsft (approved):

sdk/src/azure/core/az_base64.c
```diff
@@ -0,0 +1,521 @@
...
+  -1,
+  -1,
+  -1,
+  -1,
+  -1,
+  -1,
+  -1,
+};
+
+static AZ_NODISCARD int32_t _az_base64_encode(uint8_t* three_bytes)
```
antkmsft, line 278: Mark a bunch of these statics as AZ_INLINE, especially if they are being used once?

ahsonkhan: That's an implementation detail/compiler optimization, I can do that separately.

sdk/src/azure/core/az_base64.c
ahsonkhan: That's an implementation detail, and not sure if it is worth it from a total code size perspective. I will investigate this in a follow-up:
https://github.com/Azure/azure-sdk-for-c/issues/1817

sdk/src/azure/core/az_base64.c
ahsonkhan: I saw this pattern being used in IoT already, so decided to follow it. We can change it across the board later if there is a coding guideline :)

danewalton (approved):

sdk/inc/azure/core/az_base64.h
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
```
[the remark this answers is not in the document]
ahsonkhan: I believe we got rid of that in most places already (probably the remaining IoT files can remove it too). Let me dig up the rational as to why.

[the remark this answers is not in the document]
ahsonkhan: Here: https://github.com/Azure/azure-sdk-for-c/pull/1057

From @antkmsft 
> * Remove file name from the `@file` section - Doxygen still adds it to documentation perfectly, only that now if you rename a file, you don't need to update the comment, the name will get picked up automatically.

ahsonkhan:
Code coverage looks good (almost 100% line coverage):
https://dev.azure.com/azure-sdk/public/_build/results?buildId=990064&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab
image

The only branches missing is precondition testing:
https://github.com/Azure/azure-sdk-for-c/issues/1836

PR #1840, opened 2021-07-17
C89 compatibility (the old ANSI standard)

SamuelMarks:
Happy to up the warning and permissive level and make this even more strictly compliant, but thought this was a trivial start that should make it C89 compatible enough.

**Advantages**:

  - Support MSVC (no version of C newer than C89 is fully supported in any version of MSVC)
  - Support older compilers targeting esoteric hardware (relevant to your IoT selling-point)
  - Support FFI layers (almost no popular FFI layers in other language supports a standard newer than C89)

**Disadvantages**:

  - None really, as you're not [really] using any newer features anyway.

PS: Just say the word and I'll go strict-mode C89 support.

---

EDIT: Strict-mode C89 support is mostly done.


SamuelMarks:
@danewalton ?

SamuelMarks:
Is there interest in this support? - It'll increase cross-platform cross-toolchain support, and FFI.

danewalton:
Hi @SamuelMarks 
I will let @RickWinter give the ultimate decision here but from the beginning, being C99 compliant was a core tenet of this repository. 

image

While we appreciate the work done here, even if we take these updates, there is no guarantee that in the future, all other parts would be built to be C89 compliant; some parts would work and other parts broken. The user experience for that would not be great and we would break one of the core tenets. Again, I will let others weigh in but that is what I believe the answer will be.

RickWinter:
Hi @SamuelMarks, thank you for taking the time to propose this PR and change.  As Dane mentioned we target C99 and had lengthy discussions about that target.  Although your change is relatively straight forward it does bring complexity to the code base and future maintenance/support costs which we do not want to take on.

That said there was one part of the change which we would accept,
`typedef void (*az_precondition_failed_fn)(void);`

If you'd like to break that out into a PR we would gladly accept that contribution.

PR #1871, opened 2021-08-17
Create az_iot_provisioning_client_get_request_payload API

jspaith:
Creates `az_iot_provisioning_client_get_request_payload` API.  This can optionally be used by IoT/device provisioning clients to include a payload in MQTT publish during provisioning.  

While the custom payload is opaque to SDK, certain backend operations - especially web hooks and even more specifically Plug and Play modelId specification for IoT Central - rely on this.

Verified with unit tests of course and also that a device on the `feature/iot_pnp` branch will successfully have its model ID registered with IoT Central (which previously wasn't happening).

Tag @danewalton @hihigupt since this will let you guys fix [FreeRTOS Middleware issue 63](https://github.com/Azure/azure-iot-middleware-freertos/issues/63).

Also tag @RickWinter @ahsonkhan since this is an API change going into `main` branch, and NOT `feature/iot_pnp` which will arrive way later.  I know you guys had previously reviewed and I took the feedback about reserving the parameter.


sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,47 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
     size_t mqtt_topic_size,
     size_t* out_mqtt_topic_length);
 
+/**
+ * @brief Azure IoT Provisioning Client options for az_iot_provisioning_client_get_request_payload.
+ *        Not currently used.  Reserved for future use.
+ *
+ */
+struct az_iot_provisioning_client_payload_options;
+
```
hihigupt, line 432: how about defining typedef that points to void for now and then in latter we point that typedef to some memory type?


typedef void az_iot_provisioning_client_payload_options;

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,47 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * Can be `NULL`.
+ * @param[in] reserved __[nullable]__ Reserved field for future options to this function.
+ * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
+ * If successful, contains a null-terminated string with the topic filter that needs
+ * to be passed to the MQTT client.
+ * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
+ * @param[out] out_mqtt_payload_length __[nullable]__ Contains the string length, in bytes, of \p
+ * mqtt_payload. Can be `NULL`.
+ * @pre \p client must not be `NULL`.
+ * @pre \p reserved must be `NULL`
```
hihigupt, line 454: Will this break our API, if in future we allowed it being non-null.... Whole point of adding this feild was to not break the contract. 

Note: precondition are also contracts.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,51 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+    size_t* out_mqtt_payload_length)
+{
+  (void)reserved;
+
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_IS_NULL(reserved);
+  _az_PRECONDITION_NOT_NULL(mqtt_payload);
+  _az_PRECONDITION(mqtt_payload_size > 0);
+
+  az_json_writer xJsonWriter;
```
hihigupt, line 580: >xJsonWriter [](http://example.com/codeflow?start=17&length=11)

not sure if we follow the same naming convention in azure-sdk-for-c.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,51 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+    return AZ_ERROR_NOT_ENOUGH_SPACE;
+  }
+  else
+  {
+    mqtt_payload[json_length] = 0;
+    if (out_mqtt_payload_length != NULL)
+    {
+      *out_mqtt_payload_length = json_length;
+    }
+  }
```
hihigupt, line 608: Does this need to null terminated? I understand topic being null terminated but payload can be non-null terminated. Whole motivation to have null terminated topic was paho api that expect null terminated topic, but they have no such limitation on payload.


[https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/pubsync.html](https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/pubsync.html)


This makes me think that we should take uint8_t * as input to this API. What you think?

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: I like this - cleaner and it removes all the `struct` keywoard all over place too.  Fixing.

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: I'd personally prefer to have this just as extra safety guard though its not a make or break of coures.

RE breaking contract if we used this, I'd view this more as a "relaxing the contract" than breaking.  We used to not allow a NULL and now we would if we used this field.

That said, I think this is a great thing to discuss with broader Azure SDK team since I tend to defer on issues like this.

sdk/src/azure/iot/az_iot_provisioning_client.c
jspaith: Fixing this and also xBuffer.

sdk/src/azure/iot/az_iot_provisioning_client.c
jspaith: Yes - this makes a ton of sense.  Even though API review board will need to sign off on a change like this and just generally another look at API, this one makes so much sense I'm just moving now to it.

jspaith (commented):
Thanks @hihigupt or great feedback!  Fixing all but one item around precondition where we'll need to think a bit more.  Updated commit with fixes coming shortly.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,42 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+{
+  (void)reserved;
+
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_IS_NULL(reserved);
+  _az_PRECONDITION_NOT_NULL(mqtt_payload);
+  _az_PRECONDITION(mqtt_payload_size > 0);
+
+  az_json_writer json_writer;
+  az_span payload_buffer = az_span_create((uint8_t*)mqtt_payload, (int32_t)mqtt_payload_size);
```
hihigupt, line 581: >uint8_t [](http://example.com/codeflow?start=43&length=7)

you do not need this cast now that mqtt_payload is uint8_t*

hihigupt (approved):
:shipit:

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: Yea I think John might be right here (fwiw). I think if we went from "can be `NULL` or valid value" to "must be `NULL`" that would break contract. If we in the future let people add it, other's which had it as `NULL` wouldn't be broken. Should confirm with the C sdk team regardless

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ *        Not currently used.  Reserved for future use.
+ *
+ */
+typedef void az_iot_provisioning_client_payload_options;
+
+/**
+ * @brief Builds the optional payload for a provisioning request.
+ * @remark Use this API to build an MQTT payload during registration.
+ *         This call is optional for most scenarios.  Some service
+ *         applications may require #custom_payload_property specified during
```
danewalton, line 437: The `#` here won't link to the parameter. You can use backticks to give it a typewriter font. Links can be generated to functions, enums, typedefs, etc. Parameters can't be resolved since they aren't necessarily unique across all headers.
```suggestion
 *         applications may require `custom_payload_property` specified during
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ *         registration to take additional decisions during provisioning time.
+ *         For example, if you need to register an IoT Plug and Play device you must
+ *         specify its model_id with this API via the #custom_payload_property
```
danewalton, line 440:
```suggestion
 *         specify its model_id with this API via the `custom_payload_property`
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
     size_t mqtt_topic_size,
     size_t* out_mqtt_topic_length);
 
+/**
+ * @brief Azure IoT Provisioning Client options for az_iot_provisioning_client_get_request_payload.
+ *        Not currently used.  Reserved for future use.
+ *
+ */
+typedef void az_iot_provisioning_client_payload_options;
```
ahsonkhan, line 431: We typically create an "empty" struct for such options that we want to reserve for later. For example:

https://github.com/Azure/azure-sdk-for-c/blob/442b28d84d0349b5fec96c60a971d88f49500795/sdk/inc/azure/core/az_json.h#L265-L275

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
+ * @param[out] out_mqtt_payload_length __[nullable]__ Contains the string length, in bytes, of \p
+ * mqtt_payload. Can be `NULL`.
+ * @pre \p client must not be `NULL`.
+ * @pre \p reserved must be `NULL`
+ * @pre \p mqtt_payload must not be `NULL`.
+ * @pre \p mqtt_payload_size must be greater than 0.
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
```
ahsonkhan, line 457: nit: add a changelog entry for the new api

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
     size_t mqtt_topic_size,
     size_t* out_mqtt_topic_length);
 
+/**
+ * @brief Azure IoT Provisioning Client options for az_iot_provisioning_client_get_request_payload.
```
ahsonkhan, line 427: Does this render correctly in doxygen and finds the link to the method?
OR
Do we need to do something like:
```suggestion
 * @brief Azure IoT Provisioning Client options for #az_iot_provisioning_client_get_request_payload.
```
Or even:
```suggestion
 * @brief Azure IoT Provisioning Client options for #az_iot_provisioning_client_get_request_payload().
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ *        Not currently used.  Reserved for future use.
+ *
+ */
+typedef void az_iot_provisioning_client_payload_options;
+
+/**
+ * @brief Builds the optional payload for a provisioning request.
+ * @remark Use this API to build an MQTT payload during registration.
+ *         This call is optional for most scenarios.  Some service
```
ahsonkhan, line 436: nit:
```suggestion
 *         This call is optional for most scenarios. Some service
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ *         applications may require #custom_payload_property specified during
+ *         registration to take additional decisions during provisioning time.
+ *         For example, if you need to register an IoT Plug and Play device you must
+ *         specify its model_id with this API via the #custom_payload_property
+ *         "{\"modelId\":\"" your_model_id "\"}".
```
danewalton, line 441: should this show what it actually looks like and not the C-escaped string? Back ticks will give it that typewriter font and let it stand out.
```suggestion
 *         `{"modelId":"your_model_id"}`.
```
Example:

image

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * @pre \p client must not be `NULL`.
+ * @pre \p reserved must be `NULL`
+ * @pre \p mqtt_payload must not be `NULL`.
+ * @pre \p mqtt_payload_size must be greater than 0.
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* reserved,
```
ahsonkhan, line 460: nit: just rename it as `options`, which is what we would probably call it (though renaming parameter name is non-breaking)

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,42 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* reserved,
+    uint8_t* mqtt_payload,
+    size_t mqtt_payload_size,
+    size_t* out_mqtt_payload_length)
+{
+  (void)reserved;
+
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_IS_NULL(reserved);
```
ahsonkhan, line 576: It's probably fine, but are we sure we want this check? How likely is it we'll add options in the future?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,42 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+  _az_PRECONDITION_NOT_NULL(mqtt_payload);
+  _az_PRECONDITION(mqtt_payload_size > 0);
```
ahsonkhan, line 578: Should `out_mqtt_payload_length` be non-null too?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,42 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+
+  az_json_writer json_writer;
+  az_span payload_buffer = az_span_create((uint8_t*)mqtt_payload, (int32_t)mqtt_payload_size);
+
+  _az_RETURN_IF_FAILED(az_json_writer_init(&json_writer, payload_buffer, NULL));
+  _az_RETURN_IF_FAILED(az_json_writer_append_begin_object(&json_writer));
+  _az_RETURN_IF_FAILED(az_json_writer_append_property_name(&json_writer, prov_registration_id_label));
+  _az_RETURN_IF_FAILED(az_json_writer_append_string(&json_writer, client->_internal.registration_id));
+
+  if  (az_span_size(custom_payload_property) > 0)
```
ahsonkhan, line 588: Is writing an empty payload not valid?

For example, like this:
{
...
"payload" : ""
...
}

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,45 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * @param[in] reserved __[nullable]__ Reserved field for future options to this function.
+ * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
+ * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
+ * @param[out] out_mqtt_payload_length __[nullable]__ Contains the string length, in bytes, of \p
+ * mqtt_payload. Can be `NULL`.
+ * @pre \p client must not be `NULL`.
+ * @pre \p reserved must be `NULL`
+ * @pre \p mqtt_payload must not be `NULL`.
+ * @pre \p mqtt_payload_size must be greater than 0.
+ * @return An #az_result value indicating the result of the operation.
```
ahsonkhan, line 455: Worth documenting some of the common failure conditions/return codes?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -18,6 +18,9 @@ static const az_span str_put_iotdps_register
 static const az_span str_get_iotdps_get_operationstatus
     = AZ_SPAN_LITERAL_FROM_STR("GET/iotdps-get-operationstatus/?$rid=1&operationId=");
 
+static const az_span prov_registration_id_label = AZ_SPAN_LITERAL_FROM_STR("registrationId");
+static const az_span prov_payload_label = AZ_SPAN_LITERAL_FROM_STR("payload");
```
ahsonkhan, line 22: nit: Is there a doc that defines these string literals? Might be a good idea to link to them as a source comment.

ahsonkhan:
> since this is an API change going into `main` branch

This is just an addition, correct, no API (or breaking) change?

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: > I'd view this more as a "relaxing the contract" than breaking. 

👍

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +561,42 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
   return AZ_OK;
 }
+
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* reserved,
+    uint8_t* mqtt_payload,
+    size_t mqtt_payload_size,
+    size_t* out_mqtt_payload_length)
```
danewalton, line 571: should this actually be optional? I don't recall the json writer guaranteeing `NULL` termination and if someone doesn't provide this and the whole buffer is filled, they won't know that it isn't `NULL` terminated anymore (assuming zero'd out buffer).

danewalton (changes requested):
Blocking until the optional length part is answered.

sdk/src/azure/iot/az_iot_provisioning_client.c
danewalton: @ahsonkhan for comment here/tell me if I'm crazy

ahsonkhan: >  I don't recall the json writer guaranteeing `NULL` termination

It doesn't.

You are right, @danewalton .

Since we are using the json_writer internally as an impl detail here, when this param is optional, there wouldn't be any way for the user to know how much was actually written to the buffer.

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: you can just have the `()`. Doxygen knows if it hits a `()` to find the documented function. No `#` needed.
```suggestion
 * @brief Azure IoT Provisioning Client options for az_iot_provisioning_client_get_request_payload().
```

jspaith: Fixing.  As I said in another comment I'll more closely review the doxygen output and fixup any rough spots I see.

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: @ahsonkhan - Great question.  I can do that after... my question to you is to double check you're OK with the word "some".

In general when I see API documentation with error codes, I always assume that the API could return other stuff, too, so I'm forgiving as the caller.  I think that's what most developers would do.

I'm asking here because this API's errors are ultimately going to be coming from az_json_writer, which could in theory add new errors in future that just realistically we're not going to update in the az_prov_ space.  

Also most functions in this header don't have retval.  Once we're on same page about "some", I'll open a GitHub issue to track adding them.

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: @ahsonkhan - Yes.  

How do I do that, though?  The changelog is tied to releases whereas this will will just go in a "random" commit.  Is there some mechanism to track release IOU and when the release notes come out?

sdk/src/azure/iot/az_iot_provisioning_client.c
jspaith: Fixing.

jspaith (commented):
Addressed many of the comments (thanks yet again guys!) with some ones we need more discussion on to be resolved.  Before I push next commit that incorporates all this & would be ready for review, I'll stare at doxygen output more, run clang-format since this isn't cool with rules, etc.

danewalton (approved):
unblocking since the optional length looks to be worked out. rest of it looks good to me

sdk/src/azure/iot/az_iot_provisioning_client.c
jspaith: It's possible but fairly unlikely we'd add options.  The most likely scenario would be we wanted app to specify some custom JSON into the {...} top level payload.  The issue though is that the `custom_payload` already is the escape hatch for apps to add any arbitrary JSON they want, in a payload:<userDefined>.  So putting yet another option as a peer to "payload" feels unlikely.  Though I'm all for having an escape hatch.

I don't really have strong opinions either way on this as most devs know to leave something like this as NULL.

jspaith:
> This is just an addition, correct, no API (or breaking) change?

Yes.  This is addition only, nothing breaking.

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: Yep, I totally agree.

Trying to list and solve the "transitive closure" of all error codes possible from all callers within the call tree of a particular function could become a daunting task and wouldn't scale well from a docs perspective (and as impl changes/evolves, an untenable maintenance burden).

We had loosely discussed this in the past (at least within the core team), and since we were generally at the bottom, listing more was a bit easier.

This becomes more of an art than a science, and the implementer of the method can dictate what are "some" of the more relevant or most commonly occurring return codes the user cares about that, specific to this, and would be useful for the caller to know. If the dev doesn't have a good sense of which codes might be useful in the caller scenario, maybe "tree-pruning" to just 1 level down the call stack for certain impl details is a reasonable heuristic if we are flying blind (while keeping an eye on the list not growing too large/becoming too noisy).

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: You just add a `new feature` entry to the changelog markdown file here, for the "next" release:
https://github.com/Azure/azure-sdk-for-c/blob/main/CHANGELOG.md#120-beta2-unreleased

We try to add CL entries to that README in the same PR as the feature (or bug fix), and whenever we ship that particular version, those CL contents get auto-copied as part of the release notes (as part of CI/release process). The "iou" part is the fact that the next release is marked as "unreleased", until we decide to add a ship date to it and cut a release (which produces a git release and tag).

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: So, yes, this required a bit of a judgement call, and we should let the caller scenario drive the decision of what error codes to highlight in docs. One question that might be useful to ask is: What error cases do I expect most callers of my function to handle/write code against?

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: Adding.

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: If they don't get OK, the error overwhelmingly likely to cause this to go wrong would be a `AZ_ERROR_NOT_ENOUGH_SPACE`.  I'll document that.

Trying to fix up other places in this header that need it is well beyond scope of this PR.  Opened #1878 as our IOU so we don't lose this.

jspaith:
I've just pushed a bunch of changes which should be the final product or something close.  I'm NOT going to check in until we can discuss a bit more offline given this is an API about to be locked in stone and see if anyone else wants to come in.

I think the only controversial issue is whether we leave the `_az_PRECONDITION_IS_NULL(options);` line in or not, which is quite ironic given it'll get compiled out no matter what 99% of time anyway:).  But let's do it right even on small stuff.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 1.2.0-beta.2 (Unreleased)
 
+- Add `az_iot_provisioning_client_get_request_payload()` to create MQTT payload bodies during Device Provisioning.
```
ahsonkhan, line 5: Nice!

README.md
```diff
@@ -302,7 +302,7 @@ git clone https://github.com/Microsoft/vcpkg.git
 # (consider this path as PATH_TO_VCPKG)
 cd vcpkg
 # Checkout the vcpkg commit from the vcpkg-commit.txt file (link above)
-git checkout <vcpkg commit>
+git checkout <vcpkg commit>
```
ahsonkhan, line 305: Revert this unrelated change?
```suggestion
git checkout <vcpkg commit>
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,56 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * `NULL`.
+ * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
+ * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
+ * @param[out] out_mqtt_payload_length Contains the length, in bytes, written to \p mqtt_payload on
+ * success.
+ * @pre \p client must not be `NULL`.
+ * @pre \p options must be `NULL`.
+ * @pre \p mqtt_payload must not be `NULL`.
+ * @pre \p mqtt_payload_size must be greater than 0.
+ * @pre \p out_mqtt_payload_length must be `NULL`.
```
ahsonkhan, line 463: Hmm... I think this doc comment is wrong. It must not be null, right?
```suggestion
 * @pre \p out_mqtt_payload_length must not be `NULL`.
```

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -18,6 +18,11 @@ static const az_span str_put_iotdps_register
 static const az_span str_get_iotdps_get_operationstatus
     = AZ_SPAN_LITERAL_FROM_STR("GET/iotdps-get-operationstatus/?$rid=1&operationId=");
 
+// From the protocol described in
+// https://docs.microsoft.com/en-us/azure/iot-dps/iot-dps-mqtt-support#registering-a-device
```
ahsonkhan, line 22: 👍

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +563,43 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
+
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* options,
+    uint8_t* mqtt_payload,
+    size_t mqtt_payload_size,
+    size_t* out_mqtt_payload_length)
+{
+  (void)options;
```
ahsonkhan, line 575: nit question: should this happen after the precondition statement?

ahsonkhan (approved):
Looks good other than fixing up the doc comment and removing unnecessary space in a markdown file.

ahsonkhan:
The code coverage for unit tests looks decent, but if we want to get to higher branch coverage, maybe we can add some more "failure" tests, or file an up-for-grabs/good-first-issue issue to improve the test coverage:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1074550&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab

image

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -423,6 +423,56 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
...
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The payload was created successfully.
+ * @retval #AZ_ERROR_NOT_ENOUGH_SPACE The buffer is too small.
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* options,
+    uint8_t* mqtt_payload,
```
CIPop, line 472: We've used `char*` everywhere else.  
I see Paho has `void* `.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -558,3 +563,43 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
 
   return AZ_OK;
 }
+
+AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* options,
+    uint8_t* mqtt_payload,
```
CIPop, line 571: same as above re type.

CIPop (approved):
Discussed uint8_t issue offline. LGTM!

sdk/inc/azure/iot/az_iot_provisioning_client.h
jspaith: Resolving per offline discussion to stick with uint8_t.

sdk/src/azure/iot/az_iot_provisioning_client.c
jspaith: It doesn't seem like even in the repo we're consistent as sometimes the (unreferencedVar); thingy comes up front and others its after preconditions.  In this file FWIW though, it's following pattern.  See `az_iot_provisioning_client_parse_received_topic_and_payload` e.g.

jspaith (commented):
Responding to additional comments; updated commit to follow shortly.

jspaith:
I've created [an issue](https://github.com/Azure/azure-sdk-for-c/issues/1886) to track suggestion around some additional tests.

PR #1891, opened 2021-09-02
Merge feature/iot_pnp branch into main

jspaith:
Adding [IoT Plug and Play](https://aka.ms/iotpnp) APIs to embedded SDK.

PR #1897, opened 2021-09-03
Cleanup pre-release

jspaith:
Important fixes for PnP pre-release:
* Mark `pnp_protocol.*` files as `// DEPRECATED`.  These are the helpers that we provided before the official API that came in #1891 was available.  
  * While most vestigial samples we remove, it's possible customers are using these sample helpers in their product.  So we don't want to `git delete` in this case.
  * This also addresses @ahsonkhan feedback in [this PR thread](https://github.com/Azure/azure-sdk-for-c/pull/1887/files/ad5a9a8749d0111b53367a0e3ad746ac14f9f880#r700646754).
* Samples had vestigial files and references to files `paho_iot_hub_pnp_component_sample.c` and `paho_iot_hub_pnp_sample.c`.  These were using the `pnp_protocol.*` stuff that was deprecated AND were replaced by `paho_iot_pnp_component_sample.c` and `paho_iot_pnp_sample.c`, respectively.  
  * In this case the old samples (with "iot_hub_pnp*" and not the new "iot_pnp*") would be confusing to leave around.  They weren't being touch at all by any CMake rules, so just removing.
* Have the provisioning sample send its maximum temperature always, not just when receiving something > default.  This is *required* to make the sample [PnP Certified](https://docs.microsoft.com/en-us/azure/certification/program-requirements-pnp). 
  * While presumably no one is shipping this sample much less certifying it!, we want to make it work out of the box with backend services in any event.


jspaith:
One question for @ahsonkhan / @RickWinter  - should I update the changelog.md in this checkin, or is that only something done as a later stage of release?  We have text from PM.

danewalton (approved):

hihigupt (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## 1.2.0-beta.2 (Unreleased)
 
 - Add `az_iot_provisioning_client_get_request_payload()` to create MQTT payload bodies during Device Provisioning.
+- This version provides new APIs to follow the IoT Plug and Play convention to implement Telemetry, Commands, Properties and Components defined in a DTDL model.
+  - To read/write properties the SDK now provides functions to produce the right payload for components, as shown in the header sdk/inc/azure/iot/az_iot_hub_client_properties.h.
+  - To send telemetry messages, the required header is added to identify components.
+  - When responding to a command invocation the component name is automatically parsed and provided when available.
+  - All new samples follow the IoT Plug and Play convention and can be connected to IoT Hub (with or without DPS), or IoT Central.
```
ahsonkhan, line 10: What's DPS?

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## 1.2.0-beta.2 (Unreleased)
 
 - Add `az_iot_provisioning_client_get_request_payload()` to create MQTT payload bodies during Device Provisioning.
+- This version provides new APIs to follow the IoT Plug and Play convention to implement Telemetry, Commands, Properties and Components defined in a DTDL model.
+  - To read/write properties the SDK now provides functions to produce the right payload for components, as shown in the header sdk/inc/azure/iot/az_iot_hub_client_properties.h.
```
ahsonkhan, line 7: I made this fix in https://github.com/Azure/azure-sdk-for-c/pull/1899 if that's ok. @jspaith are you ok with dropping `sdk/inc` from the path?

The header path the customer needs to include would be as follows:
```C
#include <azure/iot/az_iot_hub_client_properties.h>
```

That said, they wouldn't need to explicitly include it if they are already including the top-level `az_iot.h`

```suggestion
  - To read/write properties, the SDK now provides functions to produce the right payload for components, as shown in the header `azure/iot/az_iot_hub_client_properties.h`.
```

jspaith: Yup I'm cool with both of these changes thanks for piggy backing into #1899.

CHANGELOG.md
jspaith: Device Provisioning Service (https://docs.microsoft.com/en-us/azure/iot-dps/).  If you want to sneak in the acrynom expansion into #1899 that'd work for me too.  Thanks

ahsonkhan: Seemed folks would have this domain understanding if they are using the service with the SDK.

PR #1902, opened 2021-09-07
Update dependency types to linux

### LINKED ISSUE #1838: Use the "linux" component type within the cgmanifest whenever appropriate (e.g. clang-format)
For example, the following is currently marked as "other":
https://github.com/Azure/azure-sdk-for-c/blob/75a4910e7e034949897e9cc2cb88f4a168391f3d/cgmanifest.json#L86-L96

https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/
image

> The component type defined the package type it can be npm, nuget, maven, etc ... in this case this is an ubuntu package for what I can see, when the packages are from ubuntu, debian alpine or other linux distribution we recommend to use the component type "linux"

cc @RickWinter, @jcfiorenzano

antkmsft:
Closes #1838

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -52,7 +52,7 @@
     },
     {
       "Component": {
-        "Type": "other",
+        "Type": "linux",
         "Other": {
           "Name": "gcc-arm-none-eabi",
           "Version": "6.3.1+svn253039-1build1",
@@ -63,7 +63,7 @@
     },
     {
       "Component": {
-        "Type": "other",
+        "Type": "linux",
         "Other": {
           "Name": "gcovr",
           "Version": "3.4-1",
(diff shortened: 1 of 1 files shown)


RickWinter (approved):

cgmanifest.json
```diff
@@ -52,7 +52,7 @@
     },
     {
       "Component": {
-        "Type": "other",
+        "Type": "linux",
```
ahsonkhan, line 55: This change alone doesn't fix the issue. The "linux" type cgmanifest has a different JSON schema and info required than that of "other".

Is cgmanifest picking this up correctly? Can you check the docs to make sure we are following the right JSON format?
https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/

ahsonkhan: See https://dev.azure.com/azure-sdk/public/_componentGovernance/132714?_a=alerts&typeId=3940835&alerts-view-option=active

image

PR #1904, opened 2021-09-07
Convert cfgmanifest linux entries to matching format

### LINKED ISSUE #1838: Use the "linux" component type within the cgmanifest whenever appropriate (e.g. clang-format)

antkmsft:
Closes #1838.

### DIFF
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -53,10 +53,11 @@
     {
       "Component": {
         "Type": "linux",
-        "Other": {
+        "linux": {
           "Name": "gcc-arm-none-eabi",
           "Version": "6.3.1+svn253039-1build1",
-          "DownloadUrl": "https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/gcc-arm-none-eabi_6.3.1+svn253039-1build1_amd64.deb.html"
+          "Distribution": "Ubuntu",
+          "Release": "18.04"
         }
       },
       "DevelopmentDependency": true
--- a/cgmanifest.json
+++ b/cgmanifest.json
@@ -64,10 +64,11 @@
     {
       "Component": {
         "Type": "linux",
-        "Other": {
+        "linux": {
           "Name": "gcovr",
           "Version": "3.4-1",
-          "DownloadUrl": "https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/gcovr_3.4-1_all.deb.html"
+          "Distribution": "Ubuntu",
+          "Release": "18.04"
         }
       },
       "DevelopmentDependency": true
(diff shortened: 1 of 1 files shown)


cgmanifest.json
```diff
@@ -64,32 +64,35 @@
     {
       "Component": {
         "Type": "linux",
-        "Other": {
+        "linux": {
```
ahsonkhan, line 67: Need to fix `gcc-arm-none-eabi` as well, above.

cgmanifest.json
```diff
@@ -64,32 +64,35 @@
...
           "Name": "clang-format",
           "Version": "9.0.0-2",
-          "DownloadUrl": "https://ubuntu.pkgs.org/18.04/ubuntu-updates-universe-amd64/clang-format-9_9-2~ubuntu18.04.2_amd64.deb.html"
+          "Distribution": "Ubuntu",
+          "Release": "18.04"
```
ahsonkhan, line 95: I wonder if we need to specify `18.04.2` for this one, or if cg finds the package correctly from the distribution. We may need to test this out.

cgmanifest.json
```diff
@@ -64,32 +64,35 @@
     {
       "Component": {
         "Type": "linux",
-        "Other": {
+        "linux": {
           "Name": "gcovr",
           "Version": "3.4-1",
-          "DownloadUrl": "https://ubuntu.pkgs.org/18.04/ubuntu-universe-amd64/gcovr_3.4-1_all.deb.html"
+          "Distribution": "Ubuntu",
+          "Release": "18.04"
```
ahsonkhan, line 71: I think we may also need Pool-URL and/or Key-URL.

Maybe follow up with cg folks or look up other repos that use the `linux` component type, to see what is the full amount of detail we need to provide.

> Valid values for distribution: Debian, Alpine, Rhel, Centos, Fedora, Ubuntu. Pool-URL Example (http://archive.ubuntu.com/ubuntu) and Key-URL Example (http://archive.ubuntu.com/ubuntu/project/ubuntu-archive-keyring.gpg)

RickWinter (approved):

ahsonkhan (approved):

cgmanifest.json
jcfiorenzano: Yes this linux dependency looks correct

cgmanifest.json
jcfiorenzano: I think that if you can specify 18.04.2 is better

cgmanifest.json
ahsonkhan: Where do we typically find "valid" values for Pool and Key URLs for the different distros?

PR #1905, opened 2021-09-08
az_json_writer_append_json_text does not take into account required buffer space for commas

hwmaier:
The `az_json_writer_append_json_text()` function does not take into account required buffer space for commas when tracking buffer size with `_az_update_json_writer_state()`.

So for example if 5 array elements are written using `az_json_writer_append_json_text()`,  the resulting JSON span is short of 4 (5-1) characters and the resulting JSON string is incomplete, typically the closing braces and a few last characters show missing in the stream.

This patch fixes this issue.

### DIFF
--- a/sdk/src/azure/core/az_json_writer.c
+++ b/sdk/src/azure/core/az_json_writer.c
@@ -782,10 +782,12 @@ az_json_writer_append_json_text(az_json_writer* ref_json_writer, az_span json_te
   az_span remaining_json = _get_remaining_span(ref_json_writer, _az_MINIMUM_STRING_CHUNK_SIZE);
   _az_RETURN_IF_NOT_ENOUGH_SIZE(remaining_json, _az_MINIMUM_STRING_CHUNK_SIZE);
 
+  int32_t required_size = az_span_size(json_text);
   if (ref_json_writer->_internal.need_comma)
   {
     remaining_json = az_span_copy_u8(remaining_json, ',');
     ref_json_writer->_internal.bytes_written++;
+    required_size++; // For the leading comma separator.
   }
 
   _az_RETURN_IF_FAILED(
@@ -799,7 +801,7 @@ az_json_writer_append_json_text(az_json_writer* ref_json_writer, az_span json_te
   // Therefore, need_comma must be true after appending the json_text.
 
   // We already tracked and updated bytes_written while writing, so no need to update it here.
-  _az_update_json_writer_state(ref_json_writer, 0, az_span_size(json_text), true, last_token_kind);
+  _az_update_json_writer_state(ref_json_writer, 0, required_size, true, last_token_kind);
   return AZ_OK;
 }
 


sdk/src/azure/core/az_json_writer.c
```diff
@@ -799,7 +801,7 @@ az_json_writer_append_json_text(az_json_writer* ref_json_writer, az_span json_te
   // Therefore, need_comma must be true after appending the json_text.
 
   // We already tracked and updated bytes_written while writing, so no need to update it here.
-  _az_update_json_writer_state(ref_json_writer, 0, az_span_size(json_text), true, last_token_kind);
+  _az_update_json_writer_state(ref_json_writer, 0, required_size, true, last_token_kind);
```
ahsonkhan, line 804: Though this fixes up `_internal.total_bytes_written`, this seems to be incrementing `_internal.bytes_written` by one more than necessary, because it is already being incremented on line 789.

sdk/src/azure/core/az_json_writer.c
```diff
@@ -782,10 +782,12 @@ az_json_writer_append_json_text(az_json_writer* ref_json_writer, az_span json_te
   az_span remaining_json = _get_remaining_span(ref_json_writer, _az_MINIMUM_STRING_CHUNK_SIZE);
   _az_RETURN_IF_NOT_ENOUGH_SIZE(remaining_json, _az_MINIMUM_STRING_CHUNK_SIZE);
 
+  int32_t required_size = az_span_size(json_text);
   if (ref_json_writer->_internal.need_comma)
   {
     remaining_json = az_span_copy_u8(remaining_json, ',');
     ref_json_writer->_internal.bytes_written++;
+    required_size++; // For the leading comma separator.
```
ahsonkhan, line 790: Can you add some unit tests for this, that showcase the error before the fix?

sdk/src/azure/core/az_json_writer.c
ahsonkhan: Nevermind, I see the bytes_written increment parameter continues to be 0.

ahsonkhan (approved):
Great catch and fix!

sdk/src/azure/core/az_json_writer.c
ahsonkhan: If you only use publicly exposed APIs in unit tests, the issue isn't visible to the end user:

```C
 {
      TEST_EXPECT_SUCCESS(az_json_writer_init(&writer, AZ_SPAN_FROM_BUFFER(array), NULL));
      TEST_EXPECT_SUCCESS(az_json_writer_append_begin_array(&writer));
      TEST_EXPECT_SUCCESS(az_json_writer_append_json_text(&writer, AZ_SPAN_FROM_STR("1")));
      TEST_EXPECT_SUCCESS(az_json_writer_append_json_text(&writer, AZ_SPAN_FROM_STR("2")));
      TEST_EXPECT_SUCCESS(az_json_writer_append_json_text(&writer, AZ_SPAN_FROM_STR("3")));
      TEST_EXPECT_SUCCESS(az_json_writer_append_end_array(&writer));

      az_span_to_str((char*)array, 200, az_json_writer_get_bytes_used_in_destination(&writer));
      assert_int_equal(7, az_span_size(az_json_writer_get_bytes_used_in_destination(&writer)));
      assert_string_equal(array, "[1,2,3]");
    }
```

hwmaier: That may be the case that the unit test does not pick this up. But we picked it up when using the API from the Azure RTOS IoT middleware's `nx_azure_iot_json_writer_append_json_text()` function. When using this SDK via Azure RTOS IoT middleware, the resulting NX_PACKETs do hold wrong length information. So without the PR some internal data structure is not updated correctly. You also can see by reviewing the code that the additional comma is not taken into account in the same way as it it with the other append functions. All other append functions take into account the additional comma when calling `_az_update_json_writer_state()`.

ahsonkhan: I agree there is a correctness issue here :)
Ideally, the upstream dependencies do not use internal state like this directly, since we don't guarantee to not change or break the internal state. In fact, this particular field was only there (atm) for testing only:
https://github.com/Azure/azure-sdk-for-c/blob/9b5ef3d889b014d6e1b777d80f495dae27fe3ae1/sdk/inc/azure/core/az_json.h#L311

ahsonkhan: > Can you add some regression unit tests for this, that showcase the error before the fix?

Added a test here using the internals, just to avoid regression in the future: https://github.com/Azure/azure-sdk-for-c/pull/1907

PR #1909, opened 2021-09-08
Update the changelog to include a recent bug fix by community member.

ahsonkhan:


ahsonkhan:
cc @hwmaier

vhvb1989 (approved):

antkmsft (approved):
You probably gonna need to update https://github.com/Azure/azure-sdk-for-c/blob/main/.vscode/cspell.json, in the similar way for at least `hwmaier`, maybe for the parts of the name as well:
https://github.com/Azure/azure-sdk-for-cpp/blob/main/.vscode/cspell.json#L128-L142

ahsonkhan:
It doesn't fail the build currently, and we have a few other warnings to fix-up. Will file an issue to fix all of them together, for the next milestone. Then, we can turn spell check to fail the build, so we stay green.

image

PR #1943, opened 2021-09-30
Azure Storage Blobs Upload

antkmsft:


jameszah:
Hi, don't know if I allowed to comment in here!

I'm trying to use this code to upload .jpg and .avi files to azure blobs from an esp32.  It generates a ...

"non-trivial designated initializers not supported" 
https://stackoverflow.com/questions/31215971/non-trivial-designated-initializers-not-supported 

... when using the arduino compiler, but can be fixed by re-ordering the parameters to the original sequence.

https://github.com/Azure/azure-sdk-for-c/blob/main/sdk/inc/azure/core/internal/az_http_internal.h

line 79-83
```
  return (_az_http_policy_apiversion_options){
    ._internal = { .name = AZ_SPAN_EMPTY,
		.version = AZ_SPAN_EMPTY,
		.option_location = _az_http_policy_apiversion_option_location_header}
  };
```

Haven't got it working yet, but assume it is much faster than the IOTHub blob-upload code.

antkmsft:
Hi @jameszah, I replied to your comment in the original issue, and as of the error that the compiler is giving, I think we can make a PR that corrects initialization to match the declaration order. Arduino should not be doing that, since any order is legal C99, but I don't think it would cost us anything to rearrange the order and make Arduino's compiler happy :)

PR #1968, opened 2021-11-01
HTTP request telemetry improvements

antkmsft:


sdk/src/azure/core/az_http_policy.c
```diff
@@ -48,11 +47,32 @@ AZ_NODISCARD az_result az_http_pipeline_policy_telemetry(
     az_http_request* ref_request,
     az_http_response* ref_response)
 {
+  // https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy
+  uint8_t telemetry_id_buffer[200] = { 0 };
+  az_span telemetry_id = AZ_SPAN_FROM_BUFFER(telemetry_id_buffer);
+  {
+    _az_http_policy_telemetry_options* options = (_az_http_policy_telemetry_options*)(ref_options);
+
+    az_span remainder = az_span_copy(telemetry_id, AZ_SPAN_FROM_STR("azsdk-c"));
```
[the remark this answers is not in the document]
kyle-patterson: I would expect, in general, `azsdk-c-<packageName>/<packageVersion>`. I'm not sure what `options->component_name` will reference, does this imply that someone can set the package name through a custom telemetry options instead?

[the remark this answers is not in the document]
antkmsft: The format spec is here: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy:
`[<application_id> ]azsdk-<sdk_language>-<package_name>/<package_version> <platform_info>`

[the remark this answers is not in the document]
antkmsft: So, the storage APIs are going to send `azsdk-c-storage-blobs/1.3.0-beta.1`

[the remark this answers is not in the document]
antkmsft: I.e. to clarify:
1. @RickWinter, `ComponentName` is a terminology from the spec. It is where the Package Name is in other SDKs.
2. 2. @kyle-patterson, exactly - blob client is going to pass `"storage-blobs"`
 to this policy, so the telemetry string we are going to see being sent by these clients is going to be `azsdk-c-storage-blobs/1.3.0-beta.1`.

sdk/src/azure/core/az_http_policy.c
```diff
@@ -48,11 +47,32 @@ AZ_NODISCARD az_result az_http_pipeline_policy_telemetry(
...
 
-  _az_http_policy_telemetry_options* options = (_az_http_policy_telemetry_options*)(ref_options);
+    {
+      az_span const component_name = options->component_name;
+      if (az_span_size(component_name) > 0)
```
RickWinter, line 60: Error out if the the component name is not specified.  That is invalid

RickWinter, line 60: Add an ASSERT for when component_name is not specified.  The component_name is required

antkmsft: Technically, that would be a breaking change, but currently no client has been using it.

antkmsft: I'll add it.

vhvb1989 (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features Added
 
 - Added `az_http_response_get_status_code()` convenience function to get HTTP status code from requests.
+- Improved HTTP request telemetry.
```
ahsonkhan, line 8: Is this a new feature or should it belong in the "Other Changes" section?

sdk/inc/azure/core/internal/az_http_internal.h
```diff
@@ -70,7 +70,7 @@ typedef struct
  */
 AZ_NODISCARD AZ_INLINE _az_http_policy_telemetry_options _az_http_policy_telemetry_options_default()
 {
-  return (_az_http_policy_telemetry_options){ .os = AZ_SPAN_FROM_STR("Unknown OS") };
+  return (_az_http_policy_telemetry_options){ 0 };
```
ahsonkhan, line 73: Should the default option value have some "placeholder" component_name? Like "package-name"?

sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,21 +42,44 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
   return _az_http_pipeline_nextpolicy(ref_policies, ref_request, ref_response);
 }
 
+#define _az_TELEMETRY_ID_PREFIX "azsdk-c-"
+#define _az_TELEMETRY_ID_PREFIX_LENGTH (sizeof(_az_TELEMETRY_ID_PREFIX) - 1)
```
ahsonkhan, line 46: Is the -1 to account for the null terminator?

nit: Might be worth adding a comment.

sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,21 +42,44 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
...
+
 AZ_NODISCARD az_result az_http_pipeline_policy_telemetry(
     _az_http_policy* ref_policies,
     void* ref_options,
     az_http_request* ref_request,
     az_http_response* ref_response)
 {
+  _az_PRECONDITION_NOT_NULL(ref_options);
+
+  // Format spec: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy
```
ahsonkhan, line 56: 👍  for the comment

sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,21 +42,44 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
...
+  uint8_t telemetry_id_buffer[200] = _az_TELEMETRY_ID_PREFIX;
```
ahsonkhan, line 57: Any particular reason to pick 200 for the size?

sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,21 +42,44 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
...
+  az_span telemetry_id = AZ_SPAN_FROM_BUFFER(telemetry_id_buffer);
+  {
+    az_span remainder = az_span_slice_to_end(telemetry_id, _az_TELEMETRY_ID_PREFIX_LENGTH);
 
-  _az_http_policy_telemetry_options* options = (_az_http_policy_telemetry_options*)(ref_options);
+    _az_http_policy_telemetry_options* options = (_az_http_policy_telemetry_options*)(ref_options);
+    az_span const component_name = options->component_name;
+    _az_PRECONDITION_VALID_SPAN(component_name, 1, false);
```
ahsonkhan, line 64: nit (only because it is currently internal): I believe, if a caller passes in the `_az_http_policy_telemetry_options_default` here, they will fail this precondition. We should make the default options usable, and not trigger an error.

ahsonkhan: We should strengthen the precondition to say the component name cannot be too large. Otherwise, if someone picked a component name that is too large (over 200 characters), `az_span_copy` later will fail.

Use `_az_PRECONDITION_RANGE` on the `component_name` span size.

Given we know the format, and what other parts of the string we create, we should be able to calculate the upper bound of the component name length that's allowed, right?

sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,21 +42,44 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
...
+    remainder = az_span_copy(remainder, component_name);
+
+    remainder = az_span_copy_u8(remainder, '/');
+    remainder = az_span_copy(remainder, AZ_SPAN_FROM_STR(AZ_SDK_VERSION_STRING));
+
+    telemetry_id = az_span_slice(
+        telemetry_id, 0, (int32_t)(az_span_ptr(remainder) - az_span_ptr(telemetry_id)));
```
ahsonkhan, line 71: What is this code doing exactly? Can you use `_az_span_diff`? 

Deserves a comment.

sdk/src/azure/core/az_http_policy.c
```diff
@@ -4,14 +4,14 @@
...
 #include <azure/core/az_http.h>
+#include <azure/core/az_precondition.h>
 #include <azure/core/az_span.h>
+#include <azure/core/az_version.h>
 #include <azure/core/internal/az_http_internal.h>
 #include <azure/core/internal/az_result_internal.h>
 
 #include <azure/core/_az_cfg.h>
 
-static const az_span AZ_HTTP_HEADER_USER_AGENT = AZ_SPAN_LITERAL_FROM_STR("User-Agent");
```
ahsonkhan, line 13: Why remove this, when we are still doing the User-Agent string creation, down below to call `az_http_request_append_header`?

sdk/src/azure/core/az_http_policy.c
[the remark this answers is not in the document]
ahsonkhan: > `ComponentName` is a terminology from the spec.

Which spec? Should we just call it `package_name`?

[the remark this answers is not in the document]
antkmsft: Spec: https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy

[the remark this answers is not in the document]
antkmsft: plus, we have no packages in this SDK.

sdk/inc/azure/core/internal/az_http_internal.h
antkmsft: No, then assert will never work.
Previous commits here had no assert, and would go with just `azsdk-c/1.3.0-beta.1` when the component name is empty. But assert was suggested, and I think it is better.

sdk/src/azure/core/az_http_policy.c
antkmsft: It is used only once, so I inlined it.

sdk/src/azure/core/az_http_policy.c
antkmsft: It will not fail, it will be trimmed.
If you say that the default options can't fail, that sounds like you would prefer the initial behavior in this PR (see previous commits). But I think that having a precondition brings more value.
Having a precondition on non-empty + having a default non-empty value is a change for worse IMO, it does not solve anything, bringing only disadvantages of both approaches.

PR #1972, opened 2021-11-02
add callout for json writer space needed

danewalton:
We had another customer curious about this behavior so thought it might be useful to have it in the docs to hopefully make it easier to find.

cc @daisukeiot

### DIFF
--- a/sdk/inc/azure/core/az_json.h
+++ b/sdk/inc/azure/core/az_json.h
@@ -389,17 +389,17 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
 
 /**
  * @brief Appends the UTF-8 text value (as a JSON string) into the buffer.
- * 
- * @note For details on possibly receiving #AZ_ERROR_NOT_ENOUGH_SPACE while appending a string for 
+ *
+ * @note For details on possibly receiving #AZ_ERROR_NOT_ENOUGH_SPACE while appending a string for
  * which there is theoretically space, please see this github issue here for an explanation.
- * 
+ *
  * https://github.com/Azure/azure-sdk-for-c/issues/1030
- * 
+ *
  * The high level reasoning copied from the issue is below.
- * 
- * "The JSON writer pessimistically requires at least 64-bytes of space when writing any string larger
- * than 10 characters because it tries to write in 64 byte chunks (10 character * 6 if all need to be
- * escaped into the unicode form \uXXXX)."
+ *
+ * "The JSON writer pessimistically requires at least 64-bytes of space when writing any string
+ * larger than 10 characters because it tries to write in 64 byte chunks (10 character * 6 if all
+ * need to be escaped into the unicode form \uXXXX)."
  *
  * @param[in,out] ref_json_writer A pointer to an #az_json_writer instance containing the buffer to
  * append the string value to.
--- a/sdk/inc/azure/core/az_json.h
+++ b/sdk/inc/azure/core/az_json.h
@@ -390,16 +390,12 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
 /**
  * @brief Appends the UTF-8 text value (as a JSON string) into the buffer.
  *
- * @note For details on possibly receiving #AZ_ERROR_NOT_ENOUGH_SPACE while appending a string for
- * which there is theoretically space, please see this github issue here for an explanation.
- *
- * https://github.com/Azure/azure-sdk-for-c/issues/1030
- *
- * The high level reasoning copied from the issue is below.
- *
- * "The JSON writer pessimistically requires at least 64-bytes of space when writing any string
- * larger than 10 characters because it tries to write in 64 byte chunks (10 character * 6 if all
- * need to be escaped into the unicode form)."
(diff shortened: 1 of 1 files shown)


sdk/inc/azure/core/az_json.h
```diff
@@ -390,6 +390,17 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
...
+ * which there is theoretically space, please see this github issue here for an explanation.
+ *
+ * https://github.com/Azure/azure-sdk-for-c/issues/1030
+ *
+ * The high level reasoning copied from the issue is below.
+ *
+ * "The JSON writer pessimistically requires at least 64-bytes of space when writing any string
+ * larger than 10 characters because it tries to write in 64 byte chunks (10 character * 6 if all
+ * need to be escaped into the unicode form)."
+ *
```
ahsonkhan, line 403: Good idea!

However, rather than referencing a GitHub issue in the doc comments, let's just mention the reason in-line.

Something like what you summarized is good (we can leave out the issue). I would add:
> The JSON writer requires at least 64-bytes of slack within the output buffer, above the theoretical minimal space needed.

I can help wordsmith it some more in a bit :)

Question: `@remarks` vs `@note`, which renders better in this case?

ahsonkhan: Also, this pessimistic approach and a bit of slack space isn't just need for strings, the same applies to any thing that's variable, including numbers and property names. Once we settle on the text, we should add such a note to those APIs too.

So, even calls to `append_int32(writer, 1)` would fail, if the buffer doesn't have enough slack, need at least 11 bytes even though 1 could, theoretically, be enough. For simplicity to users, just providing at least 64-bytes of padding above the "expected" size works across the board. The caller needs to be able to handle "not enough size" regardless, since JSON output is dynamic in nature, based on changing scenarios.

https://github.com/Azure/azure-sdk-for-c/blob/0ade924e2832ad187fc62d6a37691651a0334f0e/sdk/src/azure/core/az_json_writer.c#L861

vhvb1989 (approved):

sdk/inc/azure/core/az_json.h
danewalton: I've always used `@note` so usually stick with it. Updated the wordage let me know how that looks to you. I put it in other locations where it might be relevant as well.

danewalton:
@ahsonkhan let me know if you would like this modified further but going to merge for now.

sdk/inc/azure/core/az_json.h
```diff
@@ -390,6 +390,13 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
 /**
  * @brief Appends the UTF-8 text value (as a JSON string) into the buffer.
  *
+ * @note If you receive an #AZ_ERROR_NOT_ENOUGH_SPACE result while appending data for which there is
+ * theoretically space, note that the JSON writer requires at least 64-bytes of slack within the
```
ahsonkhan, line 394: > result while appending data for which there is theoretically space

This part is a bit awkward to read/parse.

sdk/inc/azure/core/az_json.h
```diff
@@ -390,6 +390,13 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
...
+ * output buffer, above the theoretical minimal space needed. The JSON writer pessimistically
+ * requires at least 64-bytes of space when writing any chunk of data larger than 10 characters
+ * because it tries to write in 64 byte chunks (10 character * 6 if all need to be escaped into the
+ * unicode form).
```
ahsonkhan, line 398: I think the 10*6 goes into too much detail about the specifics of the implementation, that might change and isn't relevant to the caller.

It also doesn't exactly apply to append APIs that aren't dealing with strings (like appending numbers).

sdk/inc/azure/core/az_json.h
```diff
@@ -390,6 +390,13 @@ az_json_writer_get_bytes_used_in_destination(az_json_writer const* json_writer)
...
+ *
```
ahsonkhan, line 399: We are missing this comment on append property name API.

ahsonkhan (commented):
Addressed this feedback on https://github.com/Azure/azure-sdk-for-c/pull/1983

Lmk your thoughts :)

sdk/inc/azure/core/az_json.h
ahsonkhan: Difference between `@note` and `@remark` in doxygen rendering. It has a yellow bar to highlight it some more, when it is a `@note`:
image

sdk/inc/azure/core/az_json.h
```diff
@@ -420,6 +427,13 @@ AZ_NODISCARD az_result az_json_writer_append_string(az_json_writer* ref_json_wri
  * @remarks The function validates that the provided JSON to be appended is valid and properly
  * escaped, and fails otherwise.
  *
+ * @note If you receive an #AZ_ERROR_NOT_ENOUGH_SPACE result while appending data for which there is
```
ahsonkhan, line 430: We aren't consistent on whether note should go right after `@brief` or after `@param`.

Which made me realize, we aren't consistent on order of `@remarks` and `@return` either, which might be a good thing to fix across the board.

PR #1975, opened 2021-11-03
Late PR feedback follow-up

antkmsft:


sdk/src/azure/core/az_http_policy.c
```diff
@@ -42,6 +43,7 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
   return _az_http_pipeline_nextpolicy(ref_policies, ref_request, ref_response);
 }
 
+// "-1" below is to account for the null terminator at the end of the string.
```
ahsonkhan, line 46: 👍

sdk/src/azure/core/az_http_policy.c
```diff
@@ -67,8 +69,7 @@ AZ_NODISCARD az_result az_http_pipeline_policy_telemetry(
     remainder = az_span_copy_u8(remainder, '/');
     remainder = az_span_copy(remainder, AZ_SPAN_FROM_STR(AZ_SDK_VERSION_STRING));
 
-    telemetry_id = az_span_slice(
-        telemetry_id, 0, (int32_t)(az_span_ptr(remainder) - az_span_ptr(telemetry_id)));
+    telemetry_id = az_span_slice(telemetry_id, 0, _az_span_diff(remainder, telemetry_id));
```
ahsonkhan, line 72: If I understand this correctly, this is just trimming away the extra slack space, correct, so the span length is just as large as the content we put in the buffer?

ahsonkhan (approved):
This change itself looks great. I will leave it up to you to merge now or wait until the conversations from https://github.com/Azure/azure-sdk-for-c/pull/1968 are closed.

I am still thinking about the preconditions aspect of things. Maybe we can discuss offline to close on them :)

sdk/src/azure/core/az_http_policy.c
antkmsft: yes, exactly

sdk/inc/azure/core/az_http.h
```diff
@@ -234,7 +234,8 @@ AZ_NODISCARD az_result az_http_response_get_status_line(
  *
  * @return An HTTP status code.
  */
-AZ_INLINE az_http_status_code az_http_response_get_status_code(az_http_response* ref_response)
+AZ_INLINE AZ_NODISCARD az_http_status_code
```
ahsonkhan, line 237: 🎉 :)

PR #1976, opened 2021-11-03
Telemetry options constructor and component name size constraint

antkmsft:


sdk/inc/azure/core/internal/az_http_internal.h
```diff
@@ -7,6 +7,7 @@
 #include <azure/core/az_context.h>
 #include <azure/core/az_http.h>
 #include <azure/core/az_http_transport.h>
+#include <azure/core/az_precondition.h>
```
ahsonkhan, line 10: Looks like build error:
```text
In file included from /mnt/vss/_work/1/s/sdk/src/azure/core/az_http_pipeline.c:5:0:
/mnt/vss/_work/1/s/sdk/inc/azure/core/internal/az_http_internal.h: In function ‘_az_http_policy_telemetry_options_create’:
/mnt/vss/_work/1/s/sdk/inc/azure/core/internal/az_http_internal.h:78:3: error: implicit declaration of function ‘_az_PRECONDITION_VALID_SPAN’; did you mean ‘_az_PRECONDITION_H’? [-Werror=implicit-function-declaration]
   _az_PRECONDITION_VALID_SPAN(component_name, 1, false);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
   _az_PRECONDITION_H

```

antkmsft: strange...

sdk/src/azure/core/az_http_policy.c
```diff
@@ -46,6 +46,11 @@ AZ_NODISCARD az_result az_http_pipeline_policy_apiversion(
 // "-1" below is to account for the null terminator at the end of the string.
 #define _az_TELEMETRY_ID_PREFIX "azsdk-c-"
 #define _az_TELEMETRY_ID_PREFIX_LENGTH (sizeof(_az_TELEMETRY_ID_PREFIX) - 1)
+#define _az_TELEMETRY_COMPONENT_NAME_MAX_LENGTH 40
+#define _az_TELEMETRY_VERSION_MAX_LENGTH (sizeof("12.345.6789-preview.123") - 1)
+#define _az_TELEMETRY_ID_MAX_LENGTH                                                       \
```
ahsonkhan, line 51: Roughly what's the value of this? Might be worth putting in a comment. Is it 72?
8 + 40 + 1 + 23 = 72?

Presumably if/when we add things like application id, this will grow by some 24-28 chars, right? Should we add that now, or not worth the effort?

ahsonkhan (approved):
Looks good to me. Thanks!

sdk/src/azure/core/az_http_policy.c
antkmsft: I'd avoid to put 72, if the constant may change, so comment will become misleading.

This thing came in unexpected, I don't think there's necessarily enough time to add good `application_id` support. If we're lucky - yes, but should we run into a decision fork, we may not have enough time to think it well through, and I don't want the deadlines to have pressure on the decision making.

I'd propose is to add a public telemetry options struct, with the nullable `application_id` field. That struct will go to all HTTP client options. And there will be telemetry options struct with underscore, that will contain that public struct, plus the `component_id`. We will add another constructor for it that takes two parameters - `component_id` and that new public struct.

But I think it is ok to do it later.
I also can see if we do it now, it might be worth to give it 2-3 days of PR reviews.
I.e. unless you think that it is much more beneficial to add it now, I'd leave it for the future.

ahsonkhan: Yeah, makes sense, we can address it later.

sdk/src/azure/core/az_http_policy.c
```diff
@@ -56,14 +61,19 @@ AZ_NODISCARD az_result az_http_pipeline_policy_telemetry(
...
-  uint8_t telemetry_id_buffer[200] = _az_TELEMETRY_ID_PREFIX;
+  uint8_t telemetry_id_buffer[_az_TELEMETRY_ID_MAX_LENGTH] = _az_TELEMETRY_ID_PREFIX;
   az_span telemetry_id = AZ_SPAN_FROM_BUFFER(telemetry_id_buffer);
   {
     az_span remainder = az_span_slice_to_end(telemetry_id, _az_TELEMETRY_ID_PREFIX_LENGTH);
 
     _az_http_policy_telemetry_options* options = (_az_http_policy_telemetry_options*)(ref_options);
     az_span const component_name = options->component_name;
-    _az_PRECONDITION_VALID_SPAN(component_name, 1, false);
+#ifndef AZ_NO_PRECONDITION_CHECKING
```
ahsonkhan, line 71: We typically don't do this in source files, unless there is a major cost. This macro is meant mainly for tests.

This `ifndef` shouldn't be required because it is already taken care of in:
https://github.com/Azure/azure-sdk-for-c/blob/9f5a6334863ebbf214c6e2fbd5d6e3fce52fe0de/sdk/inc/azure/core/internal/az_precondition_internal.h#L61-L63

antkmsft:
ApllicationId, part of the telemetry that customer can supply their own - https://github.com/Azure/azure-sdk-for-c/issues/1982

sdk/src/azure/core/az_http_policy.c
antkmsft: This is here so that there is no "`component_name_size` variable is not being used" warning.
And it is better to be a variable and not `_az_PRECONDITION_RANGE(az_span_size(component_name), ...)`, because otherwise when macro expands, `az_span_size(component_name)` will be called twice.

PR #1988, opened 2021-11-09
Update repo link (master => main)

### LINKED ISSUE #1978: Update links from `master` to `main`
There are still a couple instances in the SDK, if you search.

antkmsft:
Closes #1978


ahsonkhan (approved):
@danieljurek, @weshaggard do we have some strategy to keep links always pointing to main going forward? I imagine link verification doesn't flag these as invalid since GitHub re-routes it.

weshaggard:
>@danieljurek, @weshaggard do we have some strategy to keep links always pointing to main going forward? I imagine link verification doesn't flag these as invalid since GitHub re-routes it.

We don't have a strategy. The thought is once there is a one-time clean-up the most new links will just point to main.

Issue #1980, opened 2021-11-04
CI needs check hyperlinks for being valid, for not having `en-us`, etc.


antkmsft:
C++ repo has this, I think we should enable it as well.

On the other hand, most of the `en-us` and `/master/` links that I've seen are in the source code comments, so if that's not something that the tool does, and if it is already enabled, then I'm sorry for the false positive.

antkmsft:
@weshaggard: "You are correct that we don't validate random links in any text we only look at well formed links in MD or html. This has come up in .NET XML doc comments recently, [Azure/azure-sdk-for-net#22216](https://github.com/Azure/azure-sdk-for-net/issues/22216), although we don't have a solution for that currently either. It can be tricky to parse out links in random source code."

kurtzeborn:
It seems like what you're talking about here is links in comments in our source code.  We could try to do this, but the solution wouldn't hacky and specialized and probably produce a lot of false positives.  I'm going to go ahead and close this since I don't think the priority of doing this is higher than most things on our backlog.  I'm certainly open to alternate opinions though, so if you disagree with this conclusion, please reactivate.

PR #1989, opened 2021-11-09
Remove `en-us` from microsoft links, and fix a couple of dead links

### LINKED ISSUE #1979: Remove `en-us` from MSDN links
In C++ repo, CI even gives us an error if we don't.
Without `en-us`, the link works perfectly, but opens to everyone in their preferred language.

antkmsft:
Closes #1979

### DIFF
--- a/sdk/inc/azure/core/internal/az_precondition_internal.h
+++ b/sdk/inc/azure/core/internal/az_precondition_internal.h
@@ -51,7 +51,7 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
 // earlier without verifying, because if __analysis_assume() is not available on earlier compiler
 // version, there will be a compilation error.
 // For more info, see
-// https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/using-the--analysis-assume-function-to-suppress-false-defects
+// https://docs.microsoft.com/windows-hardware/drivers/devtest/using-the--analysis-assume-function-to-suppress-false-defects
 #if _MSC_VER >= 1920
 #define _az_ANALYSIS_ASSUME(statement) __analysis_assume(statement)
 #else
--- a/sdk/inc/azure/iot/az_iot_common.h
+++ b/sdk/inc/azure/iot/az_iot_common.h
@@ -68,7 +68,7 @@ enum
 /**
  * @brief Azure IoT service status codes.
  *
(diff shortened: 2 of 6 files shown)


sdk/inc/azure/core/internal/az_precondition_internal.h
```diff
@@ -51,7 +51,7 @@ az_precondition_failed_fn az_precondition_failed_get_callback();
 // earlier without verifying, because if __analysis_assume() is not available on earlier compiler
 // version, there will be a compilation error.
 // For more info, see
-// https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/using-the--analysis-assume-function-to-suppress-false-defects
```
ahsonkhan, line 54: I was going to ask why link verification didn't catch this, but guessing that since these are just links within doc comments, not an actual hyperlink within markdown or other such docs, it doesn't get validated.

@danieljurek, @weshaggard has this come up as a way to improve link verification in other languages? If we have typos in links within source code comments that could 404, should those be verified?

ahsonkhan (approved):
How did you initially notice these @antkmsft ?
And did you run some regex to find/fix them all?

sdk/inc/azure/core/internal/az_precondition_internal.h
weshaggard: You are correct that we don't validate random links in any text we only look at well formed links in MD or html. This has come up in .NET XML doc comments recently, https://github.com/Azure/azure-sdk-for-net/issues/22216, although we don't have a solution for that currently either. It can be tricky to parse out links in random source code.

ahsonkhan: That's what I figured. Thanks!

antkmsft: #1980 Is about this, but based on @weshaggard's response, I think we can close it?

antkmsft:
@ahsonkhan, no regexes, manual review. There aren't this many. Dead links I've found as I was still verifying each link I was removing `en-us` from. Initially I noticed this as I myself put a link with `en-us` during the Storage SDK development, and decided to see if there are other links like this, and found more.

ahsonkhan:
> no regexes, manual review

Did you run a Ctrl+F (or find/search in an IDE like VS Code) to capture all instances of links with `en-us` in the repo?

antkmsft:
@ahsonkhan, Ctrl+Shift+F for "en-us" in VS.

PR #1991, opened 2021-11-09
Update change log for storage feature branch for November release

ahsonkhan:


ahsonkhan:
As far as I understand git merges and auto-resolving conflicts when the commit histories diverge a bit, when the file contents don't, the approach in https://github.com/Azure/azure-sdk-for-c/pull/1990, which updates the changelog in main first, is the better approach. That allows us to maintain simpler/linear commit history, especially as main's changelog continues to evolve independent of the feature branch's version

antkmsft:
Let's use the other approach where we update core first, and then pull changes to feature branch.

PR #1997, opened 2021-11-09
Make live tests to build Storage samples with Curl transport

antkmsft:


eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -22,15 +22,15 @@ jobs:
...
           CMAKE_GENERATOR: 'Visual Studio 16 2019'
           CMAKE_GENERATOR_PLATFORM: x64
-          build.args: ' -DTRANSPORT_PAHO=ON'
+          build.args: ' -DTRANSPORT_PAHO=ON -DTRANSPORT_CURL=ON'
           test_type: 'iot'
           os: 'win'
         MacOS_x64_with_iot_samples:
           Pool: Azure Pipelines
           OSVmImage: 'macOS-10.15'
           vcpkg.deps: 'cmocka paho-mqtt'
```
ahsonkhan, line 31: I think we need curl `vcpkg.deps` dependency.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -22,15 +22,15 @@ jobs:
...
           VCPKG_DEFAULT_TRIPLET: 'x64-osx'
-          build.args: ' -DTRANSPORT_PAHO=ON'
+          build.args: ' -DTRANSPORT_PAHO=ON -DTRANSPORT_CURL=ON'
           test_type: 'iot'
           os: 'macos'
         # Disable live testing for these matrix entries until we have something to
```
ahsonkhan, line 36: We may want to try to use other legs for the "HTTP-based" samples, rather than change the IoT ones.

ahsonkhan: @danieljurek what should our strategy be here?

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -33,36 +33,34 @@ jobs:
...
+          VCPKG_DEFAULT_TRIPLET: 'x64-windows-static'
+          CMAKE_GENERATOR: 'Visual Studio 16 2019'
+          CMAKE_GENERATOR_PLATFORM: x64
+          build.args: ' -DTRANSPORT_CURL=ON -DAZ_PLATFORM_IMPL=WIN32'
+        MacOS_x64_with_samples:
+          Pool: Azure Pipelines
+          OSVmImage: 'macOS-10.15'
+          vcpkg.deps: 'curl[ssl]'
+          VCPKG_DEFAULT_TRIPLET: 'x64-osx'
+          build.args: ' -DTRANSPORT_CURL=ON -DAZ_PLATFORM_IMPL=POSIX'
```
ahsonkhan, line 63: We need `os: 'macos'`


https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1187179&view=logs&j=d1a859cc-2145-5e72-055c-d2e5fdad58d8&t=08df4c10-81fa-580a-f797-71fb278bea08

> Expanded: and(contains(' -DTRANSPORT_CURL=ON -DAZ_PLATFORM_IMPL=POSIX', '-DTRANSPORT_CURL=ON'), or(eq(Null, 'linux'), eq(Null, 'macos')))
Result: False

ahsonkhan (approved):

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -33,36 +33,38 @@ jobs:
...
-        #   vcpkg.deps: 'curl[ssl]'
-        #   VCPKG_DEFAULT_TRIPLET: 'x64-osx'
-        #   build.args: ' -DTRANSPORT_CURL=ON  -DAZ_PLATFORM_IMPL=POSIX'
+        Linux_x64_with_samples:
+          Pool: azsdk-pool-mms-ubuntu-1804-general
+          OSVmImage: MMSUbuntu18.04
+          vcpkg.deps: 'curl[ssl]'
+          VCPKG_DEFAULT_TRIPLET: 'x64-linux'
+          build.args: ' -DTRANSPORT_CURL=ON -DAZ_PLATFORM_IMPL=POSIX'
+          os: 'linux'
```
weshaggard, line 42: Do we want a os name specifically or could the image name work?

antkmsft: I think it is the `os` name: `[14:53] Ahson Khan
Expanded: and(contains(' -DTRANSPORT_CURL=ON -DAZ_PLATFORM_IMPL=POSIX', '-DTRANSPORT_CURL=ON'), or(eq(Null, 'linux'), eq(Null, 'macos'))) Result: False`

antkmsft: Also, this mirrors how the existing legs are currently configured, so I think it should work.

ahsonkhan: We have `os name` on the other live test legs already.

And it is being used in this condition:
https://github.com/Azure/azure-sdk-for-c/blob/cf8f2b1fd768aa942836e98709d34ba0ced60d5e/eng/pipelines/templates/jobs/archetype-sdk-tests.yml#L94

If it makes sense to you @weshaggard  to improve that and use an existing field, that seems reasonable, and we could do that in a separate PR. I would leave it up to engsys folks like Daniel to drive that :)

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -24,6 +24,7 @@ stages:
               - Name: Storage Blobs
                 NonWindowsPath: ./build/sdk/samples/storage/blobs/blobs_client_sample
                 WindowsPath: build\sdk\samples\storage\blobs\debug\blobs_client_sample.exe
+                RunCondition: contains(variables['build.args'], '-DTRANSPORT_CURL=ON')
```
ahsonkhan, line 27: This is a bit of a stop gap for now. Maybe @danieljurek can come up with a more appropriate/clean solution to run "independent" live tests/samples on different AzDo legs.

ahsonkhan:
@weshaggard why does link verification pass in PRs like this:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1187241&view=logs&j=a129effc-2dd1-54d1-fb5a-ad7bdc0e851d&t=7b4713e3-e110-5049-e231-cc8a1e664c68

But when run to trigger the release, they fail:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1187023&view=logs&j=a129effc-2dd1-54d1-fb5a-ad7bdc0e851d&t=617ceecd-bf48-5ce3-83e2-7468b1632a13

> 'file:///D:/a/_work/1/s/sdk/samples/storage/blobs/README.md' has 2 broken link(s):
https://github.com/Azure/azure-sdk-for-c/tree/main/sdk/docs/storage
https://github.com/Azure/azure-sdk-for-c/blob/main/sdk/samples/storage/blobs/src/blobs_client_sample.c

We are working to fix the feature-specific links to point to the feature branch, rather than main, since those files don't exist in main. but wondering how the verification/substitution works which results in this discrepancy.

Example, the following should be pointing to the feature branch:
https://github.com/Azure/azure-sdk-for-c/blob/feature/storage-blobs/sdk/samples/storage/blobs/README.md#next-steps
image

https://github.com/Azure/azure-sdk-for-c/blob/feature/storage-blobs/sdk/docs/storage/README.md

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
weshaggard: I generally like to avoid duplicating information so if OSVMImage would work here then that is more common and would be preferred.

PR #2010, opened 2021-11-12
Fix order of arguments for _az_http_policy_apiversion_options

ewertons:
The current order is different than what the structure defines, and that may cause compiler issues.
E.g.: without this fix the build fails for ESP32 xtensa compiler.

### DIFF
--- a/sdk/inc/azure/core/internal/az_http_internal.h
+++ b/sdk/inc/azure/core/internal/az_http_internal.h
@@ -83,9 +83,9 @@ AZ_NODISCARD AZ_INLINE _az_http_policy_apiversion_options
 _az_http_policy_apiversion_options_default()
 {
   return (_az_http_policy_apiversion_options){
-    ._internal = { .option_location = _az_http_policy_apiversion_option_location_header,
-                   .name = AZ_SPAN_EMPTY,
-                   .version = AZ_SPAN_EMPTY }
+    ._internal = { .name = AZ_SPAN_EMPTY,
+                   .version = AZ_SPAN_EMPTY,
+                   .option_location = _az_http_policy_apiversion_option_location_header }
   };
 }
 


sdk/inc/azure/core/internal/az_http_internal.h
```diff
@@ -83,9 +83,9 @@ AZ_NODISCARD AZ_INLINE _az_http_policy_apiversion_options
 _az_http_policy_apiversion_options_default()
 {
   return (_az_http_policy_apiversion_options){
-    ._internal = { .option_location = _az_http_policy_apiversion_option_location_header,
-                   .name = AZ_SPAN_EMPTY,
-                   .version = AZ_SPAN_EMPTY }
+    ._internal = { .name = AZ_SPAN_EMPTY,
```
ahsonkhan, line 86: For reference, here's the order in the struct:
https://github.com/Azure/azure-sdk-for-c/blob/5ad7985d50af8873c6148f867f57c487e294fb28/sdk/inc/azure/core/internal/az_http_internal.h#L42-L56

ahsonkhan:
> The current order is different than what the structure defines, and that may cause compiler issues.

I wonder if we have similar issues elsewhere in the SDK.

@ewertons how did you find out about this one?

@barcharcraz are there any tools or linters that we can setup which can help us detect this?

ahsonkhan (approved):

ewertons:
> > The current order is different than what the structure defines, and that may cause compiler issues.
> 
> I wonder if we have similar issues elsewhere in the SDK.
> 
> @ewertons how did you find out about this one?
> 
> @barcharcraz are there any tools or linters that we can setup which can help us detect this?

I am working on a project for the days of learning, using the StorageBlob client from [archive/StorageBlobs](https://github.com/azure/azure-sdk-for-c/tree/archive/StorageBlobs).

All I did was trying to compile on Arduino for Espressif ESP32. Got an error. I can get the error message if that helps.

ahsonkhan:
Thanks @ewertons. Makes sense.

> I am working on a project for the days of learning, using the StorageBlob client from [archive/StorageBlobs](https://github.com/azure/azure-sdk-for-c/tree/archive/StorageBlobs).

Consider using this:
https://github.com/Azure/azure-sdk-for-c/blob/5a0436b9bfdf46fc4549fad580781beb7e57746f/sdk/inc/azure/storage/az_storage_blobs.h#L91

PR #2052, opened 2021-12-15
Improve Storage code coverage

antkmsft:
Closes #2052

### DIFF
--- a/sdk/src/azure/storage/az_storage_blobs_blob_client.c
+++ b/sdk/src/azure/storage/az_storage_blobs_blob_client.c
@@ -54,63 +54,60 @@ AZ_INLINE az_span _az_get_host_from_url(az_span url)
   int32_t const url_length = az_span_size(url);
   uint8_t const* url_ptr = az_span_ptr(url);
 
-  if (url_length >= (int32_t)(sizeof("s://h") - 1))
+  int32_t const colon_max = url_length - (int32_t)(sizeof("//h") - 1);
+  for (int32_t colon_pos = 0; colon_pos < colon_max; ++colon_pos)
   {
-    int32_t const colon_max = url_length - (int32_t)(sizeof("//h") - 1);
-    for (int32_t colon_pos = 0; colon_pos < colon_max; ++colon_pos)
+    if (url_ptr[colon_pos] == ':')
     {
-      if (url_ptr[colon_pos] == ':')
+      if (url_ptr[colon_pos + 1] == '/' && url_ptr[colon_pos + 2] == '/')
       {
-        if (url_ptr[colon_pos + 1] == '/' && url_ptr[colon_pos + 2] == '/')
+        int32_t const authority_pos = colon_pos + 3;
+        int32_t authority_end_pos = authority_pos;
+        for (; authority_end_pos < url_length; ++authority_end_pos)
         {
-          int32_t const authority_pos = colon_pos + 3;
-          int32_t authority_end_pos = authority_pos;
-          for (; authority_end_pos < url_length; ++authority_end_pos)
+          if (url_ptr[authority_end_pos] == '/')
           {
-            if (url_ptr[authority_end_pos] == '/')
+            break;
+          }
+        }
+
+        if (authority_end_pos > authority_pos)
+        {
+          int32_t host_start_pos = authority_pos;
+          for (; host_start_pos < authority_end_pos; ++host_start_pos)
+          {
+            if (url_ptr[host_start_pos] == '@')
             {
               break;
             }
           }
 
-          if (authority_end_pos > authority_pos)
+          if (host_start_pos == authority_end_pos)
           {
-            int32_t host_start_pos = authority_pos;
-            for (; host_start_pos < authority_end_pos; ++host_start_pos)
+            host_start_pos = authority_pos;
+          }
+          else
+          {
+            ++host_start_pos;
+          }
+
+          if (host_start_pos < authority_end_pos)
+          {
+            int32_t host_end_pos = host_start_pos;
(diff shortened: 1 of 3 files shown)


sdk/src/azure/storage/az_storage_blobs_blob_client.c
```diff
@@ -54,63 +54,60 @@ AZ_INLINE az_span _az_get_host_from_url(az_span url)
   int32_t const url_length = az_span_size(url);
   uint8_t const* url_ptr = az_span_ptr(url);
 
-  if (url_length >= (int32_t)(sizeof("s://h") - 1))
```
antkmsft, line 57: GitHub is not displaying this change visually clear. What this all is about, is to remove the check that is verified by a precondition below.

image

RickWinter: Can we separate out this change into another PR.  That way this PR is purely about improving code coverage via additional tests.

antkmsft: @RickWinter, done, here it is - https://github.com/Azure/azure-sdk-for-c/pull/2062

vhvb1989 (approved):

sdk/tests/storage/blobs/test_az_storage_blobs.c
```diff
@@ -32,6 +35,23 @@ void test_storage_blobs_init(void** state)
       NULL)));
 }
 
+void test_storage_blobs_init_nonnull_options(void** state);
+void test_storage_blobs_init_nonnull_options(void** state)
```
ahsonkhan, line 39: We do we need to define the function like this first before implementing it?

Doesn't look like we do this elsewhere, unless I am mistaken. And the tests are just static.

sdk/tests/storage/blobs/test_az_storage_blobs.c
```diff
@@ -312,3 +332,258 @@ void test_storage_blobs_download(void** state)
...
+      az_storage_blobs_blob_client_init(&client, AZ_SPAN_EMPTY, AZ_CREDENTIAL_ANONYMOUS, NULL));
+}
+
+void verify_storage_blobs_upload_null_client(void** state);
+void verify_storage_blobs_upload_null_client(void** state)
+{
+  (void)state;
+  SETUP_PRECONDITION_CHECK_TESTS();
+
+  _az_http_client_set_callback(no_op_transport);
```
ahsonkhan, line 533: Unrelated question: Is this how we expect customers to set the transport?

sdk/tests/storage/blobs/test_az_storage_blobs.c
```diff
@@ -312,3 +332,258 @@ void test_storage_blobs_download(void** state)
...
+
+  ASSERT_PRECONDITION_CHECKED(az_storage_blobs_blob_download(NULL, NULL, NULL, NULL));
+
+  _az_http_client_set_callback(NULL);
+}
```
ahsonkhan, line 570: Nice!

sdk/tests/storage/blobs/test_az_storage_blobs.c
antkmsft: We do. Could be in other places, but we do. This is to avoid warning on one of the compilers that we define a function without prior declaration.

sdk/tests/storage/blobs/test_az_storage_blobs.c
antkmsft: To pick curl, you specify `TRANSPORT_CURL=ON` when building the SDK. There isn't any other transport at the moment, but in general the transport should implement the  `az_http_client_send_request()` function. If there was WinHTTP implementation, for example, I'd think there would be some switch like `TRANSPORT_WINHTTP=ON`, and then in CMakeLists.txt for core we would replace `az_curl.c` with `az_winhttp.c` (can't easily have both, although it is possible).

In this specific case, when storage tests are built, they do implement the `az_http_client_send_request()`, and `_az_http_client_set_callback()` is really a part of storage tests, not the core API: see https://github.com/Azure/azure-sdk-for-c/blob/feature/storage-blobs/sdk/tests/storage/blobs/_az_test_http_client.h and https://github.com/Azure/azure-sdk-for-c/blob/feature/storage-blobs/sdk/tests/storage/blobs/_az_test_http_client.c . It allows you to provide callback, but our core does not directly offer you that feature, although as you see, if test code can do that, core could add that in the future if needed, or even customers can add that feature themselves if they need to switch between transport (although they'll have to do some linker magic or code patching to rename az_curl's `az_http_client_send_request()`.

sdk/tests/storage/blobs/test_az_storage_blobs.c
```diff
@@ -312,3 +332,258 @@ void test_storage_blobs_download(void** state)
...
+      &client, AZ_SPAN_FROM_STR("x:///"), AZ_CREDENTIAL_ANONYMOUS, NULL)));
+
+  _az_http_client_set_callback(no_op_transport);
+
+  ASSERT_PRECONDITION_CHECKED(az_storage_blobs_blob_download(&client, NULL, NULL, NULL));
+
+  _az_http_client_set_callback(NULL);
+}
+
+#endif
```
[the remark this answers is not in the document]
antkmsft: Let's make clangformat to do that, like it does for C++. With manual comment, it may go out of sync (for instance, if you  change the ifdef).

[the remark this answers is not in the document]
antkmsft: It looks like there isn't an option in clang-format to do that. I had an impression it does that, but that's for C++ namespaces, not the preprocessor directives.

PR #2062, opened 2022-01-11
Storage: remove the condition that is already verified by preconditions

antkmsft:
GitHub is not displaying this change visually clear. What this all is about, is to remove the check that is verified by a precondition below.

`if()` line removal (plus the curly braces) is the only thing that gets modified here (the rest of the change is indentation).

image

### DIFF
--- a/sdk/src/azure/storage/az_storage_blobs_blob_client.c
+++ b/sdk/src/azure/storage/az_storage_blobs_blob_client.c
@@ -54,63 +54,60 @@ AZ_INLINE az_span _az_get_host_from_url(az_span url)
   int32_t const url_length = az_span_size(url);
   uint8_t const* url_ptr = az_span_ptr(url);
 
-  if (url_length >= (int32_t)(sizeof("s://h") - 1))
+  int32_t const colon_max = url_length - (int32_t)(sizeof("//h") - 1);
+  for (int32_t colon_pos = 0; colon_pos < colon_max; ++colon_pos)
   {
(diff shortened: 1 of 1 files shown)


ahsonkhan:
If you hide the whitespace difference, you see exactly what changed:
https://github.com/Azure/azure-sdk-for-c/pull/2062/files?diff=split&w=1

image

sdk/src/azure/storage/az_storage_blobs_blob_client.c
```diff
@@ -54,63 +54,60 @@ AZ_INLINE az_span _az_get_host_from_url(az_span url)
   int32_t const url_length = az_span_size(url);
   uint8_t const* url_ptr = az_span_ptr(url);
 
-  if (url_length >= (int32_t)(sizeof("s://h") - 1))
```
ahsonkhan, line 57: >  What this all is about, is to remove the check that is verified by a precondition below.

Can you link to where is the precondition that is verifying this? I can't find what you are referring to

ahsonkhan: Oh right here, got it :)

https://github.com/Azure/azure-sdk-for-c/blob/736b8005b41b4ed34976b37bc79b1f3270a2278e/sdk/src/azure/storage/az_storage_blobs_blob_client.c#L124

ahsonkhan (approved):
Looks good. One extra benefit of trying to get code coverage up is finding these unnecessary branches :)

PR #2063, opened 2022-01-12
Generate manifest alongside build artifacts

scbedd:
Related to Azure/azure-sdk-tools#2273

@danieljurek I have a bad bad feeling about what is going to happen if we attempt to run the manifest generation tool on some of the more esoteric platforms that C runs against. We probably will need to add a platform specific condition.

EDIT: and an additional concern. Will the `build/` folder have been supplanted by the docs pieces? I should probably place this publish _before_ the docs builds yes?


weshaggard (approved):

weshaggard (approved):

danieljurek (approved):
👍🚀‼️

PR #2092, opened 2022-02-03
Add additional guidance on iot char encodings

danewalton:


sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -75,7 +75,9 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * @param[in] global_device_hostname The device provisioning services global host name.
  * @param[in] id_scope The ID Scope.
  * @param[in] registration_id The Registration ID. This must match the client certificate name (CN
- * part of the certificate subject).
+ * part of the certificate subject). Must be alphanumeric characters plus the special characters:
```
CIPop, line 78: Can we point to the Docs page instead (or in addition)? 

https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service#registration-id

I think I've been through at least one change in requirements and if this changes again, it's unlikely we'll know to change our comment.

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -43,6 +43,12 @@ typedef struct
 {
   /**
    * The module name (if a module identity is used).
+   * If the module name contains any of the following characters,
```
CIPop, line 46: I think this isn't describing the whole problem. For example, devices could contain Unicode higher-plane characters or be characters interpreted in non-Latin alphabets.

All parts of the Topic (deviceid, moduleid, user-agent, properties, etc) _must_ follow the guidelines regarding MQTT 3.1.1 (and future versions) topic names described in the Oasis MQTT spec. E.g. [MQTT 3.1.1 Section 4.7 Topic Names](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106).

Device and module IDs that contain characters unallowed by the MQTT specification must use URI escaping as described by [RFC 3986, section 2.1](https://datatracker.ietf.org/doc/html/rfc3986/#section-2.1)

CIPop (approved):
Thanks @danewalton !

sdk/inc/azure/iot/az_iot_provisioning_client.h
danewalton: In fairness though, our implementation would have to change if that were the case....documentation won't change that.

ericwol-msft:
I think there was an ask to add a precondition that checks for unescaped chars on these parameters

drwill-ms (changes requested):
As mentioned in the bug, my worry is that sometimes the code URL encodes these properties ([az_iot_hub_client_sas_get_signature](https://github.com/Azure/azure-sdk-for-c/blob/main/sdk/src/azure/iot/az_iot_hub_client_sas.c) and az_iot_hub_client_sas_get_password) and sometimes they don't.

I don't think documentation is enough to resolve this. I figure the client is going to need to always url encode or always expect the user to do it. Unfortunately, that is probably a breaking change.

There is also the question of the model Id.

ericwol-msft:
> As mentioned in the bug, my worry is that sometimes the code URL encodes these properties ([az_iot_hub_client_sas_get_signature](https://github.com/Azure/azure-sdk-for-c/blob/main/sdk/src/azure/iot/az_iot_hub_client_sas.c) and az_iot_hub_client_sas_get_password) and sometimes they don't.
> 
> I don't think documentation is enough to resolve this. I figure the client is going to need to always url encode or always expect the user to do it. Unfortunately, that is probably a breaking change.
> 
> There is also the question of the model Id.

I assume the assumption is that model id will always need to be encoded because of the [required colons](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fdigital-twin-model-identifier&data=04%7C01%7CEric.Wolz%40microsoft.com%7C9fa7d6402a9a4db52ba708d9e85b3310%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637796300672289095%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=myVh8Wv3s%2F8iWMh7FQ7AaMdImCAXb6G1hGwQ4GUQgQw%3D&reserved=0) in the ID. Where as device ID may not (most likely) will not contain encoded chars. On embedded, you probably should avoid device ID encoding, as it requires additional memory for encoding and decoding.

drwill-ms:
> I assume the assumption is that model id will always need to be encoded because of the [required colons](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fdigital-twin-model-identifier&data=04%7C01%7CEric.Wolz%40microsoft.com%7C9fa7d6402a9a4db52ba708d9e85b3310%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637796300672289095%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=myVh8Wv3s%2F8iWMh7FQ7AaMdImCAXb6G1hGwQ4GUQgQw%3D&reserved=0) in the ID. Where as device ID may not (most likely) will not contain encoded chars. On embedded, you probably should avoid device ID encoding, as it requires additional memory for encoding and decoding.

I'm not entirely sure if any of the values should always be encoded. Device Id and module Id are also used in username and client id when connecting to MQTT. Are they supposed to be URL encoded in those cases?

For model Id, what about when used in DPS when it is in the payload? Probably not then, right?

ericwol-msft:
> I'm not entirely sure if any of the values should always be encoded. Device Id and module Id are also used in username and client id when connecting to MQTT. Are they supposed to be URL encoded in those cases?

I don't see anything documented that this must be encoded for MQTT. We should get clarity from the hub team.


> For model Id, what about when used in DPS when it is in the payload? Probably not then, right?

correct, JSON

ericwol-msft (approved):

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -43,12 +43,10 @@ typedef struct
...
-   * If the module name contains any of the following characters,
-   * they must be percent-encoded as follows:
-   *         - `/` : `%2F`
-   *         - `%` : `%25`
-   *         - `#` : `%23`
-   *         - `&` : `%26`
+   * Must conform to the requirements of the MQTT spec for topic
+   * names (listed below) and of the IoT Hub (listed below)
+   * http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106
+   * https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry#device-identity-properties
```
drwill-ms, line 49: We should avoid including the locale and language in the URLs, so people get them in the regional settings on their browser/OS.

```suggestion
   * https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#device-identity-properties
```

sdk/inc/azure/iot/az_iot_hub_client.h
```diff
@@ -100,12 +98,10 @@ AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default();
...
- * @param[in] device_id The Device ID. If the ID contains any of the following characters, they must
- * be percent-encoded as follows:
- *         - `/` : `%2F`
- *         - `%` : `%25`
- *         - `#` : `%23`
- *         - `&` : `%26`
+ * @param[in] device_id The Device ID. Must conform to the requirements of the MQTT spec for
+ * topic names (listed below) and of the IoT Hub (listed below)
+ * http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106
+ * https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry#device-identity-properties
```
drwill-ms, line 104:
```suggestion
 * https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#device-identity-properties
```

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -75,9 +75,8 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * @param[in] global_device_hostname The device provisioning services global host name.
  * @param[in] id_scope The ID Scope.
  * @param[in] registration_id The Registration ID. This must match the client certificate name (CN
- * part of the certificate subject). Must be alphanumeric characters plus the special characters:
- * '-',
- * '.', '_', ':'. The last character must be alphanumeric or dash ('-').
+ * part of the certificate subject). Must conform to the limitation listed in the link below:
+ * https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service#registration-id
```
drwill-ms, line 79:
```suggestion
 * https://docs.microsoft.com/azure/iot-dps/concepts-service#registration-id
```

drwill-ms (approved):

sdk/inc/azure/iot/az_iot_hub_client.h
danewalton: ahhh yup this is usually the thing I point out in other people's pr's damn

> Why do you see the speck that is in your brother's eye, but do not notice the log that is in your own eye?

thanks!

drwill-ms: He who is without code errors, cast the first PR comment.

PR #2162, opened 2022-03-28
Remove failure if version not present in the reported properties response

momuno:
Remove the failure if version is not present in the reported properties response. 
It will be present from IoT Hub, but absent from IoT Edge.

### DIFF
--- a/sdk/src/azure/iot/az_iot_hub_client_twin.c
+++ b/sdk/src/azure/iot/az_iot_hub_client_twin.c
@@ -173,8 +173,9 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
       {
         // Is a reported prop response
         out_response->response_type = AZ_IOT_HUB_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES;
-        _az_RETURN_IF_FAILED(az_iot_message_properties_find(
-            &props, az_iot_hub_twin_version_prop, &out_response->version));
+        // Version only present if response is from IoT Hub. Version is absent via IoT Edge.
+        az_iot_message_properties_find(
+            &props, az_iot_hub_twin_version_prop, &out_response->version);
       }
       else // 200 or 202
       {
--- a/sdk/src/azure/iot/az_iot_hub_client_twin.c
+++ b/sdk/src/azure/iot/az_iot_hub_client_twin.c
@@ -173,16 +173,16 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
       {
         // Is a reported prop response
         out_response->response_type = AZ_IOT_HUB_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES;
-        
+
         result = az_iot_message_properties_find(
             &props, az_iot_hub_twin_version_prop, &out_response->version);
         if (result == AZ_ERROR_ITEM_NOT_FOUND)
         {
-            out_response->version = AZ_SPAN_EMPTY;
+          out_response->version = AZ_SPAN_EMPTY;
         }
         else
         {
-             _az_RETURN_IF_FAILED(result);
+          _az_RETURN_IF_FAILED(result);
         }
       }
       else // 200 or 202
--- a/sdk/src/azure/iot/az_iot_hub_client_twin.c
+++ b/sdk/src/azure/iot/az_iot_hub_client_twin.c
@@ -173,9 +173,17 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
       {
         // Is a reported prop response
         out_response->response_type = AZ_IOT_HUB_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES;
-        out_response->version = AZ_SPAN_EMPTY;
-        az_iot_message_properties_find(
+        
+        az_result result = az_iot_message_properties_find(
             &props, az_iot_hub_twin_version_prop, &out_response->version);
+        if (result == AZ_ERROR_ITEM_NOT_FOUND)
+        {
+            out_response->version = AZ_SPAN_EMPTY;
+        }
+        else
+        {
+             _az_RETURN_IF_FAILED(result);
+        }
       }
       else // 200 or 202
       {
--- a/sdk/src/azure/iot/az_iot_hub_client_twin.c
+++ b/sdk/src/azure/iot/az_iot_hub_client_twin.c
@@ -173,7 +173,7 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
       {
         // Is a reported prop response
         out_response->response_type = AZ_IOT_HUB_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES;
-        // Version only present if response is from IoT Hub. Version is absent via IoT Edge.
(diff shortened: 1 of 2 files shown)


sdk/src/azure/iot/az_iot_hub_client_twin.c
```diff
@@ -173,8 +173,9 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
       {
         // Is a reported prop response
         out_response->response_type = AZ_IOT_HUB_CLIENT_TWIN_RESPONSE_TYPE_REPORTED_PROPERTIES;
-        _az_RETURN_IF_FAILED(az_iot_message_properties_find(
-            &props, az_iot_hub_twin_version_prop, &out_response->version));
+        out_response->version = AZ_SPAN_EMPTY;
+        az_iot_message_properties_find(
+            &props, az_iot_hub_twin_version_prop, &out_response->version);
```
danewalton, line 178: We should still check if this fails, and if it does with a `AZ_ERROR_ITEM_NOT_FOUND`, explicitly set `&out_response->version = AZ_SPAN_EMPTY`. Wouldn't want garbage going in there by accident.

danewalton: Also...you're going to hate me.....a test with this scenario :)

momuno: Thanks for commenting!  I was curious to know what others were thinking on this issue.  We will need to do a test with IoT Edge, not IoT Hub.  Do you know how we go about creating devices that talk to IoT Edge and not Hub?

momuno: That makes sense to switch the order to look only for the specific AZ failure of version being absent.

danewalton: > Thanks for commenting! I was curious to know what others were thinking on this issue. We will need to do a test with IoT Edge, not IoT Hub. Do you know how we go about creating devices that talk to IoT Edge and not Hub?

A unit test should be enough for it.

momuno: Side note: The underlying az_iot_message_properties_find only returns AZ_OK or AZ_ERROR_ITEM_NOT_FOUND, so for now there isn't another failure option.  However, if the logic ever changed in the future to include a different possible error return, we would be missing that scenario if we don't add the check in at this level.

CIPop (changes requested):
LGTM. Please add a UT to pin down this new behavior.

CIPop:
While not an API breaking change, this is a behavior breaking change: apps relying on finding a version if the parser passed will now find an AZ_SPAN_EMPTY instead. We already know that devices connecting to Edge will fail so I believe the risk is very small. @ahsonkhan @RickWinter FYI.

momuno:
> While not an API breaking change, this is a behavior breaking change: apps relying on finding a version if the parser passed will now find an AZ_SPAN_EMPTY instead. We already know that devices connecting to Edge will fail so I believe the risk is very small. @ahsonkhan @RickWinter FYI.

I agree Cris and was thinking about this as well.  This all works if the Hub does not stop sending version in the topic.  But if they stop, it could break apps that are expecting version and are only checking for AZ_OK.

momuno:
@CIPop @danewalton UT added.

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -26,16 +26,18 @@ static const az_span test_device_request_id = AZ_SPAN_LITERAL_FROM_STR("id_one")
 static const az_span test_twin_received_topic_desired_success
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/?$version=id_one");
 static const az_span test_twin_received_topic_fail
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/rez/200");
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/rez/200"); // rez DNE
 static const az_span test_twin_received_topic_incomplete_fail
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/200");
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/200"); // Missing rid
```
danewalton, line 31: great comments thanks for these!

danewalton (approved):

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
momuno: aw thanks dane. Glad it helps.

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -26,16 +26,18 @@ static const az_span test_device_request_id = AZ_SPAN_LITERAL_FROM_STR("id_one")
 static const az_span test_twin_received_topic_desired_success
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/PATCH/properties/desired/?$version=id_one");
 static const az_span test_twin_received_topic_fail
-    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/rez/200");
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/rez/200"); // rez DNE
```
CIPop, line 29: What does DNE mean?

CIPop (approved):

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
momuno: Does Not Exist.

momuno: Expanded :)

sdk/src/azure/iot/az_iot_hub_client_twin.c
```diff
@@ -173,8 +173,17 @@ AZ_NODISCARD az_result az_iot_hub_client_twin_parse_received_topic(
...
+        result = az_iot_message_properties_find(
+            &props, az_iot_hub_twin_version_prop, &out_response->version);
+        if (result == AZ_ERROR_ITEM_NOT_FOUND)
+        {
+          out_response->version = AZ_SPAN_EMPTY;
+        }
+        else
+        {
+          _az_RETURN_IF_FAILED(result);
+        }
```
ahsonkhan, line 186: Are we sure this is what we want? We are changing `AZ_ERROR_ITEM_NOT_FOUND` return value to `AZ_OK` on line 195 now.

Do we need to update the docs here:
https://github.com/Azure/azure-sdk-for-c/blob/cc7beafdb56f2f495d488451ca2f28f855070a17/sdk/inc/azure/iot/az_iot_hub_client.h#L625-L632

momuno: We were notified via a colleague on Azure RTOS who had a customer issue.  Through debugging, it was found that IoT Edge did not supply the $version in the reported properties response while IoT Hub did.  This was confirmed by IoT Edge.

Can you please remind me where I create the changelog entry? Thanks!

ahsonkhan: > Can you please remind me where I create the changelog entry? Thanks!

Here:
https://github.com/Azure/azure-sdk-for-c/blob/main/CHANGELOG.md#bugs-fixed

PR #2164, opened 2022-03-29
Add 400 error test for twin topic reported-properties response

momuno:
Adds a missing unit test for the 400+ status response scenario.

### DIFF
--- a/sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
+++ b/sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
@@ -40,7 +40,7 @@ static const az_span test_twin_received_topic_reported_response_no_version_succe
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/204/?$rid=id_one");
 static const az_span test_twin_received_topic_bad_request_success
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/400/?$rid=id_one");
-    
+
 static const char test_correct_twin_get_request_topic[] = "$iothub/twin/GET/?$rid=id_one";
 static const char test_correct_twin_patch_pub_topic[]
     = "$iothub/twin/PATCH/properties/reported/?$rid=id_one";
@@ -329,7 +329,7 @@ static void test_az_iot_hub_client_twin_parse_received_topic_request_error_succe
       az_iot_hub_client_init(&client, test_device_hostname, test_device_id, NULL), AZ_OK);
   az_iot_hub_client_twin_response response;
 
- assert_int_equal(
+  assert_int_equal(
       az_iot_hub_client_twin_parse_received_topic(
           &client, test_twin_received_topic_bad_request_success, &response),
       AZ_OK);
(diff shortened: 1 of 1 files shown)


sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -320,6 +322,24 @@ test_az_iot_hub_client_twin_parse_received_topic_reported_props_no_version_found
...
+  az_iot_hub_client_twin_response response;
+
+  assert_int_equal(
+      az_iot_hub_client_twin_parse_received_topic(
+          &client, test_twin_received_topic_bad_request_success, &response),
+      AZ_OK);
+
+  assert_true(az_span_is_content_equal(response.request_id, test_device_request_id));
+  assert_true(az_span_is_content_equal(response.version, AZ_SPAN_EMPTY));
+  assert_true(response.status >= AZ_IOT_STATUS_BAD_REQUEST);
```
CIPop, line 339: Shouldn't this be exactly equal?

CIPop (approved):
Thanks for the extra test @momuno !

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
momuno: Well, it could for the sake of the string and the limited scope of the unit test, but the code itself covers more than just the 400 case.  I was debating keeping a 400 case and perhaps adding a second 400+ case?  Separate the => into = and>  Perhaps I should do just that.

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -478,6 +518,8 @@ int test_az_iot_hub_client_twin()
         test_az_iot_hub_client_twin_parse_received_topic_reported_props_version_found_succeed),
     cmocka_unit_test(
         test_az_iot_hub_client_twin_parse_received_topic_reported_props_no_version_found_succeed),
+    cmocka_unit_test(test_az_iot_hub_client_twin_parse_received_topic_bad_request_succeed),
```
ewertons, line 521: Update the names if taking up the suggested changes.

ewertons (approved):

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -38,6 +38,10 @@ static const az_span test_twin_received_topic_reported_response_version_success
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/204/?$rid=id_one&$version=16");
 static const az_span test_twin_received_topic_reported_response_no_version_success
     = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/204/?$rid=id_one");
+static const az_span test_twin_received_topic_bad_request_success
```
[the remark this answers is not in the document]
momuno: its a successful parsing of the topic.  but i see how that could be confusing.

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
```diff
@@ -38,6 +38,10 @@ static const az_span test_twin_received_topic_reported_response_version_success
...
+    = AZ_SPAN_LITERAL_FROM_STR("$iothub/twin/res/400/?$rid=id_one");
+static const az_span test_twin_received_topic_status_timeout_success
```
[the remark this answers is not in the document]
momuno: same as above.  success vs failure I view as referring to the function itself -- did it succeed in parting the topic or fail?  But i can see the grey area here, so I'll remove failure from it.  Thanks!

sdk/tests/iot/hub/test_az_iot_hub_client_twin.c
momuno: will do thanks.

momuno: After thinking about it some more, I think the "success/succeed" should stay for the tests that pass with AZ_OK.  This result means a valid topic was successfully parsed, which is true in the case of the 400+ result codes.  To minimize confusion, I will change the names of those tests to use the result code number, avoiding the head scratching "bad_request_succeed" for example.

PR #2184, opened 2022-04-12
Adding API deprecation markers.

CIPop:
Adding standard deprecation documentation and compilation markers for `az_iot_provisioning_client_get_request_payload`, deprecated in favor of `az_iot_provisioning_client_register_get_request_payload`.

Tested on Linux / GCC and Windows.

PR #2188, opened 2022-04-15
Changing az_iot_provisioning_client_register_get_request_payload options behavior

CIPop:
1. Changing `az_iot_provisioning_client_register_get_request_payload` to accept `options == NULL`. 
2. Addressing PR comments from #2184.
3. Renaming UT names.

### DIFF
--- a/sdk/inc/azure/iot/az_iot_provisioning_client.h
+++ b/sdk/inc/azure/iot/az_iot_provisioning_client.h
@@ -78,7 +78,9 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * part of the certificate subject). Must conform to the limitations listed in the link below:
  * https://docs.microsoft.com/azure/iot-dps/concepts-service#registration-id
  * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_options
- * structure. Can be `NULL` for default options.
+ * structure. If `NULL` is passed, the provisioning client will use the default options. If using
+ * custom options, please initialize first by calling az_iot_provisioning_client_options_default()
+ * and then populating relevant options with your own values.
  * @pre \p client must not be `NULL`.
  * @pre \p global_device_hostname must be a valid span of size greater than 0.
  * @pre \p id_scope must be a valid span of size greater than 0.
@@ -478,9 +480,10 @@ az_iot_provisioning_client_payload_options_default();
  * @param[in] custom_payload_property __[nullable]__ Custom JSON to be added to this payload.
  * Can be `NULL`.
  * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_payload_options
- * structure. If `NULL` is passed, the function will use the default options. If using custom
- * options, please initialize first by calling az_iot_provisioning_client_payload_options_default()
- * and then populating relevant options with your own values.
+ * structure. If `NULL` is passed, the provisioning client will use the default options. If using
+ * custom options, please initialize first by calling
+ * az_iot_provisioning_client_payload_options_default() and then populating relevant options with
+ * your own values.
  * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
  * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
  * @param[out] out_mqtt_payload_length Contains the length, in bytes, written to \p mqtt_payload on
--- a/sdk/inc/azure/iot/az_iot_provisioning_client.h
+++ b/sdk/inc/azure/iot/az_iot_provisioning_client.h
@@ -477,14 +477,15 @@ az_iot_provisioning_client_payload_options_default();
  * @param[in] client The #az_iot_provisioning_client to use for this call.
  * @param[in] custom_payload_property __[nullable]__ Custom JSON to be added to this payload.
  * Can be `NULL`.
- * @param[in] options __[nullable]__ Reserved field for future options to this function.  Must be
- * `NULL`.
+ * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_payload_options
+ * structure. If `NULL` is passed, the function will use the default options. If using custom
+ * options, please initialize first by calling az_iot_provisioning_client_payload_options_default()
+ * and then populating relevant options with your own values.
  * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
  * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
  * @param[out] out_mqtt_payload_length Contains the length, in bytes, written to \p mqtt_payload on
  * success.
  * @pre \p client must not be `NULL`.
- * @pre \p options must be `NULL`.
  * @pre \p mqtt_payload must not be `NULL`.
  * @pre \p mqtt_payload_size must be greater than 0.
  * @pre \p out_mqtt_payload_length must not be `NULL`.
--- a/sdk/src/azure/iot/az_iot_provisioning_client.c
+++ b/sdk/src/azure/iot/az_iot_provisioning_client.c
@@ -641,10 +641,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_request_payload(
     size_t mqtt_payload_size,
     size_t* out_mqtt_payload_length)
 {
-  (void)options;
-
   _az_PRECONDITION_NOT_NULL(client);
-  _az_PRECONDITION_NOT_NULL(options);
   _az_PRECONDITION_NOT_NULL(mqtt_payload);
   _az_PRECONDITION(mqtt_payload_size > 0);
   _az_PRECONDITION_NOT_NULL(out_mqtt_payload_length);
@@ -665,7 +662,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_request_payload(
     _az_RETURN_IF_FAILED(az_json_writer_append_json_text(&json_writer, custom_payload_property));
   }
 
-  if (az_span_size(options->certificate_signing_request) > 0)
+  if (options != NULL && az_span_size(options->certificate_signing_request) > 0)
   {
     _az_RETURN_IF_FAILED(az_json_writer_append_property_name(&json_writer, prov_csr_request_label));
     _az_RETURN_IF_FAILED(
--- a/sdk/inc/azure/core/_az_cfg.h
+++ b/sdk/inc/azure/core/_az_cfg.h
@@ -107,10 +107,10 @@
 #endif
 
 #ifdef _MSC_VER
-#define AZ_PUSH_IGNORE_DEPRECATIONS  _Pragma("warning(push)") _Pragma("warning(disable:4996)")
+#define AZ_PUSH_IGNORE_DEPRECATIONS _Pragma("warning(push)") _Pragma("warning(disable:4996)")
 #define AZ_POP_WARNINGS _Pragma("warning(pop)")
 #else
-#define AZ_PUSH_IGNORE_DEPRECATIONS  \
+#define AZ_PUSH_IGNORE_DEPRECATIONS \
   _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
 #define AZ_POP_WARNINGS _Pragma("GCC diagnostic pop")
 #endif
--- a/sdk/inc/azure/core/_az_cfg.h
+++ b/sdk/inc/azure/core/_az_cfg.h
@@ -107,10 +107,10 @@
 #endif
 
 #ifdef _MSC_VER
-#define AZ_IGNORE_DEPRECATIONS _Pragma("warning(push)") _Pragma("warning(disable:4996)")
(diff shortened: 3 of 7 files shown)


sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -477,14 +477,15 @@ az_iot_provisioning_client_payload_options_default();
  * @param[in] client The #az_iot_provisioning_client to use for this call.
  * @param[in] custom_payload_property __[nullable]__ Custom JSON to be added to this payload.
  * Can be `NULL`.
- * @param[in] options __[nullable]__ Reserved field for future options to this function.  Must be
- * `NULL`.
+ * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_payload_options
+ * structure. If `NULL` is passed, the function will use the default options. If using custom
```
momuno, line 481:
```suggestion
 * structure. If `NULL` is passed, the provisioning client will use the default options. If using custom
```

momuno: This parallels the description used for the az_iot_hub_client.

momuno: Also, the description at line 88 for the provisining_client init does not parallel the iot hub client init in its options description if NULL... is it relevant to update that as well?

momuno (approved):

danewalton (approved):

sdk/inc/azure/iot/az_iot_provisioning_client.h
CIPop: Re "function": Initially I was thinking that the function is using payload options (and not client options) but this is really using the same provisioning client object so your proposal makes more sense. Will change.

Thanks for noticing the DPS client: will change.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -665,7 +662,7 @@ AZ_NODISCARD az_result az_iot_provisioning_client_register_get_request_payload(
     _az_RETURN_IF_FAILED(az_json_writer_append_json_text(&json_writer, custom_payload_property));
   }
 
-  if (az_span_size(options->certificate_signing_request) > 0)
+  if (options != NULL && az_span_size(options->certificate_signing_request) > 0)
```
ahsonkhan, line 665: When options is `NULL`, can we assume the `options` are the same as if the user called `az_iot_provisioning_client_payload_options_default()` and passed those.

So, we'd assume `.certificate_signing_request = AZ_SPAN_EMPTY` in that case (but without relying on knowing the implementation detail of the default options, since that can change).

Something like:
```c
if (options == NULL)
{
  options= az_iot_provisioning_client_payload_options_default();
}
```

CIPop: Yes, the behavior is the same but the proposed implementation won't work because `options` is a pointer to a pre-allocated struct.
Since we don't have a memory location for the structure, we can't call that function without using more stack (which isn't ideal for embedded systems).

ahsonkhan: The issue with supporting `NULL` options with this approach, is if we were to add new options or modify `az_iot_provisioning_client_payload_options_default()`, we'd have to make sure to modify the implementation of `az_iot_provisioning_client_register_get_request_payload()`, to make sure the right thing is happening.

How much of a burden do you think it will be to remember having to update both places?

Maybe we should add a source doc comment for future maintainers to make sure such modifications to options and default options are propagated and the impl of functions using it are updated as well.

ahsonkhan: I believe this is the first instance of a case where we have options passed in to a client **method** (rather than ctor).

The pattern of passing null options work at init because we save the options as a member, on the client. We can't really do that for methods:
https://github.com/Azure/azure-sdk-for-c/blob/285b7ce15650ed6747369bf4aed361324c5253eb/sdk/src/azure/iot/az_iot_hub_client.c#L38-L53

I wonder if we should have ctors and methods be consistent, or make an exception for methods, and follow this pattern:
- at client ctor level, null options are allowed and mean default options
- at client method/operation level, null options are not allowed

@JeffreyRichter, what do you think?

This pattern comes up in blob storage upload and download methods as well (and future SDKs), but we don't have any options for it yet:
https://github.com/Azure/azure-sdk-for-c/blob/5c6fa6eef60f39aa2e251d79c26892c6f0131bec/sdk/src/azure/storage/az_storage_blobs_blob_client.c#L286-L295

I wish there was a way to make the semantics of all options accepting APIs (ctors and methods) be the same, in what `NULL` means?

cc @antkmsft

JeffreyRichter: I'm not sure I'm following the whole conversation. If, insite our method, e check if options is NULL and then create a default options, then we are still using stack space. In fact, we we force the customer to pass a non=NULL value, then the customer can allocate the options struct globally (so it's not on the stack) and re-use over the app's/thread's lifetime.

Conceptually, I don't see a problem with having methods take an options struct to fine-tune the method's behavior just like how a client ctor can take options to fine-tune a client's behavior.

antkmsft:
FWIW, in Go, there are no overloads, and no default values. So, `NULL` is allowed in many places, and it means "default value". And I think it is pretty intuitive, that `NULL` means "default".
So, from that perspective, I do not see disadvantages in "NULL = default" idiom. In that aspect, I would not distinguish between "constructors" and "methods". Let NULL be default everywhere.

But, I understand that in *Embedded* C, there could be other considerations, such as stack allocation. And from that perspective, I see how we may not want to do this if there is extra price associated with it.

IF we are in the place where we really consider saving on stack allocations, then yes - do not make NULL to mean "default", DO require the user to always pass the parameter, and do not allow NULLs (make it a precondition). Just make sure that for every such parameter, there is an (inline) function called `az_xxx_default()` that allows the user to construct the default value easily.

I would say that IF it would be possible to use "NULL = default" idiom in a way that the user does only pay the extra stack price when they use NULL, I would do it.

for example, it could look like this:
```
typedef struct {
    int y;
} options;

options az_options_default() { return { .y = 5}; }

int f(int x, options const* o)
{
    if (o == NULL)
    {
        options oo = az_options_default(); // <-- if only we could guarantee that this stack space is only allocated when o == NULL...
        return f(x, &oo);
    }

    return x + o->y;
}
```
That would be great if it could work in a way that the stack space is only allocated when the control enters the `if()` block. But most likely, the compiler will allocate the space for it in the beginning, before any comparison. If only we could use assembler to do it, then we could guarantee the behavior that we want. We could even avoid the recursive call. But maybe there are ways to do it, I just don't know them.

I would encourage us to consult with the language experts and try to find a way to use "you only pay for it if you use it" approach for "NULL = default" idiom, so it would work on all compilers.

Another option is to allow it on a case by case basis. For example, we could allow "NULL = default" in methods for Storage-Blobs, because it does use HTTP transport, which is huge relatively to most of the IoT applications. So, if you use Storage-Blobs, you are using the hardware of such caliber, that worrying about extra 48 bytes on stack is not something you would worry about.

Other option is to embed the default values in code. For the example below, it would look like this:
```
typedef struct {
    int y;
} options;

enum { _az_OPTIONS_DEFAULT_Y = 5 };

options az_options_default() { return { .y = _az_OPTIONS_DEFAULT_Y }; }

int f(int x, options const* o)
{
    return x + (o != NULL ? o->y : _az_OPTIONS_DEFAULT_Y );
}
```
This is not bad either. Maybe it is a bit harder to maintain, but not impossible. And after all, this is why we are making the SDK - to make customer's life easier, even if it means more work for us and more attention that needs to be paid by us when developing. Having a constant (`_az_OPTIONS_DEFAULT_Y`) makes it less prone to inconsistencies. In fact, I would rather insist on using named constants and not values, because it will be very hard to find all usages.

I would pay attention that if a function is using `o` for 20 times, the compiler does not generate 20 `if (o == NULL)` comparisons. I think it is likely that the compiler will optimize it.

A variation of this approach is to have getter functions for options:
```
AZ_INLINE int _az_options_get_y(options const* o) { return (o != NULL ? o->y : _az_OPTIONS_DEFAULT_Y); }

int f(int x, options const* o)
{
    return x + _az_options_get_y(o);
}
```
Especially if they are inline, there should be no extra cost (given that the compiler won't generate redundant NULL checks).

sdk/src/azure/iot/az_iot_provisioning_client.c
antkmsft: FWIW, in Go, there are no overloads, and no default values. So, `NULL` is allowed in many places, and it means "default value". And I think it is pretty intuitive, that `NULL` means "default".
So, from that perspective, I do not see disadvantages in "NULL = default" idiom. In that aspect, I would not distinguish between "constructors" and "methods". Let NULL be default everywhere.

But, I understand that in *Embedded* C, there could be other considerations, such as stack allocation. And from that perspective, I see how we may not want to do this if there is extra price associated with it.

IF we are in the place where we really consider saving on stack allocations, then yes - do not make NULL to mean "default", DO require the user to always pass the parameter, and do not allow NULLs (make it a precondition). Just make sure that for every such parameter, there is an (inline) function called `az_xxx_default()` that allows the user to construct the default value easily.

I would say that IF it would be possible to use "NULL = default" idiom in a way that the user does only pay the extra stack price when they use NULL, I would do it.

for example, it could look like this:
```
typedef struct {
    int y;
} options;

options az_options_default() { return { .y = 5}; }

int f(int x, options const* o)
{
    if (o == NULL)
    {
        options oo = az_options_default(); // <-- if only we could guarantee that this stack space is only allocated when o == NULL...
        return f(x, &oo);
    }

    return x + o->y;
}
```
That would be great if it could work in a way that the stack space is only allocated when the control enters the `if()` block. But most likely, the compiler will allocate the space for it in the beginning, before any comparison. If only we could use assembler to do it, then we could guarantee the behavior that we want. We could even avoid the recursive call. But maybe there are ways to do it, I just don't know them.

I would encourage us to consult with the language experts and try to find a way to use "you only pay for it if you use it" approach for "NULL = default" idiom, so it would work on all compilers.

Another option is to allow it on a case by case basis. For example, we could allow "NULL = default" in methods for Storage-Blobs, because it does use HTTP transport, which is huge relatively to most of the IoT applications. So, if you use Storage-Blobs, you are using the hardware of such caliber, that worrying about extra 48 bytes on stack is not something you would worry about.

Other option is to embed the default values in code. For the example below, it would look like this:
```
typedef struct {
    int y;
} options;

enum { _az_OPTIONS_DEFAULT_Y = 5 };

options az_options_default() { return { .y = _az_OPTIONS_DEFAULT_Y }; }

int f(int x, options const* o)
{
    return x + (o != NULL ? o->y : _az_OPTIONS_DEFAULT_Y );
}
```
This is not bad either. Maybe it is a bit harder to maintain, but not impossible. And after all, this is why we are making the SDK - to make customer's life easier, even if it means more work for us and more attention that needs to be paid by us when developing. Having a constant (`_az_OPTIONS_DEFAULT_Y`) makes it less prone to inconsistencies. In fact, I would rather insist on using named constants and not values, because it will be very hard to find all usages.

I would pay attention that if a function is using `o` for 20 times, the compiler does not generate 20 `if (o == NULL)` comparisons. I think it is likely that the compiler will optimize it.

A variation of this approach is to have getter functions for options:
```
AZ_INLINE int _az_options_get_y(options const* o) { return (o != NULL ? o->y : _az_OPTIONS_DEFAULT_Y); }

// actually, in this case, we don't have to have constants, we can alternatively write it as below:
// AZ_INLINE int _az_options_get_y(options const* o) { return (o != NULL ? o->y : 5); }

int f(int x, options const* o)
{
    return x + _az_options_get_y(o);
}
```
Especially if they are inline, there should be no extra cost (given that the compiler won't generate redundant NULL checks).

ahsonkhan: @CIPop I discussed this further with @JeffreyRichter and I think the best solution here is to disallow NULL for client method options, with a precondition, like we had previously. We'll continue to allow NULL == default semantics only for client init/ctors because that doesn't require extra stack space allocation.

As in:
> - at client ctor level, null options are allowed and mean default options
> - at client method/operation level, null options are not allowed

This way, we don't have to worry about modifications of the default options impacting the implementation detail of the method. We avoid the extra stack space, along with the extra runtime performance cost of initializing up the default options in the method.

Regarding some of the interesting suggestions from @antkmsft above, like recursively calling the client method to be "pay-for-play" by trying to craft the implementation so that we only pay for the cost of the stack-space *if* the user passes in `NULL`, the issue there is we'd end up relying on compiler specific details and optimizations, which is out of our control, and since those aren't guaranteed to be consistent, introduces brittleness.

> But, I understand that in *Embedded* C, there could be other considerations, such as stack allocation. And from that perspective, I see how we may not want to do this if there is extra price associated with it.

This rationale sums up why we'd have this difference compared to GoLang, and why we'd treat client init vs client methods differently.

ahsonkhan: Unless you have some other thoughts or different perspective, I'll leave it up to you to decide how and when to make this change, @CIPop :)

CIPop: Thanks @ahsonkhan: I will roll-back the default/NULL change.

PR #2227, opened 2022-06-13
allow assumption of two padding chars with base64

danewalton:
With JWS or JWK's, padding characters are explicitly not allowed. We need to allow for the assumption of padding characters should the text be `length % 4 != 0`.

### DIFF
--- a/sdk/src/azure/core/az_base64.c
+++ b/sdk/src/azure/core/az_base64.c
@@ -404,7 +404,7 @@ AZ_NODISCARD az_result
 az_base64_decode(az_span destination_bytes, az_span source_base64_text, int32_t* out_written)
 {
   _az_PRECONDITION_VALID_SPAN(destination_bytes, 1, false);
-  _az_PRECONDITION_VALID_SPAN(source_base64_text, 4, false);
+  _az_PRECONDITION_VALID_SPAN(source_base64_text, 2, false);
   _az_PRECONDITION_NOT_NULL(out_written);
 
   int32_t source_length = az_span_size(source_base64_text);
@@ -413,8 +413,9 @@ az_base64_decode(az_span destination_bytes, az_span source_base64_text, int32_t*
   int32_t destination_length = az_span_size(destination_bytes);
   uint8_t* destination_ptr = az_span_ptr(destination_bytes);
 
-  // The input must be non-empty and a multiple of 4 to be valid.
-  if (source_length == 0 || source_length % 4 != 0)
+  // The input must be non-empty and a minimum of two characters long.
+  // There can only be two assumed padding characters.
+  if (source_length == 0 || source_length % 4 == 1)
   {
     return AZ_ERROR_UNEXPECTED_END;
   }
@@ -440,12 +441,14 @@ az_base64_decode(az_span destination_bytes, az_span source_base64_text, int32_t*
     source_index += 4;
   }
 
-  // We are guaranteed to have an input with at least 4 bytes at this point, with a size that is a
-  // multiple of 4.
-  int32_t i0 = *(source_ptr + source_length - 4);
-  int32_t i1 = *(source_ptr + source_length - 3);
-  int32_t i2 = *(source_ptr + source_length - 2);
-  int32_t i3 = *(source_ptr + source_length - 1);
+  // If length is divisible by four, do nothing. Else, we assume up to two padding characters.
+  int32_t source_length_mod_four = source_length % 4;
+  int32_t i0 = *(source_ptr + source_index);
+  int32_t i1 = *(source_ptr + source_index + 1);
+  int32_t i2 = source_length_mod_four == 2 ? _az_ENCODING_PAD : *(source_ptr + source_index + 2);
+  int32_t i3 = source_length_mod_four == 2 || source_length_mod_four == 3
+      ? _az_ENCODING_PAD
+      : *(source_ptr + source_index + 3);
 
   i0 = _az_base64_decode_array[i0];
   i1 = _az_base64_decode_array[i1];
--- a/sdk/tests/core/test_az_base64.c
+++ b/sdk/tests/core/test_az_base64.c
@@ -165,11 +165,19 @@ static void az_base64_decode_test(void** state)
 
   uint8_t expected_buffer1[1] = { 1 };
   _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQ=="), AZ_SPAN_FROM_BUFFER(expected_buffer1));
+  uint8_t expected_buffer0[1] = { 1 };
+  _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQ"), AZ_SPAN_FROM_BUFFER(expected_buffer0));
   uint8_t expected_buffer2[2] = { 1, 2 };
   _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQI="), AZ_SPAN_FROM_BUFFER(expected_buffer2));
   uint8_t expected_buffer3[3] = { 1, 2, 3 };
   _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQID"), AZ_SPAN_FROM_BUFFER(expected_buffer3));
   uint8_t expected_buffer4[4] = { 1, 2, 3, 4 };
+  // Can't have three short padding characters, so test later for "AQIDB" returns AZ_ERROR_UNEXPECTED_END
+  // Assume 2 padding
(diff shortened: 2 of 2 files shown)


sdk/src/azure/core/az_base64.c
```diff
@@ -413,8 +413,9 @@ az_base64_decode(az_span destination_bytes, az_span source_base64_text, int32_t*
   int32_t destination_length = az_span_size(destination_bytes);
   uint8_t* destination_ptr = az_span_ptr(destination_bytes);
 
-  // The input must be non-empty and a multiple of 4 to be valid.
-  if (source_length == 0 || source_length % 4 != 0)
+  // The input must be non-empty and a minimum of two characters long.
+  // There can only be two assumed padding characters.
+  if (source_length == 0 || source_length % 4 == 1)
```
ericwol-msft, line 418: if you say min of 2, than why not source_length >= 2?

sdk/tests/core/test_az_base64.c
```diff
@@ -165,11 +165,19 @@ static void az_base64_decode_test(void** state)
...
   uint8_t expected_buffer2[2] = { 1, 2 };
   _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQI="), AZ_SPAN_FROM_BUFFER(expected_buffer2));
   uint8_t expected_buffer3[3] = { 1, 2, 3 };
   _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQID"), AZ_SPAN_FROM_BUFFER(expected_buffer3));
   uint8_t expected_buffer4[4] = { 1, 2, 3, 4 };
+  // Can't have three short padding characters, so test later for "AQIDB" returns
+  // AZ_ERROR_UNEXPECTED_END Assume 2 padding
+  _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQIDBA"), AZ_SPAN_FROM_BUFFER(expected_buffer4));
+  // Assume 1 padding
+  _az_base64_decode_test_helper(AZ_SPAN_FROM_STR("AQIDBA="), AZ_SPAN_FROM_BUFFER(expected_buffer4));
```
ericwol-msft, line 179: I don't see a fail test for "" and "A"

ericwol-msft (approved):

ahsonkhan:
> With JWS or JWK's, padding characters are explicitly not allowed. We need to allow for the assumption of padding characters should the text be `length % 4 != 0`.

I don't fully understand the purpose of this PR. Do we need full Base64 Url encoding/decoding support? If so, we'd need more changes, which we can't fit directly into an API that's meant just to do base64 encoding.

Does JWS or JWK use plain Base64 encoding or does it use Base64 Url encoding? Base64 Url encoding has other subtle differences (the character set has '-' and '_' instead of '+', '/'):
https://en.wikipedia.org/wiki/Base64#URL_applications

sdk/src/azure/core/az_base64.c
```diff
@@ -404,7 +404,7 @@ AZ_NODISCARD az_result
 az_base64_decode(az_span destination_bytes, az_span source_base64_text, int32_t* out_written)
 {
   _az_PRECONDITION_VALID_SPAN(destination_bytes, 1, false);
-  _az_PRECONDITION_VALID_SPAN(source_base64_text, 4, false);
+  _az_PRECONDITION_VALID_SPAN(source_base64_text, 2, false);
```
ahsonkhan, line 407: I don't believe this is valid base64, if we are strictly adhering to the spec, which this method was meant for:
https://datatracker.ietf.org/doc/html/rfc4648#section-4

> The encoding process represents 24-bit groups of input bits as output
   strings of **4 encoded characters**.  Proceeding from left to right, a
   24-bit input group is formed by concatenating 3 8-bit input groups.
   These 24 bits are then treated as 4 concatenated 6-bit groups, each
   of which is translated into a single character in the base 64
   alphabet.

danewalton: > > With JWS or JWK's, padding characters are explicitly not allowed. We need to allow for the assumption of padding characters should the text be `length % 4 != 0`.
> 
> I don't fully understand the purpose of this PR. Do we need full Base64 **Url** encoding/decoding support? If so, we'd need more changes, which we can't fit directly into an API that's meant just to do base64 encoding.
> 
> Does JWS or JWK use plain Base64 encoding or does it use Base64 Url encoding? Base64 Url encoding has other subtle differences (the character set has '-' and '_' instead of '+', '/'): https://en.wikipedia.org/wiki/Base64#URL_applications

Good points. This comes from the fact that [JWS explicitly drops the padding characters](https://datatracker.ietf.org/doc/html/rfc7515#appendix-C). It is also base64url encoded but that's easy to swap with a helper function. We can discuss what might be a more elegant solution to this? It's definitely needed for ADU and we would rather not have to add in the padding characters since we would have to memcpy the whole buffer over.

ahsonkhan: A strawman solution is to have new base64url encode and decode APIs with an additional bool parameter for optional/required padding characters emitted or required (depending on encode vs decode). They'd use a new array for the encode and decode character set, and otherwise largely share most of the implementation detail (common layer can be refactored out into helpers, to reduce size). Presumably, adding those extra 256+64 bytes is ok from a code size perspective for the embedded space, and the compiler can optimize out unused code.

That solution is cleaner from an API design perspective, conveys intent correctly, and keeps them single responsibility (also, we want to keep the existing base64 API true to what it does, which fails on invalid input on decode). So, unless there are some concerns with that, I suggest starting with that.

The character swap/replacement logic in an helper that you were thinking of is a decent solution, but it doesn't address the padding concern and I agree doing a memcpy to add the padding can be an issue (removing padding while encoding seems easy). The method though would be slower than always encoding correctly right of the bat (more work to go through the data in the buffer again), but there are runtime cost vs code size trade-offs that we might need to consider here.

How's the data that we need to process being encoded? Would the data being decoded typically get encoded with the use of our Base64 encode APIs or is the expectation that they get encoded from somewhere else?

danewalton: > How's the data that we need to process being encoded? Would the data being decoded typically get encoded with the use of our Base64 encode APIs or is the expectation that they get encoded from somewhere else?

This is coming from the service. So we take in the payload and have to decode it before doing things with the decoded characters.

ahsonkhan: @danewalton would it be ok if the base64url decoder didn't allow padding characters at all?

@LarryOsterman given padding characters are optional in base64url, should the decoder reject any payload that has padding as invalid, or be flexible and ignore them? I was thinking of either letting the caller make the decision about being extra strict (with a param) or being more flexible in what it accepts by default.

If we decide not to have the bool parameter, we can always be strict on the decoder to start with and allow padding characters, if need be, in the future, as that will be non-breaking.

The encoder not producing any padding characters from the get-go, and not giving an option for the caller to do so, seems reasonable.

danewalton: I see no reason to not allow for the padding characters to be there. It doesn't seem like the URL doesn't allow it, but merely gives the option to not have them. I'm not sure we lose any value essentially keeping what I have in this PR, which is that it will assume padding if they are missing, and if the length is `length % 4 == 0`, then it keeps whatever is there.

LarryOsterman: @ahsonkhan Padding characters are optional in base64url encoding. If they're there, the decoder should ignore them. The encoder should *not* produce them. Adding a parameter doesn't help the situation and only complicates the API surface (and as I've mentioned elsewhere, I've yet to encounter a situation where base64url encoded content doesn't have a well-defined length (largely because protocol designers who specify base64url encoding usually ensure that there is *something* in the protocol which can be used to determine the length of the encoded data). 

@danewalton What is important is that base64 decode needs to *reject* base64 encoded data that doesn't have padding characters, and base64url decode needs to allow base64url encoded data which doesn't have padding characters. And of course, base64 encode needs to add padding characters and base64url encode should NOT add padding characters.

ahsonkhan: That makes a lot of sense to me, and I agree. So, exclude the bool parameter from the API in the solution I mentioned above (on both sides).

Thanks for clarifying :)

danewalton: > @danewalton What is important is that base64 decode needs to reject base64 encoded data that doesn't have padding characters, and base64url decode needs to allow base64url encoded data which doesn't have padding characters. And of course, base64 encode needs to add padding characters and base64url encode should NOT add padding characters.

Perfect that sounds good to me

danewalton:
Closing this for now since the solution will be different.

PR #2233, opened 2022-06-14
Remove unreachable code

RickWinter:


ewertons (approved):
Hopefully this will not cause some other compiler to fail.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+ - Removed unreachable code
```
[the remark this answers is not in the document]
ahsonkhan: nit: Add a period.
```suggestion
 - Removed unreachable code in az_http_policy_retry.c.
```

sdk/src/azure/core/az_http_policy_retry.c
```diff
@@ -220,6 +220,4 @@ AZ_NODISCARD az_result az_http_pipeline_policy_retry(
       }
     }
   }
-
-  return result;
```
ahsonkhan, line 224: How is it guaranteed that we won't be stuck in an infinite loop?

I think a comment above this line would be helpful for maintainability in the future:
> if (attempt > max_retries || az_result_failed(result))

Something like:
// We increment attempt, each iteration of the loop, and this condition guarantees the loop will terminate after hitting max retries.

ahsonkhan (approved):

sdk/src/azure/core/az_http_policy_retry.c
RickWinter: There are other conditions that would terminate the loop.  Letting the code self-document seems better in this case, less likely the comment gets out of sync.
image

Issue #2079, opened 2022-01-25
Multi-block blob transfer in Storage client 


ghost:
The 1.3.0 version preview provides the Storage client for sending blobs to the Azure Storage service. However, this client allows only for sending one HTTP request with the whole blob content as the request body. Let's take the situation where the microcontroller device is acquiring lots of data from sensors (e.g. 10 seconds of 16 kHz, 3-axis accelerometer = ~2MB) and saving those data in the external flash memory. The device has only ~100 kB of free RAM. Without the multi-block blob transfer, the only solution I see with this Storage client is to create over a dozen files for one data acquisition instead of one, which is of course possible, but very inconvenient for future processing.
 
Are there any plans to implement a multi-block transfer similar to the one in the Azure IoT C SDK?

JeffreyRichter:
We could implement this; we've been waiting to see if any customers really needed the behavior.
We could implement uploading multiple blocks to a block blob or we could implement appending blocks to an append blob.
Do you have a preference? Seems like the append blob is perfect for your scenario.

szymsta:
Support for the append blob sounds good. It will be great if the library supports sending MD5 or CRC64 hash of the append block content to verify the integrity of the block during transport.

JeffreyRichter:
Don't you think that just using HTTP(S) is good enough for verifying the integrity ?

szymsta:
Most probably, yes. I just saw this option in the append block documentation: https://docs.microsoft.com/en-us/rest/api/storageservices/append-block

JeffreyRichter:
It will take our team a while to get to this request but we'd love to do it. It also wouldn't be difficult for someone to copy/paste this [code](https://github.com/Azure/azure-sdk-for-c/blob/feature/storage-blobs/sdk/src/azure/storage/az_storage_blobs_blob_client.c#L232) and tweak it to create a new az_storage_blobs_appendblob_appendBlock function which wraps this [operation](https://docs.microsoft.com/en-us/rest/api/storageservices/append-block). Another az_storage_blobs_appendblob_create function is required that wraps this [operation](https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob).

If you wrote these 2 functions, you'd be up and running quickly and, of course, we'd love you to make a PR to this repo :)

szymsta:
[...]

PR #2265, opened 2022-07-11
Multiblock and append blobs implementation

szymsta:
Resolves issue #2079 – "Multi-block blob transfer in Storage client"

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,8 +264,8 @@ The following compilation, preprocessor options will add or remove functionality
 See [cmake options](#cmake-options) to learn about how to build an HTTP transport adapter, how to build IoT samples, and to turn logging on.
 
 
-### Storage Sample
-The storage sample expects a storage account with a container and SaS token used for authentication to be set in an environment variable `AZURE_BLOB_URL_WITH_SAS`.
+### Storage Samples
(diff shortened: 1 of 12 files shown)


antkmsft:
Thank you so much @szymsta! It is really high quality PR. We are looking at it, and it may take us up some time as we'll review mostly the API to make sure it is extensible in the future to allow us to expand without introducing breaking changes.

antkmsft:
Hi @szymsta,

We could take  `az_storage_blobs_appendblob_create()` and `az_storage_blobs_appendblob_append_block()`, but not `az_storage_blobs_multiblock_blob_upload()` as it is.

Instead of `az_storage_blobs_multiblock_blob_upload()` being a function with a callback, we would create 2 functions: one to upload a block (taking int64 block id and an az_span instead of callback), and one to commit the block list.

With current design, it is not clear what should be done if error occurs in the middle of upload.

With 2 functions, user can wrap them to essentially do what `az_storage_blobs_multiblock_blob_upload()` currently does: it could iterate over blocks, and then commit the result, if there were no errors (or with some errors it thinks are acceptable).

BTW, I saw that you are doing `snprintf()` of block id to perform "itoa". That is not necessary. THe block id is a binary-value int64 that is Base64-encoded.

PR #2287, opened 2022-07-26
Unescape

### LINKED ISSUE #2173: JSON parser escaping/un-escaping
One of the main Azure IoT Provisioning Client purposes is to provide easy access to data within payloads received from the service. Today, it does this using `az_span`s pointing within the original buffer. 

This issue is discussing solutions to zero-memory allocation escaping/un-escaping.

## Proposed solutions:
1. In-place un-escaping of strings. This will alter the original buffer and destroy the JSON document.
2. Returning raw (escaped) strings and allowing the application to un-escape them. Today, the `az_json_token_get_string` will not work without a JSON context. Since we return the string only (within an `az_span`), the function cannot be used.

----

## Example snippet from the client's implementation returning the raw JSON element:

```C
    if (az_json_token_is_text_equal(&jr->token, AZ_SPAN_FROM_STR("assignedHub")))
    {
      _az_RETURN_IF_FAILED(az_json_reader_next_token(jr));
      if (jr->token.kind != AZ_JSON_TOKEN_STRING)
      {
      }
      out_state->assigned_hub_hostname = jr->token.slice;
      found_assigned_hub = true;
    }
```
We expect that `assigned_hub_hostname` does not contain any escaped characters and that the `az_span` contained string can immediately be used to connect to the assigned IoT Hub.

----

## Example for solution 1: in-place escaping:

```C
if (az_json_token_is_text_equal(&jr->token, AZ_SPAN_FROM_STR("issuedClientCertificate")))
    {
      _az_RETURN_IF_FAILED(az_json_reader_next_token(jr));
      if (jr->token.kind != AZ_JSON_TOKEN_STRING)
      {
      }

      // Un-escape JSON string in-place.
      int32_t new_size = 0;
      _az_RETURN_IF_FAILED(az_json_token_get_string(
          &jr->token,
          (char*)az_span_ptr(jr->token.slice),
          az_span_size(jr->token.slice),
          &new_size));
      out_state->issued_client_certificate = az_span_slice(jr->token.slice, 0, new_size);
    }
```

---

## JSON Parser API request for solution 2 above:

```C
/**
 * @brief Gets the JSON token's string after unescaping it, if required.
 */
AZ_NODISCARD az_result az_json_string_unescape(
    az_span json_string,
    char* destination,
    int32_t destination_max_size,
    int32_t* out_string_length);
```

gearama:
Add method for unescaping string , since it can be used to perform in place unescape , besides with different destination, added another method and tests to simplify the in place call.

closes #2173


ahsonkhan (commented):
Partial review. Remove convenience helper.

sdk/inc/azure/core/az_json.h
```diff
@@ -792,6 +792,43 @@ AZ_NODISCARD az_result az_json_reader_next_token(az_json_reader* ref_json_reader
...
+ * \p which denote the length of the unescaped string. If `NULL` is passed, the
+ * \p parameter is ignored.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The string is returned.
+ * @retval #AZ_ERROR_NOT_ENOUGH_SPACE \p destination does not have enough size.
+ */
+
+AZ_NODISCARD az_result
+az_json_string_unescape_in_place(az_span* json_string, int32_t* out_string_length);
```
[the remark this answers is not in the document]
gearama: 1. @CIPop  liked this API 
2. @ahsonkhan  please elaborate, the API works so i'm not sure where your'going with it, we also update the size in the span to the new size.
3. Apparently they would need more the in place rather than out of place due to memory constraints. 
Removing this API saves 10s -100s of bytes in the process code when loaded in memory, then use KBs in the data due to having to duplicate buffers. that does not seem like an overall loss in memory for the sake of doctrine.

bld22/.cmake/api/v1/query/client-vscode/query.json
```diff
@@ -0,0 +1 @@
+{"requests":[{"kind":"cache","version":2},{"kind":"codemodel","version":2},{"kind":"toolchains","version":1}]}
```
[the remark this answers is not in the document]
gearama: done

gearama:
@ahsonkhan all ready

sdk/inc/azure/core/az_json.h
```diff
@@ -792,6 +792,27 @@ AZ_NODISCARD az_result az_json_reader_next_token(az_json_reader* ref_json_reader
...
+ * \p destination which denote the length of the unescaped string. If `NULL` is passed, the
+ * \p parameter is ignored.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The string is returned.
+ * @retval #AZ_ERROR_NOT_ENOUGH_SPACE \p destination does not have enough size.
+ */
+AZ_NODISCARD az_result az_json_string_unescape(
+    az_span json_string,
+    char* destination,
```
CIPop, line 812: @ahsonkhan just curious: why was this originally created as a `char*` and `out_string_length` instead of an `az_span` ?
Is it guaranteed in the original `az_json_token_get_string()` API that the output is a `NULL`-terminated string?

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,54 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+      uint8_t next_char = json_string._internal.ptr[i + 1];
+      // check that we have sometig to escape
+      if (_az_is_valid_escaped_character(next_char))
+      {
+        current_char = _az_json_unescape_single_byte(next_char);
+        i++;
+      }
+    }
+
+    // TODO Check same buffer input and dest
```
CIPop, line 357: This TODO should be addressed.

sdk/inc/azure/core/az_json.h
```diff
@@ -792,6 +792,27 @@ AZ_NODISCARD az_result az_json_reader_next_token(az_json_reader* ref_json_reader
  */
 AZ_NODISCARD az_result az_json_reader_skip_children(az_json_reader* ref_json_reader);
 
+/**
+ * @brief Gets the JSON token's string after unescaping it, if required.
+ *
+ * @param[in] json_string The #az_span that contains the string to be unnescaped.
```
CIPop, line 798: Spelling: unescaped.

sdk/inc/azure/core/az_json.h
```diff
@@ -792,6 +792,27 @@ AZ_NODISCARD az_result az_json_reader_next_token(az_json_reader* ref_json_reader
...
+ * @param destination Pointer to the buffer that will contain the oputput.
```
CIPop, line 799: Spelling: output.

CIPop, line 799: It's not clear here that the function allows _in-place_ unescaping.
I liked the idea of having 2 APIs (one in-place and one generic) to indicate setting the `json_string` and `destination` to point at the same memory location.

CIPop (approved):
A couple of spelling errors in the API docs.

sdk/inc/azure/core/az_json.h
gearama: @ahsonkhan didn't want that in order to reduce the API surface. Apparently that API does not meet the requirements. And since it was a wrapper, and the test show how to do inplace i reluctantly agreed.

sdk/inc/azure/core/az_json.h
gearama: The API signature is the one that came from the IOT team , it is in the issue associated with the PR. While i agree with the point i suspect that there might be some other reason for it

ahsonkhan: > Is it guaranteed in the original `az_json_token_get_string()` API that the output is a `NULL`-terminated string?

Yes.

sdk/inc/azure/core/az_json.h
```diff
@@ -792,6 +792,29 @@ AZ_NODISCARD az_result az_json_reader_next_token(az_json_reader* ref_json_reader
  */
 AZ_NODISCARD az_result az_json_reader_skip_children(az_json_reader* ref_json_reader);
 
+/**
+ * @brief Gets the JSON token's string after unescaping it, if required.
```
ahsonkhan, line 796: The input parameter is not a `json_token` but rather `az_span`, so consider updating this comment

```suggestion
 * @brief Unescapes the JSON string within the provided #az_span.
```

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
   return AZ_OK;
 }
 
+/**
+ * @brief Gets the JSON token's string after unescaping it, if required.
+ */
```
ahsonkhan, line 330: Duplicating the doc comment from the header into the source file isn't required.

```suggestion
```

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+AZ_NODISCARD az_result az_json_string_unescape(
+    az_span json_string,
+    char* destination,
+    int32_t destination_max_size,
+    int32_t* out_string_length)
+{
+  _az_PRECONDITION_VALID_SPAN(json_string, 1, false);
```
ahsonkhan, line 337: This is a good precondition. Even though an empty JSON string token is valid (`""`), the caller can check it up front and not have the need to unescape anything in that case.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+  _az_PRECONDITION_NOT_NULL(destination);
+  _az_PRECONDITION(destination_max_size > 0);
+
+  int32_t position = 0;
+  for (int32_t i = 0; i < json_string._internal.size; i++)
```
ahsonkhan, line 342: Avoid using references to `_internal` members of other types (like `az_span`).
In this case, I'd use the public API: `az_span_size(json_string)`, and save that size in a local.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+  {
+    uint8_t current_char;
+    current_char = json_string._internal.ptr[i];
```
ahsonkhan, line 345: Similarly here (and elsewhere), use `az_span_ptr()`, to avoid accessing `_internal`.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+    int32_t* out_string_length)
+{
+  _az_PRECONDITION_VALID_SPAN(json_string, 1, false);
+  _az_PRECONDITION_NOT_NULL(destination);
+  _az_PRECONDITION(destination_max_size > 0);
+
+  int32_t position = 0;
+  for (int32_t i = 0; i < json_string._internal.size; i++)
+  {
+    uint8_t current_char;
```
ahsonkhan, line 344: nit: Out of curiosity, why is the declaration and assignment separate?

Why not:
```c
uint8_t current_char = json_string._internal.ptr[i];
```

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+    current_char = json_string._internal.ptr[i];
+    if (current_char == '\\' && i < json_string._internal.size)
+    {
+      uint8_t next_char = json_string._internal.ptr[i + 1];
+      // check that we have sometig to escape
```
ahsonkhan, line 349:
```suggestion
      // check that we have something to escape
```

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+      if (_az_is_valid_escaped_character(next_char))
+      {
+        current_char = _az_json_unescape_single_byte(next_char);
+        i++;
+      }
```
ahsonkhan, line 354: If it isn't a valid escaped character, please return an error `az_result`, and document that (in the else case).

We shouldn't silently ignore invalid JSON as input.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+
+    destination[position] = (char)current_char;
+    position++;
+  }
+
+  if (out_string_length != NULL)
+  {
+    *out_string_length = position;
+  }
+
```
ahsonkhan, line 370: We are missing the null terminating character.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+    uint8_t current_char;
+    current_char = json_string._internal.ptr[i];
+    if (current_char == '\\' && i < json_string._internal.size)
+    {
+      uint8_t next_char = json_string._internal.ptr[i + 1];
+      // check that we have sometig to escape
+      if (_az_is_valid_escaped_character(next_char))
+      {
+        current_char = _az_json_unescape_single_byte(next_char);
+        i++;
```
ahsonkhan, line 353: What happens if the input is `"A\\\"\"Z"`? We seem to be skipping past all sequences of backslashes, which isn't correct.

The output in this case should be:
`A\""Z`


Please add test cases.

sdk/tests/core/test_az_json.c
```diff
@@ -2978,6 +2981,264 @@ static void test_az_json_reader_chunked(void** state)
...
+    assert_int_equal(AZ_OK, result);
+    assert_int_equal(expected._internal.size, final_size);
+    assert_string_equal(expected._internal.ptr, destination);
+    assert_true(az_span_is_content_equal(expected, destination_span));
+  }
+
+  // fake escapes
+  {
+    az_span original = AZ_SPAN_FROM_STR("\\9");
+    az_span expected = AZ_SPAN_FROM_STR("\\9");
```
ahsonkhan, line 3060: This is invalid JSON, and rather than copying the data verbatim, we should return an error `az_result`.

sdk/tests/core/test_az_json.c
```diff
@@ -2978,6 +2981,264 @@ static void test_az_json_reader_chunked(void** state)
...
+    az_span destination_span = AZ_SPAN_FROM_BUFFER(destination);
+
+    int final_size;
+    az_result result = az_json_string_unescape(original, destination, 59, &final_size);
+
+    destination_span._internal.size = final_size;
+
+    assert_int_equal(AZ_OK, result);
+    assert_int_equal(expected._internal.size, final_size);
+    assert_string_equal(expected._internal.ptr, destination);
```
ahsonkhan, line 3072: maintenance nit: Similar to the product code, avoid using `az_span` _internal state outside of files related to the az_span type, like these, which are `az_json` tests.

Just a nit, since it's test code.

Use the public members `az_span_size()` and `az_span_ptr()`.
This avoid brittleness and allows changes to `_internal` fields (which is not-breaking) to be scoped to just the related files, and not requiring changes all over the place.

ahsonkhan (commented):
There are some edge cases that I don't think are being handled correctly in the implementation atm. Also, we are missing some JSON string validation checks + tests.

sdk/tests/core/test_az_json.c
```diff
@@ -2978,6 +2981,264 @@ static void test_az_json_reader_chunked(void** state)
...
+    assert_true(az_span_is_content_equal(expected, destination_span));
+  }
+
+  // only escapes
+  {
+    az_span original = AZ_SPAN_FROM_STR("\\b\\f\\n\\r\\t\\\\");
```
ahsonkhan, line 3020: We are missing a test case for embedded quotes, like `\"` and forward slash `\/`.

Also look to add a test case like this:
https://github.com/Azure/azure-sdk-for-c/blob/935cb92bdbbcecfed91c02b98eec39d84c8c4297/sdk/tests/core/test_az_json.c#L2117

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+    if (position > destination_max_size)
+    {
+      return AZ_ERROR_NOT_ENOUGH_SPACE;
+    }
+
+    destination[position] = (char)current_char;
+    position++;
+  }
+
+  if (out_string_length != NULL)
```
ahsonkhan, line 366: We are missing branch coverage test for this being null.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+ */
+AZ_NODISCARD az_result az_json_string_unescape(
+    az_span json_string,
+    char* destination,
+    int32_t destination_max_size,
+    int32_t* out_string_length)
+{
+  _az_PRECONDITION_VALID_SPAN(json_string, 1, false);
+  _az_PRECONDITION_NOT_NULL(destination);
+  _az_PRECONDITION(destination_max_size > 0);
```
ahsonkhan, line 339: We have pretty decent test coverage for the current implementation of this method.

That said, we can improve it slightly by adding test coverage for the precondition validation.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=1743193&view=codecoverage-tab
image

sdk/src/azure/core/az_json_token.c
gearama: same for a 1 char string , to unescape you need at least 2 chars

sdk/src/azure/core/az_json_token.c
gearama: i can't do that how do i know that "\\9" is not what the customer intended in that string. that check only sais that the next byte is a known escape character, but i cannot determine the intent of a char that is not escapable. See the definition of the escape next byte, it cannot tell if it is an erro or not. it returns the char if it is not a known one.

sdk/src/azure/core/az_json_token.c
gearama: no, i'm not, spans are not null terminated.

gearama: especially for the in place scenario we don't want to add \0 at the end. We unescape what is there, not add extra stuffs

sdk/tests/core/test_az_json.c
gearama: now i'm supposed to get into the json validation game   ? You yourself said that the customer can send whatever

sdk/src/azure/core/az_json_token.c
gearama: actually you are wrong "A\\\"\"Z" broken into chars is 
0 A
1 \\  -> \     
2 \"  -> "    
3 \"   -> "
4 Z

char 1 is \ char 2 is " which unescaped become \" meaning " 
so the final unquted string is A""Z , quoted is "A\"\"Z"

sdk/tests/core/test_az_json.c
ahsonkhan: > Also no other methos in this file validates that the input is a valid json

That is because all other APIs accept an `az_json_token`, which the user gets from a validating `az_json_reader`.

This new API is accepting an az_span, which can come from any arbitrary slice of input or byte array.

We can discuss how we want to think about validation of input, as a follow-up, separately.

sdk/src/azure/core/az_json_token.c
ahsonkhan: I understand. That said, the API we are adding is about JSON unescaping.

> i can't do that how do i know that "\9" is not what the customer intended in that string

You know because such a string would be invalid JSON and is meaningless (or at least undefined) to try to unescape such a string. It is invalid data (or possibly a data corruption or user-side bug).

The only valid characters that can follow a `\` in a JSON string are one of this set:
https://datatracker.ietf.org/doc/html/rfc8259#section-7

We have to decide on one of the following (and document that behavior). Either:
- The user has to provide valid JSON, as an az_span
  - What happens if they don't?
- The user can pass any arbitrary input, and we validate/return an error when the input is malformed JSON.

> See the definition of the escape next byte, it cannot tell if it is an erro or not. it returns the char if it is not a known one.

That is just a helper method used to unescape already validated data. A `az_json_token` is known to be valid already, because of how it is created.

sdk/src/azure/core/az_json_token.c
ahsonkhan: Any test input that has multiple repeating `\` characters will highlight the issue with the current implementation. It is skipping through all sequences of `\` characters. That's what I am trying to point out here.

The two characters `\\` should be transformed into `\`, when unescaped, not ignored.

The test input I am talking about is the following 8 character valid JSON string:
0 A
1 \\
2 \\
3 \\
4 \"
5  \\
6 \"
7 Z

That should be transformed into the following string, once unescaped:
`A\""Z` (5 characters)

Can you share how I am wrong here?

sdk/src/azure/core/az_json_token.c
ahsonkhan: Btw, I am talking about having an else for this: `if(_az_is_valid_escaped_character(next_char))`

That check should end up catching and correctly rejecting `\9` as invalid.

sdk/src/azure/core/az_json_token.c
```diff
@@ -325,6 +325,52 @@ AZ_NODISCARD static az_result _az_json_token_get_string_helper(
...
+
+  int32_t position = 0;
+  for (int32_t i = 0; i < json_string._internal.size; i++)
+  {
+    uint8_t current_char;
+    current_char = json_string._internal.ptr[i];
+    if (current_char == '\\' && i < json_string._internal.size)
+    {
+      uint8_t next_char = json_string._internal.ptr[i + 1];
```
ahsonkhan, line 348: Looking at test coverage, I just noticed, this `i + 1` can end up accessing the address outside the bounds of the span. Please fix.

Let's say input is 1 character long: `\`

current_char = `\` and i = 0, which is less than the length (1).
next_char = json_string._internal.ptr[1]; // undefined behavior, accessing out-of-memory

sdk/src/azure/core/az_json_token.c
gearama: All nice and dandy , how do you catch exceptions in C ?

sdk/src/azure/core/az_json_token.c
gearama: there is a test for that line 3087

PR #2288, opened 2022-07-27
add adu code

danewalton:


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,532 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
```
ewertons, line 5:
```suggestion
 * @file az_iot_adu_client.h
```

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,532 @@
...
+ *
+ * @brief
```
ewertons, line 7:
```suggestion
 * @brief Definition for the Azure IoT ADU Client
```

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,532 @@
...
+ * @note
```
ewertons, line 8:
```suggestion
 * @note More details about Azure Device Update can be found online at https://docs.microsoft.com/en-us/azure/iot-hub-device-update/understand-device-update
```

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,532 @@
...
+ *
+ * @note
```
ewertons, line 10:
```suggestion
 * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
 * prefixed with an underscore ('_') directly in your application code. These symbols
 * are part of Azure SDK's internal implementation; we do not document these symbols
 * and they are subject to change in future versions of the SDK which would break your code.
```

sdk/samples/iot/paho_iot_adu_sample_common.c
```diff
@@ -0,0 +1,662 @@
...
+
+// ADU Feature Values
+static az_iot_adu_update_request xBaseUpdateRequest;
+static az_iot_adu_update_manifest xBaseUpdateManifest;
+static char adu_new_version[16];
+static bool did_parse_update = false;
+static bool did_update = false;
+static char adu_scratch_buffer[10000];
+
+#define AZ_IOT_ADU_AGENT_VERSION "DU;agent/0.8.0-rc1-public-preview"
```
ewertons, line 79: Instead of redefining we could just use the homonymous macro from az_iot_adu_client.h

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,956 @@
...
+      // TODO: Add quotes if result_details is not enclosed by quotes.
+      _az_RETURN_IF_FAILED(az_json_writer_append_property_name(
+          ref_json_writer, AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_RESULT_DETAILS)));
+      _az_RETURN_IF_FAILED(
+          az_json_writer_append_string(ref_json_writer, last_install_result->result_details));
+    }
+
+    for (int32_t i = 0; i < last_install_result->step_results_count; i++)
+    {
+      az_span remaining_buffer = get_json_writer_remaining_buffer(ref_json_writer);
```
ewertons, line 281: Not for now, but we should get the Azure SDK team opinion about this... Seems a little unsettled to just grab space from the end of the buffer.

ewertons (approved):

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: learned from victor that this isn't actually necessary (and is debatably better without it since there's isn't an error if you change the file name).

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: I'll add a TODO on it

CMakeLists.txt
```diff
@@ -113,6 +113,7 @@ if (UNIT_TESTING)
   add_subdirectory(sdk/tests/core)
 
   # IoT
+  add_subdirectory(sdk/tests/iot/adu)
```
ericwol-msft, line 116: Would this make it more discovale instead of using the acronym?
```suggestion
  add_subdirectory(sdk/tests/iot/device_update)
```

danewalton: It's a decent idea but probably something we should look at later. Just because there would be a lot of changes and we want this to be as minimal a PR and process as possible.

PR #2290, opened 2022-07-29
Unescape

gearama:
Some more updates after long Ashon discussion


sdk/tests/core/test_az_json.c
```diff
@@ -3010,11 +3010,25 @@ static void test_az_json_string_unescape(void** state)
...
-    assert_string_equal(expected._internal.ptr, destination);
     assert_true(az_span_is_content_equal(expected, destination_span));
   }
 
+  // unexpected end
+  {
+    uint8_t buffer[3];
+    az_span test_span = AZ_SPAN_FROM_BUFFER(buffer);
+    test_span._internal.ptr[0] = 'a';
+    test_span._internal.ptr[1] = '"';
```
ahsonkhan, line 3021: Generally each unit test should validate one thing.
This test case is about hitting an unexpected end. We should remove embedded quotes, as that's an invalid JSON string.

```suggestion
    test_span._internal.ptr[1] = 'b';
```

Let's separate the two test cases.

Otherwise, when we add unescaped embedded quote validation, this test will fail with unexpected char, and not unexpected end.

sdk/tests/core/test_az_json.c
```diff
@@ -3049,15 +3061,37 @@ static void test_az_json_string_unescape(void** state)
...
+
+    char destination[59] = { 0 };
+
+    int final_size;
+    az_result result = az_json_string_unescape(original, destination, 59, &final_size);
+
+    assert_int_equal(AZ_ERROR_UNEXPECTED_END, result);
+  }
+
+  // malformed escapes 2
```
ahsonkhan, line 3091: This comment doesn't match the test case ("a")

sdk/tests/core/test_az_json.c
```diff
@@ -3068,16 +3102,24 @@ static void test_az_json_string_unescape(void** state)
...
+    int final_size;
+
+    az_result result = az_json_string_unescape(json, destination, 4, &final_size);
 
+    assert_int_equal(AZ_ERROR_NOT_ENOUGH_SPACE, result);
+  }
+
+  // magic test
+  {
+    az_span original = AZ_SPAN_FROM_STR("A\\\"\"Z");
```
ahsonkhan, line 3121: Ideally, we should disallow such inputs since having an unescaped embedded quote within a JSON string token is invalid.

Rather than this succeeding, it should return az_result unexpected char.

That said, we can resolve this in a follow-up issue.

sdk/tests/core/test_az_json.c
```diff
@@ -3068,16 +3102,24 @@ static void test_az_json_string_unescape(void** state)
     destination_span._internal.size = final_size;
```
ahsonkhan, line 3102: Nit: Rather than manually modifying the internals of the az_span, which shouldn't be used outside the type, use the public slice API (here and elsewhere).

```suggestion
    destination_span = az_span_slice(destination_span, 0, final_size):
```

Just a nit, since it's test code.

sdk/tests/core/test_az_json.c
```diff
@@ -3194,49 +3295,31 @@ static void test_az_json_string_unescape_same_buffer(void** state)
...
-    original._internal.size = final_size;
+        = az_json_string_unescape(original, (char*)az_span_ptr(original), 59, &final_size);
 
-    assert_int_equal(AZ_OK, result);
-    assert_true(az_span_is_content_equal(expected, original));
+    assert_int_equal(AZ_ERROR_UNEXPECTED_END, result);
     _az_span_free(&original);
   }
-
-  // insufficient space
```
ahsonkhan, line 3229: Why are we removing this test case?

ahsonkhan (approved):
Some test fix-up suggestions. Looks good otherwise.

sdk/src/azure/core/az_json_token.c
```diff
@@ -339,19 +336,28 @@ AZ_NODISCARD az_result az_json_string_unescape(
...
       {
         current_char = _az_json_unescape_single_byte(next_char);
         i++;
       }
+      else
+      {
+        return AZ_ERROR_UNEXPECTED_CHAR;
+      }
+    }
+    else if (current_char == '\\' && i == span_size - 1)
```
ahsonkhan, line 358: Test coverage is showing this branch isn't fully covered (likely one of the 3 cases where the condition will be false is missing).

sdk/tests/core/test_az_json.c
gearama: done

sdk/tests/core/test_az_json.c
gearama: done

sdk/tests/core/test_az_json.c
gearama: it was one of the follow up issues we discussed yesterday

sdk/tests/core/test_az_json.c
gearama: it's a boring test, i put it back.

sdk/src/azure/core/az_json_token.c
gearama: i would say the code covreage tool is dumb. the branch not covered here , setting breakpoints on all the branches , get hit, there is one branch which is not explicitly written in the code but it is executed, the else on the last else if(). Since the condition returns the else branch continues the execution. i would rather not specifically put the else branch as it would make the code look blah.

ahsonkhan: Hmm, I think I know why. Parts of the condition are redundant, so they can never be covered. It's not because of a missing else.

You can remove `&& i == span_size - 1`, and the code will be identical.

The way the code is written, I don't think there's any way for `current_char == '\\' ` to be true, in this else-if but `i == span_size - 1` to be false.

The for-loop puts a bound on i to be between 0 and span_size.
If i is not less than span_size - 1 (from the if-check), then it must be equal to it.

ahsonkhan: Fixed in https://github.com/Azure/azure-sdk-for-c/pull/2291

Issue #2301, opened 2022-08-08
Add null terminator and precondition on input length for the json unescape method


ahsonkhan:
This will match the behavior more closely to `az_json_token_get_string`

https://github.com/Azure/azure-sdk-for-c/blob/c5238a6ba7ff1cdf645c2fec0cae07d762faf88c/sdk/inc/azure/core/az_json.h#L812

Issue #2302, opened 2022-08-08
Consider a json_string_unescape method that has span as the destination and how much input validation is needed


ahsonkhan:
**Current API:**
https://github.com/Azure/azure-sdk-for-c/blob/c5238a6ba7ff1cdf645c2fec0cae07d762faf88c/sdk/inc/azure/core/az_json.h#L812
```c
// Null termination
// Open question of how much validation to do
AZ_NODISCARD az_result az_json_string_unescape(
    az_span json_string,
    char* destination,
    int32_t destination_max_size,
    int32_t* out_string_length);
```

Option 1:
```c
// No need for null termination, no input validation
AZ_NODISCARD az_span az_json_string_unescape(
    az_span json_string,
    az_span destination);
```

Option 2:
```c
// No need for null termination
// Open question of how much validation to do
AZ_NODISCARD az_result az_json_string_unescape(
    az_span json_string,
    az_span destination,
    int32_t* out_string_length);
```

ahsonkhan:
The other question is: Should we call the API decode, instead of unescape, since we are decoding the JSON string?

The action happening is clearer with decode, since it's a verb:
`az_json_string_decode`

Using `unescape` could be ambigious, as a noun.

PR #2303, opened 2022-08-08
Add null terminator for char* based json string unescape, and precondition on input length

ahsonkhan:
Fixes https://github.com/Azure/azure-sdk-for-c/issues/2301

### DIFF
--- a/sdk/src/azure/core/az_json_token.c
+++ b/sdk/src/azure/core/az_json_token.c
@@ -333,7 +333,8 @@ AZ_NODISCARD az_result az_json_string_unescape(
 {
   _az_PRECONDITION_VALID_SPAN(json_string, 1, false);
   _az_PRECONDITION_NOT_NULL(destination);
-  _az_PRECONDITION(destination_max_size > 0);
+  // The destination needs to be larger than the input, for null terminator.
+  _az_PRECONDITION(destination_max_size > az_span_size(json_string));
 
   int32_t position = 0;
   int32_t span_size = az_span_size(json_string);
@@ -369,6 +370,8 @@ AZ_NODISCARD az_result az_json_string_unescape(
     destination[position] = (char)current_char;
     position++;
   }
+    
+  destination[position] = 0;
 
   if (out_string_length != NULL)
   {
--- a/sdk/src/azure/core/az_json_token.c
+++ b/sdk/src/azure/core/az_json_token.c
@@ -370,7 +370,7 @@ AZ_NODISCARD az_result az_json_string_unescape(
     destination[position] = (char)current_char;
     position++;
   }
-    
+
   destination[position] = 0;
 
   if (out_string_length != NULL)
--- a/sdk/tests/core/test_az_json.c
+++ b/sdk/tests/core/test_az_json.c
@@ -3105,17 +3105,6 @@ static void test_az_json_string_unescape(void** state)
     assert_true(az_span_is_content_equal(expected, destination_span));
   }
 
-  // insufficient space
-  {
-    az_span json = AZ_SPAN_FROM_STR(" { \"name\": \"some value string\" , \"code\" : 123456 } ");
-    char destination[5] = { 0 };
-    int final_size;
-
-    az_result result = az_json_string_unescape(json, destination, 4, &final_size);
-
-    assert_int_equal(AZ_ERROR_NOT_ENOUGH_SPACE, result);
-  }
-
   // magic test
   {
     az_span original = AZ_SPAN_FROM_STR("A\\\"\"Z");
@@ -3317,18 +3306,6 @@ static void test_az_json_string_unescape_same_buffer(void** state)
 
     assert_int_equal(AZ_ERROR_UNEXPECTED_END, result);
     _az_span_free(&original);
-  }
-
-  // insufficient space
-  {
-    az_span json = az_span_create_from_str(
-        strdup(" { \"name\": \"some value string\" , \"code\" : 123456 } "));
-    int final_size;
-
-    az_result result = az_json_string_unescape(json, (char*)az_span_ptr(json), 5, &final_size);
-
-    assert_int_equal(AZ_ERROR_NOT_ENOUGH_SPACE, result);
-    _az_span_free(&json);
   }
 }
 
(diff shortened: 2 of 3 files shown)


sdk/tests/core/test_az_json.c
```diff
@@ -3105,17 +3105,6 @@ static void test_az_json_string_unescape(void** state)
     assert_true(az_span_is_content_equal(expected, destination_span));
   }
 
-  // insufficient space
-  {
-    az_span json = AZ_SPAN_FROM_STR(" { \"name\": \"some value string\" , \"code\" : 123456 } ");
-    char destination[5] = { 0 };
-    int final_size;
-
-    az_result result = az_json_string_unescape(json, destination, 4, &final_size);
```
LarryOsterman, line 3114: Strangely, this test for unescape doesn't have an escaped string.

I would have expected to see a test which had a string with escaped values where the destination was too short for the unescaped string.

sdk/src/azure/core/az_json_token.c
```diff
@@ -333,7 +333,8 @@ AZ_NODISCARD az_result az_json_string_unescape(
 {
   _az_PRECONDITION_VALID_SPAN(json_string, 1, false);
   _az_PRECONDITION_NOT_NULL(destination);
-  _az_PRECONDITION(destination_max_size > 0);
```
LarryOsterman, line 336: Interestingly, the precondition at line 335 precludes the more traditional use of this pattern:

```c
err = az_json_string_unescape(json_string, NULL, 0, &size);
char *dest = malloc(size);
err = az_json_string_unescape(json_string, dest, size, &size);
```

Instead, you need to do something silly like:
```c
char dummy;
err = az_json_string_unescape(json_string, &dummy, 0, &size);
char *dest = malloc(size);
err = az_json_string_unescape(json_string, dest, size, &size);
```

LarryOsterman (approved):

sdk/tests/core/test_az_json.c
```diff
@@ -3105,17 +3105,6 @@ static void test_az_json_string_unescape(void** state)
     assert_true(az_span_is_content_equal(expected, destination_span));
   }
 
-  // insufficient space
-  {
-    az_span json = AZ_SPAN_FROM_STR(" { \"name\": \"some value string\" , \"code\" : 123456 } ");
```
antkmsft, line 3110: We need to be testing for `AZ_ERROR_NOT_ENOUGH_SPACE`, `AZ_ERROR_UNEXPECTED_CHAR`, `AZ_ERROR_UNEXPECTED_END`, for the null terminator at the end, for the precondition, for the correct size being written to `out_string_length`, for the precondition, and for the fact that the string is indeed unescaped where expected.

Maybe some of it is being tested, but it is not obvious so far. So far I can see you are deleting this coverage (one instance is legitimately a duplicate?) But I do not see at least the test for null terminator and for the precondition.

antkmsft (approved):
Will look good once covered by tests.
Functionality LGTM, I see no issues.

sdk/tests/core/test_az_json.c
ahsonkhan: The other cases you mentioned are covered already.

We haven't tested the other preconditions for this method, so we can improve coverage separately. The goal of the PR is just to fix the null terminator.

I will add the test case for nulls at the end.

sdk/tests/core/test_az_json.c
ahsonkhan: Agreed. That said, this existing test was only testing destination size being too small, independent of unescaping.

sdk/src/azure/core/az_json_token.c
ahsonkhan: Our existing SDK APIs don't behave like this, where the required size is returned on a failed call.

sdk/tests/core/test_az_json.c
ahsonkhan: Here's the coverage. The null terminator line is covered:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1769797&view=codecoverage-tab

And the `AZ_ERROR_NOT_ENOUGH_SPACE` case will only be possible if preconditions are off, which is why it isn't being covered now. We don't expect anyone to hit that case (and it is only there to avoid memory corruption, as an extra check).

I'll file an issue to address the test coverage for this API around preconditions, since the issue already exists in main.

ahsonkhan: https://github.com/Azure/azure-sdk-for-c/issues/2310

antkmsft: Is it even needed then, if it is a precondition? Once you violate the precondition, you're in the undefined behavior territory, and also the purposes of preconditions is for our code to not spend extra cycles checking for errors. I'd say it should be one or another.

Null terminator string will be shown as covered, but that's only because the line was executed during tests, and the tool can't see that you are not actually checking the buffer that null terminator was indeed being written. Please add that test.

sdk/src/azure/core/az_json_token.c
```diff
@@ -333,7 +333,8 @@ AZ_NODISCARD az_result az_json_string_unescape(
...
+  // The destination needs to be larger than the input, for null terminator.
+  _az_PRECONDITION(destination_max_size > az_span_size(json_string));
 
   int32_t position = 0;
   int32_t span_size = az_span_size(json_string);
```
antkmsft, line 340: Since size is precondition, `if (position > destination_max_size)` can be dropped.
It checks on every loop iteration.

Either way, even if we keep it, it would be much cheaper to just test it before the loop, just as you do with precondition.

Then, it could be a single `if (current_char == '\\') { if (i < span_size - 1) { ... } else { ... } }`, instead of checking `if (current_char == '\\')` twice (even if not for optimization, but readability as well).

ahsonkhan: You are right.

I think I am going to re-look at this whole implementation, including the optimizations you mentioned, as part of:
https://github.com/Azure/azure-sdk-for-c/issues/2302

I want to keep the code change in this PR minimal, for now, to avoid too much churn, since it will likely be throw away work.

sdk/src/azure/core/az_json_token.c
LarryOsterman: That's a shame, because the entire reason for this pattern `([in]foo*, [in]int fooSize, [out]int* sizeNeeded)` is to enable that use scenario (passing in 0 and `NULL` to do a sizing pass followed by the actual allocation).

PR #2329, opened 2022-09-06
add base64 url encoder

danewalton:


sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
   return (((source_bytes_size + 2) / 3) * 4);
 }
 
-static AZ_NODISCARD int32_t _az_base64_decode(uint8_t* encoded_bytes)
+static int32_t _get_base64_decoded_char(int32_t index, _az_base64_mode mode)
+{
+  if (mode == _az_base64_mode_url)
+  {
+    if (index == 43 || index == 47)
```
LarryOsterman, line 383: Magic numbers like 43, 45, or 95 confuse me. Could you use '-'/'_'/'+'/'/' instead of their ASCII equivalents? Or #define a value for the numeric constants?

sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
...
+    {
+      return -1; // can't use + or / with URL encoding
+    }
```
antkmsft, line 386:
```suggestion
    if (index == (int32_t)'+' || index == (int32_t)'/')
    {
      return -1;
    }
```

sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
...
+
+    if (index == 45)
+    {
+      index = 43; // - becomes a +
+    }
```
antkmsft, line 391:
```suggestion
    if (index == (int32_t)'-')
    {
      index = (int32_t)'+';
    }
```

sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
...
+    else if (index == 95)
+    {
+      index = 47; // _ becomes a /
+    }
```
antkmsft, line 395:
```suggestion
    else if (index == (int32_t)'_')
    {
      index = (int32_t)'/';
    }
```

sdk/src/azure/core/az_base64.c
danewalton: Ahhh yes absolutely

sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
...
+  }
+
+  return _az_base64_decode_array[index];
```
antkmsft, line 398: Can we get rid of the array? it is 256 bytes, and it is not very readable (bunch of `-1`s). Can we replace it with an (inline?) function doing something like this?
```c
if (c >= 'A' && c <= 'Z')
{
  return (c - 'A');
}

if (c >= 'a' && c <= 'z')
{
  return 26 + (c - 'a');
}

if (c >= '0' && c <= '9')
{
  return 52 + (c - '0');
}

if (c == '+')
{
  return 62;
}

if (c == '/')
{
  return 63;
}

return -1;
```

antkmsft (approved):

sdk/src/azure/core/az_base64.c
danewalton: Yea I can do it in another PR

sdk/src/azure/core/az_base64.c
```diff
@@ -370,17 +376,40 @@ AZ_NODISCARD int32_t az_base64_get_max_encoded_size(int32_t source_bytes_size)
...
+}
+
+static AZ_NODISCARD int32_t
+_az_base64_decode_four_bytes(uint8_t* encoded_bytes, _az_base64_mode mode)
 {
   int32_t i0 = *encoded_bytes;
```
LarryOsterman, line 404: Btw, this is a nit, but it might matter: When you convert from uint8_t to int32_t, [compiler will zero extend](https://docs.microsoft.com/en-us/cpp/c-language/conversions-from-unsigned-integral-types?view=msvc-170), but it's not clear that this sign extension makes sense here.

Also, I just noticed that lines 409-412 don't respect the check at line 385. Given that the check isn't respected, does it make sense to remove the check? IMHO you either check for errors or you don't check for errors. But checking for errors in one place and then ignoring them in another doesn't seem consistent.

danewalton: The idea there is that it replicates the number which the array would have return (aka `-1`). 

For the pointer conversion, is there a suggested way to resolve that? The bit shifting on line 414 and below would then cause issues with them as int8_t

LarryOsterman: Actually sign extending for bit shifting makes sense. 

What I'm mostly worried about is that the function won't return a value < 0 if any of the values i0..i3 are < 0. The value of i0 at line 409 (and i1 at 410, etc) is going to be 255 or some value < 128, but when you left shift the value and or in the other values, I don't see how that is always going to come up with a value < 0. 

I'm pretty sure that the code is working correctly (at a minimum, the test cases seem to handle this). But I don't fully understand *why* it works.

danewalton: Ahhhh I see what you're saying okay. I have another PR open to remove the decode array to simplify/save space. [I have this commit here to help guard against this scenario](https://github.com/Azure/azure-sdk-for-c/pull/2338/commits/8717171b7ce829d31e4b69b519602aa97222bf33). Would you mind having a look and let me know if that works?

PR #2331, opened 2022-09-07
Change ADU client update id to match service specification

ewertons:
The update-id reported by the ADU agent, as defined by the ADU service, is an escaped string, and has no need to be split into a separate struct with the individual components (provider, name, version). This change simplifies the ADU agent state generation by following this specification, as well as removes a bug-prone buffer manipulation from az_iot_adu_client.c.

### DIFF
--- a/sdk/inc/azure/iot/az_iot_adu_client.h
+++ b/sdk/inc/azure/iot/az_iot_adu_client.h
@@ -119,7 +119,7 @@ typedef struct
    * The version for the update.
    */
   az_span version;
-} az_iot_adu_client_update_id;
+} az_iot_adu_update_id;
 
 /**
  * @brief Holds any user-defined custom properties of the device.
@@ -179,7 +179,7 @@ typedef struct
   /**
    * An ID of the update that is currently installed.
    */
-  az_iot_adu_client_update_id update_id;
+  az_span update_id;
 } az_iot_adu_client_device_properties;
 
 /**
@@ -417,7 +417,7 @@ typedef struct
   /**
    * User-defined identity of the update manifest.
    */
-  az_iot_adu_client_update_id update_id;
+  az_iot_adu_update_id update_id;
   /**
    * Instructions of the update manifest.
    */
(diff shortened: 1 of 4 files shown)


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -119,7 +119,7 @@ typedef struct
    * The version for the update.
    */
   az_span version;
-} az_iot_adu_client_update_id;
+} az_iot_adu_update_id;
```
danewalton, line 122: Should we really drop the `_client`?

ewertons: We should, because this is to represent the update id in the update manifest. It's not specific to the client.

danewalton: already precedent with the device information

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -72,7 +72,10 @@
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_CREATED_DATE_TIME "createdDateTime"
 
 #define NULL_TERM_CHAR_SIZE 1
-#define UPDATE_ID_ESCAPING_CHARS_LENGTH 24
+
+#define RESULT_STEP_ID_PREFIX "step_"
+#define MAX_UINT32_NUMBER_OF_DIGITS 10
+#define RESULT_STEP_ID_MAX_SIZE (sizeof(RESULT_STEP_ID_PREFIX) + MAX_UINT32_NUMBER_OF_DIGITS)
```
danewalton, line 78: 1. are any of these used?
2. Suggestion:

```suggestion
#define RESULT_STEP_ID_PREFIX "step_"
#define MAX_UINT32_NUMBER_OF_DIGITS 10
#define RESULT_STEP_ID_MAX_SIZE (sizeof(RESULT_STEP_ID_PREFIX) - 1 + MAX_UINT32_NUMBER_OF_DIGITS)
```

Minus one needed?

danewalton (approved):

vaavva (approved):

sdk/src/azure/iot/az_iot_adu_client.c
ewertons: Yes, it's used to generate the results for the instructions steps in the Agent state.
Your suggestion for the -1 is good. Taken.

PR #2340, opened 2022-09-15
ADU request keeps span slice instead of copying string

danewalton:
This change doesn't require the user to pass a buffer to the ADU properties parsing. The values in the JSON are passed by value, with the ability for the user to then unescape them as needed (possibly in place to further save memory).

Ultimately, we require less from the user and they get more control.

### DIFF
--- a/sdk/src/azure/iot/az_iot_adu_client.c
+++ b/sdk/src/azure/iot/az_iot_adu_client.c
@@ -844,23 +844,3 @@ AZ_NODISCARD az_result az_iot_adu_client_parse_update_manifest(
 
   return AZ_OK;
 }
-
-/* --- az_core extensions --- */
-// static az_span split_az_span(az_span span, int32_t size, az_span* remainder)
-// {
-//   az_span result = az_span_slice(span, 0, size);
-
-//   if (remainder != NULL)
-//   {
-//     if (az_span_is_content_equal(AZ_SPAN_EMPTY, result))
-//     {
-//       *remainder = AZ_SPAN_EMPTY;
-//     }
-//     else
-//     {
-//       *remainder = az_span_slice(span, size, az_span_size(span));
-//     }
-//   }
-
-//   return result;
-// }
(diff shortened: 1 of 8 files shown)


sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -689,19 +655,22 @@ static void test_az_iot_adu_client_parse_service_properties_with_null_file_url_v
...
-      AZ_OK);
+      az_iot_adu_client_parse_service_properties(&adu_client, &reader, &request), AZ_OK);
 
   // Workflow
   assert_int_equal(request.workflow.action, workflow_action);
   assert_memory_equal(az_span_ptr(request.workflow.id), workflow_id, sizeof(workflow_id) - 1);
 
   // Update Manifest
+  int32_t out_size;
+  assert_int_equal(
```
danewalton, line 666: lets not to the in place here. Put in buffer for now until a possible az_span api instead of string/null terminated.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -953,21 +846,21 @@ AZ_NODISCARD az_result az_iot_adu_client_parse_update_manifest(
...
-    }
-    else
-    {
-      *remainder = az_span_slice(span, size, az_span_size(span));
-    }
-  }
-
-  return result;
-}
+// static az_span split_az_span(az_span span, int32_t size, az_span* remainder)
```
ericwol-msft, line 849: delete this?

ericwol-msft (approved):

sdk/tests/core/test_az_json.c
```diff
@@ -2978,6 +2981,378 @@ static void test_az_json_reader_chunked(void** state)
...
+
+    char destination[59] = { 0 };
+
+    int final_size;
+    az_result result = az_json_string_unescape(original, destination, 59, &final_size);
+
+    assert_int_equal(AZ_ERROR_UNEXPECTED_END, result);
+  }
+
+  // single char
```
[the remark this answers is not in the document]
danewalton: that's a precondition check I believe which would have to go into a different PR into main.

ericwol-msft (approved):

PR #2342, opened 2022-09-16
Move dtmi to public header

danewalton:
The parsing APIs are developed to be hardcoded for a specific version of the DTMI. If the DTMI were to change, our underlying parsing would also have to change. I see no reason why we shouldn't make this public, since users can then reference it in their own published DTMI (and not have to create this string themselves).

For example, in the middleware sample, I don't see a reason why the user should have to provide this string. We should make this available to them to use.

https://github.com/Azure-Samples/iot-middleware-freertos-samples/blob/aa3d65a6e930062319bc4802d7cc1a00d0169570/demos/sample_azure_iot_adu/sample_azure_iot_adu.c#L216-L219

### DIFF
--- a/sdk/inc/azure/iot/az_iot_adu_client.h
+++ b/sdk/inc/azure/iot/az_iot_adu_client.h
@@ -30,6 +30,11 @@
 #include <azure/core/_az_cfg_prefix.h>
 
 /**
+ * @brief  Define the ADU agent interface ID.
+ */
+#define AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID "dtmi:azure:iot:deviceUpdate;1"
+
+/**
  * @brief ADU Agent Version
  */
 #define AZ_IOT_ADU_CLIENT_AGENT_VERSION "DU;agent/0.8.0-rc1-public-preview"
--- a/sdk/src/azure/iot/az_iot_adu_client.c
+++ b/sdk/src/azure/iot/az_iot_adu_client.c
@@ -11,9 +11,6 @@
 /* Define the ADU agent component name.  */
 #define AZ_IOT_ADU_CLIENT_AGENT_COMPONENT_NAME "deviceUpdate"
 
-/* Define the ADU agent interface ID.  */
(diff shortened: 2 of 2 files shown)


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -29,6 +29,11 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
+/**
+ * @brief  Define the ADU agent interface ID.
+ */
+#define AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID "dtmi:azure:iot:deviceUpdate;1"
```
CIPop, line 35: Instead of exposing the interface ID, should we rather offer a more flexible API that would alter the `az_iot_hub_client_options`? 

That would allow us in the future to change other things and not tie in the application to this particular constant.
The new function would not alter an already existing PnP model and would only set it if the application didn't configure it before.

CIPop (approved):
Left one comment but I don't have strong feelings either way.

jspaith:
Just so I understand scenario, this is so that an application can pass in `AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID` when registering an interface ModelId, correct?  So it's a convenience to them to avoid them duplicating this since they'd need anyway?

In this case the change makes sense.  And I think not having it be configurable is best since as you said customer cannot change this to like version2 and expect anything to work since there'd need to be .c changes.

jspaith (approved):

danewalton:
> Just so I understand scenario, this is so that an application can pass in `AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID` when registering an interface ModelId, correct? So it's a convenience to them to avoid them duplicating this since they'd need anyway?
> 
> In this case the change makes sense. And I think not having it be configurable is best since as you said customer cannot change this to like version2 and expect anything to work since there'd need to be .c changes.

Perfect yes this is exactly right. The DTMI is tied closely to how we parse the document. The only way customers would want this to change is if someone wants to extend it (or have it as a subcomponent in the future), in which case people are able to use this as a base or actually have in their DTMI itself. But the core of our parsing is still based on this value.

PR #2349, opened 2022-09-27
[ADU] Update request manifest parsing to ignore Delta Updates

vaavva:
- ignore delta update fields when parsing an adu request manifest.
- add test for a request manifest that has these fields (keeping tests with manifests that _don't_ have these fields).
- combine shared code to test request manifests (since this change explicitly _doesn't_ expose or process these fields, there are no additional fields to check against, we just want to ensure that everything still works as expected).

Note: I did modify all of the tests to use request manifests that have delta update fields to verify that they all pass (they did), but the other tests don't explicitly test the code changes, so I only duplicated the relevant test.

### DIFF
--- a/sdk/src/azure/iot/az_iot_adu_client.c
+++ b/sdk/src/azure/iot/az_iot_adu_client.c
@@ -808,24 +808,24 @@ AZ_NODISCARD az_result az_iot_adu_client_parse_update_manifest(
             }
           }
           /*
-          * C SDK will not support delta updates at this time, so relatedFiles,
-          * downloadHandler, and mimeType are not exposed or processed.
-          */
+           * C SDK will not support delta updates at this time, so relatedFiles,
+           * downloadHandler, and mimeType are not exposed or processed.
+           */
           else if (az_json_token_is_text_equal(
-                  &ref_json_reader->token,
-                  AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_RELATED_FILES)))
+                       &ref_json_reader->token,
+                       AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_RELATED_FILES)))
           {
             _az_RETURN_IF_FAILED(az_json_reader_skip_children(ref_json_reader));
           }
           else if (az_json_token_is_text_equal(
-                  &ref_json_reader->token,
-                  AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_DOWNLOAD_HANDLER)))
+                       &ref_json_reader->token,
+                       AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_DOWNLOAD_HANDLER)))
           {
             _az_RETURN_IF_FAILED(az_json_reader_skip_children(ref_json_reader));
           }
           else if (az_json_token_is_text_equal(
-                  &ref_json_reader->token,
-                  AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_MIME_TYPE)))
+                       &ref_json_reader->token,
+                       AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_MIME_TYPE)))
           {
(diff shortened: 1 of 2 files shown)


sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -804,6 +807,28 @@ AZ_NODISCARD az_result az_iot_adu_client_parse_update_manifest(
...
+                  &ref_json_reader->token,
+                  AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_RELATED_FILES)))
+          {
+            _az_RETURN_IF_FAILED(az_json_reader_skip_children(ref_json_reader));
+          }
+          else if (az_json_token_is_text_equal(
+                  &ref_json_reader->token,
+                  AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_DOWNLOAD_HANDLER)))
+          {
+            _az_RETURN_IF_FAILED(az_json_reader_skip_children(ref_json_reader));
```
danewalton, line 824: I think these should all have 
```suggestion
            _az_RETURN_IF_FAILED(az_json_reader_skip_children(ref_json_reader));
            _az_RETURN_IF_FAILED(az_json_reader_next_token(ref_json_reader));
```

The skip children is a protection against the possibility of there being an array. It's a no-op if there isn't an array.

sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -890,7 +902,7 @@ test_az_iot_adu_client_parse_service_properties_payload_no_deployment_null_file_
   assert_int_equal(request.file_urls_count, 0);
 }
 
-static void test_az_iot_adu_client_parse_update_manifest_succeed(void** state)
+static void parse_update_manifest_succeed(void** state, uint8_t request_manifest[], int32_t request_manifest_size)
```
danewalton, line 905:
```suggestion
static void parse_update_manifest_succeed(uint8_t * request_manifest, int32_t request_manifest_size)
```

sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -890,7 +902,7 @@ test_az_iot_adu_client_parse_service_properties_payload_no_deployment_null_file_
...
 {
   (void)state;
```
danewalton, line 907: can remove this once the state is removed

sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -973,90 +985,19 @@ static void test_az_iot_adu_client_parse_update_manifest_succeed(void** state)
...
-  assert_int_equal(
-      az_span_size(update_manifest.files[0].hashes[0].hash_value), sizeof(files_hashes_sha) - 1);
-  assert_memory_equal(
-      az_span_ptr(update_manifest.create_date_time),
-      created_date_time,
-      sizeof(created_date_time) - 1);
-  assert_int_equal(az_span_size(update_manifest.create_date_time), sizeof(created_date_time) - 1);
+static void test_az_iot_adu_client_parse_update_manifest_payload_reverse_order_succeed(void** state)
+{
+  parse_update_manifest_succeed(state, adu_request_manifest_reverse_order, sizeof(adu_request_manifest_reverse_order));
```
danewalton, line 1000: wonderful simplification :)

sdk/src/azure/iot/az_iot_adu_client.c
vaavva: Having both causing parsing issues when it is an object - I'll take a look at what happens if it's not an object though

sdk/tests/iot/adu/test_az_iot_adu.c
vaavva: Moved back to the parent functions

sdk/src/azure/iot/az_iot_adu_client.c
vaavva: The following scenarios work with just `skip_children`, so it looks like it parses through fine since we want to ignore the contents anyways! There's a `az_json_reader_next_token` at the end of the switch statement that processes the manifest which either moves past the contents of a non-object or the closing bracket of an object.

`"relatedFiles": null` 
`"relatedFiles": "testString"`

danewalton: Ahhh I see. Checked on the skip children API and if it's a property name then it advances to the next token before skipping object/arrays.

danewalton (approved):

PR #2353, opened 2022-09-28
fix todos

danewalton:
For items not explicitly fixed:

>  // TODO: verify if this needs to be exposed as an option.

  - No we do not since this is tied to the ADU parsing. Worst case scenario, if this is ever an issue, we can add an option in the ADU client options field.
  
>  // TODO: Add quotes if result_details is not enclosed by quotes.
  - append_string adds the quotes already

### DIFF
--- a/sdk/src/azure/iot/az_iot_adu_client.c
+++ b/sdk/src/azure/iot/az_iot_adu_client.c
@@ -117,8 +117,6 @@ const uint8_t azure_iot_adu_root_key_n[385]
         0x77, 0x3e, 0x40, 0x87, 0x18, 0xc9, 0xab, 0xd9, 0xf7, 0x79 };
 const uint8_t azure_iot_adu_root_key_e[3] = { 0x01, 0x00, 0x01 };
 
-uint8_t azure_iot_step_id_scratch_buffer[7];
-
 const az_span default_compatibility_properties
     = AZ_SPAN_LITERAL_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_DEFAULT_COMPATIBILITY_PROPERTIES);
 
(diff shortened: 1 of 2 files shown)


sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -427,7 +414,9 @@ AZ_NODISCARD az_result az_iot_adu_client_parse_service_properties(
         }
         else
         {
-          // TODO: log unexpected property.
+          _az_LOG_WRITE(
+              AZ_LOG_IOT_ADU,
+              AZ_SPAN_FROM_STR("Unexpected property found in ADU manifest workflow"));
```
vaavva, line 419: Should we log what the unexpected property was as well here?

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -116,6 +117,8 @@ const uint8_t azure_iot_adu_root_key_n[385]
         0x77, 0x3e, 0x40, 0x87, 0x18, 0xc9, 0xab, 0xd9, 0xf7, 0x79 };
 const uint8_t azure_iot_adu_root_key_e[3] = { 0x01, 0x00, 0x01 };
 
+uint8_t azure_iot_step_id_scratch_buffer[7];
```
CIPop, line 120: This won't work if multi-threaded systems. If memory is required, we need to either:
- allocate from the stack (7 bytes is not a lot).
- require the app to pass it into the APIs that need it
- require the app to pass it into the client / handle initializer.

CIPop (changes requested):
Left one comment about internal memory allocation.

ericwol-msft (approved):

CIPop (approved):
Thanks @danewalton !

vaavva (approved):

PR #2354, opened 2022-09-29
distinguish between model id and interface id

danewalton:


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -30,9 +30,9 @@
 #include <azure/core/_az_cfg_prefix.h>
 
 /**
- * @brief  Define the ADU agent interface ID.
+ * @brief  Define the ADU agent model ID.
  */
-#define AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID "dtmi:azure:iot:deviceUpdateModel;1"
+#define AZ_IOT_ADU_CLIENT_AGENT_MODEL_ID "dtmi:azure:iot:deviceUpdateModel;1"
```
vaavva, line 35: Is this being used somewhere?

danewalton: No not here but would be in samples which want to use ADU. [Link here for docs.](https://learn.microsoft.com/en-us/azure/iot-hub-device-update/device-update-plug-and-play#model-id)

Copy important part here:

> Device Update for IoT Hub requires the IoT Plug and Play smart device to announce a model ID with a value of "dtmi:azure:iot:deviceUpdateModel;1" as part of the device connection.

wduraes (approved):

PR #2371, opened 2022-10-27
Add Azure Device Update APIs

danewalton:


danewalton:
closing to merge main first so we don't have files that looked changed

PR #2372, opened 2022-10-27
Fix minimum log buffer size calculation

### LINKED ISSUE #2224: Incorrect Calculation  _az_http_policy_logging_append_http_request_msg()
File : sdk\src\azure\core\az_http_policy_logging.c
• API : _az_http_policy_logging_append_http_request_msg( )
• Line : 73
• Issue : Incorrect calculation of required_length.
• The value needs to be incremented, not reassigned.
• Required length is being calculated using four values:
• “HTTP Request : ” 15 in length (http_request_string)
• HTTP Method used (_internal.method)
• Space
• Actual length of URL (_internal.url_length)
The length of http_request_string is missing from the calculation.
• Effect : Machine goes to Infinite Loop.
• **Recommendation: Modify “=” to “+=” in L:73**

image

antkmsft:
Closes #2224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bugs Fixed
 
+[[#2372](https://github.com/Azure/azure-sdk-for-c/pull/2372)] Incorrect minimum buffer size calculation when logging an HTTP request.
+
 ### Other Changes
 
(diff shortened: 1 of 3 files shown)


ahsonkhan (approved):
CL entry?

PR #2375, opened 2022-10-27
Add Azure Device Update code to main

danewalton:


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+
+/**
+ * @brief  Maximum Number of Files Handled by this ADU Agent (Steps)
+ */
+#define AZ_IOT_ADU_CLIENT_MAX_INSTRUCTIONS_STEPS 10
+
+/**
+ * @brief  Maximum Number of Files Handled by this ADU Agent (File Hashes)
+ */
+#define AZ_IOT_ADU_CLIENT_MAX_FILE_HASH_COUNT 2
```
CIPop, line 95: Should this be the same as the URL count?

sdk/inc/azure/core/az_span.h
```diff
@@ -76,17 +76,24 @@ AZ_NODISCARD az_span az_span_create(uint8_t* ptr, int32_t size);
 #endif // AZ_NO_PRECONDITION_CHECKING
 
 /**
- * @brief An empty #az_span.
+ * @brief An empty #az_span literal.
```
CIPop, line 79: (adding the comment again - not sure why it's not listed).

I don't think we should add another public API for AZ_SPAN_EMPTY. We should instead use the existing API.

Even if we do decide to add the new API, please separate the PR / discussion.

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include <azure/core/_az_cfg_prefix.h>
+
+/**
+ * @brief  Define the ADU agent model ID.
+ */
+#define AZ_IOT_ADU_CLIENT_AGENT_MODEL_ID "dtmi:azure:iot:deviceUpdateModel;1"
```
CIPop, line 35: I assume these will change in time. Can we instead create public API functions instead of `#define`s ?
I would instead define _AGENT_MODEL_DEVICEUPDATE_MODEL_V1 (internally) then expose a public API function.

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+
+/**
+ * @brief ADU Agent Version
+ */
+#define AZ_IOT_ADU_CLIENT_AGENT_VERSION "DU;agent/0.8.0-rc1-public-preview"
```
CIPop: Same as above if we want this public. Having a function will indicate that this value will be changing in the future.

CIPop: Should the agent version still be 0.8.0 RC1 public preview or a GA version?
/cc @wduraes

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+#define AZ_IOT_ADU_CLIENT_AGENT_DEFAULT_COMPATIBILITY_PROPERTIES "manufacturer,model"
+
+/* The following key is used to validate the Azure Device Update update manifest signature */
+/* For more details, please see
+ * https://docs.microsoft.com/azure/iot-hub-device-update/device-update-security */
+
+/**
+ * @brief The root key id used to identify the key.
+ */
+extern const uint8_t azure_iot_adu_root_key_id[13];
```
CIPop, line 114: I don't think these should be part of the public API. Instead, we should allow parameters passed to ADU.

Shouldn't this API support a _set_ of keys? https://learn.microsoft.com/en-us/azure/iot-hub-device-update/device-update-security#root-keys

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+   *         the delivery optimization agent. If unused, set to #AZ_SPAN_EMPTY.
+   *
+   * https://docs.microsoft.com/azure/iot-hub-device-update/device-update-plug-and-play#device-properties
+   */
+  az_span delivery_optimization_agent_version;
+  /**
+   * An ID of the update that is currently installed.
+   */
+  az_span update_id;
+} az_iot_adu_client_device_properties;
```
CIPop, line 207: I think this is large enough to warrant initialization APIs. That way, if the app doesn't provide some of the values, the constructor will be able to fill-in defaults.

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+
+/**
+ * @brief The update result reported by the agent.
+ *
+ * This details the result for the overall update.
+ */
+typedef struct
+{
+  /**
+   * A code that contains information about the result of the last update action.
```
CIPop, line 241: I think this should be based on one of the enums above.

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+ */
+typedef struct
+{
+  /**
+   * An integer that corresponds to an action the agent should perform.
+   * @remark Refer to the following defines for the expected values:
+   *         AZ_IOT_ADU_CLIENT_SERVICE_ACTION_APPLY_DEPLOYMENT
+   *         AZ_IOT_ADU_CLIENT_SERVICE_ACTION_CANCEL
+   */
+  int32_t action;
```
CIPop, line 278: enum instead of #defines

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+ * @brief User-defined properties for handling an update request.
+ *
+ */
+typedef struct
+{
+  /**
+   * The installed criteria which defines a successful installation.
+   */
+  az_span installed_criteria;
+} az_iot_adu_client_update_manifest_instructions_step_handler_properties;
```
CIPop, line 347: Can this be removed and directly define `handler_properties` as an az_span below?

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -0,0 +1,620 @@
...
+   */
+  az_span files[AZ_IOT_ADU_CLIENT_MAX_FILE_URL_COUNT];
+  /**
+   * Number of items in \p files.
+   */
+  uint32_t files_count;
+  /**
+   * Additional user-defined properties for the update step handler.
+   */
+  az_iot_adu_client_update_manifest_instructions_step_handler_properties handler_properties;
```
CIPop, line 370: See above: `az_span` type instead of above type?

sdk/samples/iot/paho_iot_adu_sample_common.c
```diff
@@ -0,0 +1,663 @@
...
+    if (az_iot_adu_is_component_device_update(component_name))
+    {
+      // ADU Component
+      rc = az_iot_adu_parse_service_properties(
+          &jr, xScratchBufferSpan, &xBaseUpdateRequest, &xScratchBufferSpan);
+
+      if (az_result_failed(rc))
+      {
+        IOT_SAMPLE_LOG_ERROR("az_iot_adu_parse_service_properties failed: 0x%08x.", rc);
+        /* TODO: return individualized/specific errors. */
```
CIPop, line 525: Please address the TODOs - these should not be in or should be tracked by a GitHub issue.

sdk/samples/iot/paho_iot_adu_sample_common.c
```diff
@@ -0,0 +1,663 @@
...
+      }
+      else
+      {
+        rc = az_iot_adu_parse_update_manifest(
+            xBaseUpdateRequest.update_manifest, &xBaseUpdateManifest);
+
+        if (az_result_failed(rc))
+        {
+          IOT_SAMPLE_LOG_ERROR("az_iot_adu_parse_update_manifest failed: 0x%08x", rc);
+          /* TODO: return individualized/specific errors. */
```
CIPop, line 536: Please address the TODOs - these should not be in or should be tracked by a GitHub issue.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,872 @@
...
+    return AZ_ERROR_JSON_INVALID_STATE;                        \
+  }
+
+#define RETURN_IF_JSON_TOKEN_NOT_TEXT(jr_ptr, literal_text)                         \
+  if (!az_json_token_is_text_equal(&jr_ptr->token, AZ_SPAN_FROM_STR(literal_text))) \
+  {                                                                                 \
+    return AZ_ERROR_JSON_INVALID_STATE;                                             \
+  }
+
+const uint8_t azure_iot_adu_root_key_id[13] = "ADU.200702.R";
```
CIPop, line 95: Please remove from the SDK - application should be able to pass this in.
A set of keys is always required - we should understand what the keys are today and add them to our samples.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,872 @@
...
+{
+  _az_PRECONDITION_NOT_NULL(client);
+
+  (void)client;
+
+  return az_span_is_content_equal(
+      AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_COMPONENT_NAME), component_name);
+}
+
+static az_result generate_step_id(az_span buffer, uint32_t step_index, az_span* step_id)
```
CIPop, line 156: Naming: statics should always start with `_`.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,872 @@
...
+
+  _az_RETURN_IF_FAILED(
+      az_iot_hub_client_properties_writer_end_response_status(NULL, ref_json_writer));
+  _az_RETURN_IF_FAILED(az_iot_hub_client_properties_writer_end_component(NULL, ref_json_writer));
+  _az_RETURN_IF_FAILED(az_json_writer_append_end_object(ref_json_writer));
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_iot_adu_client_parse_update_manifest(
```
CIPop, line 534: Please split this function into smaller (inline?) functions to reduce complexity and increase readability.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,872 @@
...
+AZ_NODISCARD az_result az_iot_adu_client_parse_service_properties(
+    az_iot_adu_client* client,
+    az_json_reader* ref_json_reader,
+    az_iot_adu_client_update_request* update_request)
+{
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_NOT_NULL(ref_json_reader);
+  _az_PRECONDITION_NOT_NULL(update_request);
+
+  (void)client;
```
CIPop, line 372: Almost all functions seem to be related to a JSON reader.
Looks like it may be a good idea to move the reader as a private member of the `client` instead of a parameter to the function.

sdk/inc/azure/core/az_span.h
```diff
@@ -76,17 +76,24 @@ AZ_NODISCARD az_span az_span_create(uint8_t* ptr, int32_t size);
...
  *
  * @remark There is no guarantee that the pointer backing this span will be `NULL` and the caller
  * shouldn't rely on it. However, the size will be 0.
  */
-#define AZ_SPAN_EMPTY                      \
-  (az_span)                                \
+#define AZ_SPAN_LITERAL_EMPTY              \
```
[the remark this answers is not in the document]
danewalton: Good callout. I believe we needed this when we had nested literal structs but we have since changed some fields to no longer need this. Let me double check in this.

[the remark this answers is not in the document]
danewalton: I stand corrected. This is needed for the same reason we have `AZ_SPAN_LITERAL_FROM_STR()`

image

It will complain that the value is not constant since it has a cast to an `az_span`.

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: no this is the number of allowed hashes per file (url). In case there is a hash that is not just sha256

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: this has been updated to GA

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: We can't (or shouldn't). The user has to initialize the JSON reader, which allows Azure RTOS to initialize it with discontinuous memory.

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: image

As is formatted in the JSON, this is an object which could theoretically have more values. We attempt to model our structs based on the JSON structure.

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: This I believe we discussed before. We would advance this with a new version as the parsing would change. cc @ewertons | @vaavva for thoughts

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: these are usually statically allocated and remain unchanged for the device.

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: For time, not going to do this right now but I'll leave this open so that we may do this later.

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: Just to note, this follows the precedent already set with PnP APIs

CIPop (approved):

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: https://github.com/Azure/azure-sdk-for-c/pull/2392

wduraes (approved):

vaavva (approved):

ericwol-msft (approved):

ericwol-msft (approved):

ericwol-msft (approved):

sdk/inc/azure/iot/internal/az_iot_adu_internal.h
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/*
+ * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
+ * prefixed with an underscore ('_') directly in your application code. These symbols
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
+ */
```
ahsonkhan, line 9: Since it is an internal file, consider this a nit (but add a double star to start, along with `@file` and a `@brief` for consistency with all other files in the repo):
```suggestion
/**
 * @file
 *
 * @brief Defines internal macros used by the ADU client.
 *
 * @note You MUST NOT use any symbols (macros, functions, structures, enums, etc.)
 * prefixed with an underscore ('_') directly in your application code. These symbols
 * are part of Azure SDK's internal implementation; we do not document these symbols
 * and they are subject to change in future versions of the SDK which would break your code.
 */
```

sdk/src/azure/iot/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ add_library (az_iot_hub
   ${CMAKE_CURRENT_LIST_DIR}/az_iot_hub_client_methods.c
   ${CMAKE_CURRENT_LIST_DIR}/az_iot_hub_client_commands.c
   ${CMAKE_CURRENT_LIST_DIR}/az_iot_hub_client_properties.c
+  ${CMAKE_CURRENT_LIST_DIR}/az_iot_adu_client.c
```
ahsonkhan, line 39: nit: sort in alphabetical order

sdk/src/azure/iot/az_iot_hub_client_properties.c
```diff
@@ -65,7 +65,6 @@ AZ_NODISCARD az_result az_iot_hub_client_properties_writer_begin_component(
     az_json_writer* ref_json_writer,
     az_span component_name)
 {
-  _az_PRECONDITION_NOT_NULL(client);
```
ahsonkhan, line 68: What motivated weakening this precondition?

danewalton: The ADU APIs need to use it without passing a hub client (since the adu client doesn't have access to it)

sdk/inc/azure/iot/internal/az_iot_adu_internal.h
danewalton: https://github.com/Azure/azure-sdk-for-c/pull/2414

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+#include <azure/core/internal/az_precondition_internal.h>
+#include <azure/core/internal/az_result_internal.h>
+#include <stdio.h>
+
+/* Define the ADU agent component name.  */
+#define AZ_IOT_ADU_CLIENT_AGENT_COMPONENT_NAME "deviceUpdate"
+
+#define AZ_IOT_ADU_CLIENT_AGENT_INTERFACE_ID "dtmi:azure:iot:deviceUpdate;1"
+
+/* Define the ADU agent property name "agent" and sub property names.  */
```
ahsonkhan, line 17: When do you decide some `#define` macros belong in the public header file and are publicly accessible, vs when they are "private implementation details" like this, and only exist in the source file?

We use `_az_` for such non-public `#defines`:
https://github.com/Azure/azure-sdk-for-c/blob/31e832e14a8e77f843639a02eca79de95309a454/sdk/src/azure/core/az_span.c#L22

https://github.com/Azure/azure-sdk-for-c/blob/5ad7985d50af8873c6148f867f57c487e294fb28/sdk/src/azure/core/az_http_policy.c#L47-L50

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+    az_iot_adu_client* client,
+    az_span component_name)
+{
+  _az_PRECONDITION_NOT_NULL(client);
+
+  (void)client;
+
+  return az_span_is_content_equal(
+      AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_COMPONENT_NAME), component_name);
+}
```
ahsonkhan, line 135: Do we really need this API at all? It seems very much a simple convenience helper. Is this only to avoid having to make this `AZ_IOT_ADU_CLIENT_AGENT_COMPONENT_NAME` string value public?

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: I guess that's fair. Those could be `_az` since they aren't used by the developer. Only internal to the `.c` file

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+static az_result _generate_step_id(az_span buffer, uint32_t step_index, az_span* step_id)
+{
+  az_result result;
+  *step_id = buffer;
+  buffer = az_span_copy(buffer, AZ_SPAN_FROM_STR(RESULT_STEP_ID_PREFIX));
+
+  result = az_span_u32toa(buffer, step_index, &buffer);
+  _az_RETURN_IF_FAILED(result);
+
+  *step_id = az_span_slice(*step_id, 0, az_span_size(*step_id) - az_span_size(buffer));
```
ahsonkhan, line 146: nit: you could consider using `_az_span_diff()` here

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: Yes it's mainly to hide the inner workings of the component name.

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: noted. For now we'll leave it since the resulting behavior should be the same.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+      _az_RETURN_IF_FAILED(az_json_writer_append_property_name(
+          ref_json_writer, AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_RESULT_DETAILS)));
+      _az_RETURN_IF_FAILED(
+          az_json_writer_append_string(ref_json_writer, last_install_result->result_details));
+    }
+
+    for (int32_t i = 0; i < last_install_result->step_results_count; i++)
+    {
+      az_span step_id = AZ_SPAN_FROM_BUFFER(step_id_scratch_buffer);
+      _az_RETURN_IF_FAILED(_generate_step_id(step_id, (uint32_t)i, &step_id));
```
ahsonkhan, line 266: nit:
It doesn't seem necessary for `_generate_step_id` to accept `uint32_t` as an input parameter (and hence this cast). You could just pass the `int32_t` directly and use `az_span_u32toa`. Unless there is something I am missing about uint8_t or the requirement of this being unsigned.

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+    }
+    else if (az_json_token_is_text_equal(
+                 &ref_json_reader->token,
+                 AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_UPDATE_MANIFEST)))
+    {
+      _az_RETURN_IF_FAILED(az_json_reader_next_token(ref_json_reader));
+
+      if (ref_json_reader->token.kind != AZ_JSON_TOKEN_NULL)
+      {
+        update_request->update_manifest = ref_json_reader->token.slice;
```
ahsonkhan, line 428: And if it is null ,we leave the `update_manifest` as-is, unchanged?

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+      ref_json_writer,
+      AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_SERVICE),
+      (int32_t)status,
+      version,
+      AZ_SPAN_EMPTY));
+
+  // It is not necessary to send the properties back in the acknowledgement.
+  // We opt not to send them to reduce the size of the payload.
+  _az_RETURN_IF_FAILED(az_json_writer_append_begin_object(ref_json_writer));
+  _az_RETURN_IF_FAILED(az_json_writer_append_end_object(ref_json_writer));
```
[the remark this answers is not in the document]
ahsonkhan: Is it the same reason as here?
https://github.com/Azure/azure-sdk-for-c/blob/6d3efa09ac2502c605039186192019aec966ee6e/sdk/src/azure/iot/az_iot_adu_client.c#L731-L736

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -0,0 +1,871 @@
...
+                  = ref_json_reader->token.slice;
+
+              update_manifest->files[files_index].hashes_count++;
+
+              _az_RETURN_IF_FAILED(az_json_reader_next_token(ref_json_reader));
+            }
+          }
+          /*
+           * C SDK will not support delta updates at this time, so relatedFiles,
+           * downloadHandler, and mimeType are not exposed or processed.
```
ahsonkhan, line 815: Should this be a remark in the API docs?

ahsonkhan (approved):
Didn't dive too deep into each of the implementation details, but overall, looks good to me!

sdk/src/azure/iot/az_iot_adu_client.c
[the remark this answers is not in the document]
danewalton: no it's just a mechanic of plug and play to acknowledge the properties. They can view the twin and see what was acknowledged in the "desired" field

sdk/src/azure/iot/az_iot_adu_client.c
ahsonkhan: Since it is just naming, as part of non-breaking implementation detail, it can be addressed later :)

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: That's a good question. We should probably set this and the signature to `AZ_SPAN_NULL`. I'll make a note and update in a subsequent update.

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: noted

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: Good point. I'll add this since it's documentation updates only.

danewalton: https://github.com/Azure/azure-sdk-for-c/pull/2421

ahsonkhan:
The line coverage for ADU is pretty great. For branch coverage though, other IoT code has set a much higher bar (80%+) :)
image

Consider adding some edge case tests *in the future* to improve code coverage (file an issue to track it). At a quick glance, many of the uncovered branches aren't functionally relevant at the moment, or are about invalid/incomplete JSON reading/writing:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1980826&view=codecoverage-tab
image
image

There are a couple places that could benefit from unit tests though, unless those branches are unreachable:
image

danewalton:
@ahsonkhan We can definitely look to get some of those numbers up. Branches might be tough because of the complexity of the JSON but we can do the simpler ones for sure.

PR #2377, opened 2022-10-28
Fix zlib 1.2.12 vcpkg failure

antkmsft:
Unfortunately, we can't easily pick same SHA that we picked for updating the C++ SDK (https://github.com/Azure/azure-sdk-for-cpp/pull/4036), because `paho-mqtt` has a breaking change at that commit SHA (`fatal error C1083: Cannot open include file: 'paho-mqtt/MQTTClient.h': No such file or directory`).

So, `overrides` if not the final step, is definitely the first fix.

### DIFF
--- a/cmake-modules/AzureVcpkg.cmake
+++ b/cmake-modules/AzureVcpkg.cmake
@@ -21,7 +21,7 @@ macro(az_vcpkg_integrate)
       if(NOT DEFINED ENV{AZURE_SDK_DISABLE_AUTO_VCPKG})
         # GET VCPKG FROM SOURCE
         #  User can set env var AZURE_SDK_VCPKG_COMMIT to pick the VCPKG commit to fetch
-        set(VCPKG_COMMIT_STRING f0aa678b7471497f1adedcc99f40e1599ad22f69) # default SDK tested commit
+        set(VCPKG_COMMIT_STRING 6ca56aeb457f033d344a7106cb3f9f1abf8f4e98) # default SDK tested commit
         if(DEFINED ENV{AZURE_SDK_VCPKG_COMMIT})
           set(VCPKG_COMMIT_STRING "$ENV{AZURE_SDK_VCPKG_COMMIT}") # default SDK tested commit
         endif()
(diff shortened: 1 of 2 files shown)


ahsonkhan:
How come CI has been passing even without this change over the last few weeks?

It makes me wonder if we actually need this change in Embedded C, or if we aren't building the expected configuration in CI to hit the issue.

ahsonkhan:
> because `paho-mqtt` has a breaking change at that commit SHA (`fatal error C1083: Cannot open include file: 'paho-mqtt/MQTTClient.h': No such file or directory`).

FYI @danewalton, @CIPop - luckily this should only impact samples

ahsonkhan (approved):
Looks good (as long as it is actually needed to build the SDK, following our steps/CI to pass).

P.S. I just noticed our SDK version in the vcpkg.json file is out-dated (1.0.0), but we don't actually ship on vcpkg, but rather use it to get our dependencies, so it isn't crucial to keep in sync.

antkmsft:
@ahsonkhan, we do also have versions in the C++ sdk, but we have two sets of files - first are the ones that are used for vcpkg export. Those should have version updated. But they keep it updated by using a placeholder, and then the script inserts the release version. Second set of `vcpkg.json` files is for "vcpkg manifest mode" - and there, version number is not important at all.

LarryOsterman:
Do we still need this change given that it appears that the fix is available in vcpkg?

ahsonkhan:
That's what I am wondering too. @antkmsft do we actually need this change?

antkmsft:
Yes, we need the fix, because baseline is at `f0aa678b7471497f1adedcc99f40e1599ad22f69` vcpkg commit:
https://github.com/Azure/azure-sdk-for-c/blob/cd62e2b78a396c3d5478d3b60c046adb72687274/vcpkg.json#L4

And at that commit, `1.2.12#2` (the version uses link that does not 404 + security patch) or `1.2.13` does not exist.

LarryOsterman:
Why not move C to a newer baseline? That baseline is back in April.

antkmsft:
@LarryOsterman, one of the dependencies, `paho-mqtt` appears to have a breaking change at that SHA :(

LarryOsterman:
What is the plan to move the baseline forward? We can't remain stuck on a 6 month old package baseline for all packages.

CIPop:
> > because `paho-mqtt` has a breaking change at that commit SHA (`fatal error C1083: Cannot open include file: 'paho-mqtt/MQTTClient.h': No such file or directory`).
> 
> FYI @danewalton, @CIPop - luckily this should only impact samples

Was this already fixed? They've just changed the include paths (which was incorrect initially). 
I have all of these fixed in my investigation branch here: https://github.com/CIPop/azure-sdk-for-c/tree/hfsm using a VCPKG from sometime in June/July.

PR #2388, opened 2022-11-03
remove adu root keys

danewalton:
- Removes root keys from API. If we want to put them in later we can.


sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -104,25 +104,6 @@
  */
 #define AZ_IOT_ADU_CLIENT_AGENT_DEFAULT_COMPATIBILITY_PROPERTIES "manufacturer,model"
 
-/* The following key is used to validate the Azure Device Update update manifest signature */
```
vaavva, line 107: Do we need to add some documentation for the user on how to add the root keys themselves? (both for this PR and the other one) Will this break ADU functionality if they don't?

danewalton: Good question. The keys will be added to the samples (and places where we test ADU functionality). There are currently no places where customers can get the keys publicly from ADU, so we will have to point customers to the samples for now while we talk with the ADU folks to hopefully get an online place for them.

ewertons: So how are we going to point the customers to the samples? Are you thinking about adding a comment to docs and headers here pointing customers to?

CIPop (approved):

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: Yes for the time being, once those samples are updated, we can add a link to the samples (maybe Arduino) for the root keys. We should also reach out to the ADU folks to release public documentation on the JWS validation as well since that is also private at the moment.

PR #2391, opened 2022-11-03
Typedef the enums

danewalton:


vaavva (approved):

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -287,7 +287,7 @@ AZ_NODISCARD az_result az_iot_adu_client_get_agent_state_payload(
   /* Fill the agent state.   */
   _az_RETURN_IF_FAILED(az_json_writer_append_property_name(
       ref_json_writer, AZ_SPAN_FROM_STR(AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_STATE)));
-  _az_RETURN_IF_FAILED(az_json_writer_append_int32(ref_json_writer, agent_state));
+  _az_RETURN_IF_FAILED(az_json_writer_append_int32(ref_json_writer, (int32_t)agent_state));
```
ewertons, line 290: Ok, so I think this will preserve whatever value the service is sending (e.g., 4). Then it's up to the customer to verify if the value matches any of the expected enum values.

ewertons (approved):

sdk/src/azure/iot/az_iot_adu_client.c
danewalton: Yup correct 👍

PR #2392, opened 2022-11-03
add initializer for device props

danewalton:


sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -457,6 +453,20 @@ static uint8_t file_url_contoso[]
...
 
+static int setup(void** state)
+{
+  (void)state;
+
+  adu_device_properties = az_iot_adu_client_device_properties_default();
+  adu_device_properties.manufacturer = AZ_SPAN_FROM_STR(TEST_ADU_DEVICE_MANUFACTURER);
+  adu_device_properties.model = AZ_SPAN_FROM_STR(TEST_ADU_DEVICE_MODEL);
+  adu_device_properties.adu_version = AZ_SPAN_FROM_STR(TEST_AZ_IOT_ADU_CLIENT_AGENT_VERSION);
+  adu_device_properties.delivery_optimization_agent_version = AZ_SPAN_EMPTY;
```
vaavva, line 464: Should this be AZ_SPAN_LITERAL_EMPTY? Looks like that's what it used to be and what we set it to in the device_properties_default

vaavva (approved):

sdk/tests/iot/adu/test_az_iot_adu.c
danewalton: The `_LITERAL` variants of span initializers are used for setting values in struct literals. So anything with some 
```c
my_struct = { .myvalue = blah };
```
kind of initialization.

For assigning individual values of a struct, we can set using the non-literal version which adds a cast to that value.

sdk/inc/azure/iot/az_iot_adu_client.h
```diff
@@ -455,6 +459,15 @@ typedef struct
...
 AZ_NODISCARD az_iot_adu_client_options az_iot_adu_client_options_default();
 
+/**
+ * @brief Gets the default #az_iot_adu_client_device_properties.
+ * @details Call this to obtain an initialized #az_iot_adu_client_device_properties structure that
+ * can be afterwards modified and passed to necessary APIs.
+ *
+ * @return #az_iot_adu_client_device_properties.
+ */
+AZ_NODISCARD az_iot_adu_client_device_properties az_iot_adu_client_device_properties_default();
```
CIPop, line 469: Nit - since this is not an "options" it should be named `az_iot_adu_client_device_properties_init` similar to `az_iot_message_properties_init`.

CIPop (approved):

sdk/inc/azure/iot/az_iot_adu_client.h
danewalton: I think here it should be `_default()`. The `_init()` is used to initialize a struct with parameters/values which from then on become private fields of the initialized struct. The interactions on that initialized struct are more or less as though it were an object.

Here, we're giving people a preconfigured default struct which they then directly act on. Thoughts?

ericwol-msft (approved):

PR #2395, opened 2022-11-04
remove config | move to internal

danewalton:


sdk/inc/azure/iot/internal/az_iot_adu_internal.h
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ewertons, line 1: From one side this should be named az_iot_adu_config_internal.h, but on the other naming it like this will make it harder for customers to find and try to change this file.
But they could still do it anyway.
Should we name this file az_iot_adu_config_internal.h ?

danewalton: I think refraining from naming anything "config" for now is reasonable. These aren't meant to be configurable right now. When they are, we can give it the "config" wording.

CIPop (approved):
Thanks @danewalton !

PR #2397, opened 2022-11-07
Change enums to present tense

danewalton:


vaavva (approved):

CIPop (approved):

ahsonkhan (commented):
What motivated this change? Are there any service-side docs that set precedence or showcase present tense being more readable?

danewalton:
> What motivated this change? Are there any service-side docs that set precedence or showcase present tense being more readable?

No it's giving words to a number value. In this case, because the device is signaling "I'd like the accept/reject this update", it's more of an active directive rather than passive. It has to signal it rejects or accepts before it's settled as such.

ericwol-msft (approved):

PR #2412, opened 2022-11-09
Update az_json_string_unescape to accept az_span as destination to improve usability of API.

ahsonkhan:
Fixes: https://github.com/Azure/azure-sdk-for-c/issues/2302

Addressing usability feedback from IoT and simplifying the API signature.

This gives flexibility to the caller and allows the user, if they really choose to, to unescape in-place. Because we are returning spans, the destination doesn't need to be null terminated, and hence extra buffer space isn't required. Added remarks within the docs so the user is aware of the implications of doing in-place unescape.

I decided to stick with the verb "unescape" instead of "decode" in the function name since the RFC uses that term and cursory searches on the web show folks seem to refer to the act of transforming the JSON string as "unescape" more so than decode (which seems to be overloaded with deserialize sometimes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,15 @@
 
 ### Features Added
 
-[[#2329](https://github.com/Azure/azure-sdk-for-c/pull/2329)] Add Base64 URL decoder.
+- [[#2329](https://github.com/Azure/azure-sdk-for-c/pull/2329)] Add Base64 URL decoder.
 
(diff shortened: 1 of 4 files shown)


CIPop (approved):
Thanks @ahsonkhan !

danewalton (approved):
Usage in adu looks clean
https://github.com/Azure/azure-sdk-for-c/pull/2413

PR #2413, opened 2022-11-09
Merge main | Use new unescape API

danewalton:
- Merge main changes
- Add one commit to update ADU tests to use new json unescape API which do it in place.


danewalton:
**ONLY MERGE WITHOUT SQUASH**

sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -814,12 +817,8 @@ static void test_az_iot_adu_client_parse_service_properties_succeed(void** state
...
 
   // Update Manifest
-  int32_t out_size;
-  assert_int_equal(
-      az_json_string_unescape(
-          request.update_manifest, (char*)scratch_buffer, sizeof(scratch_buffer), &out_size),
-      AZ_OK);
-  request.update_manifest = az_span_create(scratch_buffer, out_size);
+  request.update_manifest
+      = az_json_string_unescape(request.update_manifest, request.update_manifest);
```
vaavva, line 821: Why are we sending the same parameter twice? It looks like we do that in every place we're using this except the tests

danewalton: The signature for the unescape is 
```c
<az_span with resulting buffer and size> az_json_string_unescape(
                   <input to be unescaped>,
                   <output buffer where to put the unescaped string>);
```
So in this case, because we're doing it in place, the input is the same as the output.

vaavva: oh okay, so the function works for in place or not in place unescaping - that makes sense!

danewalton: Yup makes it flexible!

CIPop (approved):
Assuming everything outside the "fix adu test" was just picked from main for integration (not reviewed independently).

Thanks @danewalton!

ewertons (approved):

PR #2437, opened 2022-11-30
Update ADU device props to GA fields

danewalton:


sdk/tests/iot/adu/test_az_iot_adu.c
```diff
@@ -29,14 +29,16 @@
 
 static uint8_t expected_agent_state_payload[]
```
vaavva, line 30: Do we need to update the manifest version in these tests as well for this change?

danewalton: Good question. We can. The tests are just parsing out a number so it realistically could be `10000` and it wouldn't matter. As long as the format of the payload is correct.

danewalton:
Verified with @ewertons that the only other change is the manifest version going from 4 to 5. Updated it in the unit tests.

vaavva (approved):

ewertons (commented):
Approved

sdk/src/azure/iot/az_iot_adu_client.c
```diff
@@ -12,15 +12,15 @@
...
+#define AZ_IOT_ADU_CLIENT_AGENT_CONTRACT_MODEL_ID "dtmi:azure:iot:deviceUpdateContractModel;2"
 
 /* Define the ADU agent property name "agent" and sub property names.  */
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_AGENT "agent"
 
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_DEVICEPROPERTIES "deviceProperties"
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_MANUFACTURER "manufacturer"
 #define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_MODEL "model"
-#define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_INTERFACE_ID "interfaceId"
+#define AZ_IOT_ADU_CLIENT_AGENT_PROPERTY_NAME_CONTRACT_MODEL_ID "contractModelId"
```
ahsonkhan, line 23: Add a bug fix Changelog entry?

danewalton: done
https://github.com/Azure/azure-sdk-for-c/pull/2451

Issue #2310, opened 2022-08-10
Improve precondition testing az_json_string_unescape()


ahsonkhan:
From https://github.com/Azure/azure-sdk-for-c/pull/2303#discussion_r941910653

See https://dev.azure.com/azure-sdk/public/_build/results?buildId=1769797&view=codecoverage-tab

Issue #1836, opened 2021-07-14
Improve branch coverage testing for base 64, specifically around preconditions


ahsonkhan:
We are at ~80% branch coverage, which could be improved to get closer to ~100%
![image](https://user-images.githubusercontent.com/6527137/125548276-6871ce9f-050b-40f6-8c1e-718b38744b3f.png)

https://dev.azure.com/azure-sdk/public/_build/results?buildId=990064&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab
![image](https://user-images.githubusercontent.com/6527137/125548410-f1db5a49-282c-4608-b9a9-d5bc19414368.png)

https://github.com/Azure/azure-sdk-for-c/blob/d9ef1bf5c5bdc775fcf5a86f144d27a272054a49/sdk/src/azure/core/az_base64.c#L319-L324

See examples on how to add tests for preconditions here:
https://github.com/Azure/azure-sdk-for-c/blob/9f5a6334863ebbf214c6e2fbd5d6e3fce52fe0de/sdk/tests/iot/hub/test_az_iot_hub_client_c2d.c#L41-L54

PR #2443, opened 2022-12-06
Azure IoT Provisioning Client Receiving Custom Allocation Payloads

CIPop:
Port from https://github.com/Azure/azure-sdk-for-c/tree/feature/iot-certificate-management.

- Feature: Azure IoT Provisioning Client Receiving Custom Allocation Payloads
- Includes unit tests, sample and documentation.
- Adding standardized API deprecation.
- Forcing clang v9 for verification.
- Roll-back from feature branch: member function `options` parameter cannot be `NULL`. (https://github.com/Azure/azure-sdk-for-c/pull/2188#discussion_r860141070)

**Deprecation notice:** `az_iot_provisioning_client_get_request_payload()` is deprecated in favor of `az_iot_provisioning_client_register_get_request_payload()`. The new API _requires_ an initialized `az_iot_provisioning_client_payload_options` structure (using `az_iot_provisioning_client_payload_options_default()`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,18 @@
 
 ### Features Added
 
+- Adding IoT Device Provisioning support for [Custom Allocation Payload](https://docs.microsoft.com/azure/iot-dps/how-to-send-additional-data).
+  - New APIs: `az_iot_provisioning_client_register_get_request_payload()` and  `az_iot_provisioning_client_registration_state.payload`.
+
 ### Breaking Changes
 
 ### Bugs Fixed
 
+- Azure IoT Samples now display errors following the _source file:line number_ format, allowing Visual Studio Code to automatically create links from the output log.
+
 ### Other Changes
+
+- Azure IoT Device Provisioning Client API __deprecation__: `az_iot_provisioning_client_get_request_payload()` is deprecated in favor of `az_iot_provisioning_client_register_get_request_payload()`. The new API _requires_ an initialized `az_iot_provisioning_client_payload_options` structure (using `az_iot_provisioning_client_payload_options_default()`).
 
 ## 1.4.0 (2022-11-11)
 
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -391,16 +391,16 @@ jobs:
 
   # Validate all the files are formatted correctly according to the .clang-format file.
   - bash: |
-      # Run clang-format recursively on each source and header file within the repo.
-      clang-format --version
-      find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;
+      # Run clang-format-9 recursively on each source and header file within the repo.
+      clang-format-9 --version
+      find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;
 
       git status --untracked-files=no --porcelain
 
       if [[ `git status --untracked-files=no --porcelain` ]]; then
         echo Some files were not formatted correctly according to the .clang-format file.
-        echo Please run clang-format to fix the issue by using this bash command at the root of the repo:
-        echo "find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;"
+        echo Please run clang-format-9 to fix the issue by using this bash command at the root of the repo:
+        echo "find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;"
         exit 1
       fi
 
--- a/sdk/inc/azure/core/_az_cfg.h
+++ b/sdk/inc/azure/core/_az_cfg.h
@@ -88,4 +88,31 @@
 // Get the number of elements in an array
 #define _az_COUNTOF(array) (sizeof(array) / sizeof((array)[0]))
 
+/**
+ * @brief Deprecate functions.
+ *
+ */
+#ifdef __has_c_attribute
+#if __has_c_attribute(deprecated)
+#define AZ_DEPRECATED [[deprecated]]
+#endif
+#endif
+
+#ifndef AZ_DEPRECATED
+#if defined(_MSC_VER)
+#define AZ_DEPRECATED __declspec(deprecated)
+#elif defined(__GNUC__)
+#define AZ_DEPRECATED __attribute__((deprecated))
+#endif
+#endif
+
+#ifdef _MSC_VER
+#define AZ_PUSH_IGNORE_DEPRECATIONS _Pragma("warning(push)") _Pragma("warning(disable:4996)")
+#define AZ_POP_WARNINGS _Pragma("warning(pop)")
+#else
+#define AZ_PUSH_IGNORE_DEPRECATIONS \
+  _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#define AZ_POP_WARNINGS _Pragma("GCC diagnostic pop")
+#endif
+
 #endif // _az_CFG_H
--- a/sdk/inc/azure/core/_az_cfg.h
+++ b/sdk/inc/azure/core/_az_cfg.h
@@ -95,24 +95,24 @@
 #ifdef __has_c_attribute
 #if __has_c_attribute(deprecated)
 #define AZ_DEPRECATED [[deprecated]]
-#endif
-#endif
+#endif // __has_c_attribute(deprecated)
+#endif // __has_c_attribute
 
 #ifndef AZ_DEPRECATED
 #if defined(_MSC_VER)
 #define AZ_DEPRECATED __declspec(deprecated)
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) // !defined(_MSC_VER)
 #define AZ_DEPRECATED __attribute__((deprecated))
-#endif
-#endif
+#endif // defined(_MSC_VER)
+#endif // AZ_DEPRECATED
 
 #ifdef _MSC_VER
 #define AZ_PUSH_IGNORE_DEPRECATIONS _Pragma("warning(push)") _Pragma("warning(disable:4996)")
 #define AZ_POP_WARNINGS _Pragma("warning(pop)")
-#else
+#else // !_MSC_VER
 #define AZ_PUSH_IGNORE_DEPRECATIONS \
   _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
 #define AZ_POP_WARNINGS _Pragma("GCC diagnostic pop")
-#endif
+#endif // _MSC_VER
 
 #endif // _az_CFG_H
--- a/sdk/inc/azure/iot/az_iot_provisioning_client.h
+++ b/sdk/inc/azure/iot/az_iot_provisioning_client.h
@@ -78,8 +78,9 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * part of the certificate subject). Must conform to the limitations listed in the link below:
  * https://docs.microsoft.com/azure/iot-dps/concepts-service#registration-id
  * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_options
- * structure. Can be `NULL` for default options.
- * @pre \p client must not be `NULL`.
+ * structure. If `NULL` is passed, the provisioning client will use the default options. If using
+ * custom options, please initialize first by calling az_iot_provisioning_client_options_default()
+ * and then populating relevant options with your own values.* @pre \p client must not be `NULL`.
  * @pre \p global_device_hostname must be a valid span of size greater than 0.
  * @pre \p id_scope must be a valid span of size greater than 0.
  * @pre \p registration_id must be a valid span of size greater than 0.
@@ -271,6 +272,12 @@ typedef struct
    * Submit this timestamp when asking for Azure IoT service-desk help.
    */
   az_span error_timestamp;
+
+  /**
+   * An optional custom payload received from the service.
+   */
+  az_span payload;
+
 } az_iot_provisioning_client_registration_state;
 
 /**
@@ -341,6 +348,7 @@ typedef struct
    * and device id in case of success.
    */
   az_iot_provisioning_client_registration_state registration_state;
+
 } az_iot_provisioning_client_register_response;
 
 /**
@@ -439,8 +447,8 @@ AZ_NODISCARD az_result az_iot_provisioning_client_query_status_get_publish_topic
     size_t* out_mqtt_topic_length);
 
 /**
- * @brief Azure IoT Provisioning Client options for
- * az_iot_provisioning_client_get_request_payload(). Not currently used.  Reserved for future use.
+ * @brief Azure IoT Provisioning Client options for 
+ * az_iot_provisioning_client_get_request_payload().
  *
  */
 typedef struct
@@ -453,6 +461,57 @@ typedef struct
 } az_iot_provisioning_client_payload_options;
 
 /**
+ * @brief Gets the default az_iot_provisioning_client_payload_options.
+ * @details Call this to obtain an initialized #az_iot_provisioning_client_payload_options structure
+ * that can be afterwards modified and passed to
+ * #az_iot_provisioning_client_register_get_request_payload.
+ *
+ * @return #az_iot_provisioning_client_payload_options.
+ */
+AZ_NODISCARD az_iot_provisioning_client_payload_options
+az_iot_provisioning_client_payload_options_default();
+
+/**
+ * @brief Builds the optional payload for a provisioning request.
+ * @remark Use this API to build an MQTT payload during registration.
+ *         This call is optional for most scenarios. Some service
+ *         applications may require `custom_payload_property` specified during
+ *         registration to take additional decisions during provisioning time.
+ *         For example, if you need to register an IoT Plug and Play device you must
+ *         specify its model_id with this API via the `custom_payload_property`
+ *         `{"modelId":"your_model_id"}`.
+ *
+ * @param[in] client The #az_iot_provisioning_client to use for this call.
+ * @param[in] custom_payload_property __[nullable]__ Custom JSON to be added to this payload.
+ * Can be `NULL`.
+ * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_payload_options
+ * structure. If `NULL` is passed, the provisioning client will use the default options. If using
+ * custom options, please initialize first by calling
+ * az_iot_provisioning_client_payload_options_default() and then populating relevant options with
+ * your own values.
+ * @param[out] mqtt_payload A buffer with sufficient capacity to hold the MQTT payload.
+ * @param[in] mqtt_payload_size The size, in bytes of \p mqtt_payload.
+ * @param[out] out_mqtt_payload_length Contains the length, in bytes, written to \p mqtt_payload on
+ * success.
+ * @pre \p client must not be `NULL`.
+ * @pre \p mqtt_payload must not be `NULL`.
+ * @pre \p mqtt_payload_size must be greater than 0.
+ * @pre \p out_mqtt_payload_length must not be `NULL`.
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The payload was created successfully.
+ * @retval #AZ_ERROR_NOT_ENOUGH_SPACE The buffer is too small.
+ */
+AZ_NODISCARD az_result az_iot_provisioning_client_register_get_request_payload(
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* options,
+    uint8_t* mqtt_payload,
+    size_t mqtt_payload_size,
+    size_t* out_mqtt_payload_length);
+
+/**
+ * @deprecated since 1.5.0-beta.1.
+ * @see az_iot_provisioning_client_register_get_request_payload 
  * @brief Builds the optional payload for a provisioning request.
  * @remark Use this API to build an MQTT payload during registration.
  *         This call is optional for most scenarios. Some service
@@ -480,7 +539,7 @@ typedef struct
  * @retval #AZ_OK The payload was created successfully.
  * @retval #AZ_ERROR_NOT_ENOUGH_SPACE The buffer is too small.
  */
-AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+AZ_DEPRECATED az_result az_iot_provisioning_client_get_request_payload(
     az_iot_provisioning_client const* client,
     az_span custom_payload_property,
     az_iot_provisioning_client_payload_options const* options,
--- a/sdk/inc/azure/iot/az_iot_provisioning_client.h
+++ b/sdk/inc/azure/iot/az_iot_provisioning_client.h
@@ -473,7 +473,7 @@ az_iot_provisioning_client_payload_options_default();
 
 /**
  * @brief Builds the optional payload for a provisioning request.
- * @remark Use this API to build an MQTT payload during registration.
+ * @remark Use this API to build an MQTT payload during device registration.
  *         This call is optional for most scenarios. Some service
  *         applications may require `custom_payload_property` specified during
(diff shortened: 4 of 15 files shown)


sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -453,6 +461,57 @@ typedef struct
...
+ * #az_iot_provisioning_client_register_get_request_payload.
+ *
+ * @return #az_iot_provisioning_client_payload_options.
+ */
+AZ_NODISCARD az_iot_provisioning_client_payload_options
+az_iot_provisioning_client_payload_options_default();
+
+/**
+ * @brief Builds the optional payload for a provisioning request.
+ * @remark Use this API to build an MQTT payload during registration.
```
ericwol-msft, line 476:
```suggestion
 * @remark Use this API to build an MQTT payload during device registration.
```

sdk/samples/iot/paho_iot_provisioning_sample.c
```diff
@@ -323,6 +323,16 @@ static void handle_device_registration_status_message(
           "Hub Hostname:", register_response->registration_state.assigned_hub_hostname);
       IOT_SAMPLE_LOG_AZ_SPAN("Device Id:", register_response->registration_state.device_id);
       IOT_SAMPLE_LOG(" "); // Formatting
+
+      if (az_span_size(register_response->registration_state.payload) > 0)
+      {
+        IOT_SAMPLE_LOG_SUCCESS("Optional payload (Custom Allocation Policy) received:");
```
ericwol-msft, line 329: might not want to tie this to "Custom Allocation Policy" because in the future they might allow azure functions that is not tied to Custom Allocation Policies

sdk/samples/iot/paho_iot_provisioning_sample.c
```diff
@@ -323,6 +323,16 @@ static void handle_device_registration_status_message(
...
+        IOT_SAMPLE_LOG_AZ_SPAN("\t", register_response->registration_state.payload);
+      }
+      else
+      {
+        IOT_SAMPLE_LOG_SUCCESS("Optional payload (Custom Allocation Policy) not received.");
```
ericwol-msft, line 334: log not really needed

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -276,6 +277,35 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
   return AZ_ERROR_ITEM_NOT_FOUND;
 }
 
+AZ_INLINE az_result
+_az_iot_provisioning_client_get_json_object_span(az_json_reader* jr, az_span* out_span)
+{
+  *out_span = jr->token.slice;
```
ericwol-msft, line 283: ugh, this is so generic there seams like a helper func somewhere?

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
```diff
@@ -640,6 +645,168 @@ static void test_az_iot_provisioning_client_no_logging_succeed()
...
+static void
+test_az_iot_provisioning_client_parse_received_topic_and_payload_json_custom_payload_succeed()
+{
+  az_iot_provisioning_client client = { 0 };
+  az_result ret = az_iot_provisioning_client_init(
+      &client, test_global_device_hostname, test_id_scope, test_registration_id, NULL);
+  assert_int_equal(AZ_OK, ret);
+
+  az_span received_topic = AZ_SPAN_FROM_STR("$dps/registrations/res/200/?$rid=1");
+  char received_payload[] = "{\"operationId\":\"" TEST_OPERATION_ID
```
ericwol-msft, line 657: payload term is used in multiple places here. received_response?

ericwol-msft (approved):

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
CIPop: This is used throughout the tests: MQTT payload is indeed easily confused with DPS payload.
Will leave as-is.

sdk/src/azure/iot/az_iot_provisioning_client.c
CIPop: @ericwol-msft maybe I'm not interpreting this correctly: you mean in the JSON parser?

@ahsonkhan do you have anything like this available that I missed or was added after the feature branch was created?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -276,6 +277,35 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
...
+  {
+    return AZ_ERROR_UNEXPECTED_CHAR;
+  }
+
+  _az_RETURN_IF_FAILED(az_json_reader_skip_children(jr));
+
+  if (jr->token.kind != AZ_JSON_TOKEN_END_OBJECT)
+  {
+    return AZ_ERROR_UNEXPECTED_CHAR;
+  }
```
ahsonkhan, line 301: I don't believe this check is necessary, since `az_json_reader_skip_children` should validate that it reached "end of container" (at the same depth), or fail otherwise.

Unless you observed some test case which shows otherwise, I think this branch is unreachable and can be removed.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -276,6 +277,35 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
...
 
+AZ_INLINE az_result
+_az_iot_provisioning_client_get_json_object_span(az_json_reader* jr, az_span* out_span)
+{
+  *out_span = jr->token.slice;
+
+  if (jr->token.kind == AZ_JSON_TOKEN_NULL)
+  {
+    *out_span = AZ_SPAN_EMPTY;
+    return AZ_OK;
```
ahsonkhan, line 288: Interesting, why set `out_span` as empty, rather than what it is set to, above (on line 283), which is the characters from the JSON payload, i.e. `null`?

sdk/src/azure/iot/az_iot_provisioning_client.c
ahsonkhan: I am assuming you are trying to get a span that contains the entire sub-object from within the JSON.

So, for example, in the following:
```json
{ "sub-object": { "name": "Cris", "complex-object": {}, "numberOfEmbeddedDevices": 25}}
```
You want to extract out:
```json
{ "name": "Cris", "complex-object": {}, "numberOfEmbeddedDevices": 25}
```

Correct?

What you have is the way to go about it, building on top of the provided methods. There isn't a built-in API to extract multi-token sub-objects out.

We could expose the internal property like `bytes_consumed` to indicate at what offset/index the token started, but that doesn't help too much here, since you'd still have to do the following to get the length of the sub-object and create a span: `end_object_index - start_object_index  + 1`.

CIPop: Thanks @ahsonkhan . Since extracting sub-objects is something fairly common, it may be useful to add this functionality to the JSon parser.

sdk/src/azure/iot/az_iot_provisioning_client.c
CIPop: Replacing with a precondition.

sdk/src/azure/iot/az_iot_provisioning_client.c
CIPop: The purpose of this function is to extract a complex JSON object that can be parsed by the application.
`null` isn't a valid Json and so we return an empty span.

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -453,6 +461,57 @@ typedef struct
...
+ */
+AZ_NODISCARD az_iot_provisioning_client_payload_options
+az_iot_provisioning_client_payload_options_default();
+
+/**
+ * @brief Builds the optional payload for a provisioning request.
+ * @remark Use this API to build an MQTT payload during device registration.
+ *         This call is optional for most scenarios. Some service
+ *         applications may require `custom_payload_property` specified during
+ *         registration to take additional decisions during provisioning time.
```
danewalton, line 479:
```suggestion
 *         registration to make additional decisions during provisioning time.
```

sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
```diff
@@ -250,13 +250,15 @@ static void register_device_with_provisioning_service(void)
...
 
+  AZ_PUSH_IGNORE_DEPRECATIONS
   rc = az_iot_provisioning_client_get_request_payload(
       &provisioning_client,
       custom_registration_payload_property,
       NULL,
       mqtt_payload,
       sizeof(mqtt_payload),
       &mqtt_payload_length);
+  AZ_POP_WARNINGS
```
danewalton, line 261: are we able to use the new API instead?

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -564,22 +595,27 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
   return AZ_OK;
 }
 
-AZ_NODISCARD az_result az_iot_provisioning_client_get_request_payload(
+AZ_NODISCARD az_iot_provisioning_client_payload_options
+az_iot_provisioning_client_payload_options_default()
+{
+  return (az_iot_provisioning_client_payload_options){ 0 };
+}
```
danewalton, line 602: nit: other iot files initialize each element explicitly. consider making this similar

https://github.com/Azure/azure-sdk-for-c/blob/399f1dbd09c2208f044cef4a4b4b97d115817de3/sdk/src/azure/iot/az_iot_hub_client.c#L29-L36

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -564,22 +595,27 @@ AZ_NODISCARD az_result az_iot_provisioning_client_parse_received_topic_and_paylo
...
     az_iot_provisioning_client_payload_options const* options,
     uint8_t* mqtt_payload,
     size_t mqtt_payload_size,
     size_t* out_mqtt_payload_length)
 {
-  (void)options;
-
   _az_PRECONDITION_NOT_NULL(client);
-  _az_PRECONDITION_IS_NULL(options);
+  _az_PRECONDITION_NOT_NULL(options);
```
danewalton, line 613: please remove. the API docs state this may be `NULL`.

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
```diff
@@ -682,6 +848,12 @@ int test_az_iot_provisioning_client_parser()
     cmocka_unit_test(test_az_iot_provisioning_client_operation_complete_translate_succeed),
     cmocka_unit_test(test_az_iot_provisioning_client_logging_succeed),
     cmocka_unit_test(test_az_iot_provisioning_client_no_logging_succeed),
+    cmocka_unit_test(
+        test_az_iot_provisioning_client_parse_received_topic_and_payload_json_custom_payload_succeed),
+    cmocka_unit_test(
+        test_az_iot_provisioning_client_parse_received_topic_and_payload_json_custom_payload_empty_succeed),
+    cmocka_unit_test(
+        test_az_iot_provisioning_client_parse_received_topic_and_payload_json_custom_payload_null_succeed),
```
danewalton, line 856: Are you able to double check that we have a test for when the options are `NULL` and when they're not `NULL`?

danewalton (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -391,16 +391,16 @@ jobs:
...
+      clang-format-9 --version
+      find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;
 
       git status --untracked-files=no --porcelain
 
       if [[ `git status --untracked-files=no --porcelain` ]]; then
         echo Some files were not formatted correctly according to the .clang-format file.
-        echo Please run clang-format to fix the issue by using this bash command at the root of the repo:
-        echo "find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format -i {} \;"
+        echo Please run clang-format-9 to fix the issue by using this bash command at the root of the repo:
```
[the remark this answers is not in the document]
CIPop: This works on Linux where separate executables can be installed side-by-side. Let me know if you are aware of any other solution.
Running clang-format on latest Ubuntu won't work today as the newer versions format in a different way that fails build validation.

JeffreyRichter (approved):
I mostly just looked at function signatures and they all looked good to me.

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -78,8 +78,9 @@ AZ_NODISCARD az_iot_provisioning_client_options az_iot_provisioning_client_optio
  * part of the certificate subject). Must conform to the limitations listed in the link below:
  * https://docs.microsoft.com/azure/iot-dps/concepts-service#registration-id
  * @param[in] options __[nullable]__ A reference to an #az_iot_provisioning_client_options
- * structure. Can be `NULL` for default options.
- * @pre \p client must not be `NULL`.
+ * structure. If `NULL` is passed, the provisioning client will use the default options. If using
+ * custom options, please initialize first by calling az_iot_provisioning_client_options_default()
+ * and then populating relevant options with your own values.* @pre \p client must not be `NULL`.
```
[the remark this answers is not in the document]
CIPop: Missed this: likely auto-formatting.

sdk/inc/azure/iot/az_iot_provisioning_client.h
CIPop: Fixed in both places.

sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
CIPop: Yes, we could if we needed to in the future. For now, we'll leave these as-is for reference (i.e., existing applications picking up the SDK with the deprecation).

sdk/tests/iot/provisioning/test_az_iot_provisioning_client_parser.c
CIPop: Options can't be NULL:

> Roll-back from feature branch: member function options parameter cannot be NULL. (https://github.com/Azure/azure-sdk-for-c/pull/2188#discussion_r860141070)

sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
```diff
@@ -250,13 +250,15 @@ static void register_device_with_provisioning_service(void)
   uint8_t mqtt_payload[MQTT_PAYLOAD_BUFFER_LENGTH];
   size_t mqtt_payload_length;
 
+  AZ_PUSH_IGNORE_DEPRECATIONS
```
ahsonkhan, line 253: We may want to update our samples to no longer use deprecated APIs and provide alternatives that exist. Otherwise, customers will just copy/paste this.

ericwol-msft: DSP v2 is not GA or publicly available yet. Moving to the new APIs will happen with the HSFM work.

CHANGELOG.md
```diff
@@ -4,12 +4,19 @@
...
+
 ### Breaking Changes
 
 ### Bugs Fixed
 
+- Azure IoT Samples now display errors following the _source file:line number_ format, allowing Visual Studio Code to automatically create links from the output log.
+
 ### Other Changes
 
+- Azure IoT Device Provisioning Client API __deprecation__: `az_iot_provisioning_client_get_request_payload()` is deprecated in favor of `az_iot_provisioning_client_register_get_request_payload()`. The new API _requires_ an initialized `az_iot_provisioning_client_payload_options` structure (using `az_iot_provisioning_client_payload_options_default()`).
```
ahsonkhan, line 18: @CIPop what was your intention in how you wanted `__deprecation__` rendered in markdown? Just bolded?

It shows up as:
> Azure IoT Device Provisioning Client API __deprecation__:

Did you want something like this, instead?
> Azure IoT Device Provisioning Client API `__deprecation__`:

sdk/inc/azure/core/_az_cfg.h
```diff
@@ -88,4 +88,31 @@
...
+#endif // AZ_DEPRECATED
+
+#ifdef _MSC_VER
+#define AZ_PUSH_IGNORE_DEPRECATIONS _Pragma("warning(push)") _Pragma("warning(disable:4996)")
+#define AZ_POP_WARNINGS _Pragma("warning(pop)")
+#else // !_MSC_VER
+#define AZ_PUSH_IGNORE_DEPRECATIONS \
+  _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+#define AZ_POP_WARNINGS _Pragma("GCC diagnostic pop")
+#endif // _MSC_VER
```
ahsonkhan, line 116: I just noticed that these are prefixed with `AZ_` which are meant to indicate for public use (especially when accessible from a header).

These helper macros are only used in tests and as a stop-gap in samples. Ideally we shouldn't have them at all, as there isn't substantive usage of it.

But, if we want to keep them, they should be for private use within the SDK only, which is indicated with the prefix `_az_`.

So, either we remove them, or rename them as:
`_az_POP_WARNINGS`

@CIPop let me know and I can make the fix, if you need me to, before we release 1.5.0 GA today.

sdk/src/azure/iot/az_iot_provisioning_client.c
```diff
@@ -276,6 +277,31 @@ AZ_INLINE az_result _az_iot_provisioning_client_parse_payload_error_code(
...
+  if (jr->token.kind == AZ_JSON_TOKEN_NULL)
+  {
+    *out_span = AZ_SPAN_EMPTY;
+    return AZ_OK;
+  }
+
+  if (jr->token.kind != AZ_JSON_TOKEN_BEGIN_OBJECT)
+  {
+    return AZ_ERROR_UNEXPECTED_CHAR;
+  }
```
ahsonkhan, line 294: Code coverage for `_az_iot_provisioning_client_get_json_object_span` can be improved. Line coverage for this file is great at over 95%, but branch coverage can be better (around 70%).

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2104943&view=codecoverage-tab

image

sdk/inc/azure/iot/az_iot_provisioning_client.h
```diff
@@ -453,11 +462,61 @@ typedef struct
...
+    az_iot_provisioning_client const* client,
+    az_span custom_payload_property,
+    az_iot_provisioning_client_payload_options const* options,
+    uint8_t* mqtt_payload,
+    size_t mqtt_payload_size,
+    size_t* out_mqtt_payload_length);
+
+/**
+ * @deprecated since 1.5.0-beta.1.
+ * @see az_iot_provisioning_client_register_get_request_payload
```
ahsonkhan, line 514: This doesn't render the best in doxygen docs:
https://azuresdkdocs.blob.core.windows.net/$web/c/az_iot/1.5.0/deprecated.html

We should, at least update this to the following.
```suggestion
 * @deprecated since 1.5.0
 * @see az_iot_provisioning_client_register_get_request_payload
```

 I will leave it open for others to figure out other improvements :)

CHANGELOG.md
CIPop: Yes, bolded.

sdk/src/azure/iot/az_iot_provisioning_client.c
ericwol-msft: What is the expected branch coverage for this repo? For example az_http_XXX, along with base64 parsing and json parser components has very low code and branch coverage. Are you guys addressing this???

sdk/inc/azure/core/_az_cfg.h
CIPop: Yes, these are intended for public use (they are used in our samples). This was intentional, to allow existing applications to work around the deprecation until the API will be removed in V2. No changes required.

sdk/src/azure/iot/az_iot_provisioning_client.c
CIPop: Thanks @ahsonkhan. We'll consider improving our code coverage in the future.

sdk/inc/azure/iot/az_iot_provisioning_client.h
CIPop: This was the time the changes were deprecated. Feel free to submit a new PR to update.

CIPop: > This doesn't render the best in doxygen docs:

Can you be more specific about what doesn't render correctly? Looks acceptable to me:

image

https://azuresdkdocs.blob.core.windows.net/$web/c/az_iot/1.5.0/az__iot__provisioning__client_8h.html#a82fdc4e466e89adf377f1f5c15756d2f

ahsonkhan: It looks good in the API list.

The link I shared leads to the `Deprecated List` page, which formats the function signature in a way that is difficult to parse/read.

It also don't understand what is meant by `Global`. Apparently there's a way to remove that page all together (https://stackoverflow.com/a/25065592), but maybe we should just keep it, as is.

sdk/src/azure/iot/az_iot_provisioning_client.c
ahsonkhan: The previous IoT changes/PRs from folks on the team had set a much higher bar for coverage, at close to 100% line coverage and 80%+ branch coverage, so I just raised my expectations accordingly for anything IoT related 😄 

Look at IoT Hub Client files, for example. They are pretty good!

> What is the expected branch coverage for this repo?
> If there are specific numbers, please enforce them through the CI system.
 
Agreed! **They are already here, enforced in CI**, repo-wide at 90%+ for line coverage and 70%+ for branch coverage. This file meets that target, but is at the border.

https://github.com/Azure/azure-sdk-for-c/blob/c551dfe9bd1e3b4152613066ec626d7c4901f421/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L369-L387

We are above that currently, in aggregate/repo-wide, but whenever I see some low hanging/useful test coverage improvement, I point it out to leave slack for places which are more challenging (and less fruitful) to test. We don't want each PR to incrementally trickle down towards that bar, and then a small change trips the CI failure for test coverage.

> For example az_http_XXX, along with base64 parsing and json parser components has very low code and branch coverage. Are you guys addressing this???

I wouldn't consider these "very low", in fact they are well above the bar:
- base64 has 97.7 and 81.7 respectively
- json reader has 97.5 and 91.5 respectively

Both base64 and json have pretty good overall coverage. Some areas can be improved, but most of the lack of branch coverage is around preconditions. Even then, we do have [specific issues to track that in the backlog](https://github.com/Azure/azure-sdk-for-c/issues?q=is%3Aopen+is%3Aissue+label%3Atest-enhancement).
For example:
https://github.com/Azure/azure-sdk-for-c/issues/1836
https://github.com/Azure/azure-sdk-for-c/issues/2310

That said, you correctly point out that the `az_http_policy_XXX` are the only ones that are lagging, but not enough attention has been put on those in the `main` branch for a while now, since we don't have any service being supported/shipping out of main that uses HTTP. For example, the entire `az_http_pipeline_policy_credential()` is uncovered. It also currently has no callers. Once there is a motivating need to revisit and modify those files (possibly with storage APIs being added), we should definitely up the coverage there!

sdk/inc/azure/iot/az_iot_provisioning_client.h
ahsonkhan: > This was the time the changes were deprecated. Feel free to submit a new PR to update.

Fixed in https://github.com/Azure/azure-sdk-for-c/pull/2480

ahsonkhan: > This was the time the changes were deprecated. Feel free to submit a new PR to update.

Fixed in https://github.com/Azure/azure-sdk-for-c/pull/2480

PR #2474, opened 2023-01-10
Rename push ignore deprecations and pop warnings macro from AZ_ to _az_ to indicate they are for internal use only.

ahsonkhan:
Follow-up from https://github.com/Azure/azure-sdk-for-c/pull/2443#discussion_r1066391277

### DIFF
--- a/sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
+++ b/sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
@@ -250,7 +250,7 @@ static void register_device_with_provisioning_service(void)
   uint8_t mqtt_payload[MQTT_PAYLOAD_BUFFER_LENGTH];
   size_t mqtt_payload_length;
 
-  AZ_PUSH_IGNORE_DEPRECATIONS
+  _az_PUSH_IGNORE_DEPRECATIONS
   rc = az_iot_provisioning_client_get_request_payload(
       &provisioning_client,
       custom_registration_payload_property,
@@ -258,7 +258,7 @@ static void register_device_with_provisioning_service(void)
       mqtt_payload,
       sizeof(mqtt_payload),
       &mqtt_payload_length);
-  AZ_POP_WARNINGS
+  _az_POP_WARNINGS
   if (az_result_failed(rc))
   {
     IOT_SAMPLE_LOG_ERROR(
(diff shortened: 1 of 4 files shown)


JeffreyRichter (approved):

sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
```diff
@@ -250,15 +250,15 @@ static void register_device_with_provisioning_service(void)
   uint8_t mqtt_payload[MQTT_PAYLOAD_BUFFER_LENGTH];
   size_t mqtt_payload_length;
 
-  AZ_PUSH_IGNORE_DEPRECATIONS
+  _az_PUSH_IGNORE_DEPRECATIONS
```
danewalton, line 253: Should these not be available for customers should they want to ignore the deprecations? Or are we saying they should find their own mechanism to (if they want) since they are risking using deprecated functions?

ahsonkhan: > Or are we saying they should find their own mechanism to (if they want) since they are risking using deprecated functions?

That.

Using a deprecated function is not supported, so we shouldn't make it convenient and easy to circumvent that warning to discourage use. If the user wants to do so, they can use whatever mechanism makes sense to them.

danewalton (approved):

sdk/samples/iot/paho_iot_pnp_with_provisioning_sample.c
CIPop: > Using a deprecated function is not supported

I don't think that's correct. V1 must have support for all deprecated functions for the entire lifetime if we are to be backward compatible.

> so we shouldn't make it convenient and easy to circumvent that warning

This is debatable: if an existing application is already using the function, it is safe for them to continue to do so until they will be forced to use something else in V2 or they require additional support / functionality offered by the new, non-deprecated function.

/cc @RickWinter , @JeffreyRichter

JeffreyRichter: Azure SDKs can make breaking changes we just strongly discourage it due to the customer pain.
I haven't been involved in the deprecation macro stuff and so I'm not sure what its purpose/goal is. 
Seems to me that there isn't much need to deprecate anything - we just ship a new version that is (publicly) missing the thing we don't want anymore, and notify customers of the break and how to best migrate from the old thing to the new thing.

PR #2483, opened 2023-01-18
Fix init

gearama:
#closes 2318

Fix filed initializer. causes issues with  -WMissing-fields-initializer flag

### DIFF
--- a/sdk/inc/azure/core/_az_cfg.h
+++ b/sdk/inc/azure/core/_az_cfg.h
@@ -40,6 +40,7 @@
 
 #ifdef __GNUC__
 
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
 #pragma GCC diagnostic ignored "-Wmissing-braces"
 
 #endif // __GNUC__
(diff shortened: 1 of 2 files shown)


antkmsft (commented):
We should not be receiving this warning. It should be disabled somewhere in the options. If you are still getting it, let's find out why. We don't do this trick in other places.
`{ 0 }` is valid C99, and standard says it is the syntax for everything being zeroed.

antkmsft:
Wait, `_az_cfg.h`, which are getting included in every header, shoud be disabling this: https://github.com/Azure/azure-sdk-for-c/blob/main/sdk/inc/azure/core/_az_cfg.h#L41-L52

antkmsft:
And, IIRC, C99 does not necessarily support unnamed initializers, but designated initializers, so without pragmas it should look like this:
`az_http_response_status_line status_line = { .reason_phrase = AZ_SPAN_EMPTY, .status_code = AZ_HTTP_STATUS_CODE_NONE, .major_version = 0, .minor_version = 0 }`

gearama:
pragmas it is , i prefer that to changing the actual code

antkmsft (approved):

PR #2537, opened 2023-03-28
Adding posix mutex and timers and x509 cred header

RLeclair:
In this PR:

- Adding header to support x509 credentials.
- Adding platform headers
- Adding extra functions to platform to support timers and mutexes. Implemented posix functions.
- Adding unit tests for platform.
- Adding team to CODEOWNERS in relevant files.


.github/CODEOWNERS
```diff
@@ -30,7 +30,9 @@
 /sdk/tests/iot/                @CIPop @ericwol-msft @ewertons @jspaith @preethi826 @RLeclair @vaavva
```
vaavva, line 30:
```suggestion
/sdk/docs/iot/                 @CIPop @ericwol-msft @ewertons @jspaith @RLeclair @vaavva
/sdk/inc/azure/iot/            @CIPop @ericwol-msft @ewertons @jspaith @RLeclair @vaavva
/sdk/samples/iot/              @CIPop @ericwol-msft @ewertons @jspaith @RLeclair @vaavva
/sdk/src/azure/iot/            @CIPop @ericwol-msft @ewertons @jspaith @RLeclair @vaavva
/sdk/tests/iot/                @CIPop @ericwol-msft @ewertons @jspaith @RLeclair @vaavva
```

While you're already in here editing the codeowners file

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +40,124 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  }
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_platform_mutex_init(az_platform_mutex* mutex_handle)
+{
+  _az_PRECONDITION_NOT_NULL(mutex_handle);
+
+  if (0 != pthread_mutex_init(mutex_handle, NULL))
```
CIPop, line 121: This is incorrect within the new design. The mutex must be re-entrant.

I've updated the code in my latest iteration of the proof-of-concept: https://github.com/CIPop/azure-sdk-for-c/blob/hfsm/sdk/src/azure/platform/az_posix.c#L96

sdk/tests/platform/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+add_cmocka_test(az_platform_test SOURCES
+                main.c
+                test_az_platform.c
+                COMPILE_OPTIONS ${DEFAULT_C_COMPILE_FLAGS} ${NO_CLOBBERED_WARNING}
+                LINK_LIBRARIES ${CMOCKA_LIB} az_core
+                # include cmocka headers and private folder headers
+                INCLUDE_DIRECTORIES ${CMOCKA_INCLUDE_DIR} ${CMAKE_SOURCE_DIR}/sdk/src/azure/core/
+                )
+
+create_map_file(az_platform_test az_platform_test.map)
```
CIPop, line 21: Do we care about map files for tests?

sdk/tests/platform/test_az_platform.c
```diff
@@ -0,0 +1,274 @@
...
+
+  az_platform_mutex test_mutex_handle;
+
+  assert_int_equal(az_platform_mutex_init(&test_mutex_handle), AZ_OK);
+  assert_int_equal(az_platform_mutex_acquire(&test_mutex_handle), AZ_OK);
+  assert_int_equal(az_platform_mutex_release(&test_mutex_handle), AZ_OK);
+  assert_int_equal(az_platform_mutex_destroy(&test_mutex_handle), AZ_OK);
+}
+
+static void test_az_platform_mutex_double(void** state)
```
CIPop, line 234: Please add a new test to verify recursive (reentrant) mutex aquire (i.e. 2 or more aquires in the same thread should work - number of releases must still match number of acquires).

CIPop (changes requested):
There are a few blocking changes:
1. Mutex must be made recursive `PTHREAD_MUTEX_RECURSIVE`
2. We should clarify the usage of `CLOCK_MONOTONIC` on Linux systems.

CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ option(ADDRESS_SANITIZER "Build with address sanitizer" OFF)
 # vcpkg integration
 include(AzureVcpkg)
 az_vcpkg_integrate()
+set(AZ_PLATFORM_IMPL "POSIX" CACHE STRING "Platform")
```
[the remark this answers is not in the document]
RLeclair: If we make AZ_PLATFORM_IMPL into `option` then we won't be able to set it to `"POSIX"`, `"WIN32"`, etc. since option will make it into a bool.

https://cmake.org/cmake/help/latest/command/option.html

sdk/inc/azure/core/az_platform.h
```diff
@@ -16,9 +16,13 @@
 #define _az_PLATFORM_H
 
 #include <azure/core/az_result.h>
-
 #include <stdbool.h>
 #include <stdint.h>
+#ifdef PLATFORM_POSIX
+#include "azure/platform/az_platform_posix.h"
+#else
```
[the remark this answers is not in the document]
RLeclair: Done, we only have posix and win32 for now. I will get clarification on how custom should be included into these changes.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +40,124 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  timer_handle->platform_timer._internal.callback(timer_handle->platform_timer._internal.sdk_data);
+}
+
+AZ_NODISCARD az_result az_platform_timer_create(
+    az_platform_timer* timer_handle,
+    az_platform_timer_callback callback,
+    void* sdk_data)
+{
+  _az_PRECONDITION_NOT_NULL(timer_handle);
+  _az_PRECONDITION_NOT_NULL(callback);
```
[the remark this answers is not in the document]
RLeclair: I am not sure I understand, I do check for it not being null in the precondition above: `_az_PRECONDITION_NOT_NULL(callback);`

sdk/tests/platform/CMakeLists.txt
RLeclair: The tests for `core` and `iot` use map files for their tests, I included it for consistency with the other tests

sdk/inc/azure/core/az_platform.h
```diff
@@ -53,6 +57,122 @@ AZ_NODISCARD az_result az_platform_clock_msec(int64_t* out_clock_msec);
...
+AZ_NODISCARD az_result az_platform_timer_destroy(az_platform_timer* timer_handle);
+
+/**
+ * @brief Initializes a mutex.
+ * 
+ * @param[in] mutex_handle The mutex handle.
+ * 
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK Success.
+ * @retval #AZ_ERROR_DEPENDENCY_NOT_PROVIDED No platform implementation was supplied to support this
```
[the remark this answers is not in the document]
RLeclair: Added retvals and new return codes

sdk/inc/azure/core/az_result.h
```diff
@@ -99,6 +99,24 @@ enum az_result_core
   /// Dynamic memory allocation request was not successful.
   AZ_ERROR_OUT_OF_MEMORY = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE_PLATFORM, 1),
 
+  /// Temporary error, resource unavailable temporarily.
```
CIPop, line 102: We cannot simply copy everything in POSIX. I assume that other OSs will be different. We should instead have a minimal common set of return values for the PAL that will be determined from the original errors (possible many more) across all PALs.

sdk/src/azure/platform/az_posix.c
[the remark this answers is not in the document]
CIPop: Not sure - maybe old revision? Looks good now.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  timer_handle->_internal.sev.sigev_notify_function = &_timer_callback_handler;
+  timer_handle->_internal.sev.sigev_value.sival_ptr = timer_handle;
+
+  if (0
+      != timer_create(
+          _az_PLATFORM_POSIX_CLOCK_ID,
+          &timer_handle->_internal.sev,
+          &timer_handle->_internal.timerid))
+  {
+    if (EAGAIN == errno)
```
CIPop, line 93: Please don't add if without braces {}.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  _az_PRECONDITION_NOT_NULL(timer_handle);
+
+  timer_handle->_internal.trigger.it_value.tv_sec = milliseconds / _az_TIME_MILLISECONDS_PER_SECOND;
+  timer_handle->_internal.trigger.it_value.tv_nsec
+      = (milliseconds % _az_TIME_MILLISECONDS_PER_SECOND) * _az_TIME_NANOSECONDS_PER_MILLISECOND;
+
+  if (0
+      != timer_settime(timer_handle->_internal.timerid, 0, &timer_handle->_internal.trigger, NULL))
+  {
+    if (EFAULT == errno)
```
CIPop, line 120: Same here : please don't add if without braces {}.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+{
+  _az_PRECONDITION_NOT_NULL(mutex_handle);
+
+  pthread_mutexattr_t attr;
+  pthread_mutexattr_init(&attr);
+  pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+
+  int mutex_init_result = pthread_mutex_init(mutex_handle, &attr);
+
+  if (EAGAIN == mutex_init_result)
```
CIPop, line 153: Please don't add if without braces {}.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_platform_mutex_acquire(az_platform_mutex* mutex_handle)
+{
+  _az_PRECONDITION_NOT_NULL(mutex_handle);
+
+  int mutex_lock_result = pthread_mutex_lock(mutex_handle);
+
+  if (EINVAL == mutex_lock_result)
```
CIPop, line 175: Please don't add if without braces {}.

sdk/src/azure/platform/az_posix.c
CIPop, line 93: I don't think any of these are actionable. We could log the error for debugging but why would we convert from one type of error to another if our code doesn't use these return codes (which will need to be the same on Windows, Linux, Mac, Embedded RTOSes, etc)?

Please add _only_ error codes that will be generic across PALs and _must_ be used by our code.

sdk/src/azure/platform/az_posix.c
CIPop: Also, we should likely remove the error adapter code.

sdk/src/azure/platform/az_posix.c
CIPop: Also, we should likely remove the error adapter code.

sdk/src/azure/platform/az_posix.c
CIPop: Also, we should likely remove the error adapter code.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_platform_mutex_release(az_platform_mutex* mutex_handle)
+{
+  _az_PRECONDITION_NOT_NULL(mutex_handle);
+
+  int mutex_unlock_result = pthread_mutex_unlock(mutex_handle);
+
+  if (EINVAL == mutex_unlock_result)
```
CIPop, line 195: Please remove unused adapter code. We may add actionable error codes later, as needed.

sdk/src/azure/platform/az_posix.c
```diff
@@ -26,3 +57,165 @@ AZ_NODISCARD az_result az_platform_sleep_msec(int32_t milliseconds)
...
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result az_platform_mutex_destroy(az_platform_mutex* mutex_handle)
+{
+  _az_PRECONDITION_NOT_NULL(mutex_handle);
+
+  int mutex_destroy_result = pthread_mutex_destroy(mutex_handle);
+
+  if (EBUSY == mutex_destroy_result)
```
CIPop, line 213: Please remove unused adapter code. We may add actionable error codes later, as needed.

CIPop (changes requested):
Please remove error adapter code for error codes that are not used by our SDK.
PAL is to be used only by our SDK and the extra adapter isn't providing any actionable information.

My original code was only to add all error codes returned by the PAL. We should use an error `az_result` as generic as possible.

sdk/inc/azure/core/az_platform.h
```diff
@@ -16,9 +16,18 @@
 #define _az_PLATFORM_H
 
 #include <azure/core/az_result.h>
-
+#ifndef __APPLE__
+#include <azure/platform/internal/az_platform_internal.h>
```
CIPop, line 20: Instead of this, can't we just have null implementations on Apple for now with an explanation that callback timers aren't supported?

sdk/src/azure/platform/az_posix.c
```diff
@@ -5,18 +5,53 @@
...
 
 #include <azure/core/_az_cfg.h>
 
+#ifdef __linux__
+#define _az_PLATFORM_POSIX_CLOCK_ID CLOCK_BOOTTIME
+#else
+#define _az_PLATFORM_POSIX_CLOCK_ID CLOCK_MONOTONIC
+#endif
+
+#ifndef __APPLE__
```
CIPop, line 23: I think that we should instead have a different file for Apple. The POSIX version that they support can be extracted in a common file and the rest (i.e. timers) can be separated into az_apple maybe, with no implementation right now.

CIPop (approved):
LGTM. Left a few nitpicks.

PR #2541, opened 2023-04-05
Re-designing credentials, adding x509

RLeclair:
PR contains design changes for az_credentials. Includes x509 credential support.


CIPop (approved):

sdk/inc/azure/core/az_credentials_common.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @brief Pointer to a credential implementation.
+ */
+typedef void* az_platform_credential;
+
+/**
+ * @brief Defines the supported credential types.
+ */
+typedef enum
```
RickWinter, line 44: This should be removed.  Core does not know about credential types it only knows about the AccessToken and the TokenCredential type.  Types expanding on TokenCredential are identity constructs.

RickWinter (changes requested):

sdk/inc/azure/core/az_credentials_common.h
RickWinter: You can use the C++ impl to reference the Azure SDK design patterns for credentials.
https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/inc/azure/core/credentials/credentials.hpp

CIPop: The change is based from an offline discussion we had with @LarryOsterman . Sharing the summary here:

> We will create pal_x509certifiate which represents the certificate for the selected platform (e.g. MosquittoMQTT).
> - az_credential  will become a public type.  
> - We will add a constructor (_init function) that accepts a pal_x509certificate 
> - A separate constructor will be used for SAS az_credentials

CIPop: @RickWinter what would be the recommended way to add a public API for X.509 credentials?
Should we create our own `az_iot_` type instead?

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+#include <azure/core/_az_cfg_prefix.h>
+
+/**
+ * @brief Defines the key storage types.
+ */
+typedef enum
+{
+  AZ_CREDENTIALS_X509_KEY_MEMORY = 0,
+  AZ_CREDENTIALS_X509_KEY_SECURITY_MODULE = 1,
+} az_platform_credential_x509_key_type;
```
ewertons, line 33: Maybe a name like `az_platform_credential_x509_key_source` would be more appropriate?

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+  struct
+  {
+    /// Contains x509 certificate
+    az_span cert;
+
+    /// Contains x509 key
+    az_span key;
+
+    /// Specifies the storage type for the key.
+    az_platform_credential_x509_key_type key_type;
```
ewertons, line 49: Here too, key type relates to the type of the key itself. 
The concept being considered here is: where is this key coming from?
So maybe `storage_source` or `key_source` would be more correct.

ewertons (changes requested):

sdk/inc/azure/platform/az_platform_x509_credential.h
ericwol-msft: also 
AZ_CREDENTIALS_X509_KEY_STORAGE_MEMORY
AZ_CREDENTIALS_X509_KEY_STORAGE_SECURITY_MODULE
?

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @brief x509 credential definition.
+ */
+typedef struct
+{
+  struct
+  {
+    /// Contains x509 certificate
+    az_span cert;
```
ericwol-msft, line 43:
```suggestion
    az_span certificate;
```

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Contains x509 key
```
ericwol-msft, line 45:
```suggestion
    /// Contains x509 private key
```

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+    az_span key;
+
+    /// Specifies the storage type for the key.
+    az_platform_credential_x509_key_type key_type;
+  } _internal;
+} az_platform_x509_credential;
```
ericwol-msft, line 51: So how is this different from a ECC cert?

sdk/inc/azure/platform/az_platform_x509_credential.h
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @brief x509 credential constructor.
+ */
+AZ_NODISCARD AZ_INLINE az_credential az_credential_x509_init(az_platform_x509_credential* cred)
```
ericwol-msft, line 56:
```suggestion
AZ_NODISCARD AZ_INLINE az_credential az_credential_x509_init(az_platform_x509_certificate* certificate)
```

ericwol-msft: specific credential in (cert), generic credential out.

sdk/inc/azure/platform/az_platform_x509_credential.h
RLeclair: Changed to `key_source` thanks for pointing it out!

sdk/inc/azure/platform/az_platform_x509_credential.h
RLeclair: Added both changes, thanks!

sdk/inc/azure/core/az_credentials_common.h
RLeclair: Hi @RickWinter, I've gone ahead and removed `az_credentials_common.h` and moved my changes to `az_iot_certificate_x509`.

RLeclair:
Updated the location of the credential type. Changed the naming to `az_iot_certificate_x509.h` and removed the abstraction layer.

PR #2567, opened 2023-05-17
Adding az_mqtt and integration tests

RLeclair:
Also adds mosquitto as a default package for auto vcpkg. 

Ammended instructions on sample to include mosquitto.


CMakeLists.txt
```diff
@@ -114,7 +138,7 @@ if (UNIT_TESTING)
   add_subdirectory(sdk/tests/iot/provisioning)
 
   # Platform
-  if(AZ_PLATFORM_IMPL STREQUAL "POSIX")
+  if(AZ_PLATFORM_IMPL STREQUAL "POSIX" AND AZ_MQTT_TRANSPORT_IMPL STREQUAL "MOSQUITTO") # Temporary
```
CIPop, line 141: Nit: I wouldn't add the "Temporary" comment. 
These tests do not exist today, and the plan of record is to add integration testing _only_ for Posix/MosquittoMQTT client.

sdk/inc/azure/core/internal/az_mqtt_policy.h
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
CIPop, line 1: Should the filename be `az_mqtt_policy_internal.h` too?

sdk/inc/azure/iot/az_iot_common.h
```diff
@@ -45,10 +45,10 @@ enum az_result_iot
 enum az_log_classification_iot
 {
   AZ_LOG_MQTT_RECEIVED_TOPIC
-  = _az_LOG_MAKE_CLASSIFICATION(_az_FACILITY_IOT_MQTT, 1), ///< Accepted MQTT topic received.
```
CIPop, line 48: Breaking change: these are currently in-use. The classification is correct: this is IoT-specific MQTT usage (not generic).
These are used when the Hub/DPS clients parse the topics/payloads only.

Please revert changes in this file.

sdk/inc/azure/core/az_mqtt.h
```diff
@@ -0,0 +1,391 @@
...
+#endif
+
+#include <azure/core/_az_cfg_prefix.h>
+
+/**
+ * @brief Defines the key storage types.
+ */
+typedef enum
+{
+  AZ_MQTT_X509_CLIENT_CERTIFICATE_KEY_MEMORY = 0,
```
CIPop, line 58: Throughout the file: please add doxygen documentation to all enum and struct fields.

sdk/inc/azure/platform/az_mqtt_mosquitto.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef _az_MQTT_MOSQUITTO_H
+#define _az_MQTT_MOSQUITTO_H
+
+#include "mosquitto.h"
+#include <azure/core/internal/az_mqtt_internal.h>
+
+#include <azure/core/_az_cfg_prefix.h>
+
+typedef struct
```
CIPop, line 23: Public API: please add doxygen documentation to all fields.

sdk/inc/azure/platform/az_mqtt_mosquitto.h
```diff
@@ -0,0 +1,47 @@
...
+{
+  az_mqtt_options_common platform_options;
+  /**
+   * The CA Trusted Roots span interpretable by the underlying MQTT implementation.
```
CIPop, line 27: We should also add that we recommend applications avoid configuring this and instead relying on the default OpenSSL default `ca-certificates` package.
Looks like what we need to do according to @vaavva 's tests:

> I could get this to work by setting MOSQ_OPT_TLS_USE_OS_CERTS and passing in a dummy value in ca_path that would then be ignored (ex. 's')

CIPop: Ref: https://github.com/eclipse/mosquitto/issues/1824

sdk/src/azure/core/CMakeLists.txt
```diff
@@ -94,10 +94,18 @@ target_include_directories(az_core
...
+if (APPLE) # temporary
+  target_link_libraries(az_core
+    PUBLIC
+      ${PAL}
+  )
+else()
+  target_link_libraries(az_core
+    PUBLIC
+      ${PAL}
+      ${MQTT_TRANSPORT}
```
CIPop, line 106: Nit: please ensure this isn't built on Windows either for now since we don't have an Schannel-based transport available yet (Mosquitto/Paho are both using OpenSSL AFAIK).

sdk/src/azure/iot/CMakeLists.txt
```diff
@@ -19,10 +19,18 @@ target_include_directories (az_iot_common
     ${az_SOURCE_DIR}/sdk/inc
 )
 
-target_link_libraries(az_iot_common
+if (APPLE) # temporary
```
CIPop, line 22: I would remove the "temporary" comment and ensure this applies to all except Linux for now. (No Windows support for now.)

sdk/src/azure/iot/CMakeLists.txt
```diff
@@ -19,10 +19,18 @@ target_include_directories (az_iot_common
...
+  target_link_libraries(az_iot_common
   PUBLIC
     az::core
-)
+  )
+else()
+  target_link_libraries(az_iot_common
+    PUBLIC
+      az::core
+      az::mqtt_impl
```
CIPop, line 31: An MQTT Impl should not be included for Windows.

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+        me); // callback context - i.e. user data.
+  }
+
+  if (me->mosquitto_handle == NULL)
+  {
+    ret = AZ_ERROR_OUT_OF_MEMORY;
+    return ret;
+  }
+
+  /* Configure callbacks. This should be done before connecting ideally. */
```
CIPop, line 208: Nit: please change comment style to //

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+
+  _az_RETURN_IF_FAILED(_az_result_from_mosq(mosquitto_loop_start(me->mosquitto_handle)));
+
+  return ret;
+}
+
+AZ_NODISCARD az_result
+az_mqtt_outbound_sub(az_mqtt* mqtt, az_context* context, az_mqtt_sub_data* sub_data)
+{
+  (void)context;
```
CIPop, line 247: Please apply context checking prior to all operations. 
If the application cancelled the operation, we should not call the I/O function.

See `az_context_has_expired()` - will require getting the current time.

CIPop (changes requested):
Left a few comments for changes (1/2 files reviewed for now only)

sdk/inc/azure/core/az_mqtt.h
```diff
@@ -0,0 +1,391 @@
...
+  /// MQTT Subscribe Request event.
+  AZ_MQTT_EVENT_SUB_REQ = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT, 17),
+
+  /// MQTT Subscribe Acknowledge Response event.
+  AZ_MQTT_EVENT_SUBACK_RSP = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT, 18),
+};
+
+enum
+{
+  AZ_IOT_DEFAULT_MQTT_CONNECT_KEEPALIVE_SECONDS = 240
```
vaavva, line 203: should this use the value from sdk/inc/azure/core/az_config.h (or vice versa)?

sdk/samples/iot/README.md
```diff
@@ -124,7 +124,7 @@ To run the samples, ensure you have the following programs and tools installed o
     cd vcpkg
     git checkout <vcpkg commit> # Checkout the vcpkg commit per vcpkg-commit.txt above.
     ./bootstrap-vcpkg.sh
-    ./vcpkg install --triplet x64-linux curl cmocka paho-mqtt
+    ./vcpkg install --triplet x64-linux curl cmocka paho-mqtt mosquitto
```
vaavva, line 127: Do you need libmosquitto-dev and libssl-dev as well? (you might not since you're using vcpkg)

sdk/src/azure/platform/CMakeLists.txt
```diff
@@ -82,3 +82,58 @@ if (TRANSPORT_CURL)
...
+    LIBMOSQUITTO_STATIC
+  )
+
+  target_link_libraries(az_mqtt_mosquitto PRIVATE az_core)
+
+  target_link_libraries(az_mqtt_mosquitto 
+    PUBLIC 
+      ${MOSQUITTO_LIBRARIES}
+      ${PTHREAD_LIBRARIES}
+      #${MOSQUITTO_LIBRARY}
```
vaavva, line 116:
```suggestion
```
Do we need this?

sdk/src/azure/platform/CMakeLists.txt
```diff
@@ -82,3 +82,58 @@ if (TRANSPORT_CURL)
...
+
+  target_include_directories(az_mqtt_mosquitto INTERFACE ${MOSQUITTO_INCLUDE_DIR})
+
+  # make sure that users can consume the project as a library.
+  add_library (az::mqtt_impl ALIAS az_mqtt_mosquitto)
+endif() # TRANSPORT_MOSQUITTO
+
+# No platform
+
+if (AZ_MQTT_TRANSPORT_IMPL STREQUAL "NONE" AND NOT APPLE)
```
vaavva, line 129: Does there need to be a fallthrough case for apple?

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+  return (az_mqtt_options){ .certificate_authority_trusted_roots = NULL,
+                            .openssl_engine = NULL,
+                            .mosquitto_handle = NULL };
+}
+
+AZ_NODISCARD az_result az_mqtt_init(az_mqtt* mqtt, az_mqtt_options const* options)
+{
+  _az_PRECONDITION_NOT_NULL(mqtt);
+  mqtt->_internal.options = options == NULL ? az_mqtt_options_default() : *options;
+  mqtt->mosquitto_handle = mqtt->_internal.options.mosquitto_handle;
```
vaavva, line 179: Why do we need a copy of the mosquitto_handle in the obj? Wouldn't we always have this if we have access to the obj from mosquitto?

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+  az_mqtt* me = (az_mqtt*)mqtt;
+
+  // IMPORTANT: application must call mosquitto_lib_init() before any Mosquitto clients are created.
+
+  if (me->mosquitto_handle == NULL)
+  {
+    me->mosquitto_handle = mosquitto_new(
+        (char*)az_span_ptr(connect_data->client_id),
+        false, // clean-session
+        me); // callback context - i.e. user data.
```
vaavva, line 199: Related to the question above - Isn't this sort of recursively containing `me`? Since then the mosquitto mosq holds me in the obj, which holds the mosquitto mosq?

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+
+AZ_NODISCARD az_result az_mqtt_outbound_disconnect(az_mqtt* mqtt, az_context* context)
+{
+  (void)context;
+  return _az_result_from_mosq(mosquitto_disconnect(mqtt->mosquitto_handle));
+}
+
+AZ_NODISCARD az_result az_mqtt_wait_for_event(az_mqtt* mqtt, int32_t timeout)
+{
+  return _az_result_from_mosq(mosquitto_loop(mqtt->mosquitto_handle, timeout, 1));
```
vaavva, line 277: It looks like you shouldn't be using mosquitto_loop with mosquitto_connect_async
https://mosquitto.org/api/files/mosquitto-h.html#mosquitto_connect_async

vaavva: Also I don't think you should need to use mosquitto_loop if you're using mosquitto_loop_start

sdk/tests/platform/test_az_mqtt_policy.c
```diff
@@ -0,0 +1,254 @@
...
+#include <cmocka.h>
+
+void az_platform_critical_error(void) { assert_true(false); }
+
+// MQTT information for testing
+#define TEST_MQTT_ENDPOINT "test.mosquitto.org"
+#define TEST_MQTT_PORT 1884 // Unencrypted, authenticated port.
+#define TEST_MQTT_USERNAME "rw"
+#define TEST_MQTT_PASSWORD "readwrite"
+#define TEST_MQTT_CLIENT_ID "mqtt_client_id_0505"
```
vaavva, line 30: Why this specific client id?

sdk/src/azure/platform/az_mosquitto.c
CIPop: We don't always have access: our public API can be called on any thread (application thread, timer callback thread, mosquitto callback thread, etc.)
How would `az_mqtt_outbound_connect` work for example?

sdk/src/azure/platform/az_mosquitto.c
CIPop: Not sure I understand what you are referring to as "obj". This is attaching our az_mqtt object to the Mosquitto handle so that we can retrieve it later in the callback. 

Maybe this might clarify things a bit more: we support any number of `az_mqtt` objects within the application context. This means that the callbacks can occur at the same time, on different threads, for different `az_mqtt`/mosquitto objects. We need to ensure we are operating on the correct pipeline, and that is done by passing the equivalent of the `this` pointer (`me` here) into the context.

sdk/src/azure/platform/az_mosquitto.c
CIPop: Good point! This is to support the blocking (SYNC) mode which we decided to remove for now (originally under a separate compilation flag).

@RLeclair please remove this function - it shouldn't be used. We may decide to add it back if there is any demand for a blocking / "DoEvents" type of API interface.

sdk/tests/platform/test_az_mqtt_policy.c
```diff
@@ -0,0 +1,254 @@
...
+#define TEST_MQTT_TOPIC "test_topic"
+#define TEST_MQTT_PAYLOAD "test_payload"
+
+#define TEST_MAX_RESPONSE_CHECKS 3
+#define TEST_RESPONSE_DELAY_MS 500 // Response from endpoint is slow. Temporary.
```
CIPop, line 35: Nit: I would remove temporary - we can tune these later.

sdk/tests/platform/test_az_mqtt_policy.c
CIPop: This is indeed a problem: we won't be able to run these tests in parallel (the broker will disconnect the older client with the same client ID). We will solve this issue by deploying a local MQTT broker to test against.

sdk/tests/platform/test_az_mqtt_policy.c
```diff
@@ -0,0 +1,254 @@
...
+          (az_event_policy*)&test_mqtt_policy,
+          (az_event_policy*)&test_inbound_hfsm),
+      AZ_OK);
+}
+
+#ifndef AZ_NO_PRECONDITION_CHECKING
+
+ENABLE_PRECONDITION_CHECK_TESTS()
+
+static void test_az_mqtt_init_null(void** state)
```
CIPop, line 131: Please rename all tests to use our UT convention: e.g.
`test_az_iot_provisioning_client_get_connect_info_insufficient_space_fails`

test_object_action_expectedresult

CIPop (changes requested):
Reviewed the second half (left some new comments).
Maintaining "Request changes" for comments left yesterday.

sdk/samples/iot/README.md
CIPop: The new changes shouldn't use OpenSSL in any way. 
Dependencies of Mosquitto will be addressed by VCPKG.

sdk/src/azure/platform/az_mosquitto.c
vaavva: obj is the third parameter of mosquitto_new: "A user pointer that will be passed as an argument to any callbacks that are specified."
We get it back from mosquitto in functions like `static void _az_mosquitto_on_disconnect(struct mosquitto* mosq, void* obj, int rc)`

It makes sense that the az_mqtt would contain the mosquitto_handle because of threads/different implementations, but I don't think we should be passing in the mosquitto_handle as part of the obj for the mosquitto implemetation, since every callback also gives us the mosquitto handle as the `mosq` parameter.

For mosquitto, the `mosquitto_handle` (returned on `mosquitto_new`), contains the obj (`me` for our wording), accessible with [mosquitto_userdata](https://mosquitto.org/api/files/mosquitto-h.html#mosquitto_userdata). But because we put `me` in the obj, it's a little weird where
`mosquitto_userdata(me->mosquitto_handle)` returns me, which you could then use to get the userdata on that mosquitto_handle property of that value and get `me` again.

Let me know if this doesn't clarify things!

sdk/inc/azure/core/az_mqtt.h
```diff
@@ -0,0 +1,391 @@
...
+
+/**
+ * @brief Waits for an event to be posted to the MQTT event pipeline.
+ *
+ * @param mqtt The MQTT instance.
+ * @param timeout The maximum time to wait for an event, in milliseconds.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD az_result az_mqtt_wait_for_event(az_mqtt* mqtt, int32_t timeout);
```
CIPop, line 387: Should be removed.

sdk/inc/azure/platform/az_mqtt_mosquitto.h
```diff
@@ -0,0 +1,47 @@
...
+
+struct az_mqtt
+{
+  struct
+  {
+    az_mqtt_common platform_mqtt;
+    az_mqtt_options options;
+  } _internal;
+
+  struct mosquitto* mosquitto_handle;
```
vaavva, line 42: Should this be internal?

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,278 @@
...
+
+  return ret;
+}
+
+static void _az_mosquitto_on_connect(struct mosquitto* mosq, void* obj, int reason_code)
+{
+  az_result ret;
+  az_mqtt* me = (az_mqtt*)obj;
+
+  _az_PRECONDITION(mosq == me->mosquitto_handle);
```
vaavva, line 51: Sync'd with Chris on the obj being `me`, and we discussed that it should be fine, but all of the callbacks should have this _az_PRECONDITION check and then not use `mosq` past creating the `me` var for consistency (ex. the mosquitto_disconnect should pass in me->mosquitto_handle instead of mosq)

vaavva (commented):
Sync'd with Chris on the `me`/obj discussion and these were our takeaways

CMakeLists.txt
RLeclair: Got it, removed the temporary, I had set it there more for our reference once we merge with the main branch (in case we were to test for other mqtt clients). Given that we're **only** doing Mosquitto I'll remove the temporary.

sdk/inc/azure/core/az_mqtt.h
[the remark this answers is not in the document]
RLeclair: Removed the default configurations from common and moved to `az_mqtt.h`

sdk/inc/azure/core/az_config.h
```diff
@@ -30,6 +30,9 @@ enum
 
   /// The maximum buffer size for a log message.
   AZ_LOG_MESSAGE_BUFFER_SIZE = 1024,
+
+  /// The MQTT keepalive in seconds.
+  AZ_MQTT_KEEPALIVE_SECONDS = 240,
```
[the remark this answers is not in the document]
RLeclair: Removed!

sdk/inc/azure/core/az_result.h
```diff
@@ -26,10 +26,11 @@ enum
   _az_FACILITY_CORE_PLATFORM = 0x2,
   _az_FACILITY_CORE_JSON = 0x3,
   _az_FACILITY_CORE_HTTP = 0x4,
-  _az_FACILITY_IOT = 0x5,
-  _az_FACILITY_IOT_MQTT = 0x6,
-  _az_FACILITY_ULIB = 0x7,
-  _az_FACILITY_HFSM = 0x8,
+  _az_FACILITY_CORE_MQTT = 0x5,
```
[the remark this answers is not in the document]
RLeclair: Moved `_az_FACILITY_CORE_MQTT` to the bottom

sdk/inc/azure/core/internal/az_mqtt_policy.h
RLeclair: Renamed, thanks for catching that

sdk/src/azure/platform/CMakeLists.txt
RLeclair: No we don't, thanks for catching that!

sdk/src/azure/platform/CMakeLists.txt
RLeclair: We don't want to build any of the `az_mqtt` files for apple (even `az_nomqtt`) since they depend on `az_platform` which we have not implemented for it. This made me think that I should replace the `NOT APPLE` conditional with `(UNIX AND NOT APPLE)` since we only want to build this for linux

sdk/src/azure/platform/az_mosquitto.c
RLeclair: Removed!

sdk/src/azure/platform/az_mosquitto.c
RLeclair: Added precondition checks + removed usage of mosq

sdk/inc/azure/core/az_mqtt.h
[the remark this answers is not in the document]
RLeclair: I think we should not do this change, there are many samples (like `paho_iot_hub_sas_telemetry_sample` that need `AZ_IOT_DEFAULT_...` in common. We can't include `az_mqtt.h` on those samples given that they would then need to be aware of `az_mqtt` and consequently platform changes so Windows and Mac OS won't be able to run these samples

sdk/inc/azure/platform/az_mqtt_mosquitto.h
```diff
@@ -20,25 +20,58 @@
...
+   */
   az_mqtt_options_common platform_options;
+
   /**
-   * The CA Trusted Roots span interpretable by the underlying MQTT implementation.
+   * @brief The CA Trusted Roots span interpretable by the underlying MQTT implementation.
+   *
+   * @details We recommend avoiding configuring this option and instead using the default
+   * OpenSSL trusted roots. You can configure the trusted roots by setting
+   * `MOSQ_OPT_TLS_USE_OS_CERTS`.
```
vaavva, line 39: Does this work yet in the code?

sdk/inc/azure/core/az_mqtt.h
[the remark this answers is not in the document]
CIPop: I agree if by not doing the change you mean removing it from az_config.h for now.
`AZ_IOT_DEFAULT_*` is for IoT Hub / DPS _only_ and does not apply to all MQTT connections (i.e., those values would override this setting).

CIPop (approved):
Thanks @RLeclair !

vaavva (approved):

sdk/inc/azure/core/az_mqtt.h
RLeclair: Moved values to `az_mqtt_config.h` as per disucssions

sdk/inc/azure/platform/az_mqtt_mosquitto.h
RLeclair: Added changes to rely on OS certificates

sdk/src/azure/platform/az_mosquitto.c
RLeclair: Removed `az_context` from `az_mqtt.h` moved to `az_mqtt_policy` instead

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -0,0 +1,290 @@
...
+        NULL)));
+  }
+  else
+  {
+    _az_RETURN_IF_FAILED(_az_result_from_mosq(
+        mosquitto_int_option(me->_internal.mosquitto_handle, MOSQ_OPT_TLS_USE_OS_CERTS, 1)));
+    // Passing any string to mosquitto_tls_set() will use the OS certs when
+    // MOSQ_OPT_TLS_USE_OS_CERTS is set.
+    _az_RETURN_IF_FAILED(_az_result_from_mosq(
+        mosquitto_tls_set(me->_internal.mosquitto_handle, NULL, "L", NULL, NULL, NULL)));
```
CIPop, line 244: Nit: please #define "L" to mean REQUIRED_TLS_SET_CERT_PATH.
When you have some time, please file a bug with Mosquitto MQTT to describe this as a bug and ask if they could fix it so that we can pass in NULL (or, if necessary, some other #defined value).

CIPop (approved):
Thanks @RLeclair !

PR #2591, opened 2023-06-20
Adding mqtt connection with unit tests, subclient

RLeclair:
Ignore TODO_L comments. They represent unimplemented features.

Adds: 
- mqtt_connection + unit tests
- subclients
- mqtt mock class


sdk/src/azure/core/az_event_policy_subclients.c
```diff
@@ -0,0 +1,105 @@
...
+
+  // The client is added to the end of the list.
+  subclient->next = NULL;
+
+  _az_event_subclient* last = subclients_policy->subclients;
+  if (last == NULL)
+  {
+    subclients_policy->subclients = subclient;
+  }
+  else
```
vaavva, line 94: Does the order of subclients matter? If not, we could just add the new one to the front of the list

sdk/src/azure/core/az_mqtt_connection_hfsm.c
```diff
@@ -0,0 +1,384 @@
...
+  }
+
+  return parent_state;
+}
+
+static az_result root(az_event_policy* me, az_event event)
+{
+  az_result ret = AZ_OK;
+
+  if (_az_LOG_SHOULD_WRITE(event.type))
```
vaavva, line 55: Probably not the place for this comment, but if we're passing the event.type into _az_LOG_WRITE, why isn't should_write checked within there?

sdk/src/azure/core/az_mqtt_connection_hfsm.c
```diff
@@ -0,0 +1,384 @@
...
+      break;
+  }
+
+  return ret;
+}
+
+static az_result connected(az_event_policy* me, az_event event)
+{
+  az_result ret = AZ_OK;
+  (void)me;
```
vaavva, line 313:
```suggestion
```
Should this be removed since we use `me` lower down?

sdk/src/azure/platform/CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ if (AZ_MQTT_TRANSPORT_IMPL STREQUAL "NONE" AND (UNIX AND NOT APPLE))
     az_mqtt_notransport 
       STATIC
       ${CMAKE_CURRENT_LIST_DIR}/az_nomqtt.c
+      ${CMAKE_CURRENT_LIST_DIR}/az_mqtt_policy.c
```
vaavva, line 133: is this needed when there's no mqtt implementation?

sdk/src/azure/platform/az_mosquitto.c
```diff
@@ -229,22 +229,44 @@ AZ_NODISCARD az_result az_mqtt_outbound_connect(az_mqtt* mqtt, az_mqtt_connect_d
...
   if (az_span_size(me->_internal.options.certificate_authority_trusted_roots) != 0)
   {
-    _az_RETURN_IF_FAILED(_az_result_from_mosq(mosquitto_tls_set(
-        me->_internal.mosquitto_handle,
-        (const char*)az_span_ptr(me->_internal.options.certificate_authority_trusted_roots),
-        NULL,
-        (const char*)az_span_ptr(connect_data->certificate.cert),
-        (const char*)az_span_ptr(connect_data->certificate.key),
-        NULL)));
+    if (connect_data->use_client_certificate)
```
vaavva, line 232: Why do we need a separate boolean for this instead of just checking if they've provided connect_data->certificate.cert/connect_data->certificate.key?

I'd also vote to combine these three `mosquitto_tls_set` calls into one since all of the options are something vs NULL

for example, I do this, where ca_file, cert_file, and key_file are NULL if there isn't a value:
```
    if (use_OS_certs)
    {
      MQTT_RETURN_IF_FAILED(mosquitto_int_option(mosq, MOSQ_OPT_TLS_USE_OS_CERTS, true));
    }
    MQTT_RETURN_IF_FAILED(mosquitto_tls_set(
        mosq,
        connection_settings.ca_file,
        use_OS_certs ? REQUIRED_TLS_SET_CERT_PATH : NULL,
        connection_settings.cert_file,
        connection_settings.key_file,
        NULL));
```

sdk/tests/core/CMakeLists.txt
```diff
@@ -24,13 +31,15 @@ endif()
 if (AZ_PLATFORM_IMPL STREQUAL "POSIX" AND NOT APPLE)
     add_cmocka_test(az_core_test SOURCES
                     main.c
+                    mock_az_mqtt.c
```
vaavva, line 34: does `_az_MOCK_ENABLED` need to be checked here as well?

sdk/tests/core/mock_az_mqtt.c
```diff
@@ -0,0 +1,74 @@
...
+#include <azure/core/internal/az_result_internal.h>
+#include <azure/core/internal/az_span_internal.h>
+
+#include <mosquitto.h>
+#include <pthread.h>
+#include <stdlib.h>
+
+#include <azure/core/_az_cfg.h>
+
+#ifdef _az_MOCK_ENABLED
```
vaavva, line 26: Should this surround the #includes too?

sdk/tests/core/test_az_mqtt_connection.c
```diff
@@ -0,0 +1,240 @@
...
+#include <azure/core/internal/az_event_policy_subclients_internal.h>
+#include <azure/core/internal/az_hfsm_internal.h>
+
+#include <setjmp.h>
+#include <stdarg.h>
+#include <stddef.h>
+
+#include <cmocka.h>
+
+#if defined(_az_MOCK_ENABLED)
```
vaavva, line 18: same here

sdk/tests/core/test_az_mqtt_connection.c
```diff
@@ -0,0 +1,240 @@
...
+
+  test_connection_options.hostname = AZ_SPAN_FROM_STR(TEST_HOSTNAME);
+  test_connection_options.client_id_buffer = AZ_SPAN_FROM_STR(TEST_CLIENT_ID);
+  test_connection_options.username_buffer = AZ_SPAN_FROM_STR(TEST_USERNAME);
+  test_connection_options.password_buffer = AZ_SPAN_FROM_STR(TEST_PASSWORD);
+  test_connection_options.port = 8883;
+  test_connection_options.use_username_password = true;
+  test_connection_options.connection_management = true;
+  az_mqtt_x509_client_certificate test_cert = {
+    .cert = AZ_SPAN_FROM_STR("test-cert"),
```
vaavva, line 123: why aren't `test-cert`/the other connection options macros too?

sdk/tests/core/test_az_mqtt_connection.c
```diff
@@ -0,0 +1,240 @@
...
+      AZ_OK);
+
+  assert_int_equal(ref_conn_rsp, 2);
+}
+
+static void test_az_mqtt_connection_enabled_pub_send_success(void** state)
+{
+  (void)state;
+
+  ref_pub_rsp = 0;
```
vaavva, line 181: these tests can't run in parallel, right?

vaavva (approved):

sdk/src/azure/core/az_event_policy_subclients.c
CIPop: Good point: my original plan was to order them then stop processing once one of the clients accepted the message. 
The way it's written now, ordering does not matter, and we could add to the front and avoid the enumeration. We can think of these optimizations later, as needed.

sdk/src/azure/core/az_mqtt_connection_hfsm.c
CIPop: This is following logging best practices for az_sdk (and other SDKs).
The main reason is to avoid extra computation for logging when there is no listener attached (this is not the case but imagine that a function call is made to compute one of the logging parameters).
The second reason is that if the compiler can determine that no listener is ever attached, this code should be pruned.

sdk/src/azure/core/az_mqtt_connection_hfsm.c
```diff
@@ -0,0 +1,384 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
CIPop, line 1: Nit: This code is preliminary. We will update after

sdk/src/azure/platform/CMakeLists.txt
CIPop: Yeah... it's kind of odd but az_nomqtt.c _is_ an interface implementation. If there really isn't any MQTT we won't be compiling the stateful clients at all (since they require the interface).

sdk/tests/core/CMakeLists.txt
```diff
@@ -11,7 +11,14 @@ include(AddCMockaTest)
 
 # -ld link option is only available for gcc
 if(UNIT_TESTING_MOCKS)
-    set(WRAP_FUNCTIONS "-Wl,--wrap=az_platform_clock_msec -Wl,--wrap=az_platform_sleep_msec")
+    set(WRAP_FUNCTIONS "-Wl,--wrap=az_platform_clock_msec \
```
CIPop, line 14: Why do we need these `--wrap` statements?

sdk/tests/core/mock_az_mqtt.c
CIPop: +1 - please check with various CMake options to ensure we can still build with or without mocks enabled.

sdk/tests/core/mock_az_mqtt.c
```diff
@@ -0,0 +1,74 @@
...
+AZ_NODISCARD az_result __wrap_az_mqtt_init(az_mqtt* mqtt, az_mqtt_options const* options)
+{
+  (void)mqtt;
+  (void)options;
+
+  return AZ_OK;
+}
+
+AZ_NODISCARD az_result
+__wrap_az_mqtt_outbound_connect(az_mqtt* mqtt, az_mqtt_connect_data* connect_data);
```
CIPop, line 38: I'm not sure I understand why this is needed: can't we simply provide an az_mqtt.h implementation and compile that into the UT instead of using GCC's mocks?

sdk/tests/core/test_az_mqtt_connection.c
CIPop: We only run separate test processes in parallel.

CIPop (approved):
I don't really like the new dependency on GCC Mocks. We should avoid that in order to allow running UTs on various compilers.

CIPop:
> I don't really like the new dependency on GCC Mocks. We should avoid that in order to allow running UTs on various compilers.

@vhvb1989 looks like the SDK is already using CMocka and GCC-specific flags. If we want to continue doing so, the tests here are correct (and we will limit UTs to GCC).

sdk/tests/core/CMakeLists.txt
RLeclair: As discussed, we're sticking with GCC mock functions, they require these --wrap statements to be linked

sdk/tests/core/CMakeLists.txt
RLeclair: Thanks for pointing it out, only `UNIT_TESTING_MOCKS` should be checked, `az_MOCK_ENABLED` is a compile time definition

sdk/tests/core/mock_az_mqtt.c
RLeclair: Resolving as part of our discussion regarding gcc mocks.

sdk/inc/azure/core/az_mqtt_connection.h
```diff
@@ -0,0 +1,250 @@
...
+   * @brief The port of the MQTT broker.
+   *
+   */
+  int16_t port;
+
+  /**
+   * @brief Denotes whether the MQTT connection should be managed by the SDK.
+   *
+   */
+  bool connection_management;
```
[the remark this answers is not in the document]
RLeclair: Repurposed to the above suggestion.

sdk/inc/azure/core/az_mqtt_connection.h
```diff
@@ -0,0 +1,250 @@
...
+   * @brief The password for the MQTT connection. REQUIRED if connection_management is true.
+   *
+   */
+  az_span password_buffer;
+
+  /**
+   * @brief Specifies whether the MQTT connection should use a username and password. REQUIRED if
+   * connection_management is true.
+   */
+  bool use_username_password;
```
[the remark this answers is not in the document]
RLeclair: Removed bool and now checking for username and password instead

sdk/inc/azure/core/internal/az_event_policy_subclients_internal.h
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * @brief The next subclient in the orthogonal region.
+   *
+   */
+  _az_event_subclient* next;
+};
+
+/**
+ * @brief The first subclient in an orthogonal region.
```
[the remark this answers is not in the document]
RLeclair: Changed the name of the file and provided more clarification, thanks!

sdk/src/azure/core/az_mqtt_connection_hfsm.c
RLeclair: Removed!

sdk/src/azure/platform/az_mosquitto.c
RLeclair: Removed `use_client_certificate` and added a similar approach to the above.

sdk/tests/core/mock_az_mqtt.c
RLeclair: I removed the `_az_MOCK_ENABLED`. Instead, if `_az_MOCK_ENABLED` is not set, this file will not compile

sdk/tests/core/test_az_mqtt_connection.c
RLeclair: Removed for same reasons as above

sdk/tests/core/test_az_mqtt_connection.c
RLeclair: Thanks for pointing this out! Added them as macros

PR #2616, opened 2023-07-27
Removing pw requirement when user is present and setting to null after free

RLeclair:
Addressing comments made to the MQTTv5 PR.

### DIFF
--- a/sdk/src/azure/platform/az_mosquitto5.c
+++ b/sdk/src/azure/platform/az_mosquitto5.c
@@ -313,7 +313,7 @@ az_mqtt5_outbound_connect(az_mqtt5* mqtt5, az_mqtt5_connect_data* connect_data)
         NULL)));
   }
 
-  if (az_span_ptr(connect_data->username) != NULL && az_span_ptr(connect_data->password) != NULL)
+  if (az_span_ptr(connect_data->username) != NULL)
   {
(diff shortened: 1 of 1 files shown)


vaavva (approved):

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -584,6 +584,8 @@ void az_mqtt5_property_string_free(az_mqtt5_property_string* prop_str)
   _az_PRECONDITION_NOT_NULL(prop_str);
 
   free((void*)prop_str->str._internal.ptr);
+  prop_str->str._internal.ptr = NULL;
```
CIPop, line 587: Code outside the `az_span*.c` should _never_ access _internal.

Instead, use:

```C
prop_str->str = AZ_SPAN_EMPTY
```

Same goes for free, etc. In that case you must use `az_span_ptr`.  (Don't assume anything about how a span is implemented: anything in _internal as well as any function behavior may change in the future.)

CIPop (commented):
@RLeclair, the PR is not following https://azure.github.io/azure-sdk/clang_design.html#repository-guidelines

sdk/src/azure/platform/az_mosquitto5.c
RLeclair: Applied those changes, thanks for pointing it out

PR #2619, opened 2023-08-08
Addressing comments for properties v5

RLeclair:
- Adding `az_mqtt5_property_bag_destroy`

Addressed comments on:
#2613 
#2616

### DIFF
--- a/sdk/inc/azure/core/az_mqtt5.h
+++ b/sdk/inc/azure/core/az_mqtt5.h
@@ -30,6 +30,59 @@
  * and they are subject to change in future versions of the SDK which would break your code.
  */
 
+/**
+ * @page mqtt5port Porting a MQTTv5 implementation
+ *
+ * @section section1 Creating an implementation specific header file:
+ * Create a implementation specific header file (i.e az_mqtt5_[mymqtt].h)
+ *
+ * The following structs must be implemented:
+ * - #az_mqtt5_options
+ * - #az_mqtt5
+ * - #az_mqtt5_property_bag
+ * - #az_mqtt5_property_string
+ * - #az_mqtt5_property_stringpair
+ * - #az_mqtt5_property_binarydata
+ *
+ * The following function must be created:
+ * - #az_mqtt5_property_bag_init with the last argument being specific to the implementation.
+ *
+ * @section section2 Functions that must be implemented:
+ * The following functions must be implemented and will be called by the SDK to send data:
+ *
+ * In az_mqtt5.h:
+ * - #az_mqtt5_options_default
+ * - #az_mqtt5_init
+ * - #az_mqtt5_outbound_connect
+ * - #az_mqtt5_outbound_sub
+ * - #az_mqtt5_outbound_pub
+ * - #az_mqtt5_outbound_disconnect
+ *
+ * The following functions must be implemented and will be called by the SDK to interact with the
+ * MQTT5 properties.
+ *
+ * In az_mqtt5_property_bag.h:
+ * - #az_mqtt5_property_bag_empty
+ * - #az_mqtt5_property_bag_string_append
+ * - #az_mqtt5_property_bag_stringpair_append
+ * - #az_mqtt5_property_bag_byte_append
+ * - #az_mqtt5_property_bag_int_append
+ * - #az_mqtt5_property_bag_binary_append
+ * - #az_mqtt5_property_bag_string_read
+ * - #az_mqtt5_property_bag_stringpair_find
+ * - #az_mqtt5_property_bag_byte_read
+ * - #az_mqtt5_property_bag_int_read
+ * - #az_mqtt5_property_bag_binarydata_read
+ * - #az_mqtt5_property_string_get
+ * - #az_mqtt5_property_stringpair_key_get
+ * - #az_mqtt5_property_stringpair_value_get
+ * - #az_mqtt5_property_binarydata_get
+ * - #az_mqtt5_property_string_free
+ * - #az_mqtt5_property_stringpair_free
+ * - #az_mqtt5_property_binarydata_free
+ *
+ */
+
 #ifndef _az_MQTT5_H
 #define _az_MQTT5_H
 
@@ -324,9 +377,6 @@ enum az_event_type_mqtt5
   AZ_MQTT5_EVENT_SUBACK_RSP = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT5, 18),
 };
 
-// Porting az_mqtt 1. The following functions must be called by the implementation when data is
-// received:
-
 /**
  * @brief Posts a MQTT 5 publish receive indication event to the event pipeline.
  *
@@ -432,9 +482,6 @@ az_mqtt5_inbound_disconnect(az_mqtt5* mqtt5, az_mqtt5_disconnect_data* disconnec
   return _az_event_pipeline_post_inbound_event(
       pipeline, (az_event){ .type = AZ_MQTT5_EVENT_DISCONNECT_RSP, .data = disconnect_data });
 }
-
-// Porting az_mqtt 2. The following functions must be implemented and will be called by the SDK to
-//            send data:
 
 /**
  * @brief The default MQTT 5 options.
--- a/sdk/inc/azure/core/az_mqtt5.h
+++ b/sdk/inc/azure/core/az_mqtt5.h
@@ -63,23 +63,23 @@
  *
  * In az_mqtt5_property_bag.h:
  * - #az_mqtt5_property_bag_empty
- * - #az_mqtt5_property_bag_string_append
- * - #az_mqtt5_property_bag_stringpair_append
- * - #az_mqtt5_property_bag_byte_append
- * - #az_mqtt5_property_bag_int_append
- * - #az_mqtt5_property_bag_binary_append
- * - #az_mqtt5_property_bag_string_read
- * - #az_mqtt5_property_bag_stringpair_find
- * - #az_mqtt5_property_bag_byte_read
- * - #az_mqtt5_property_bag_int_read
- * - #az_mqtt5_property_bag_binarydata_read
- * - #az_mqtt5_property_string_get
- * - #az_mqtt5_property_stringpair_key_get
- * - #az_mqtt5_property_stringpair_value_get
- * - #az_mqtt5_property_binarydata_get
- * - #az_mqtt5_property_string_free
- * - #az_mqtt5_property_stringpair_free
- * - #az_mqtt5_property_binarydata_free
+ * - #az_mqtt5_property_bag_append_string
+ * - #az_mqtt5_property_bag_append_stringpair
+ * - #az_mqtt5_property_bag_append_byte
+ * - #az_mqtt5_property_bag_append_int
+ * - #az_mqtt5_property_bag_append_binary
+ * - #az_mqtt5_property_bag_read_string
+ * - #az_mqtt5_property_bag_find_stringpair
+ * - #az_mqtt5_property_bag_read_byte
+ * - #az_mqtt5_property_bag_read_int
+ * - #az_mqtt5_property_bag_read_binarydata
+ * - #az_mqtt5_property_get_string
+ * - #az_mqtt5_property_stringpair_get_key
+ * - #az_mqtt5_property_stringpair_get_value
+ * - #az_mqtt5_property_get_binarydata
+ * - #az_mqtt5_property_free_string
+ * - #az_mqtt5_property_free_stringpair
+ * - #az_mqtt5_property_free_binarydata
  *
  */
 
--- a/sdk/inc/azure/core/az_mqtt5.h
+++ b/sdk/inc/azure/core/az_mqtt5.h
@@ -74,6 +74,8 @@ typedef struct
 {
   /**
    * @brief The properties of the publish request.
+   *
+   * @details Set to NULL if no properties are required.
    */
   az_mqtt5_property_bag* properties;
 
@@ -157,6 +159,8 @@ typedef struct
 {
   /**
    * @brief The properties of subscribe request.
+   *
+   * @details Set to NULL if no properties are required.
    */
   az_mqtt5_property_bag* properties;
   /**
@@ -200,6 +204,8 @@ typedef struct
 {
   /**
    * @brief The properties of the connect request.
+   *
+   * @details Set to NULL if no properties are required.
    */
   az_mqtt5_property_bag* properties;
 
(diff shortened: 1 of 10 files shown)


sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -54,6 +54,8 @@ static int ref_puback = 0;
 static int ref_recv = 0;
 static int ref_disconnect = 0;
 
+static int expect_msg_properties = 0;
```
vaavva, line 57: bool instead of int? Or change checks to `if (expect_msg_properties != 0)` or something like that

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -211,6 +216,7 @@ static void test_az_mqtt5_policy_init_success(void** state)
   ref_puback = 0;
   ref_recv = 0;
   ref_disconnect = 0;
+  expect_msg_properties = 0;
```
vaavva, line 219: does this need to be set here?

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -284,11 +290,44 @@ static void test_az_mqtt5_policy_outbound_sub_success(void** state)
...
+
+  int retries = TEST_MAX_RESPONSE_CHECKS;
+  while ((ref_puback == 0 || ref_recv == 0) && retries > 0)
+  {
+    assert_int_equal(az_platform_sleep_msec(TEST_RESPONSE_DELAY_MS), AZ_OK);
+    retries--;
+  }
+
+  assert_int_equal(ref_puback, 1);
+  assert_true(ref_recv > 0);
```
vaavva, line 322: I think we need `az_mqtt5_property_bag_destroy(&test_mqtt5_property_bag);` here too

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -104,79 +106,82 @@ static az_result test_inbound_hfsm_root(az_event_policy* me, az_event event)
...
-          az_mqtt5_property_bag_binarydata_read(
-              test_recv_data->properties,
-              AZ_MQTT5_PROPERTY_TYPE_CORRELATION_DATA,
-              &test_mqtt5_property_binary_data),
-          AZ_OK);
-      assert_true(az_span_is_content_equal(
-          test_mqtt5_property_binary_data.bindata,
-          AZ_SPAN_FROM_BUFFER(TEST_MQTT_PROPERTY_CORRELATION_DATA)));
-      az_mqtt5_property_binarydata_free(&test_mqtt5_property_binary_data);
+      if (expect_msg_properties)
```
vaavva, line 109: If we're not exposing a check to see if the property bag is empty, then these functions should still be able to run and return a failure if there aren't any properties - I would propose the test to have an else on this if that looks like this:
```suggestion
if (!expect_msg_properties)
      az_mqtt5_property_string test_mqtt5_property_string;
      assert_int_equal(
          az_mqtt5_property_bag_string_read(
              test_recv_data->properties,
              AZ_MQTT5_PROPERTY_TYPE_CONTENT_TYPE,
              &test_mqtt5_property_string),
          AZ_ERROR_ITEM_NOT_FOUND);
      // not sure what this should be checked against here
      assert_true(az_span_is_content_equal(
          az_mqtt5_property_string_get(&test_mqtt5_property_string),
          AZ_SPAN_EMPTY));
      az_mqtt5_property_string_free(&test_mqtt5_property_string);
```

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -216,6 +216,8 @@ static void _az_mosquitto5_on_message(
                              .topic = az_span_create_from_str(message->topic),
                              .properties = &property_bag });
 
+  az_mqtt5_property_bag_destroy(&property_bag);
```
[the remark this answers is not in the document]
RLeclair: Removed it in latest iteration, it should not be there

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -284,11 +290,44 @@ static void test_az_mqtt5_policy_outbound_sub_success(void** state)
   assert_int_equal(ref_suback, 1);
 }
 
+static void test_az_mqtt5_policy_outbound_pub_no_properties_success(void** state)
+{
```
vaavva, line 294: If we're requiring an empty property bag on a pub without properties, should we also be requiring an empty property bag on a sub with no properties? Internally, we also have
> sub_data->properties ? sub_data->properties->properties : NULL));

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -345,20 +347,20 @@ AZ_NODISCARD az_result az_mqtt5_outbound_sub(az_mqtt5* mqtt5, az_mqtt5_sub_data*
...
+  return _az_result_from_mosq5(mosquitto_publish_v5(
       mqtt5->_internal.mosquitto_handle,
       &pub_data->out_id,
       (char*)az_span_ptr(pub_data->topic), // Assumes properly formed NULL terminated string.
       az_span_size(pub_data->payload),
       az_span_ptr(pub_data->payload),
       pub_data->qos,
       false,
-      pub_data->properties ? pub_data->properties->properties : NULL));
+      pub_data->properties->properties));
```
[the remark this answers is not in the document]
RLeclair: Yes, we are letting the implementation handle the case where there are no properties

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: Changed check to `!=0`

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: I set all of the "reference" variables at the start of the test suite, but it is not necessary

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: Added it, thanks for pointing that out

sdk/src/azure/platform/az_mosquitto5.c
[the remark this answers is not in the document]
RLeclair: Updating thread from discussion, we will let the application specify properties as `NULL`

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: Changed implementation to not require a property bag if it will not be used.

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -340,7 +340,7 @@ AZ_NODISCARD az_result az_mqtt5_outbound_sub(az_mqtt5* mqtt5, az_mqtt5_sub_data*
       (char*)az_span_ptr(sub_data->topic_filter),
       sub_data->qos,
       0,
-      sub_data->properties ? sub_data->properties->properties : NULL));
+      (sub_data->properties == NULL) ? NULL : sub_data->properties->properties));
```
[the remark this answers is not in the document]
RLeclair: We are currently calling `mosquitto_connect_async` to connect which does not accept properties. From what I checked, there is not a `mosquitto_connect_async_v5` in the mosquitto API. The closest would be `mosquitto_connect_bind_v5` which takes different parameters. Since at the moment we are not using properties on connect I am keeping it as is.

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -318,7 +324,8 @@ enum az_event_type_mqtt5
   AZ_MQTT5_EVENT_SUBACK_RSP = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT5, 18),
 };
 
-// Porting 1. The following functions must be called by the implementation when data is received:
+// Porting az_mqtt 1. The following functions must be called by the implementation when data is
```
CIPop, line 327: We should extract this into a separate doxygen section.

sdk/inc/azure/core/az_mqtt5_property_bag.h
```diff
@@ -148,6 +103,13 @@ AZ_NODISCARD az_mqtt5_property_bag_options az_mqtt5_property_bag_options_default
  */
 AZ_NODISCARD az_result az_mqtt5_property_bag_empty(az_mqtt5_property_bag* property_bag);
 
+/**
+ * @brief Destroys a MQTT 5 property bag instance.
+ *
+ * @param[in] property_bag A pointer to an #az_mqtt5_property_bag instance to destroy.
+ */
+void az_mqtt5_property_bag_destroy(az_mqtt5_property_bag* property_bag);
```
CIPop, line 111: I don't think we should have this: our SDK should not be able to allocate or de-allocate resources.

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
```diff
@@ -105,6 +105,50 @@ typedef struct
   mosquitto_property* properties;
```
CIPop, line 105: I think this should be renamed to mosq_properties or something more specific.

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -315,50 +315,50 @@ az_mqtt5_outbound_connect(az_mqtt5* mqtt5, az_mqtt5_connect_data* connect_data)
...
+  return _az_result_from_mosq5(mosquitto_publish_v5(
       mqtt5->_internal.mosquitto_handle,
       &pub_data->out_id,
       (char*)az_span_ptr(pub_data->topic), // Assumes properly formed NULL terminated string.
       az_span_size(pub_data->payload),
       az_span_ptr(pub_data->payload),
       pub_data->qos,
       false,
-      pub_data->properties ? pub_data->properties->properties : NULL));
+      (pub_data->properties == NULL) ? NULL : pub_data->properties->properties));
```
CIPop, line 356: This is hard to parse: as noted above, I would rename the Mosquitto-specific properties making this `pub_data->properties->mosq_property_handler` or similar.

sdk/src/azure/platform/az_mosquitto5.c
```diff
@@ -390,6 +390,15 @@ AZ_NODISCARD az_result az_mqtt5_property_bag_empty(az_mqtt5_property_bag* proper
   return AZ_OK;
 }
 
+void az_mqtt5_property_bag_destroy(az_mqtt5_property_bag* property_bag)
+{
+  _az_PRECONDITION_NOT_NULL(property_bag);
+
+  mosquitto_property_free_all(&property_bag->properties);
```
CIPop, line 397: As above: app should directly call `mosquitto_property_free_all`.

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -54,13 +54,16 @@ static int ref_puback = 0;
 static int ref_recv = 0;
```
CIPop, line 54: Nit: shouldn't this test be called `test_az_mosquitto5_policy.c`? It could also be general but it would need lots of `#ifdef` to switch between the various MQTT implementations we will support.

CIPop (changes requested):
We need to remove the az_mqtt5_property_destroy() as it breaks our overall SDK guidance to avoid allocations/deallocations.

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: We discussed that the tests in the platform folder would be integration tests in the same way we have the ones for POSIX and Win32 (future implementation).

sdk/inc/azure/core/az_mqtt5.h
RLeclair: Not sure how that is meant to look, is this something like:

```
/**
* @brief Porting Guide
* 
* Porting 1: The following functions in az_mqtt5.h must be implemented 
* and will be called by the SDK to send data:
* #az_mqtt5_init, ... , #az_mqtt5_outbound_disconnect.
* 
* Porting 2: The following functions in az_mqtt5_property_bag.h must be ...
**/
```

Wouldn't it be better to have a readme documenting the porting process? It could live in sdk/docs/platform

CIPop: I don't like having documentation in 2 different places. I would prefer everything is in one place: doxygen.
We can also have a readme.md that points to the doxygen documentation.

CIPop: Also, you would need to generate the doxygen html and check to see if what you have looks good.

RLeclair: I'll keep everything in doxygen, what I am still unsure about is what you mean by separate? Is it just to change the porting comments from `//` to `/** **/`? Or a doxygen comment like what I've written above which can be placed at the top of `az_mqtt.h`

CIPop: I was looking at the Doxygen docs: We could use either
1. @page https://www.doxygen.nl/manual/commands.html#cmdpage (and @subpage)
2. @module - https://www.doxygen.nl/manual/grouping.html

RLeclair: Used @ page

RLeclair:
@vaavva I made minor changes to the MQTT rpc server sample, please take a look when you have the time!

CIPop (approved):
Thanks @RLeclair !

vaavva:
> @vaavva I made minor changes to the MQTT rpc server sample, please take a look when you have the time!

These changes are good with me!

vaavva (approved):

CIPop (approved):

sdk/inc/azure/core/az_mqtt5_property_bag.h
```diff
@@ -146,7 +133,7 @@ AZ_NODISCARD az_mqtt5_property_bag_options az_mqtt5_property_bag_options_default
  *
  * @return An #az_result value indicating the result of the operation.
  */
-AZ_NODISCARD az_result az_mqtt5_property_bag_empty(az_mqtt5_property_bag* property_bag);
+AZ_NODISCARD az_result az_mqtt5_property_bag_clear(az_mqtt5_property_bag* property_bag);
```
vaavva, line 136: From the review, should this be az_mqtt5_clear_property_bag?

vaavva (approved):

Issue #1918, opened 2021-09-09
Document more explicitly to not use _internal only fields with in-source comments


ahsonkhan:
Upstream dependencies shouldn't be relying on or referencing fields that are marked `_internal`. Similarly, types and methods prefixed with `_az` are private implementation (that aren't documented) and shouldn't be used.

For example, add a comment for each nested `_internal` struct within the repo:
https://github.com/Azure/azure-sdk-for-c/blob/52b7d3c4a005924739b420d2acffe275f0a6c9a5/sdk/inc/azure/core/az_json.h#L306-L318

Example comment:
```C
// These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
```

Related context: https://github.com/azure-rtos/netxduo/issues/47#issuecomment-915642478

This is similar to the note we have above each header file:
https://github.com/Azure/azure-sdk-for-c/blob/84bfe57b50091d88dc993ce80e95ca435aaa0d16/sdk/inc/azure/az_iot.h#L9-L12

cc @JeffreyRichter, @RickWinter

rmknan:
Hello, I would like to work on this. I need to add these comments only in the header file mentioned?

ahsonkhan:
That's great, thanks @rmknan.

Please **search** for all header files that have nested `_internal` structs and add the comment above them, not just the ones mentioned in the issue. You can try searching in your IDE of choice to catch the various instances within the `.h` files (such as Ctrl+Shift+F in VSCode).

>// These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.

For example, the recommended comment should be added on top of this too:
https://github.com/Azure/azure-sdk-for-c/blob/59b721bb33a6a0653824e5c52f1458016a515044/sdk/inc/azure/core/internal/az_http_internal.h#L38-L41

Same here/elsewhere:
https://github.com/Azure/azure-sdk-for-c/blob/59b721bb33a6a0653824e5c52f1458016a515044/sdk/inc/azure/core/az_context.h#L38-L44

rmknan:
PR #2622

PR #2622, opened 2023-08-13
Added the recommended comment to the structs containing internal

rmknan:
I have added the required comments to the header files.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -146,7 +146,7 @@ jobs:
         OSVmImage: ubuntu-20.04
         vcpkg.deps: 'cmocka'
         VCPKG_DEFAULT_TRIPLET: 'x64-linux'
-        AptDependencies: 'gcovr lcov clang-format-9'
+        AptDependencies: 'gcovr lcov clang-format-14'
         build.args: '-DUNIT_TESTING=ON -DUNIT_TESTING_MOCKS=ON'
         AZ_SDK_CODE_COV: 1
         AZ_SDK_C_NO_SAMPLES: 'true'
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -146,7 +146,7 @@ jobs:
         OSVmImage: ubuntu-20.04
         vcpkg.deps: 'cmocka'
         VCPKG_DEFAULT_TRIPLET: 'x64-linux'
-        AptDependencies: 'gcovr lcov clang-format-14'
+        AptDependencies: 'gcovr lcov clang-format-12'
         build.args: '-DUNIT_TESTING=ON -DUNIT_TESTING_MOCKS=ON'
         AZ_SDK_CODE_COV: 1
         AZ_SDK_C_NO_SAMPLES: 'true'
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -360,16 +360,16 @@ jobs:
 
   # Validate all the files are formatted correctly according to the .clang-format file.
   - bash: |
-      # Run clang-format-9 recursively on each source and header file within the repo.
-      clang-format-9 --version
-      find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;
+      # Run clang-format-12 recursively on each source and header file within the repo.
+      clang-format-12 --version
+      find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-12 -i {} \;
 
       git status --untracked-files=no --porcelain
 
       if [[ `git status --untracked-files=no --porcelain` ]]; then
         echo Some files were not formatted correctly according to the .clang-format file.
-        echo Please run clang-format-9 to fix the issue by using this bash command at the root of the repo:
-        echo "find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;"
+        echo Please run clang-format-12 to fix the issue by using this bash command at the root of the repo:
+        echo "find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-12 -i {} \;"
         exit 1
       fi
 
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -41,9 +41,9 @@ struct az_context
     int64_t expiration; // Time when context expires
     void const* key; // Pointers to the key & value (usually NULL)
     void const* value;
-    // These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
+    // These are internal only fields and are subject to change without notice. They are not meant
+    // to be used by anyone externally.
   } _internal;
-
 };
 
 #define _az_CONTEXT_MAX_EXPIRATION 0x7FFFFFFFFFFFFFFF
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -34,15 +34,14 @@ typedef struct az_context az_context;
  * @details The root node in the tree (ultimate parent).
  */
 struct az_context
-{
+{ // These are internal only fields and are subject to change without notice. They are not meant
+  // to be used by anyone externally.
   struct
   {
     az_context const* parent; // Pointer to parent context (or NULL); immutable after creation
     int64_t expiration; // Time when context expires
     void const* key; // Pointers to the key & value (usually NULL)
     void const* value;
-    // These are internal only fields and are subject to change without notice. They are not meant
-    // to be used by anyone externally.
   } _internal;
 };
 
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -33,7 +33,6 @@ typedef struct az_context az_context;
  *
  * @details The root node in the tree (ultimate parent).
  */
-
 struct az_context
 {
   struct
@@ -42,10 +41,9 @@ struct az_context
     int64_t expiration; // Time when context expires
     void const* key; // Pointers to the key & value (usually NULL)
     void const* value;
+    // These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
   } _internal;
-/**
-* These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
-*/
+
 };
 
 #define _az_CONTEXT_MAX_EXPIRATION 0x7FFFFFFFFFFFFFFF
--- a/sdk/inc/azure/core/az_context.h
+++ b/sdk/inc/azure/core/az_context.h
@@ -33,7 +33,7 @@ typedef struct az_context az_context;
  *
  * @details The root node in the tree (ultimate parent).
  */
-// These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
+
 struct az_context
 {
   struct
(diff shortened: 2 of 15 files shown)


rmknan:
For the Issue #1918

rmknan:
@microsoft-github-policy-service agree

CIPop (changes requested):
I think the comments are useful but they should be modified to be attached to `_internal`.
Also, I think it would be better if they are converted to doxygen (i.e. the following pattern:
```C
/**
 * These are ...
 * /```

**Comment by rmknan** [comment]
> I think the comments are useful but they should be modified to be attached to `_internal`. Also, I think it would be better if they are converted to doxygen (i.e. the following pattern:
> 
> ```c
> /**
>  * These are ...
>  * /```
> ```

Attached to meaning that when one places the cursor on top of internal the comment should pop up? could you clarify, please? Thank you

**Comment by rmknan** [comment]
I have modified the comments. Please let me know if they are alright.

File: sdk/inc/azure/iot/az_iot_adu_client.h [hunk 1]
```diff
@@ -456,6 +456,7 @@ typedef struct
  * @brief Structure that holds the state of the Azure IoT ADU client.
  *
  */
+// These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
```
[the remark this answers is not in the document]
rmknan: > The comments are not place correctly. In this and other structs, the only fields this statement applies to is the `_internal` as well as anything that starts with underscore (`_`).
> 
> See the top of the include files:
> 

They were on top of struct in my previous commit but I changed it after this. I thought you meant to change the comment location and I placed it under the `_internal`.

rmknan:
> The comments are not place correctly. In this and other structs, the only fields this statement applies to is the `_internal` as well as anything that starts with underscore (`_`).
> 
> See the top of the include files:
> 

They were on top of struct in my previous commit but I changed it after this. I thought you meant to change the comment location and I placed it under the `_internal`.

CIPop:
> You mean like this:? [...]

I think so but please make a single change then check the Doxygen generated HTML files to ensure that your comment is correctly attached to the `_internal` struct element only. If that change shows up correctly, you could alter all comments.

> Also, the command gives out an error : find: ‘clang-format-9’: No such file or directory

As with `doxygen` you probably need to install `clang-format-9`. On my Ubuntu 20, I did:
```sh
sudo apt install clang-format-9 doxygen
```

rmknan:
I generated the doxygen html file. But i do not see the `_internal` struct anywhere in the page. I can see various other comments with various tags such as @brief , @note etc but it does not show up. I changed the position of the comment such that it would be above the `_internal` but I do not see it in the HTML file.

My comment is ```
```
/**
* 
* @remark These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
*/
} _internal;
```

Can you suggest how to go about this ? thank you.

sdk/inc/azure/core/az_context.h
```diff
@@ -33,6 +33,7 @@ typedef struct az_context az_context;
  *
  * @details The root node in the tree (ultimate parent).
  */
+
 struct az_context
 {
   struct
```
CIPop, line 39: I think that the comment should go above the `struct` in line 39.

rmknan: I have made the changes and now the the comment can be seen in the doxygen html under `Note`. As for the clang format My ubuntu version is 22.04 and this command does not seem to work. I installed clang, but clang format-9 gives me error same as previous one.

`find . \( -iname '*.h' -o -iname '*.c' \) -exec clang-format-9 -i {} \;`

CIPop: It's unclear if you've installed version 9 of clang-format. "clang-format" might be a different version and won't format the code the same way.

Did this work? 

`sudo apt install clang-format-9`

Afterwards, test using

```sh
clang-format-9 --version
```

ahsonkhan:
> Also, I think it would be better if they are converted to doxygen

@CIPop FWIW we try to exclude any internal/private and `_` prefixed types from the doxygen docs:
https://github.com/Azure/azure-sdk-for-c/blob/59b721bb33a6a0653824e5c52f1458016a515044/doc/Doxyfile#L884-L896

CIPop:
> > Also, I think it would be better if they are converted to doxygen
> 
> @CIPop FWIW we try to exclude any internal/private and `_` prefixed types from the doxygen docs:

I thought that's what you were asking for in https://github.com/Azure/azure-sdk-for-c/issues/1918: 

> image

The example given is the _doxygen_ note stating that internals shouldn't be used. The requirement was to add more (I assumed Doxygen) documentation for all `_internal` members.

rmknan:
@ahsonkhan So, the first commit is fine? Please let me know what changes have to be made. Thank you. @CIPop @ahsonkhan

ahsonkhan:
Let me take a look at how our current doxygen docs look like and get back to you, @rmknan.

If adding a doxygen style comment (instead of regular // comments) results in more docs to be generated which include `_internal` when they weren't before.

But I do think we should **get the location right**, regardless of the doc style. Unless there's a reason not to, putting it right above the nested _internal struct makes most sense, which is what @CIPop is recommending as well.

For background:
The rationale here is that because internal and private types are meant to be implementation detail, we don't want them advertise in the published doxygen docs.

That said, since customers do consume sources and look at the files as well, the purpose of the issue was to highlight the note in source comments, so if someone happens to consume sources directly and was accidentally taking a dependency on `_internal` fields, they would get a local hint not to (in case they missed the file-level doc comment).

sdk/inc/azure/core/az_context.h
rmknan: > It's unclear if you've installed version 9 of clang-format. "clang-format" might be a different version and won't format the code the same way.
> 
> Did this work?
> 
> `sudo apt install clang-format-9`

 Nope, it says `Unable to locate package clang-format-9`

ahsonkhan:
@rmknan your first commit puts the comment above the entire struct. I would suggest putting it right above the **internal** struct definition that's nested, not the whole struct.

sdk/inc/azure/core/az_context.h
```diff
@@ -33,6 +33,7 @@ typedef struct az_context az_context;
  *
  * @details The root node in the tree (ultimate parent).
  */
+
```
ahsonkhan, line 36: Revert this new line additions introduced in this PR between the existing docs and the struct definitions.

rmknan:
Okay and as a normal comments right? because doxygen comments inside the struct do not show up on the HTML. And If it has to be seen in the source code then normal comments would be the way to go, I presume.

ahsonkhan:
As long as they don't cause undocumented `_internal` structs and their fields to show up in doc pages like these, any style is fine.

https://azuresdkdocs.blob.core.windows.net/$web/c/az_core/1.5.0/structaz__json__token.html#details

It looks like we see nested struct fields within doxygen docs, when they are documented.

Unless @CIPop has a strong preference on the comment style, I think normal style comments is a reasonable choice here.

The important thing is to get that comment visible. We can adjust the style later, if needed :)

Thanks @rmknan for making sure we get the details right, here. Appreciate it!

I might try and see if and how the generated docs differ if we use `//` comment vs `///` comment for the note (3 slashes). I believe `///` will make it visible in Doxygen ([like this](https://github.com/Azure/azure-sdk-for-c/blob/59b721bb33a6a0653824e5c52f1458016a515044/sdk/inc/azure/core/az_json.h#L82)). If you are curious and want to try and experiment yourself, feel free to do and let us know what you find.

rmknan:
I did not revert the additional spaces in this commit, should I revert this commit and make a new commit without the spaces?

rmknan:
@CIPop As for clang-format-9 it seems it's not supported on 22.04, what should be done? thank you.

CIPop:
> @CIPop As for clang-format-9 it seems it's not supported on 22.04, what should be done? thank you.

@danieljurek /cc @RickWinter : I think this is a blocking development issue for the repo.
My apologizes @rmknan: I thought I did this before on 22 but it turns out I might have just installed it on Ubuntu 20...

The package is not available on Ubuntu 22:
image

This issue is tracked in https://github.com/Azure/azure-sdk-for-c/issues/2177

CIPop:
@rmknan the simplest way to pass the CI tests is to install an Ubuntu 20 VM and run the formatter there. Alternatively, you would need to find a way to install and run clang-format-9 on your system (This article pointing out the llvm toolchain package repositories might help: https://stackoverflow.com/questions/20756924/how-can-i-install-clang-format-in-ubuntu).

From my experience (see #2177), running with a newer clang-format will produce different formatting that would still fail the check.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -146,7 +146,7 @@ jobs:
         OSVmImage: ubuntu-20.04
         vcpkg.deps: 'cmocka'
         VCPKG_DEFAULT_TRIPLET: 'x64-linux'
-        AptDependencies: 'gcovr lcov clang-format-9'
+        AptDependencies: 'gcovr lcov clang-format-14'
```
danieljurek, line 149: Should also include changes to the invocations of `clang-format` here to either `clang-format` or `clang-format-14`: https://github.com/Azure/azure-sdk-for-c/blob/bbe6c1cff6483ebab650a2b6a8087512107cd77f/eng/pipelines/templates/jobs/archetype-sdk-client.yml#L363-L365

danieljurek: Indeed `clang-format-14` is not available on Ubuntu 20.04. A couple of options: 

1. Upgrade this matrix entry to Ubuntu 22.04
2. Use a different version of clang-format 

Cursory examination suggests clang-format-12 is available Ubuntu 20.04: 

image

rmknan: Got it, I will try again using clang-format-12.

danieljurek: Gotta use clang-format-12 in the command, not clang-format-9... The script as it was written fails silently. I refactored and made it fail more loudly if the expected command isn't present.

rmknan: When I went to the terminal in the SDK folder, I executed the command with clang-format-12. I did not realize that the .yml file also had to be changed. Now is it resolved or do I commit again with those changes? Thank you.

rmknan:
@CIPop Is there anything that needs to be changed here? or is it fine?

sdk/inc/azure/core/az_json.h
```diff
@@ -402,6 +410,8 @@ AZ_NODISCARD az_result az_json_writer_chunked_init(
  * first provided destination span, this function only returns the text written into the last
  * provided destination buffer from the allocator callback.
  */
+// These are internal only fields and are subject to change without notice. They are not meant to be
```
RickWinter, line 413: This comment is not valid.  The function is public.

sdk/inc/azure/core/az_context.h
```diff
@@ -41,6 +41,8 @@ struct az_context
     int64_t expiration; // Time when context expires
     void const* key; // Pointers to the key & value (usually NULL)
     void const* value;
+    // These are internal only fields and are subject to change without notice. They are not meant
```
RickWinter, line 44: This comment should be at the start of the structure definition.  Ditto for all other locations

RickWinter (changes requested):

sdk/inc/azure/core/az_context.h
rmknan: `struct az_context
{ 1)
  struct
  { 2)
    az_context const* parent; // Pointer to parent context (or NULL); immutable after creation
    int64_t expiration; // Time when context expires
    void const* key; // Pointers to the key & value (usually NULL)
    void const* value;
    // These are internal only fields and are subject to change without notice. They are not meant to be used by anyone externally.
  } _internal;
};`

At location 1) or 2) ?

RickWinter: 1 is where we'd typically place the comment describing the structure.

rmknan:
@RickWinter Do let me know if further changes are required.Thank you.

rmknan:
Does anything else need to be changed?

CIPop (approved):
Thanks @rmknan !

PR #2623, opened 2023-08-14
RPC Server Address comments

vaavva:
- Addresses comments from https://github.com/Azure/azure-sdk-for-c/pull/2621
- Removes pending command concept from the rpc server hfsm policy
    - Simplifies the rpc server hfsm policy to a single state
    - this enables handling multiple commands under the same subscription with the same rpc server policy. This still requires one rpc server instance per subscription.
- Updates initializer to use standardized options pattern


sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -105,7 +104,7 @@ static void timer_callback(union sigval sv)
   }
 
   pending_command.content_type = AZ_SPAN_FROM_BUFFER(content_type_buffer);
-  pending_command.response_topic = AZ_SPAN_FROM_BUFFER(response_topic_buffer);
+  pending_command.request_topic = AZ_SPAN_FROM_BUFFER(request_topic_buffer);
```
CIPop, line 107: Note: (didn't check - we might already do this) - it would be preferable if we could verify the size of the buffer in the `_init()` API (and return `AZ_ERROR_NOT_ENOUGH_SPACE` if these buffers are too small).

I was also thinking that now, with multiple buffers, detection of which buffer is too small could get complicated. We might want to have some preconditions when there isn't sufficient space, so that the app blocks on the offending line / buffer.

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -362,32 +383,28 @@ int main(int argc, char* argv[])
...
   LOG_AND_EXIT_IF_FAILED(az_mqtt5_property_bag_init(&property_bag, &mqtt5, NULL));
 
-  rpc_server_options
-      = (az_mqtt5_rpc_server_options){ .sub_topic = AZ_SPAN_FROM_BUFFER(sub_topic_buffer),
-                                       .command_name = command_name,
-                                       .model_id = model_id };
-
   az_mqtt5_rpc_server_memory rpc_server_memory
-      = (az_mqtt5_rpc_server_memory){ .property_bag = property_bag };
+      = (az_mqtt5_rpc_server_memory){ .property_bag = property_bag,
```
CIPop, line 398: API naming - I would either rename this to rpc_server_context or completely remove it and accept the property bag and subscribe_topic as arguments.

Also, per API design guidelines, the rpc_server_options must exist on all '_init()' functions (constructors). 
- In this case it could be just a struct `_internal { uint8_t reserved }`. 
- A corresponding `az_mqtt5_rpc_server_memory_options_default()` function should also exist. (When we do have options, the application side pattern would be to call this function first, then modify non-internal members before passing the options to the _init function.)

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -400,54 +309,62 @@ static az_result waiting(az_event_policy* me, az_event event)
       // No-op
       break;
 
+    case AZ_MQTT5_EVENT_SUBACK_RSP:
+    {
+      // if get suback that matches the sub we sent, stop waiting for the suback
+      az_mqtt5_suback_data* data = (az_mqtt5_suback_data*)event.data;
+      if (data->id
+          == this_policy->_internal.rpc_server_memory._internal._az_mqtt5_rpc_server_pending_sub_id)
```
CIPop, line 317: The long full-struct name here is more reason to simply attach the contents of the "memory" directly to the state machine's object (i.e. this_policy->_internal.subscription_id).

Also - the name of the internal shouldn't be this long (instead of `_az_mqtt5_rpc_server_pending_sub_id` it could be just `id` or `subscription_packet_id`).

CIPop (approved):
Left a few comments, mostly about naming and simplifying the internal structs.

RLeclair (approved):
LGTM! Some minor comments only

sdk/samples/core/mosquitto_rpc_server_sample.c
vaavva: Removed the memory struct and moved the property bag and subscribe topic to be arguments on init

sdk/inc/azure/core/az_mqtt5_rpc_server.h
```diff
@@ -25,6 +27,10 @@
  * @brief The default timeout in seconds for subscribing.
  */
 #define AZ_MQTT5_RPC_SERVER_DEFAULT_TIMEOUT_SECONDS 10
+/**
+ * @brief The default QOS to use for subscribing/publishing.
+ */
+#define AZ_MQTT5_RPC_DEFAULT_QOS 1
```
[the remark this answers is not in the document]
vaavva: Let me know if the change I made is what you had in mind!

sdk/inc/azure/core/az_mqtt5_rpc_server.h
```diff
@@ -263,36 +241,57 @@ typedef struct
...
- * @param[in] rpc_server_memory The allocated az_mqtt5_rpc_server_memory to use for the RPC Server.
+ * @param[in] rpc_server_memory The application allocated az_mqtt5_rpc_server_memory to use for the
+ * RPC Server.
+ * @param[in] model_id The model id to use for the subscription topic.
+ * @param[in] client_id The client id to use for the subscription topic.
+ * @param[in] command_name The command name to use for the subscription topic.
  * @param[in] options Any az_mqtt5_rpc_server_options to use for the RPC Server.
  *
- * @return #az_result
+ * @return An #az_result value indicating the result of the operation.
```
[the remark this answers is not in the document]
vaavva: I've tracked this as a work item - will address it once we're settled on the APIs

sdk/samples/core/mosquitto_rpc_server_sample.c
vaavva: Will investigate this once I have the public API

PR #2635, opened 2023-09-07
Paho Async Implementation

RLeclair:
Changes:
- Adding Paho Async implementation of az_mqtt5.
- Adding an extra pipeline for Paho Async.

TODO_L tags denote changes to be made for a follow up PR once clarification is obtained on those comments.


sdk/src/azure/platform/CMakeLists.txt
```diff
@@ -124,6 +123,41 @@ if (TRANSPORT_MOSQUITTO AND AZ_PLATFORM_IMPL STREQUAL "POSIX" AND (UNIX AND NOT
   add_library (az::mqtt5_impl ALIAS az_mqtt5_mosquitto)
 endif() # TRANSPORT_MOSQUITTO
 
+# Paho Async Platform
+
+if (TRANSPORT_PAHO AND AZ_PLATFORM_IMPL STREQUAL "POSIX" AND (UNIX AND NOT APPLE))
+  message(STATUS "Using Paho Async MQTT transport")
```
RLeclair, line 129: Remove this
```suggestion
  
```

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+  }
+
+  char port_buffer[SERVER_URI_PORT_MAX_LENGTH];
+  sprintf(port_buffer, "%d", connect_data->port);
+  az_span port_span = az_span_create_from_str(port_buffer);
+
+  int32_t server_uri_length
+      = az_span_size(connect_data->host) + SERVER_URI_COLON_LENGTH + SERVER_URI_PORT_MAX_LENGTH + 1;
+  char server_uri_buffer[server_uri_length];
+  az_span server_uri = az_span_create((uint8_t*)server_uri_buffer, server_uri_length);
```
vaavva, line 272:
```suggestion
  az_span server_uri = az_span_create_from_buffer((uint8_t*)server_uri_buffer);
```

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+    ret = AZ_ERROR_OUT_OF_MEMORY;
+    return ret;
+  }
+
+  _az_RETURN_IF_FAILED(_az_result_from_pahoasync(MQTTAsync_setCallbacks(
+      *me->_internal.pahoasync_handle,
+      mqtt5,
+      _az_pahoasync_on_disconnect,
+      _az_pahoasync_on_message,
+      NULL)));
```
vaavva, line 299: is this the puback callback, and do we want to set it?

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+  (void)context;
+  (void)response;
+
+  // TODO_L: Implement logic to handle failing to publish.
+  _az_pahoasync_critical_error();
+}
+
+static void _az_pahoasync_on_disconnect_success(void* context, MQTTAsync_successData5* response)
+{
+  az_mqtt5* me = (az_mqtt5*)context;
```
vaavva, line 185: Is this called in addition to _az_pahoasync_on_disconnect? If so, we probably shouldn't send the inbound event twice

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+  conn_opts.context = me;
+  conn_opts.ssl = &ssl_opts;
+
+  _az_RETURN_IF_FAILED(
+      _az_result_from_pahoasync(MQTTAsync_connect(*me->_internal.pahoasync_handle, &conn_opts)));
+
+  return ret;
+}
+
+AZ_NODISCARD az_result az_mqtt5_outbound_sub(az_mqtt5* mqtt5, az_mqtt5_sub_data* sub_data)
```
vaavva, line 314: Will we still get the mid filled out in the sub_data once this function returns?

vaavva: (same question for pub)

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+{
+  MQTTAsync_responseOptions opts = MQTTAsync_responseOptions_initializer;
+  opts.onSuccess5 = _az_pahoasync_on_subscribe_success;
+  opts.onFailure5 = _az_pahoasync_on_subscribe_failure;
+  opts.context = mqtt5;
+
+  return _az_result_from_pahoasync(MQTTAsync_subscribe(
```
vaavva, line 321: Should we add functionality now to add properties?

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,169 @@
...
+   * @brief Paho Async specific property array.
+   *
+   */
+  MQTTProperties* pahoasync_properties;
+
+  /**
+   * @brief Paho Async property structure used for constructing the property array.
+   *
+   */
+  MQTTProperty pahoasync_property;
```
vaavva, line 85: If I understand this correctly, this property struct is used as a temporary place to create a property before it's added to the bag, and then would be overwritten/reused the next time you create a property and add it to the bag. Is this correct? If so, can you add something like this in the comment so it's clear this shouldn't be used as a stable reference at any point?

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+{
+  _az_PRECONDITION_NOT_NULL(property_bag);
+  _az_PRECONDITION_NOT_NULL(prop_str);
+
+  property_bag->pahoasync_property.identifier = (enum MQTTPropertyCodes)type;
+  property_bag->pahoasync_property.value.data.data = (char*)az_span_ptr(prop_str->str);
+  property_bag->pahoasync_property.value.data.len = az_span_size(prop_str->str);
+
+  return _az_result_from_pahoasync(
+      MQTTProperties_add(property_bag->pahoasync_properties, &(property_bag->pahoasync_property)));
```
vaavva, line 393: Should the values set in pahoasync_property be cleared after they're added to the bag so you don't accidentally carry any values over later?

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+
+  prop = MQTTProperties_getPropertyAt(
+      property_bag->pahoasync_properties, (enum MQTTPropertyCodes)type, stringpair_index);
+
+  while (prop != NULL)
+  {
+    az_span prop_key = az_span_create((uint8_t*)prop->value.data.data, prop->value.data.len);
+    if (az_span_is_content_equal(key, prop_key))
+    {
+      out_prop_strpair->key = az_span_create((uint8_t*)prop->value.data.data, prop->value.data.len);
```
vaavva, line 499: I feel like there should be a way to not recreate this same span as before, but I can't think of one - just leaving this comment in case you can

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+{
+  _az_PRECONDITION_NOT_NULL(property_bag);
+  _az_PRECONDITION_NOT_NULL(out_prop_strpair);
+  MQTTProperty* prop;
+  int stringpair_index = 0;
+
+  prop = MQTTProperties_getPropertyAt(
+      property_bag->pahoasync_properties, (enum MQTTPropertyCodes)type, stringpair_index);
+
+  while (prop != NULL)
```
vaavva, line 494:
```suggestion
  for (prop =  MQTTProperties_getPropertyAt(property_bag->pahoasync_properties, (enum MQTTPropertyCodes)type, stringpair_index); prop != NULL; prop =  MQTTProperties_getPropertyAt(property_bag->pahoasync_properties, (enum MQTTPropertyCodes)type, stringpair_index))
```

not sure if something like this makes things more concise about switching to the next prop or not

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+}
+
+AZ_NODISCARD az_span az_mqtt5_property_get_binarydata(az_mqtt5_property_binarydata* prop_bindata)
+{
+  _az_PRECONDITION_NOT_NULL(prop_bindata);
+
+  return prop_bindata->bindata;
+}
+
+void az_mqtt5_property_free_string(az_mqtt5_property_string* prop_str) { (void)prop_str; }
```
vaavva, line 609: Does the application need to free this themselves because they allocate it? If so, would it make sense to still provide the free in here or not?
Or is it that nothing gets allocated so nothing needs to be free'd?

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -74,7 +76,6 @@ static az_mqtt5_connect_data test_mqtt5_connect_data = {
 
 static az_mqtt5_options options = {
```
vaavva, line 77: Should this struct be defined within separate ifdefs for the different mqtt clients if they have different values?

sdk/tests/platform/test_az_mqtt5_policy.c
```diff
@@ -93,6 +94,8 @@ static az_result test_inbound_hfsm_root(az_event_policy* me, az_event event)
   {
     case AZ_MQTT5_EVENT_CONNECT_RSP:
       ref_connack++;
+      az_mqtt5_connack_data* test_connack_data = (az_mqtt5_connack_data*)event.data;
+      assert_int_equal(test_connack_data->connack_reason, 0);
```
vaavva, line 98: Would we ever test a case that has a different connack reason?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         MapFileArtifactSuffix: 'lnx-x64-rel-noprc-log-samples'
         BuildType: Release
 
-      Linux_Logging_UnitTests_Mocks_Samples:
+      Linux_MosquittoMQTT_Logging_UnitTests_Mocks_Samples:
         Pool: azsdk-pool-mms-ubuntu-2004-general
         OSVmImage: ubuntu-20.04
         vcpkg.deps: 'curl[ssl] paho-mqtt cmocka mosquitto'
```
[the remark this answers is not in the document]
RLeclair: The test still had `TRANSPORT_PAHO=ON` as an argument, so we need to include paho-mqtt for the iot tests to compile

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,169 @@
...
+   * @brief The binary data value of the property.
+   */
+  az_span bindata;
+} az_mqtt5_property_binarydata;
+
+/**
+ * @brief Initializes the MQTT 5 instance specific to Eclipse Paho Async.
+ *
+ * @param mqtt5 The MQTT 5 instance.
+ * @param pahoasync_handle The Paho Async handle, can't be NULL.
```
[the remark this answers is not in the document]
RLeclair: The memory for the client struct (`MQTTAsync`) has to be allocated by the application, Paho does not do it for us

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -128,11 +128,18 @@ AZ_NODISCARD AZ_INLINE bool az_span_topic_matches_sub(az_span sub, az_span topic
...
+#if defined(PLATFORM_MOSQUITTO)
   if (MOSQ_ERR_SUCCESS
       != mosquitto_topic_matches_sub((char*)az_span_ptr(sub), (char*)az_span_ptr(topic), &ret))
   {
     ret = false;
   }
+#else // PLATFORM_MOSQUITTO
+  (void)sub;
+  (void)topic;
+  ret = false;
```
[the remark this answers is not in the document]
RLeclair: I will create a card for adding this function to az_mqtt which I think is where it should belong, I also added a TODO right above this comment

sdk/src/azure/platform/CMakeLists.txt
```diff
@@ -124,6 +123,41 @@ if (TRANSPORT_MOSQUITTO AND AZ_PLATFORM_IMPL STREQUAL "POSIX" AND (UNIX AND NOT
...
+
+  find_package(OpenSSL CONFIG)
+  if(NOT OpenSSL_FOUND)
+    find_package(OpenSSL REQUIRED)
+  endif()
+
+  add_library (
+    az_mqtt5_paho
+      STATIC
+      ${CMAKE_CURRENT_LIST_DIR}/az_pahoasync.c
```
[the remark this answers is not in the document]
RLeclair: Changed the name!

sdk/src/azure/platform/az_pahoasync.c
```diff
@@ -0,0 +1,619 @@
...
+
+static void _az_pahoasync_on_disconnect(void* context, char* cause)
+{
+  (void)cause; // cause is always set to NULL currently based on the Paho Async documentation.
+  az_mqtt5* me = (az_mqtt5*)context;
+
+  az_result ret = az_mqtt5_inbound_disconnect(
+      me,
+      &(az_mqtt5_disconnect_data){ .tls_authentication_error = false,
+                                   .disconnect_requested = false });
```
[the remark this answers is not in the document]
RLeclair: That callback would be `_az_pahoasync_on_disconnect_success`

sdk/src/azure/platform/az_pahoasync.c
RLeclair: There is no `az_span_create_from_buffer` but if you're referring to `AZ_SPAN_FROM_BUFFER`, I already tried that approach but obtain a compiler error on Ubuntu 20 because of the conversion I do for the length of the buffer (from unsigned int to int)

sdk/src/azure/platform/az_pahoasync.c
RLeclair: The puback callback is set when we publish in `az_mqtt5_outbound_pub`, I already tested setting this callback and I am unable to receive PUBACKs there

sdk/src/azure/platform/az_pahoasync.c
RLeclair: This covers the case where we send a disconnect ourselves, `_az_pahoasync_on_disconnect` covers the case where the broker sends a disconnect to the client

sdk/src/azure/platform/az_pahoasync.c
RLeclair: Thanks for pointing that out! I've added the changes to fill out the message id

sdk/src/azure/platform/az_pahoasync.c
RLeclair: I can add that in a separate PR

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
RLeclair: Correct, I could have added it to each of the property structs we have (`az_mqtt5_property_string_pair`) but we would have had to add property structs for byte, int and other primitive types. I clarified my comment!

sdk/src/azure/platform/az_pahoasync.c
RLeclair: I added a helper function to clear the property

sdk/src/azure/platform/az_pahoasync.c
RLeclair: Switched it to your approach, it makes it more concise :)

sdk/src/azure/platform/az_pahoasync.c
RLeclair: Nothing gets allocated, and if they do allocate a MQTTProperty they would need to free it themselves, not us since Paho itself doesn't allocate those except when it receives a message

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: I can add that once we make use of the different parts of the structure, we currently don't use the `openssl_engine` in az_mosquitto's `az_mqtt_options` nor do we test it

sdk/tests/platform/test_az_mqtt5_policy.c
RLeclair: We could do so once we expand the tests, in this case we want to ensure that the connection is accepted

CMakeLists.txt
```diff
@@ -106,6 +106,7 @@ add_subdirectory(sdk/src/azure/platform)
 if(NOT DEFINED ENV{AZ_SDK_C_NO_SAMPLES})
   if(TRANSPORT_PAHO)
     add_subdirectory(sdk/samples/iot)
+    add_subdirectory(sdk/samples/core)
```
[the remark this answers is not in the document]
RLeclair: I'm going to edit the CMakeLists.txt file at the samples/core level to only include mosquitto samples when the `AZ_MQTT_IMPLEMENTATION` is mosquitto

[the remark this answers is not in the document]
RLeclair: The `CMakeLists.txt` at the `samples/core` level does not compile the Mosquitto specific samples when the MQTT implementation is paho

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -208,6 +210,11 @@ typedef struct
    * @brief The publish request ID.
    */
   int32_t id;
+
+  /**
+   * @brief Publish acknowledgement (PUBACK) reason code. Indicates success or reason for failure.
+   */
+  int32_t puback_reason;
```
ericwol-msft, line 217: Humm do we need puback_ inside puback struct?

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -255,6 +262,46 @@ typedef struct
...
+  az_mqtt5_property_bag* properties;
+  /**
+   * @brief Topic filter to unsubscribe from.
+   */
+  az_span topic_filter;
+
+  /**
+   * @brief Id to correlate the unsubscribe request with the response acknowledgement.
+   */
+  int32_t out_id;
```
ericwol-msft, line 285: out?

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -447,6 +500,27 @@ az_mqtt5_inbound_suback(az_mqtt5* mqtt5, az_mqtt5_suback_data* suback_data)
       pipeline, (az_event){ .type = AZ_MQTT5_EVENT_SUBACK_RSP, .data = suback_data });
 }
 
+/**
+ * @brief Posts a MQTT 5 unsubscribe acknowledgement event to the event pipeline.
```
ericwol-msft, line 504:
```suggestion
 * @brief Posts a MQTT unsubscribe acknowledgement event to the event pipeline.
```

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -447,6 +500,27 @@ az_mqtt5_inbound_suback(az_mqtt5* mqtt5, az_mqtt5_suback_data* suback_data)
...
+ *
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD AZ_INLINE az_result
+az_mqtt5_inbound_unsuback(az_mqtt5* mqtt5, az_mqtt5_unsuback_data* unsuback_data)
+{
+  _az_event_pipeline* pipeline = mqtt5->_internal.platform_mqtt5.pipeline;
+  if (!pipeline)
+  {
+    return AZ_ERROR_NOT_IMPLEMENTED;
```
ericwol-msft, line 517: AZ_ERROR_ARG ?

sdk/inc/azure/core/az_mqtt5.h
```diff
@@ -518,7 +592,17 @@ az_mqtt5_outbound_connect(az_mqtt5* mqtt5, az_mqtt5_connect_data* connect_data);
 AZ_NODISCARD az_result az_mqtt5_outbound_sub(az_mqtt5* mqtt5, az_mqtt5_sub_data* sub_data);
 
 /**
- * @brief Sends a MQTT 5 publish data packet to broker.
+ * @brief Sends a MQTT 5 unsubscribe data packet to the broker.
```
ericwol-msft, line 595:
```suggestion
 * @brief Sends a MQTT unsubscribe data packet to the broker.
```

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
```diff
@@ -89,7 +89,7 @@ typedef struct
 /**
  * @brief MQTT 5 property string.
  *
- * @note String should always be freed using #az_mqtt5_property_free_string.
+ * @note String should always be freed after reading using #az_mqtt5_property_free_string.
```
ericwol-msft, line 92:
```suggestion
 * @note String should always be freed, after reading, using #az_mqtt5_property_free_string.
```

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
```diff
@@ -102,7 +102,7 @@ typedef struct
 /**
  * @brief MQTT 5 property string pair.
  *
- * @note String should always be freed using #az_mqtt5_property_free_stringpair.
+ * @note String should always be freed after reading using #az_mqtt5_property_free_stringpair.
```
ericwol-msft, line 105:
```suggestion
 * @note String should always be freed, after reading, using #az_mqtt5_property_free_stringpair.
```

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
```diff
@@ -120,7 +120,7 @@ typedef struct
 /**
  * @brief MQTT 5 property binary data.
  *
- * @note Binary data should always be freed using #az_mqtt5_property_free_binarydata.
+ * @note Binary data should always be freed after reading using #az_mqtt5_property_free_binarydata.
```
ericwol-msft, line 123:
```suggestion
 * @note Binary data should always be freed, after reading, using #az_mqtt5_property_free_binarydata.
```

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+/**
+ * @file
+ *
+ * @brief Defines MQTT 5 constructs for Eclipse Paho Async.
```
ericwol-msft, line 7:
```suggestion
 * @brief Defines MQTT 5 constructs for Paho Asynchronous MQTT C Client Library

```

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,172 @@
...
+ */
+typedef struct
+{
+  /**
+   * @brief Platform options that are common across all MQTT 5 implementations.
+   */
+  az_mqtt5_options_common platform_options;
+
+  /**
+   * @brief The CA Trusted Roots path span interpretable by the underlying MQTT 5 implementation.
```
ericwol-msft, line 35:
```suggestion
   * @brief The CA Trusted Roots certificate path span interpretable by the underlying MQTT 5 implementation.
```

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * @details We recommend avoiding configuring this option and instead using the default
+   * OpenSSL trusted roots instead.
```
ericwol-msft, line 38:
```suggestion
   * OpenSSL trusted roots instead if available.
```

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,172 @@
...
+   * constructing the property array.
+   *
+   */
+  MQTTProperty pahoasync_property;
+} az_mqtt5_property_bag;
+
+/**
+ * @brief MQTT 5 property string.
+ *
+ * @note String should always be freed after reading using #az_mqtt5_property_free_string.
```
ericwol-msft, line 94:
```suggestion
 * @note String should always be freed, after reading, using #az_mqtt5_property_free_string.
```

sdk/src/azure/platform/az_pahoasync5.c
```diff
@@ -0,0 +1,711 @@
...
+  {
+    ret = _az_RESULT_MAKE_ERROR(_az_FACILITY_CORE_MQTT5, pahoasync_ret);
+  }
+
+  return ret;
+}
+
+AZ_INLINE void _clear_pahoasync5_property(az_mqtt5_property_bag* property_bag)
+{
+  property_bag->pahoasync_property.identifier = 0;
```
ericwol-msft, line 48: Use memset(&property_bag->pahoasync_property, 0, sizeof(property_bag->pahoasync_property));

sdk/inc/azure/platform/az_mqtt5_pahoasync.h
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * @details We recommend avoiding configuring this option and instead using the default
+   * OpenSSL trusted roots instead.
+   */
+  az_span certificate_authority_trusted_roots;
+
+  /**
+   * @brief Whether to use TLS for the underlying MQTT 5 implementation.
+   */
+  bool disable_tls;
```
ericwol-msft, line 45:
```suggestion
  bool disable_tls_validation;
```

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -129,11 +129,18 @@ AZ_NODISCARD AZ_INLINE bool az_span_topic_matches_sub(az_span sub, az_span topic
 {
   bool ret;
   // TODO: have this not be mosquitto specific
+  // TODO_L: Create this function for things to work with paho async.
+#if defined(PLATFORM_MOSQUITTO)
```
vaavva, line 133:
```suggestion
#if defined(TRANSPORT_MOSQUITTO)
```

sdk/inc/azure/core/az_mqtt5.h
RLeclair: Removed the `puback_` suffix and replaced with `reason_code`

sdk/inc/azure/core/az_mqtt5.h
RLeclair: The MQTT stack will set this variable when executing an unsubscribe, the application can then use it to correlate messages.

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
RLeclair: Changed to:
```suggestion
 * @note After reading using #az_mqtt5_property_bag_read_string, string should always be freed
 * using #az_mqtt5_property_free_string.
```

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
RLeclair: Changed to: 
```suggestion
 * @note After reading using #az_mqtt5_property_bag_find_stringpair, string pair should always be freed using #az_mqtt5_property_free_stringpair.
```

sdk/inc/azure/platform/az_mqtt5_mosquitto.h
RLeclair: Changed to: 
```suggestion
 * @note After reading using #az_mqtt5_property_bag_read_binarydata, binary data should always be freed using #az_mqtt5_property_free_binarydata.
```

sdk/tests/core/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ if (UNIX)
     set(MATH_LIB_UNIX "m")
 endif()
 
-if (AZ_PLATFORM_IMPL STREQUAL "POSIX" AND NOT APPLE AND UNIT_TESTING_MOCKS)
+if (AZ_PLATFORM_IMPL STREQUAL "POSIX" AND NOT APPLE AND UNIT_TESTING_MOCKS AND NOT AZ_MQTT_TRANSPORT_IMPL STREQUAL "PAHO")
```
CIPop, line 31: Why are these tests disabled when PAHO MQTT is used?

sdk/tests/core/main.c
```diff
@@ -21,7 +21,8 @@ int main()
   // negative numbers
   result += test_az_base64();
   result += test_az_context();
-#if !defined(__APPLE__) && defined(PLATFORM_POSIX) && defined(_az_MOCK_ENABLED)
+#if !defined(__APPLE__) && defined(PLATFORM_POSIX) && defined(_az_MOCK_ENABLED) \
+    && !defined(TRANSPORT_PAHO)
```
CIPop, line 25: It feels a bit too complicated to keep both TRANSPORT_PAHO and AZ_MQTT_TRANSPORT_IMPL. Can we remove TRANSPORT_PAHO?

CIPop (approved):
I assume that MQTT testing is disabled temporarily for Paho Async. If yes, please add a TODO comment to ensure we don't forget to re-enable them.

sdk/tests/core/CMakeLists.txt
RLeclair: MQTT testing is disabled for rpc tests on Paho. We should not be using any underlying MQTT implementation for the tests in core but leaving as a TODO for later

sdk/tests/core/main.c
RLeclair: TRANSPORT_PAHO is used for the IoT samples unfortunately

CIPop: Yes, I understand that: I'm proposing we replace TRANSPORT_PAHO with AZ_MQTT_TRANSPORT_IMPL=PAHO_SYNC, even for IoT Hub samples.

sdk/tests/core/CMakeLists.txt
CIPop: I'm confused - then why is it disabling the tests if AZ_MQTT_TRANSPORT_IMPL is set to "PAHO" but not "MOSQUITTO"?

PR #2638, opened 2023-09-12
RPC Custom Topics & Server + Client Sample supporting multiple commands

vaavva:
- Adds support for custom topic formats
- Adds new sample that has both a client and a server on the same device that support multiple commands under a single subscription (one for client and one for server)
- Creates an az_mqtt5_rpc.c file since it was getting to be too much for just an .h file


sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+
+// User-defined parameters
+static const az_span client_id = AZ_SPAN_LITERAL_FROM_STR("application-id");
+static const az_span username = AZ_SPAN_LITERAL_FROM_STR("application-id");
+static const az_span hostname = AZ_SPAN_LITERAL_FROM_STR("localhost");
+static const az_span server_client_id = AZ_SPAN_LITERAL_FROM_STR("controller-id");
+static const az_span content_type = AZ_SPAN_LITERAL_FROM_STR("application/json");
+static const az_span start_module_command_name = AZ_SPAN_LITERAL_FROM_STR("startModule");
+static const az_span stop_module_command_name = AZ_SPAN_LITERAL_FROM_STR("stopModule");
+static const az_span server_subscription_topic_format
```
ewertons, line 39: Do we require customer to pass the string null-terminator? Can't we avoid that?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    = AZ_SPAN_LITERAL_FROM_STR("device/{executorId}/command/{name}\0");
```
ewertons, line 40: Same here for the null-terminator

ewertons: And everywhere down.

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+      break;
+    }
+
+    case AZ_HFSM_EVENT_ERROR:
+      printf(LOG_APP_ERROR "Error Event\n");
+      // az_platform_critical_error();
+      break;
+
+    default:
+      // TODO:
```
ewertons, line 428: Empty TODO

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+    topic_length += az_span_size(command_name);
+    topic_length -= az_span_size(command_name_key);
+    max_temp_length = topic_length > max_temp_length ? topic_length : max_temp_length;
+  }
+  if (az_span_find(format, executor_id_key) >= 0)
+  {
+    has_substitution = true;
+    _az_PRECONDITION_VALID_SPAN(executor_client_id, 1, false);
+    topic_length += az_span_size(executor_client_id);
+    topic_length -= az_span_size(executor_id_key);
```
ewertons, line 80:
```suggestion
    topic_length -= az_span_size(executor_client_id_key);
```

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+    max_temp_length = topic_length > max_temp_length ? topic_length : max_temp_length;
+  }
+  if (az_span_find(format, invoker_id_key) >= 0)
+  {
+    has_substitution = true;
+    _az_PRECONDITION_VALID_SPAN(invoker_client_id, 1, false);
+    topic_length += az_span_size(invoker_client_id);
+    topic_length -= az_span_size(invoker_id_key);
```
ewertons, line 88:
```suggestion
    topic_length -= az_span_size(invoker_client_id_key);
```

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+    max_temp_length = topic_length > max_temp_length ? topic_length : max_temp_length;
+  }
+
+  // Must be large enough to fit the entire format as items are substituted throughout the function
+  // even if that's larger than the output topic
+  _az_PRECONDITION_VALID_SPAN(out_topic, max_temp_length, true);
```
ewertons, line 94: So at this point `max_temp_length` will be equal to `topic_length`. Why not use `topic_length` instead then? and not declare `max_temp_length`

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+  {
+    az_span_copy(out_topic, format);
+    if (out_topic_length != NULL)
+    {
+      *out_topic_length = format_size;
+    }
+    return AZ_OK;
+  }
+
+  // Must be large enough to fit the entire format even if that's larger than the output topic
```
ewertons, line 107: Suggestion:
create a _az_span_replace(az_span in, az_span from, az_span to, az_span* out_replaced);
and shorten the code below.

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+AZ_NODISCARD az_result az_rpc_get_topic_from_format(
+    az_span model_id,
+    az_span executor_client_id,
+    az_span invoker_client_id,
+    az_span command_name,
+    az_span format,
+    az_span out_topic,
+    int32_t* out_topic_length)
+{
+  const az_span model_id_key = AZ_SPAN_FROM_STR("{serviceId}");
```
ewertons, line 48: Is there any situation where any of these keys would appear more than once in the topic format? Or is it documented that that is not supported?

sdk/src/azure/core/az_mqtt5_rpc_client.c
```diff
@@ -10,100 +10,65 @@
...
 {
-  return (az_mqtt5_rpc_client_options){ .subscribe_timeout_in_seconds
-                                        = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
-                                        .publish_timeout_in_seconds
-                                        = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS };
+  return (az_mqtt5_rpc_client_options){
+    .subscribe_timeout_in_seconds = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
+    .publish_timeout_in_seconds = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
+    .subscription_topic_format
+    = AZ_SPAN_FROM_STR("vehicles/{serviceId}/commands/{executorId}/{name}/__for_{invokerId}\0"),
```
ewertons, line 17: It feels like this mixes a sample with the actual SDK. Maybe the default formats should be AZ_SPAN_EMPTY?

sdk/src/azure/core/az_mqtt5_rpc_server.c
```diff
@@ -12,39 +12,23 @@ AZ_NODISCARD az_mqtt5_rpc_server_options az_mqtt5_rpc_server_options_default()
 {
   return (az_mqtt5_rpc_server_options){
     .subscribe_timeout_in_seconds = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
+    .subscription_topic_format
```
ewertons, line 15: It feels like this mixes a sample with the actual SDK. Maybe the default formats should be AZ_SPAN_EMPTY?

ewertons (approved):

sdk/src/azure/core/az_mqtt5_rpc.c
vaavva: Not necessarily - topic_length is the final topic length. If we create the topic in place in the out_topic span, then it needs to be able to have some keys and some substituted values. Ex. format is "type/{invokerId}/{executorId}/{name}/{serviceId}\0" and invokerId is "1", executorId is "2", etc, then the out_topic needs to be able to hold 
"type/{invokerId}/{executorId}/{name}/{serviceId}\0"  and then 
"type/1/{executorId}/{name}/{serviceId}\0"  and then 
"type/1/2/{name}/{serviceId}\0"   and then 
"type/1/2/3/{serviceId}\0" and then finally the topic_length span:
"type/1/2/3/4\0"

I might be able to get around this by creating another temporary az_span to have that longest topic in and then out_topic wouldn't need to be as large - I was having trouble getting that working the other day, but I might have some more ideas to try.

sdk/src/azure/core/az_mqtt5_rpc.c
vaavva: ooh, interesting question. I'll have to find out if that's supported or not

sdk/src/azure/core/az_mqtt5_rpc_client.c
vaavva: that works for me! If there ends up being a default topic format in the future, we could always add that back in.

sdk/inc/azure/core/az_mqtt5_rpc_client.h
```diff
@@ -43,6 +43,22 @@ typedef struct
    */
   uint32_t publish_timeout_in_seconds;
 
+  /**
+   * @brief The topic format to use for the subscription topic.
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, {executorId} for the
+   * server's client_id, and/or {invokerId} for the client's client_id
```
ericwol-msft, line 50: Some of these end with a period, and others do not. What is the consistency here?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+
+// Static memory allocation
+static char client_response_topic_buffer[256];
+static char client_request_topic_buffer[256];
+static char client_subscription_topic_buffer[256];
+static uint8_t client_correlation_id_buffers[RPC_CLIENT_MAX_PENDING_COMMANDS]
+                                            [AZ_MQTT5_RPC_CORRELATION_ID_LENGTH];
+
+static char server_subscription_topic_buffer[256];
+// static char response_payload_buffer[256];
```
ericwol-msft, line 54: needed?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    // TODO: deserialize request
+
+    rc = execute_command();
+
+    // if command hasn't timed out, send result back
+    if (az_span_is_content_equal(correlation_id_copy, pending_server_command.correlation_id))
+    {
+      stop_timer();
+      az_span response_payload = AZ_SPAN_EMPTY;
+      // if (rc == AZ_MQTT5_RPC_STATUS_OK)
```
ericwol-msft, line 242: Ifdef this

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+static az_mqtt5_rpc_client rpc_client;
+static az_mqtt5_rpc_client_policy rpc_client_policy;
+
+volatile bool sample_finished = false;
+
+static pending_commands_array client_pending_commands;
+static az_mqtt5_rpc_server_execution_req_event_data pending_server_command;
+
+#ifdef _WIN32
+static timer_t timer; // placeholder
```
ericwol-msft, line 78: why is timer in both defs?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    printf(LOG_APP_ERROR "Failed sending execution response to HFSM\n");
+    return;
+  }
+
+  pending_server_command.content_type = AZ_SPAN_FROM_BUFFER(server_content_type_buffer);
+  pending_server_command.request_topic = AZ_SPAN_FROM_BUFFER(server_request_topic_buffer);
+  pending_server_command.correlation_id = AZ_SPAN_EMPTY;
+
+#ifdef _WIN32
+  return AZ_ERROR_DEPENDENCY_NOT_PROVIDED;
```
ericwol-msft, line 134: How does this compile under win32??

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+#else
+  if (0 != timer_delete(timer))
```
ericwol-msft, line 136: delete timer as first step in func?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+  trigger.it_value.tv_sec = delay_milliseconds / 1000;
+  trigger.it_value.tv_nsec = (delay_milliseconds % 1000) * 1000000;
+
+  if (0 != timer_settime(timer, 0, &trigger, NULL))
+  {
+    return AZ_ERROR_ARG;
+  }
+#endif
+
+  return AZ_OK;
```
ericwol-msft, line 170: move return AZ_O into #else -> #endif

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+#ifdef _WIN32
+  return AZ_ERROR_DEPENDENCY_NOT_PROVIDED;
+#else
+  if (0 != timer_delete(timer))
+  {
+    return AZ_ERROR_ARG;
+  }
+#endif
+
+  return AZ_OK;
```
ericwol-msft, line 187: move into else

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+  // for now, just print details from the command
+  printf(
+      LOG_APP "Executing command from topic: %s\n",
+      az_span_ptr(pending_server_command.request_topic));
+  return AZ_MQTT5_RPC_STATUS_OK;
+}
+
+/**
+ * @brief Check if there is a pending command and execute it. On completion, if the command hasn't
+ * timed out, send the result back to the hfsm
```
ericwol-msft, line 205:
```suggestion
 * timed out, send the result back to the HFSM
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+ * @note Result to be sent back to the hfsm needs to be modified for your solution
```
ericwol-msft, line 206:
```suggestion
 * @note Result to be sent back to the HFSM needs to be modified for your solution
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
ericwol-msft: other places this is in CAPS

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+ */
+az_result check_for_commands_to_execute()
+{
+  if (az_span_ptr(pending_server_command.correlation_id) != NULL)
+  {
+    // copy correlation id to a new span so we can compare it later
+    uint8_t copy_buffer[az_span_size(pending_server_command.correlation_id)];
```
ericwol-msft, line 213:
```suggestion
    uint8_t correlation_id_buffer[az_span_size(pending_server_command.correlation_id)];
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    az_span correlation_id_copy
+        = az_span_create(copy_buffer, az_span_size(pending_server_command.correlation_id));
```
ericwol-msft, line 215:
```suggestion
        = az_span_create(copy_buffer, correlation_id_size(pending_server_command.correlation_id));
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+az_result mqtt_callback(az_mqtt5_connection* client, az_event event)
+{
+  (void)client;
+  az_app_log_callback(event.type, AZ_SPAN_FROM_STR("APP/callback"));
+  switch (event.type)
+  {
+    case AZ_MQTT5_EVENT_CONNECT_RSP:
+    {
+      az_mqtt5_connack_data* connack_data = (az_mqtt5_connack_data*)event.data;
+      printf(LOG_APP "CONNACK: %d\n", connack_data->connack_reason);
```
ericwol-msft, line 307:
```suggestion
      printf(LOG_APP "CONNACK: reason=%d\n", connack_data->connack_reason);
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    {
+      az_mqtt5_rpc_server_execution_req_event_data data
+          = *(az_mqtt5_rpc_server_execution_req_event_data*)event.data;
+      // can check here for the expected request topic to determine which command to execute
+      if (az_span_ptr(pending_server_command.correlation_id) != NULL)
+      {
+        // can add this command to a queue to be executed if the application supports executing
+        // multiple commands at once.
+        printf(LOG_APP
+               "Received command while another command is executing. Sending error response.\n");
```
ericwol-msft, line 331:
```suggestion
               "Received new command while another command is executing. Sending error response.\n");
```

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+                az_mqtt5_rpc_server_execution_finish(&rpc_server_policy, &return_data)))
+        {
+          printf(LOG_APP_ERROR "Failed sending execution response to HFSM\n");
+        }
+      }
+      else
+      {
+        // Mark that there's a pending command to be executed
+        LOG_AND_EXIT_IF_FAILED(copy_execution_event_data(&pending_server_command, data));
+        start_timer(NULL, 10000);
```
ericwol-msft, line 350: humm #define for 10000

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+  az_log_set_classification_filter_callback(az_sdk_log_filter_callback);
+
+  connection_context = az_context_create_with_expiration(
+      &az_context_application, az_context_get_expiration(&az_context_application));
+
+  az_mqtt5 mqtt5;
+  struct mosquitto* mosq = NULL;
+
+  az_mqtt5_options mqtt5_options = az_mqtt5_options_default();
+  mqtt5_options.disable_tls = true;
```
ericwol-msft, line 459: make an IFDEF for this?

ericwol-msft: does MosquittoLib support cert store?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+
+  LOG_AND_EXIT_IF_FAILED(az_mqtt5_init(&mqtt5, &mosq, &mqtt5_options));
+
+  az_mqtt5_connection_options connection_options = az_mqtt5_connection_options_default();
+  connection_options.client_id_buffer = client_id;
+  connection_options.username_buffer = username;
+  connection_options.password_buffer = AZ_SPAN_EMPTY;
+  connection_options.hostname = hostname;
+  connection_options.port = 1883;
```
ericwol-msft, line 468: #define

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+
+  LOG_AND_EXIT_IF_FAILED(
+      pending_commands_array_init(&client_pending_commands, client_correlation_id_buffers));
+
+  LOG_AND_EXIT_IF_FAILED(az_mqtt5_connection_open(&mqtt_connection));
+
+  az_result rc;
+
+  // infinite execution loop
+  for (int i = 20; !sample_finished && i > 0; i++)
```
ericwol-msft, line 528: why start at 20??

ericwol-msft: i > 0
why this check?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+  {
+    // check for commands for the server
+    LOG_AND_EXIT_IF_FAILED(check_for_commands_to_execute());
+
+    // remove any expired commands from the client
+    pending_command* expired_command = get_first_expired_command(client_pending_commands);
+    while (expired_command != NULL)
```
ericwol-msft, line 535: refactor loop
func expired_command()
func invoke_begin(start_module_command_name)
func invoke_begin(stop_module_command_name)

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+      print_correlation_id(expired_command->correlation_id);
+      printf(" expired\n");
+      az_result ret = remove_command(&client_pending_commands, expired_command->correlation_id);
+      if (ret != AZ_OK)
+      {
+        printf(LOG_APP_ERROR "Expired command not a pending command\n");
+      }
+      expired_command = get_first_expired_command(client_pending_commands);
+    }
+#ifdef _WIN32
```
ericwol-msft, line 547: sleep at end of loop?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+    Sleep((DWORD)1000);
+#else
+    sleep(1);
+#endif
+    printf(LOG_APP "Waiting...\r");
```
ericwol-msft, line 552: move above sleep?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,629 @@
...
+      uuid_generate(new_uuid);
+      az_mqtt5_rpc_client_invoke_req_event_data command_data
+          = { .correlation_id
+              = az_span_create((uint8_t*)new_uuid, AZ_MQTT5_RPC_CORRELATION_ID_LENGTH),
+              .content_type = content_type,
+              .rpc_server_client_id = server_client_id,
+              .command_name = stop_module_command_name,
+              // TODO: Payload should be generated and serialized
+              .request_payload = AZ_SPAN_FROM_STR(
+                  "{\"RequestTimestamp\":1691530585198,\"RequestedFrom\":\"mobile-app\"}") };
```
ericwol-msft, line 593: add get time for time stamp

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+    topic_length += az_span_size(invoker_client_id);
+    topic_length -= az_span_size(invoker_id_key);
+    max_temp_length = topic_length > max_temp_length ? topic_length : max_temp_length;
+  }
+
+  // Must be large enough to fit the entire format as items are substituted throughout the function
+  // even if that's larger than the output topic
+  _az_PRECONDITION_VALID_SPAN(out_topic, max_temp_length, true);
+
+  // if there are no substitutions, just copy the format to the output
```
ericwol-msft, line 96:
```suggestion
  // if there are no topic substitutions, just copy the format to the output
```

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+      = format_size > az_span_size(out_topic) ? format_size : az_span_size(out_topic);
+  uint8_t format_buf[format_buf_size];
+  az_span temp_format_buf = az_span_create(format_buf, format_buf_size);
+  az_span_copy(temp_format_buf, format);
+  temp_format_buf = az_span_slice(temp_format_buf, 0, format_size);
+
+  az_span temp_span = out_topic;
+
+  int32_t index = az_span_find(temp_format_buf, model_id_key);
+  if (index >= 0)
```
ericwol-msft, line 118: error if index == 0?

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,191 @@
...
+    temp_format_buf = az_span_create(format_buf, format_buf_size);
+    az_span_copy(temp_format_buf, out_topic);
+    temp_format_buf = az_span_slice(temp_format_buf, 0, format_size);
+    temp_span = out_topic;
+  }
+
+  index = az_span_find(temp_format_buf, command_name_key);
+  if (index >= 0)
+  {
+    format_size += az_span_size(command_name);
```
ericwol-msft, line 137: can all these IF's be refactored into one func()?

ericwol-msft (approved):

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -103,10 +93,38 @@ AZ_NODISCARD AZ_INLINE bool az_span_topic_matches_sub(az_span sub, az_span topic
...
+ * @param[in] format the format string to use to generate the topic. Can include {name} for command
+ * name, {serviceId} for model id, {executorId} for the server's client_id, and/or {invokerId} for
+ * the client's client_id
+ * @param[out] out_topic the buffer to write the topic to
+ * @param[out] out_topic_length the length of the topic written to the buffer. Can pass in NULL if
+ * you don't need this value
+ *
+ * @return #az_result indicating the result of the operation
+ */
+AZ_NODISCARD az_result az_rpc_get_topic_from_format(
```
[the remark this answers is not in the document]
vaavva: fixed!

sdk/src/azure/core/az_mqtt5_rpc_client.c
vaavva: changed to AZ_SPAN_EMPTY and added setting the formats in all samples

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Let me know if the change I pushed for this was what you had in mind

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Let me know if the change I pushed for this was what you had in mind

sdk/samples/core/mosquitto_rpc_client_sample.c
```diff
@@ -247,55 +289,27 @@ int main(int argc, char* argv[])
...
-#ifdef _WIN32
-    Sleep((DWORD)1000);
-#else
-    sleep(1);
-#endif
     printf(LOG_APP "Waiting...\r");
     fflush(stdout);
+
+    // remove any expired commands from the client
+    remove_expired_commands();
```
CIPop, line 298: This expiration is computed on the client-side. We could add a comment that, even for expired commands, we will be calling the callback with results that are received from the other end.

CIPop, line 298: There is a race condition if the MQTT stack operates on a different thread (e.g. Mosquitto, AsyncPaho).
The `pending_commands` object is now a shared resource and it must be guarded by a mutex.

CIPop (changes requested):
Race condition on all operations on `pending_commands` between the `mqtt_callback` thread and the main thread. This is applicable to both MosquittoMQTT and AsyncPaho in threaded mode.

sdk/samples/core/mosquitto_rpc_client_sample.c
vaavva: I have plans to move the pending_commands functionality within the sdk, where we could use the platform mutex. I'm still figuring out the level of data that might be needed for different solutions since it won't be customer modifiable if it's in the sdk. This isn't a new issue in this PR, so I'll address this in a follow-up PR either moving the pending commands functionality to the sdk or implementing mutexes in the samples.

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Fixed! We now add the null-terminator in the topic generation

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -207,47 +207,60 @@ _handle_request(az_mqtt5_rpc_server_policy* this_policy, az_mqtt5_recv_data* dat
...
-  // {
-  // az_event_policy_send_inbound_event((az_event_policy*)this_policy, (az_event){.type =
-  // AZ_MQTT5_EVENT_RPC_SERVER_EXECUTE_COMMAND_REQ, .data = data});
-  // }
-  _az_RETURN_IF_FAILED(_az_mqtt5_connection_api_callback(
-      this_policy->_internal.connection,
-      (az_event){ .type = AZ_MQTT5_EVENT_RPC_SERVER_EXECUTE_COMMAND_REQ, .data = &command_data }));
+    if (!az_result_failed(ret))
+    {
+      // validate request isn't expired?
```
RLeclair, line 227: Should these comments be marked as TODO?

RLeclair (approved):
LGTM!

sdk/samples/core/mosquitto_rpc_client_sample.c
vaavva: Added mutexes to the server sample pending_command struct and the client_pending_commands file

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Resolved windows concerns by using platform implementations

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Resolved windows concerns by using platform implementations

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -77,36 +77,54 @@ typedef enum
...
-  bool ret;
-  // TODO: have this not be mosquitto specific
-  if (MOSQ_ERR_SUCCESS
-      != mosquitto_topic_matches_sub((char*)az_span_ptr(sub), (char*)az_span_ptr(topic), &ret))
-  {
-    ret = false;
-  }
-  return ret;
-}
+AZ_NODISCARD bool az_span_topic_matches_sub(az_span sub, az_span topic);
```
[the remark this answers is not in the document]
vaavva: Will address this in a follow-up PR

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Will add this when we do proper payload generation/serialization

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: moved this sample to have the same default setup as other samples

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: yep!

sdk/src/azure/core/az_mqtt5_rpc.c
vaavva: It's not explicitly disallowed right now - I added a comment for now calling out the limitation in our code, but we might have to support this scenario in the future

sdk/inc/azure/core/az_mqtt5_rpc.h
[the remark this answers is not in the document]
CIPop: Likely `az_mqtt5_rpc_client_parse_received_topic` and `az_mqtt5_rpc_server_parse_received_topic` (different for client and server). 
The implementation can be the same.

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,603 @@
...
+
+  az_mqtt5_property_bag server_property_bag;
+  mosquitto_property* server_mosq_prop = NULL;
+  LOG_AND_EXIT_IF_FAILED(
+      az_mqtt5_property_bag_init(&server_property_bag, &mqtt5, &server_mosq_prop));
+
+  az_mqtt5_rpc_server_options server_options = az_mqtt5_rpc_server_options_default();
+  server_options.subscription_topic_format = server_subscription_topic_format;
+
+  LOG_AND_EXIT_IF_FAILED(az_rpc_server_policy_init(
```
CIPop, line 524: Policies shouldn't leak through the public API.
Could we rename `az_mqtt5_rpc_server_policy` to `az_mqtt5_rpc_server` and the current `az_mqtt5_rpc_server` to `az_mqtt5_rpc_server_codec`? (same for client)

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -83,14 +73,20 @@ void az_platform_critical_error()
...
+static void timer_callback(void* callback_context)
 {
-#ifdef _WIN32
-  return; // AZ_ERROR_DEPENDENCY_NOT_PROVIDED
-#else
-  // void* callback_context = sv.sival_ptr;
-  (void)sv;
-#endif
+  (void)callback_context;
+  if (az_result_failed(az_platform_timer_destroy(&timer)))
```
CIPop: Same as above: should be under the lock.

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -184,11 +131,12 @@ az_mqtt5_rpc_status execute_command(unlock_request req)
 
 /**
  * @brief Check if there is a pending command and execute it. On completion, if the command hasn't
- * timed out, send the result back to the hfsm
- * @note Result to be sent back to the hfsm needs to be modified for your solution
+ * timed out, send the result back to the HFSM
```
CIPop, line 134: We should be consistent with naming and avoid using HFSM which is an implementation detail.
Maybe "send results back to the mqtt_callback?" or rpc_server callback?

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -385,16 +345,24 @@ int main(int argc, char* argv[])
   LOG_AND_EXIT_IF_FAILED(az_mqtt5_connection_init(
       &mqtt_connection, &connection_context, &mqtt5, mqtt_callback, &connection_options));
 
+  LOG_AND_EXIT_IF_FAILED(az_platform_mutex_init(&pending_command_mutex));
+  LOG_AND_EXIT_IF_FAILED(az_platform_mutex_acquire(&pending_command_mutex));
```
CIPop, line 355: Since 'open' wasn't yet called, acquiring the mutex here isn't necessary.

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -0,0 +1,180 @@
...
+
+#include <azure/core/_az_cfg.h>
+
+static const az_span model_id_key = AZ_SPAN_LITERAL_FROM_STR("{serviceId}");
+static const az_span command_name_key = AZ_SPAN_LITERAL_FROM_STR("{name}");
+static const az_span executor_client_id_key = AZ_SPAN_LITERAL_FROM_STR("{executorId}");
+static const az_span invoker_client_id_key = AZ_SPAN_LITERAL_FROM_STR("{invokerId}");
+
+AZ_NODISCARD int32_t
+_az_rpc_calculate_substitution_length(az_span format, az_span key, az_span value);
```
CIPop, line 23: Should at least some of these `_az' functions be made `static` (or even `AZ_INLINE` for those used in a single place)?

sdk/src/azure/core/az_mqtt5_rpc_client.c
```diff
@@ -13,97 +13,60 @@ AZ_NODISCARD az_mqtt5_rpc_client_options az_mqtt5_rpc_client_options_default()
...
+      client->_internal.model_id,
+      AZ_SPAN_FROM_STR("+"),
+      client->_internal.client_id,
+      _az_span_is_valid(client->_internal.command_name, 1, 0) ? client->_internal.command_name
+                                                              : AZ_SPAN_FROM_STR("+"),
+      out_subscription_topic,
+      out_topic_length);
 }
 
+// Replaces + in subscription topic with server client id
```
CIPop, line 37:
```suggestion
// Replaces + in subscription topic with RPC server's client_id.
```

sdk/src/azure/core/az_mqtt5_rpc_client.c
```diff
@@ -13,97 +13,60 @@ AZ_NODISCARD az_mqtt5_rpc_client_options az_mqtt5_rpc_client_options_default()
...
 AZ_NODISCARD az_result az_rpc_client_get_response_topic(
     az_mqtt5_rpc_client* client,
     az_span server_client_id,
```
CIPop, line 40: Can we instead pass an `az_mqtt5_rpc_server_codec` object? We shouldn't be leaking internals within the codec.

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -556,6 +556,8 @@ static az_result ready(az_event_policy* me, az_event event)
       _az_RETURN_IF_FAILED(az_rpc_client_get_response_topic(
           this_policy->_internal.rpc_client,
           event_data->rpc_server_client_id,
+          _az_span_is_valid(event_data->command_name, 1, 0) ? event_data->command_name
```
CIPop, line 559: Is there a case where the span is invalid on purpose?
If not, I think the checks should be done as preconditions rather than product code.

If, for example, the span _was_ valid but got corrupted (so that the size is now negative for example), this code will ignore the corruption and pass an empty span.
Instead, we should only accept _valid_ spans (including empty) and perform the validation within the codec function.

sdk/src/azure/core/az_mqtt5_rpc_server.c
```diff
@@ -10,41 +10,23 @@
...
-  return (az_mqtt5_rpc_server_options){
-    .subscribe_timeout_in_seconds = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
-  };
+  return (az_mqtt5_rpc_server_options){ .subscribe_timeout_in_seconds
+                                        = AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS,
+                                        .subscription_topic_format = AZ_SPAN_EMPTY };
 }
 
 AZ_NODISCARD az_result
 az_rpc_server_get_subscription_topic(az_mqtt5_rpc_server* client, az_span out_subscription_topic)
```
CIPop, line 19: Functions must be named according to the object.
Since the object is `az_mqtt5_rpc_server` the function name must be `az_mqtt5_rpc_server_get_subscription_topic`. Same with all other codec public APIs.

CIPop (approved):
Left a few comments regarding mutexes and naming convention.

sdk/src/azure/core/az_mqtt5_rpc_client.c
vaavva: This is the code server object - this will be clearer after the rename

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
vaavva: yep! The application has the option of passing in the command name for this specific invocation or passing AZ_SPAN_EMPTY and using the one that's on the client codec object.

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -0,0 +1,603 @@
...
+    // TODO: deserialize request
+
+    LOG_AND_EXIT_IF_FAILED(az_platform_mutex_release(&pending_server_command_mutex));
+    rc = execute_command();
+    LOG_AND_EXIT_IF_FAILED(az_platform_mutex_acquire(&pending_server_command_mutex));
+
+    // if command hasn't timed out, send result back
+    if (az_span_is_content_equal(correlation_id_copy, pending_server_command.correlation_id))
+    {
+      LOG_AND_EXIT_IF_FAILED(az_platform_timer_destroy(&timer));
```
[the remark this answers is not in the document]
vaavva: moved both to be under the mutex

sdk/samples/core/mosquitto_multiple_rpc_sample.c
vaavva: Will address in follow-up PR

PR #2647, opened 2023-09-26
Topic filter matching helper function

RLeclair:
- Adding internal helper function to eliminate reliance on MQTT implementation's topic filter matching function.

### DIFF
--- a/sdk/src/azure/core/az_mqtt5_rpc.c
+++ b/sdk/src/azure/core/az_mqtt5_rpc.c
@@ -30,21 +30,137 @@ AZ_NODISCARD static az_result _az_rpc_substitute_key_for_value(
     az_span value,
     az_span out_topic);
 
-AZ_NODISCARD bool az_span_topic_matches_sub(az_span sub, az_span topic)
+/**
+ * @brief Helper function to obtain the next level of a topic.
+ *
+ * @param topic Topic to get the next level of.
+ * @return An #az_span value containing the next level of the topic including the backslash.
+ */
+AZ_NODISCARD AZ_INLINE az_span _get_next_topic_level(az_span topic)
 {
-  bool ret;
-#if defined(TRANSPORT_MOSQUITTO)
-  if (MOSQ_ERR_SUCCESS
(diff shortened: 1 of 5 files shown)


sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -30,21 +30,137 @@ AZ_NODISCARD static az_result _az_rpc_substitute_key_for_value(
...
+  if (az_span_find(topic, AZ_SPAN_FROM_STR("#")) > 0
+      || az_span_find(topic, AZ_SPAN_FROM_STR("+")) > 0)
+  {
+    return false;
+  }
+
+  az_span filter_remaining = topic_filter;
+  az_span topic_remaining = topic;
+
+  az_span filter_level = _get_next_topic_level(filter_remaining);
```
CIPop, line 87: We don't allow recursion in the embedded SDK.

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -30,21 +30,137 @@ AZ_NODISCARD static az_result _az_rpc_substitute_key_for_value(
...
+AZ_NODISCARD AZ_INLINE bool _has_backslash_in_topic(az_span topic)
+{
+  if (az_span_size(topic) == 0)
+  {
+    return false;
+  }
+  return az_span_find(topic, AZ_SPAN_FROM_STR("/")) == az_span_size(topic) - 1;
+}
+
+AZ_NODISCARD bool _az_span_topic_matches_filter(az_span topic_filter, az_span topic)
```
CIPop, line 70: This is too complicated: the RPC client and RPC server should comply with a fixed template instead of having a generic topic matcher.
The function shouldn't be more complicated than https://github.com/Azure/azure-sdk-for-c/blob/feature/v2/sdk/src/azure/iot/az_iot_hub_client_methods.c#L24C24-L24C70.

CIPop (changes requested):
The implementation is overly complicated for an embedded system and it relies on recursion.
Please follow the example of IoT Hub methods/twins that are simplified by having their own pattern.

PR #2657, opened 2023-10-09
Add context to az_mqtt5_connection_callback + additional fixes

ewertons:
- Added an user-defined context to az_mqtt5_connection_callback
- Added check for available memory size before copying correlation_id in az_mqtt5_rpc_client_hfsm.c
- Added missing reference to mosquitto.h in az_mqtt5_rpc.h

### DIFF
--- a/sdk/inc/azure/core/az_mqtt5_connection.h
+++ b/sdk/inc/azure/core/az_mqtt5_connection.h
@@ -64,7 +64,7 @@ enum az_event_type_mqtt5_connection
  * @brief Callback for MQTT 5 connection events.
  *
  */
-typedef az_result (*az_mqtt5_connection_callback)(az_mqtt5_connection* client, az_event event);
+typedef az_result (*az_mqtt5_connection_callback)(az_mqtt5_connection* client, az_event event, const void* event_callback_context);
 
 /**
  * @brief MQTT 5 connection options.
@@ -155,6 +155,12 @@ struct az_mqtt5_connection
     az_mqtt5_connection_callback event_callback;
 
     /**
+     * @brief Context for MQTT 5 connection events callback.
+     *
+     */
+    const void* event_callback_context;
+
+    /**
      * @brief Options for the MQTT 5 connection.
      *
      */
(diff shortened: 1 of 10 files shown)


sdk/inc/azure/core/az_mqtt5_connection.h
```diff
@@ -64,7 +64,7 @@ enum az_event_type_mqtt5_connection
  * @brief Callback for MQTT 5 connection events.
  *
  */
-typedef az_result (*az_mqtt5_connection_callback)(az_mqtt5_connection* client, az_event event);
+typedef az_result (*az_mqtt5_connection_callback)(az_mqtt5_connection* client, az_event event, const void* event_callback_context);
```
ericwol-msft, line 67:
```suggestion
typedef az_result (*az_mqtt5_connection_callback)(az_mqtt5_connection* client, az_event event, const void* callback_context);
```

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -20,6 +20,8 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
+#include "mosquitto.h"
```
ericwol-msft, line 23: Why here??

ewertons: Thanks for asking. I double checked and this is no longer needed here.
Before #8ccc3ac5d6db61e74ecb106e7ca8549728ad7a83, there was a reference in `az_mqtt5_rpc.h` to `MOSQ_ERR_SUCCESS`, which comes from `mosquitto.h`.
Since it's no longer needed I am removing it from this PR.

sdk/inc/azure/core/az_mqtt5_connection.h
```diff
@@ -154,6 +154,12 @@ struct az_mqtt5_connection
      */
     az_mqtt5_connection_callback event_callback;
 
+    /**
+     * @brief Context for MQTT 5 connection events callback.
+     *
+     */
+    const void* event_callback_context;
```
vaavva, line 161: since there isn't an example in the samples - what is this expected to be used for?

vaavva (approved):

sdk/inc/azure/core/az_mqtt5_connection.h
ewertons: The use case is when a customer wants to add this SDK into their solution and they do not want to define global variables to access the application state. They want to invoke the _init function from main and receive a reference to their app state as a context in the callback.

sdk/inc/azure/core/az_mqtt5_connection.h
CIPop, line 67: I'm not convinced on why this would be `const`. If the application wants to change the data, it should be able to do so - it's not the SDK's data to begin with.

PR #2662, opened 2023-10-12
Adding client certificate count

RLeclair:


sdk/src/azure/core/az_mqtt5_connection.c
```diff
@@ -16,6 +16,7 @@ AZ_NODISCARD az_mqtt5_connection_options az_mqtt5_connection_options_default()
   return (az_mqtt5_connection_options){
     .hostname = AZ_SPAN_EMPTY,
     .port = AZ_MQTT5_DEFAULT_CONNECT_PORT,
+    .client_certificate_count = 0,
```
CIPop, line 19: Can we also initialize the array of structs with `{ 0 }` even if we don't know its size ? (Not sure that compiles.)

CIPop (approved):

vaavva (approved):

PR #2668, opened 2023-10-27
[RPC Server] fix segfault on incoming publish with missing properties

vaavva:
Problem: On an incoming RPC execution request, if the required MQTT5 properties are missing, an error is sent back to the MQTT client, causing a seg fault

Fix: Instead of sending this error to the MQTT client, which can't do anything about the problem, instead just log and ignore the publish

### DIFF
--- a/sdk/inc/azure/core/az_event_policy.h
+++ b/sdk/inc/azure/core/az_event_policy.h
@@ -78,7 +78,7 @@ az_event_policy_send_inbound_event(az_event_policy* policy, az_event const event
   {
     // Replace the original event with an error event that is flowed to the application.
     ret = inbound->inbound_handler(
-        policy,
+        inbound,
         (az_event){ AZ_HFSM_EVENT_ERROR,
                     &(az_hfsm_event_data_error){
                         .error_type = ret,
--- a/sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
+++ b/sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
@@ -59,9 +59,20 @@ static az_result root(az_event_policy* me, az_event event)
 
     case AZ_HFSM_EVENT_ERROR:
     {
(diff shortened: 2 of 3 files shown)


sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -337,7 +337,13 @@ static az_result waiting(az_event_policy* me, az_event event)
         }
 
         // parse the request details and send it to the application for execution
-        _az_RETURN_IF_FAILED(_handle_request(this_policy, recv_data));
+        if (az_result_failed(_handle_request(this_policy, recv_data)))
```
CIPop, line 340: There should never be a seg-fault: while I do agree with the fix to log in this case and ignore the message, inbound posts that cause faults should be handled by the MQTT implementation.

CIPop (approved):

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
vaavva: Addressed in az_event_policy - before, if a message had failed to be sent inbound, it would send a failure message to the current policy, while the failure message should also be sent inbound (which will be a policy collection, unlike the current policy)

sdk/inc/azure/core/az_event_policy.h
```diff
@@ -78,7 +78,7 @@ az_event_policy_send_inbound_event(az_event_policy* policy, az_event const event
   {
     // Replace the original event with an error event that is flowed to the application.
     ret = inbound->inbound_handler(
-        policy,
+        inbound,
```
CIPop, line 81: Excellent find!

CIPop (approved):
Thanks @vaavva for finding the pipeline bug!

RLeclair (approved):
LGTM!

PR #2671, opened 2023-10-30
RPC Codec Update

RLeclair:
RPC Codec update
- Includes updates to the RPC HFSM code to integrate with updated codec APIs.
- Please look at first commit to see changes without file rename


sdk/inc/azure/core/az_mqtt5_rpc_server_hfsm.h
```diff
@@ -15,119 +14,19 @@
...
 
 #include <azure/core/az_mqtt5_connection.h>
 #include <azure/core/az_mqtt5_rpc.h>
+#include <azure/core/az_mqtt5_rpc_server_codec.h>
 #include <azure/core/az_result.h>
 #include <azure/core/az_span.h>
 
 #include <azure/core/_az_cfg_prefix.h>
 
 // ~~~~~~~~~~~~~~~~~~~~ Codec Public API ~~~~~~~~~~~~~~~~~
```
vaavva, line 28:
```suggestion
```

sdk/inc/azure/core/az_mqtt5_rpc_server_hfsm.h
```diff
@@ -262,42 +166,42 @@ typedef struct
...
- * @param[out] client The #az_mqtt5_rpc_server to initialize.
+ * @param[out] client The #az_mqtt5_rpc_server_hfsm to initialize.
  * @param[in] rpc_server_codec The #az_mqtt5_rpc_server_codec to initialize and use within the RPC
  * Server.
  * @param[in] connection The #az_mqtt5_connection to use for the RPC Server.
  * @param[in] property_bag The application allocated #az_mqtt5_property_bag to use for the
  * RPC Server.
  * @param[in] subscription_topic The application allocated #az_span to use for the subscription
  * topic.
+ * @param[in] service_group_id The service group id to use for the subscription topic. May be
```
vaavva, line 186: What is the service_group_id used for?

sdk/inc/azure/core/az_mqtt5_rpc_client_codec.h
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The client was initialized successfully.
+ * @retval #AZ_ERROR_ARG An invalid argument was provided.
+ */
+AZ_NODISCARD az_result az_mqtt5_rpc_client_codec_init(
+    az_mqtt5_rpc_client_codec* client,
+    az_span client_id,
+    az_span model_id,
+    az_span command_name,
```
CIPop, line 232: Same as above: command_name should be a variable passed when requesting and parsed from a response.

sdk/inc/azure/core/az_mqtt5_rpc_server_codec.h
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * @brief MQTT5 RPC Server Codec options.
+ *
+ */
+typedef struct
+{
+  /**
+   * @brief Timeout in seconds for subscribing acknowledgement.
+   */
+  int32_t subscribe_timeout_in_seconds;
```
CIPop, line 32: Same as with client: timeouts should not exist in the codec layer since they cannot be used.
We can though have some recommendations (through `#define` statements) on what these timeouts should be by default.

sdk/inc/azure/core/az_mqtt5_rpc_server_codec.h
```diff
@@ -0,0 +1,161 @@
...
+   * @brief An #az_span containing the service group id to use for the subscription topic, can be
+   * AZ_SPAN_EMPTY if not using a shared subscription topic.
+   */
+  az_span service_group_id;
+
+  /**
+   * @brief The topic format to use for the subscription topic.
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, and/or {executorId} for
+   * the server's client_id.
```
CIPop, line 44: Please add the default value in the documentation as an example.

sdk/inc/azure/core/az_mqtt5_rpc_client_codec.h
```diff
@@ -0,0 +1,237 @@
...
+   * @brief Timeout in seconds for publishing (must be > 0).
+   */
+  int32_t publish_timeout_in_seconds;
+
+  /**
+   * @brief The topic format to use for the subscription topic.
+   *
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, {executorId} for the
+   * server's client_id, and/or {invokerClientId} for the client's client_id.
```
CIPop, line 44: Please add the default value in the documentation as an example.

sdk/inc/azure/core/az_mqtt5_rpc_client_codec.h
```diff
@@ -0,0 +1,237 @@
...
+   */
+  az_span subscription_topic_format;
+
+  /**
+   * @brief The topic format to use for the request topic.
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, and/or {executorId} for
+   * the server's client_id.
+   */
+  az_span request_topic_format;
```
CIPop, line 54: Please add the default value in the documentation as an example.

sdk/inc/azure/core/az_mqtt5_rpc_server_hfsm.h
CIPop: This is used in $shared subscriptions.
@RLeclair please clarify within the documentation statement and add an example of that this is for (e.g. `$shared/mygroup/<my_service_topic>`)

sdk/inc/azure/core/az_mqtt5_rpc_server_hfsm.h
```diff
@@ -262,42 +166,42 @@ typedef struct
...
  * @param[in] client_id The client id to use for the subscription topic. May be AZ_SPAN_EMPTY if not
  * used in the subscription topic.
- * @param[in] command_name The command name to use for the subscription topic or AZ_SPAN_EMPTY to
- * have this rpc server handle all commands for this topic.
  * @param[in] options Any #az_mqtt5_rpc_server_codec_options to use for the RPC Server.
  *
  * @return An #az_result value indicating the result of the operation.
  */
 AZ_NODISCARD az_result az_mqtt5_rpc_server_init(
-    az_mqtt5_rpc_server* client,
```
CIPop, line 293: Is there a need to call these `_hfsm`? Can we have:
- `rpc_server_codec` for the codec.
- `rpc_server` the stateful implementation?

sdk/inc/azure/core/az_mqtt5_rpc_server_hfsm.h
```diff
@@ -262,42 +166,42 @@ typedef struct
...
+    az_mqtt5_rpc_server_hfsm* client,
     az_mqtt5_rpc_server_codec* rpc_server_codec,
     az_mqtt5_connection* connection,
     az_mqtt5_property_bag property_bag,
     az_span subscription_topic,
+    az_span service_group_id,
```
CIPop, line 202: This should be initialized once, within the rpc_server_codec. Using the pre-configured codec here should make service_group_ids work as expected.

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -35,7 +35,7 @@ static const az_span stop_module_command_name = AZ_SPAN_LITERAL_FROM_STR("stopMo
 static const az_span server_subscription_topic_format
     = AZ_SPAN_LITERAL_FROM_STR("device/{executorId}/command/{name}");
 static const az_span client_subscription_topic_format
-    = AZ_SPAN_LITERAL_FROM_STR("device/{executorId}/command/{name}/__for_{invokerId}");
+    = AZ_SPAN_LITERAL_FROM_STR("device/{executorId}/command/{name}/for/{invokerClientId}");
```
CIPop, line 38: Can't we use the defaults and avoid creating special topics for the samples? (i.e. remove all `*_topic_*` definitions, simplifying the code a bit).

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -368,17 +367,16 @@ int main(int argc, char* argv[])
...
       = az_mqtt5_rpc_server_codec_options_default();
-  server_codec_options.subscription_topic_format = subscription_topic_format;
 
   LOG_AND_EXIT_IF_FAILED(az_mqtt5_rpc_server_init(
       &rpc_server,
       &rpc_server_codec,
       &mqtt_connection,
       property_bag,
       AZ_SPAN_FROM_BUFFER(subscription_topic_buffer),
+      service_group_id,
```
CIPop, line 377: Same as above - I think the codec should keep this stored (by passing it as an option). The RPC server should simply use the pre-configured codec.

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -14,21 +14,11 @@
 
 #include <azure/core/_az_cfg.h>
 
+static const az_span service_group_id_key = AZ_SPAN_LITERAL_FROM_STR("{serviceGroupId}");
```
CIPop, line 17: Enhancement: the codec implementation should be separated from the HFSM one. An app-developer should be able to pick the codec .h/.c files (and core dependencies) and use them within their project, without the HFSM portions being included.

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
 #include <azure/core/internal/az_log_internal.h>
+#include <azure/core/internal/az_precondition_internal.h>
+#include <azure/core/internal/az_result_internal.h>
+#include <azure/iot/az_iot_common.h>
 #include <stdlib.h>
 
 #include <azure/core/_az_cfg.h>
 
+static const az_span az_mqtt5_rpc_client_codec_default_publish_topic_format
+    = AZ_SPAN_LITERAL_FROM_STR("services/{serviceId}/{executorId}/command/{name}/request");
```
CIPop, line 16: Please create `#define` statements within the include file with the default strings.

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
-      client->_internal.client_id,
-      _az_span_is_valid(client->_internal.command_name, 1, 0) ? client->_internal.command_name
-                                                              : command_name,
-      out_response_topic,
-      out_topic_length);
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_NOT_NULL(mqtt_topic);
+  _az_PRECONDITION(mqtt_topic_size > 0);
+
+  az_span topic_format = client->_internal.options.subscription_topic_format;
```
CIPop, line 238: Nit: this is very inefficient as it scans the strings several times.
A single-scan state machine would be advised, one that can be reused by both of the `get_*_topic` functions given that the tokens are always the same.

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
-      client->_internal.client_id,
-      _az_span_is_valid(client->_internal.command_name, 1, 0) ? client->_internal.command_name
-                                                              : command_name,
-      out_request_topic,
-      NULL);
+  _az_PRECONDITION_NOT_NULL(client);
+  _az_PRECONDITION_VALID_SPAN(received_topic, 1, false);
+  _az_PRECONDITION_NOT_NULL(out_response);
+
+  az_span topic_format = client->_internal.options.subscription_topic_format;
```
CIPop, line 315: This is the same code as above - it should be refactored so that the same function is reused.

CIPop (approved):

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -23,11 +23,10 @@ static const az_span key_path1 = AZ_SPAN_LITERAL_FROM_STR("<path to cert key fil
 static const az_span client_id = AZ_SPAN_LITERAL_FROM_STR("vehicle03");
 static const az_span username = AZ_SPAN_LITERAL_FROM_STR("vehicle03");
 static const az_span hostname = AZ_SPAN_LITERAL_FROM_STR("<hostname>");
-static const az_span command_name = AZ_SPAN_LITERAL_FROM_STR("unlock");
+// static const az_span command_name = AZ_SPAN_LITERAL_FROM_STR("unlock");
```
vaavva, line 26:
```suggestion
```

sdk/samples/core/paho_rpc_server_sample.c
```diff
@@ -23,11 +23,10 @@ static const az_span key_path1 = AZ_SPAN_LITERAL_FROM_STR("<path to cert key fil
 static const az_span client_id = AZ_SPAN_LITERAL_FROM_STR("vehicle03");
 static const az_span username = AZ_SPAN_LITERAL_FROM_STR("vehicle03");
 static const az_span hostname = AZ_SPAN_LITERAL_FROM_STR("<hostname>");
-static const az_span command_name = AZ_SPAN_LITERAL_FROM_STR("unlock");
+// static const az_span command_name = AZ_SPAN_LITERAL_FROM_STR("unlock");
```
vaavva, line 26:
```suggestion
```

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -67,234 +57,72 @@ AZ_NODISCARD AZ_INLINE bool _has_backslash_in_topic(az_span topic)
...
-AZ_NODISCARD static az_result _az_rpc_substitute_key_for_value(
-    az_span format,
-    uint8_t* format_buf,
-    az_span key,
-    az_span value,
-    az_span out_topic)
+AZ_NODISCARD bool _az_mqtt5_rpc_valid_topic_format(az_span topic_format)
 {
-  az_span temp_span = out_topic;
+  int32_t service_group_id_index = az_span_find(topic_format, service_group_id_key);
```
vaavva, line 81: Why not do each of these right before their if statement so if the function returns false on one of the first ones, we don't need to calculate the rest?

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
-  return az_rpc_get_topic_from_format(
-      client->_internal.options.subscription_topic_format,
-      client->_internal.model_id,
-      AZ_SPAN_FROM_STR("+"),
-      client->_internal.client_id,
-      _az_span_is_valid(client->_internal.command_name, 1, 0) ? client->_internal.command_name
-                                                              : AZ_SPAN_FROM_STR("+"),
-      out_subscription_topic,
-      out_topic_length);
+  _az_PRECONDITION_NOT_NULL(client);
```
vaavva, line 44: Could these get topic functions use a single underlying function with different parameters? They seem very similar

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
vaavva: Response topic and publish topic could be unique per invocation (although those two might be able to be created together?), and are generated at a different time than the subscribe topic, which would only be generated once

sdk/inc/azure/core/az_mqtt5_rpc_client_codec.h
```diff
@@ -0,0 +1,237 @@
...
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, {executorId} for the
+   * server's client_id, and/or {invokerClientId} for the client's client_id.
+   */
+  az_span subscription_topic_format;
+
+  /**
+   * @brief The topic format to use for the request topic.
+   *
+   * @note Can include {name} for command name, {serviceId} for model id, and/or {executorId} for
```
vaavva, line 51: If there are other requirements now with the slashes, can you include those here as well?

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -126,15 +126,16 @@ static az_result root(az_event_policy* me, az_event event)
...
     az_event event,
     az_event_policy_handler source_state)
 {
   az_mqtt5_recv_data* recv_data = (az_mqtt5_recv_data*)event.data;
+  az_mqtt5_rpc_client_codec_request_response request_response;
 
   // Ensure pub is of the right topic
-  if (_az_span_topic_matches_filter(
-          this_policy->_internal.rpc_client_codec->_internal.subscription_topic, recv_data->topic))
+  if (az_mqtt5_rpc_client_codec_parse_received_topic(
```
vaavva, line 137: Does this return a bool?

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -526,38 +528,44 @@ static az_result ready(az_event_policy* me, az_event event)
...
       _RETURN_AND_CLEAR_PROPERTY_BAG_IF_FAILED(
-          az_mqtt5_rpc_client_codec_get_request_topic(
+          az_mqtt5_rpc_client_codec_get_publish_topic(
               this_policy->_internal.rpc_client_codec,
               event_data->rpc_server_client_id,
               _az_span_is_valid(event_data->command_name, 1, 0) ? event_data->command_name
                                                                 : AZ_SPAN_EMPTY,
-              this_policy->_internal.rpc_client_codec->_internal.request_topic_buffer),
+              (char*)az_span_ptr(this_policy->_internal.request_topic_buffer),
+              (size_t)az_span_size(this_policy->_internal.request_topic_buffer),
```
vaavva, line 562: Does this cause any problems with the null terminator for mosquitto/paho?

vaavva: (since there's no slice)

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
+          az_span extracted_client_id
+              = az_span_slice(received_topic, i, i + az_span_size(client->_internal.client_id));
+          if (az_span_is_content_equal(extracted_client_id, client->_internal.client_id))
+          {
+            i += az_span_size(client->_internal.client_id) - 1;
+            format_idx += az_span_size(az_mqtt5_rpc_client_id_key) - 1;
+          }
+          else
+          {
+            return AZ_ERROR_IOT_TOPIC_NO_MATCH;
```
vaavva, line 343: Should this be an IoT specific error? Would this work right if the IoT portion of the sdk wasn't compiled?

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -1,83 +1,420 @@
...
 // SPDX-License-Identifier: MIT
 
-#include <azure/core/az_mqtt5_rpc_client.h>
+#include <azure/core/az_mqtt5_rpc.h>
+#include <azure/core/az_mqtt5_rpc_client_codec.h>
 #include <azure/core/az_result.h>
 #include <azure/core/internal/az_log_internal.h>
+#include <azure/core/internal/az_precondition_internal.h>
+#include <azure/core/internal/az_result_internal.h>
+#include <azure/iot/az_iot_common.h>
```
vaavva, line 10: I don't think this should be included here

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
RLeclair: Thanks for the suggestions! I will improve the parser in a future PR

sdk/src/azure/core/az_mqtt5_rpc.c
RLeclair: Good point, changed it to calculate before if statements

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
RLeclair: Will implement in a follow-up PR!

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
RLeclair: It returns AZ_OK, added the explicit check for AZ_OK

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
RLeclair: It is currently not causing any problems, but that's a good catch, I'll look into it!

sdk/src/azure/core/az_mqtt5_rpc_server_codec.c
```diff
@@ -1,60 +1,258 @@
...
+  az_span topic_format = server->_internal.options.subscription_topic_format;
+  az_span mqtt_topic_span = az_span_create((uint8_t*)mqtt_topic, (int32_t)mqtt_topic_size);
+  int32_t model_id_index = az_span_find(topic_format, az_mqtt5_rpc_service_id_key);
+  int32_t executor_id_index = az_span_find(topic_format, az_mqtt5_rpc_executor_id_key);
+  int32_t command_id_index = az_span_find(topic_format, az_mqtt5_rpc_command_id_key);
+
+  int32_t required_length = az_span_size(server->_internal.options.subscription_topic_format)
+      + (int32_t)sizeof((uint8_t)'\0');
+
+  required_length
```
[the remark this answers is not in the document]
RLeclair: Will address in a follow-up PR once the logic is re-worked

sdk/src/azure/core/az_mqtt5_rpc_server_codec.c
```diff
@@ -1,60 +1,258 @@
...
+
+  return AZ_OK;
 }
 
 AZ_NODISCARD az_result az_mqtt5_rpc_server_codec_init(
-    az_mqtt5_rpc_server_codec* client,
+    az_mqtt5_rpc_server_codec* server,
     az_span model_id,
     az_span client_id,
-    az_span command_name,
```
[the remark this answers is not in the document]
RLeclair: That case would be handled by the HFSM implementation

sdk/tests/core/test_az_mqtt5_rpc_client_codec.c
```diff
@@ -70,166 +76,400 @@ static void test_az_mqtt5_rpc_client_codec_init_no_options_success(void** state)
...
+          &test_rpc_client_codec,
+          AZ_SPAN_FROM_STR(TEST_CLIENT_ID),
+          AZ_SPAN_FROM_STR(TEST_MODEL_ID),
+          AZ_SPAN_FROM_STR(TEST_COMMAND_NAME),
+          NULL),
       AZ_OK);
 
-  assert_true(az_span_is_content_equal(
-      az_span_slice(sub_topic, 0, topic_length), AZ_SPAN_FROM_STR(TEST_SUBSCRIPTION_TOPIC)));
+  char test_response_property_topic_buffer[96]; // Exact size
```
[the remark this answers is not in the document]
RLeclair: Will implement this in the follow-up PR!

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
RLeclair: I believe we'll be using this error for now, I asked @CIPop about creating a `AZ_ERROR_MQTT_TOPIC_NO_MATCH` but he suggested to keep using the IoT specific error

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
RLeclair: We need it for the `AZ_ERROR_IOT_TOPIC_NO_MATCH` error

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
RLeclair: Will implement in a follow-up PR

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -98,37 +85,12 @@ AZ_NODISCARD bool _az_span_topic_matches_filter(az_span topic_filter, az_span to
...
  */
-AZ_NODISCARD az_result az_rpc_get_topic_from_format(
-    az_span format,
-    az_span model_id,
-    az_span executor_client_id,
-    az_span invoker_client_id,
-    az_span command_name,
-    az_span out_topic,
-    int32_t* out_topic_length);
+AZ_NODISCARD bool _az_mqtt5_rpc_valid_topic_format(az_span topic_format);
```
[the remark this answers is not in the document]
RLeclair: Both server and client codecs use this function, unless we add a `az_mqtt5_rpc_codec.h` we would have to add this function to both `az_mqtt5_rpc_client_codec.h` and `az_mqtt_rpc_server_codec.h`

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
CIPop: If we chose to keep things in the MQTT namespace it's likely we'll need another constant but I think most of these functions should be in a separate namespace: either IoT or something else.

sdk/src/azure/core/az_mqtt5_rpc_server_codec.c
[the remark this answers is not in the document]
vaavva: That wouldn't enable having the command_name in the subscription topic though, would it?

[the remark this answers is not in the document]
CIPop: If that's a valid scenario, command_name should be set in codec_options (defaulting to `AZ_SPAN_EMPTY`).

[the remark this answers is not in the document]
vaavva: Yes, this was stored in the codec before
```typedef struct az_mqtt5_rpc_server_codec
{
  struct
  {
    /**
     * @brief The model id to use for the subscription topic. May be AZ_SPAN_EMPTY if not used in
     * the subscription topic.
     */
    az_span model_id;
    /**
     * @brief The server client id to use for the subscription topic. May be AZ_SPAN_EMPTY if not
     * used in the subscription topic.
     */
    az_span client_id;
    /**
     * @brief The command name to use for the subscription topic. May be AZ_SPAN_EMPTY to have this
     * rpc server handle all commands for this topic.
     */
    az_span command_name;
    /**
     * @brief The topic to subscribe to for commands.
     */
    az_span subscription_topic;

    /**
     * @brief Options for the MQTT5 RPC Server Codec.
     */
    az_mqtt5_rpc_server_codec_options options;
  } _internal;
} az_mqtt5_rpc_server_codec;```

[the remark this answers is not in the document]
Comment (CIPop) [reply sdk/src/azure/core/az_mqtt5_rpc_server_codec.c:37]: @vaavva and I discussed this offline: @RLeclair we could implement the case where the RPC server really wants to accept only one command by allowing topic patterns that do not contain the `{command_name}` token.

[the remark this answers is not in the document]
Comment (RLeclair) [reply sdk/src/azure/core/az_mqtt5_rpc_server_codec.c:37]: Makes sense, then I will keep `{command_name}` out of the server codec options and adding documentation for how we can have a server accept only one command (via topic patterns not containing `{command_name}`

File: sdk/inc/azure/core/az_mqtt5_rpc.h [hunk 29]
```diff
@@ -20,6 +20,41 @@
 
 #include <azure/core/_az_cfg_prefix.h>
 
+/**
+ * @brief The default topic format for making RPC requests.
+ */
+#define AZ_MQTT5_RPC_SERVER_DEFAULT_SUBSCRIPTION_TOPIC_FORMAT \
```
vaavva, line 26: I this is used for the client and the server, should the naming be more generic? (same for others)

sdk/inc/azure/core/az_mqtt5_rpc_server_codec.h
```diff
@@ -27,21 +27,21 @@
...
 {
   /**
-   * @brief Timeout in seconds for subscribing acknowledgement.
-   */
-  int32_t subscribe_timeout_in_seconds;
-
-  /**
-   * @brief An #az_span containing the service group id to use for the subscription topic, can be
-   * AZ_SPAN_EMPTY if not using a shared subscription topic.
+   * @brief The service group id to use for the subscription topic, specifying this allows the
```
vaavva, line 30: Can you keep the part of the comment where this can be AZ_SPAN_EMPTY if you don't want to specify it?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -32,12 +32,6 @@ static const az_span server_client_id = AZ_SPAN_LITERAL_FROM_STR("controller-id"
 static const az_span content_type = AZ_SPAN_LITERAL_FROM_STR("application/json");
 static const az_span start_module_command_name = AZ_SPAN_LITERAL_FROM_STR("startModule");
 static const az_span stop_module_command_name = AZ_SPAN_LITERAL_FROM_STR("stopModule");
-static const az_span server_subscription_topic_format
-    = AZ_SPAN_LITERAL_FROM_STR("device/{executorId}/command/{name}");
-static const az_span client_subscription_topic_format
```
vaavva, line 37: Should we keep an example somewhere of how to specify custom topic formats?

sdk/samples/core/mosquitto_rpc_server_sample.c
```diff
@@ -367,16 +366,17 @@ int main(int argc, char* argv[])
 
   az_mqtt5_rpc_server_codec_options server_codec_options
       = az_mqtt5_rpc_server_codec_options_default();
+  server_codec_options.service_group_id = service_group_id;
```
vaavva, line 369: Does this need to be specified if it's AZ_SPAN_EMPTY?

sdk/samples/core/mosquitto_multiple_rpc_sample.c
RLeclair: From the above comments it seems we want to use default topics for all our samples, it would be good to add this in our readme though

sdk/samples/core/mosquitto_rpc_server_sample.c
RLeclair: Good point, removed it!

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -58,10 +60,15 @@ AZ_NODISCARD az_result az_mqtt5_rpc_client_codec_get_publish_topic(
       : az_span_size(client->_internal.model_id) - az_span_size(az_mqtt5_rpc_service_id_key);
   if (command_id_index != -1)
   {
-    required_length += (az_span_is_content_equal(command_name, AZ_SPAN_EMPTY))
-        ? az_span_size(client->_internal.command_name)
-        : az_span_size(command_name);
-    required_length -= az_span_size(az_mqtt5_rpc_command_id_key);
+    if (az_span_is_content_equal(command_name, AZ_SPAN_EMPTY))
```
vaavva, line 63:
```suggestion
    if (_az_span_is_valid(command_name, 1, false))
```
nit: just to catch if the size is zero but the pointer isn't NULL

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -13,21 +13,23 @@
 #include <azure/core/_az_cfg.h>
 
 static const az_span az_mqtt5_rpc_client_codec_default_publish_topic_format
-    = AZ_SPAN_LITERAL_FROM_STR("services/{serviceId}/{executorId}/command/{name}/request");
+    = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_SERVER_DEFAULT_SUBSCRIPTION_TOPIC_FORMAT);
```
vaavva, line 16: If some of these are only used once, why not define the az_span there (ex. line 34 for this one)

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
vaavva: nit: az_result_succeeded

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -323,10 +321,10 @@ static az_result waiting(az_event_policy* me, az_event event)
...
     {
       az_mqtt5_recv_data* recv_data = (az_mqtt5_recv_data*)event.data;
-      // Ensure pub is of the right topic
-      if (_az_span_topic_matches_filter(
-              this_policy->_internal.rpc_server_codec->_internal.subscription_topic,
-              recv_data->topic))
+      az_mqtt5_rpc_server_codec_request req_event_data;
+
+      if (az_mqtt5_rpc_server_codec_parse_received_topic(
+              this_policy->_internal.rpc_server_codec, recv_data->topic, &req_event_data))
```
vaavva, line 327:
```suggestion
      if (az_result_succeeded(az_mqtt5_rpc_server_codec_parse_received_topic(
              this_policy->_internal.rpc_server_codec, recv_data->topic, &req_event_data)))
```

sdk/tests/core/test_az_mqtt5_rpc_client_codec.c
```diff
@@ -45,16 +45,6 @@
   "vehicles/" TEST_MODEL_ID "/commands/" TEST_SERVER_ID "/" TEST_COMMAND_NAME "/" TEST_CLIENT_ID \
   "\0"
 
-static void test_az_mqtt5_rpc_client_codec_options_default_success(void** state)
-{
-  (void)state;
-
-  az_mqtt5_rpc_client_codec_options options = az_mqtt5_rpc_client_codec_options_default();
-
-  assert_int_equal(options.subscribe_timeout_in_seconds, AZ_MQTT5_RPC_DEFAULT_TIMEOUT_SECONDS);
```
vaavva, line 43: Should this test be kept for the other default values? (subscription_topic_format and request_topic_format)

sdk/tests/core/test_az_mqtt5_rpc_server_codec.c
```diff
@@ -40,15 +40,6 @@
 
 static az_mqtt5_rpc_server_codec test_rpc_server_codec; // TODO_L: Move this to each function
 
-static void test_az_mqtt5_rpc_server_codec_options_default_success(void** state)
-{
-  (void)state;
-
-  az_mqtt5_rpc_server_codec_options options = az_mqtt5_rpc_server_codec_options_default();
```
vaavva, line 32: Same here

sdk/inc/azure/core/az_mqtt5_rpc_server.h
```diff
@@ -283,8 +190,8 @@ AZ_NODISCARD az_result az_mqtt5_rpc_server_register(az_mqtt5_rpc_server* client)
  * used in the subscription topic.
  * @param[in] client_id The client id to use for the subscription topic. May be AZ_SPAN_EMPTY if not
  * used in the subscription topic.
- * @param[in] command_name The command name to use for the subscription topic or AZ_SPAN_EMPTY to
- * have this rpc server handle all commands for this topic.
+ * @param[in] subscribe_timeout_in_seconds Timeout in seconds for subscribing acknowledgemen (must
```
vaavva, line 193:
```suggestion
 * @param[in] subscribe_timeout_in_seconds Timeout in seconds for subscribing acknowledgement (must
```

sdk/tests/core/test_az_mqtt5_rpc_client_codec.c
RLeclair: The following test (`...init_no_options_succeed`) already tests the default values

sdk/tests/core/test_az_mqtt5_rpc_server_codec.c
RLeclair: See my previous comment

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -526,38 +526,44 @@ static az_result ready(az_event_policy* me, az_event event)
...
           &this_policy->_internal.property_bag);
 
+      size_t publish_topic_length = 0;
+
       _RETURN_AND_CLEAR_PROPERTY_BAG_IF_FAILED(
-          az_mqtt5_rpc_client_codec_get_request_topic(
+          az_mqtt5_rpc_client_codec_get_publish_topic(
               this_policy->_internal.rpc_client_codec,
               event_data->rpc_server_client_id,
               _az_span_is_valid(event_data->command_name, 1, 0) ? event_data->command_name
```
vaavva, line 557: and here

vaavva (approved):

PR #2674, opened 2023-11-03
Improvements to RPC parser and get topic functions

RLeclair:
Addressing remaining comments from: https://github.com/Azure/azure-sdk-for-c/pull/2671

### DIFF
--- a/sdk/inc/azure/core/az_mqtt5_rpc.h
+++ b/sdk/inc/azure/core/az_mqtt5_rpc.h
@@ -135,6 +135,36 @@ AZ_NODISCARD az_result _az_mqtt5_rpc_replace_tokens_in_format(
     uint32_t* required_length);
 
 /**
+ * @brief Helper function to extract tokens from a topic.
+ *
+ * @param topic_format The topic format to reference when extracting tokens.
+ * @param received_topic The topic to extract tokens from.
+ * @param client_id #az_span containing the client id or #AZ_SPAN_EMPTY.
+ * @param service_id #az_span containing the service id or #AZ_SPAN_EMPTY.
+ * @param executor_id #az_span containing the executor id or #AZ_SPAN_EMPTY.
+ * @param extracted_client_id Pointer to an #az_span to write the extracted client id to or NULL if
+ * not needed.
+ * @param extracted_service_id Pointer to an #az_span to write the extracted service id to or NULL
+ * if not needed.
+ * @param extracted_executor_id Pointer to an #az_span to write the extracted executor id to or NULL
+ * if not needed.
+ * @param extracted_command_name Pointer to an #az_span to write the extracted command name to or
+ * NULL if not needed.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD az_result _az_mqtt5_rpc_extract_tokens_from_topic(
+    az_span topic_format,
+    az_span received_topic,
+    az_span client_id,
+    az_span service_id,
+    az_span executor_id,
+    az_span* extracted_client_id,
+    az_span* extracted_service_id,
+    az_span* extracted_executor_id,
+    az_span* extracted_command_name);
+
+/**
  * @brief Helper function to check if an #az_mqtt5_rpc_status indicates failure.
  *
  * @param[in] status the status to check.
--- a/sdk/inc/azure/core/az_mqtt5_rpc.h
+++ b/sdk/inc/azure/core/az_mqtt5_rpc.h
@@ -21,16 +21,6 @@
 #include <azure/core/_az_cfg_prefix.h>
 
 /**
- * @brief The default topic format for making RPC requests.
- */
-#define AZ_MQTT5_RPC_DEFAULT_REQUEST_TOPIC_FORMAT \
-  "services/{serviceId}/{executorId}/command/{name}/request"
-/**
- * @brief The default topic format where RPC responses are published.
- */
-#define AZ_MQTT5_RPC_DEFAULT_RESPONSE_TOPIC_FORMAT \
-  "clients/{invokerClientId}/services/{serviceId}/{executorId}/command/{name}/response"
-/**
  * @brief Key appended to the topic format to indicate a shared subscription.
  */
 #define AZ_MQTT5_RPC_SERVICE_GROUP_ID_KEY "$share/"
@@ -54,6 +44,18 @@
  * @brief Key used to indicate the invoker client id in the topic format.
  */
 #define AZ_MQTT5_RPC_CLIENT_ID_KEY "{invokerClientId}"
+/**
+ * @brief The default topic format for making RPC requests.
+ */
+#define AZ_MQTT5_RPC_DEFAULT_REQUEST_TOPIC_FORMAT                          \
+  "services/" AZ_MQTT5_RPC_SERVICE_ID_KEY "/" AZ_MQTT5_RPC_EXECUTOR_ID_KEY \
+  "/command/" AZ_MQTT5_RPC_COMMAND_ID_KEY "/request"
+/**
+ * @brief The default topic format where RPC responses are published.
+ */
+#define AZ_MQTT5_RPC_DEFAULT_RESPONSE_TOPIC_FORMAT                               \
+  "clients/" AZ_MQTT5_RPC_CLIENT_ID_KEY "/services/" AZ_MQTT5_RPC_SERVICE_ID_KEY \
+  "/" AZ_MQTT5_RPC_EXECUTOR_ID_KEY "/command/" AZ_MQTT5_RPC_COMMAND_ID_KEY "/response"
 
 /**
  * @brief The default timeout in seconds for subscribing/publishing.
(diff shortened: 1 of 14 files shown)


sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -21,6 +23,249 @@ static const az_span command_name_key = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_CO
 static const az_span executor_client_id_key
     = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_EXECUTOR_ID_KEY);
 static const az_span invoker_client_id_key = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_CLIENT_ID_KEY);
+static const az_span token_format_end_character = AZ_SPAN_LITERAL_FROM_STR("}");
+static const az_span token_topic_end_character = AZ_SPAN_LITERAL_FROM_STR("/");
+
+static const uint32_t az_mqtt5_rpc_client_id_hash = 3426466449;
```
vaavva, line 29: Can these be calculated in case the keys change again?

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -110,6 +110,60 @@ typedef enum
   AZ_MQTT5_RPC_STATUS_TIMEOUT = 504,
 } az_mqtt5_rpc_status;
 
+/**
+ * @brief Helper function to replace tokens in a topic format.
+ *
+ * @param mqtt_topic_span Buffer to write the result to.
+ * @param topic_format The topic format to replace tokens in.
+ * @param required_length The required length of the buffer to write the result to.
```
vaavva, line 118: nit: order

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -110,6 +110,60 @@ typedef enum
   AZ_MQTT5_RPC_STATUS_TIMEOUT = 504,
 } az_mqtt5_rpc_status;
 
+/**
+ * @brief Helper function to replace tokens in a topic format.
+ *
+ * @param mqtt_topic_span Buffer to write the result to.
```
vaavva, line 116: Can you specify in/out parameters in this file?

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -21,6 +23,249 @@ static const az_span command_name_key = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_CO
...
+{
+  int32_t pos = az_span_find(format, token_format_end_character);
+  if (pos == -1)
+  {
+    return AZ_SPAN_EMPTY;
+  }
+  return az_span_slice(format, 0, pos);
+}
+
+AZ_INLINE az_span _az_mqtt5_rpc_extract_next_token_from_topic(az_span format)
```
vaavva, line 54:
```suggestion
AZ_INLINE az_span _az_mqtt5_rpc_extract_next_token_from_topic(az_span topic)
```
nit

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -21,6 +23,249 @@ static const az_span command_name_key = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_CO
...
+    _az_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, (int32_t)*required_length);
+    remainder = az_span_copy(remainder, service_group_id_key);
+    remainder = az_span_copy(remainder, service_group_id);
+    remainder = az_span_copy_u8(remainder, '/');
+  }
+
+  for (int32_t i = 0; i < az_span_size(topic_format); i++)
+  {
+    uint8_t c = az_span_ptr(topic_format)[i];
+    _az_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, (int32_t)(++*required_length));
```
vaavva, line 127: (Adding this just for discussion) I know it isn't possible with this solution, but I think it would be better to calculate the required length before copying into the out memory instead of the possibility of partially writing the topic and then failing and returning.

sdk/src/azure/core/az_mqtt5_rpc_client_codec.c
```diff
@@ -12,14 +12,6 @@
 
 #include <azure/core/_az_cfg.h>
```
vaavva, line 13: can this include be removed now? `#include <azure/iot/az_iot_common.h>`

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -110,6 +110,60 @@ typedef enum
...
+ * @param required_length The required length of the buffer to write the result to.
+ * @param service_group_id #az_span containing the service group id or #AZ_SPAN_EMPTY.
+ * @param client_id #az_span containing the client id or #AZ_SPAN_EMPTY.
+ * @param service_id #az_span containing the service id.
+ * @param executor_id #az_span containing the executor id or #AZ_SPAN_EMPTY.
+ * @param command_id #az_span containing the command id.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ */
+AZ_NODISCARD az_result _az_mqtt5_rpc_replace_tokens_in_format(
```
vaavva, line 127: any new unit tests needed for these?

sdk/src/azure/core/az_mqtt5_rpc.c
RLeclair: Yes, you'd need to use `_az_mqtt5_rpc_calculate_hash` to re-calculate them

sdk/src/azure/core/az_mqtt5_rpc.c
RLeclair: Adding @CIPop to the discussion, if we were to calculate the required length prior to doing any copying we'd have to do two passes at which point we wouldn't need the hashing function. On the first pass I can identify where all the tokens are and on the second replace them.

sdk/inc/azure/core/az_mqtt5_rpc.h
RLeclair: These functions are covered by the codec tests for both the RPC client and server

sdk/src/azure/core/az_mqtt5_rpc.c
CIPop: Please add clear comments on what exactly each of these mean as well as how to calculate them.

CIPop: Also, please add simple UTs to ensure the hashes correspond to the expected strings.

sdk/src/azure/core/az_mqtt5_rpc.c
CIPop: As long as we guarantee safety, it's OK to attempt to fill part of the buffer.

sdk/inc/azure/core/az_mqtt5_rpc.h
```diff
@@ -110,6 +110,60 @@ typedef enum
   AZ_MQTT5_RPC_STATUS_TIMEOUT = 504,
 } az_mqtt5_rpc_status;
 
+/**
```
CIPop, line 113: Please keep internal functions in internal headers. (e.g. follow the pattern from `az_config_internal.h`).

sdk/src/azure/core/az_mqtt5_rpc.c
```diff
@@ -21,6 +23,249 @@ static const az_span command_name_key = AZ_SPAN_LITERAL_FROM_STR(AZ_MQTT5_RPC_CO
...
+        return AZ_ERROR_ARG;
+      }
+    }
+    else
+    {
+      remainder = az_span_copy_u8(remainder, c);
+    }
+  }
+
+  az_span_copy_u8(remainder, '\0');
```
CIPop, line 177: I'm not sure the `required_length` is accurate. I think it may be missing the remainder and the NULL terminator.

The usage pattern for the codec is that:
- If the buffer is large enough, then we'll fill it in and return AZ_OK. `required_length` will contain how much we've filled in.
- If the buffer is too small, then we
  1. Guarantee that we do not overrun the buffer
  2. Guarantee that we correctly calculate the `required_length`

The usage pattern for an _allocating_ application is to start with the existing buffer: if the call fails then the application can reallocate a buffer of `required_length` and try again.

sdk/src/azure/core/az_mqtt5_rpc_server_codec.c
```diff
@@ -39,38 +33,24 @@ AZ_NODISCARD az_result az_mqtt5_rpc_server_codec_get_subscribe_topic(
...
-                ? az_span_size(az_mqtt5_rpc_any_executor_id)
-                : az_span_size(server->_internal.client_id));
-    required_length -= az_span_size(az_mqtt5_rpc_executor_id_key);
-  }
-  required_length += (command_id_index == -1)
-      ? 0
-      : (int32_t)sizeof((uint8_t)'+') - az_span_size(az_mqtt5_rpc_command_id_key);
-
-  _az_RETURN_IF_NOT_ENOUGH_SIZE(mqtt_topic_span, required_length);
+  az_span single_level_wildcard = AZ_SPAN_FROM_STR("+");
```
CIPop, line 39: Nit: move this as a top-of-file `static`?

sdk/src/azure/core/az_mqtt5_rpc_server_codec.c
```diff
@@ -14,12 +14,6 @@
 
 static const az_span az_mqtt5_rpc_service_group_id_key
```
CIPop, line 15: Nit: I think all of these need to start with `_` ?

CIPop (changes requested):
There are a few items that need addressing:
1. Please do not mix public APIs and internal functions in header files.
2. The implementation does not appear to correctly calculate the required buffer size if the original buffer is too small.
3. We really need UTs for the hashes.

sdk/src/azure/core/az_mqtt5_rpc.c
RLeclair: I have added some tests to ensure that `required_length` is accurate, I add the null terminator size at the start of the function.

sdk/tests/core/test_az_mqtt5_rpc.c
```diff
@@ -0,0 +1,137 @@
...
+
+#include <cmocka.h>
+
+#define TEST_SAMPLE_FORMAT                                           \
+  "test/" AZ_MQTT5_RPC_CLIENT_ID_KEY "/" AZ_MQTT5_RPC_SERVICE_ID_KEY \
+  "/" AZ_MQTT5_RPC_EXECUTOR_ID_KEY "/" AZ_MQTT5_RPC_COMMAND_ID_KEY "/endoftest"
+#define TEST_SAMPLE_FORMAT_REPLACED                                                    \
+  "$share/test_service_group_id/test/test_client_id/test_service_id/test_executor_id/" \
+  "test_command_id/endoftest\0"
+static const az_span test_sample_format = AZ_SPAN_LITERAL_FROM_STR(TEST_SAMPLE_FORMAT);
```
CIPop, line 22: These should be defining spans over the #defines in the include file.

sdk/tests/core/test_az_mqtt5_rpc.c
```diff
@@ -0,0 +1,137 @@
...
+static const az_span test_service_group_id = AZ_SPAN_LITERAL_FROM_STR("test_service_group_id");
+static const az_span test_sample_client_id = AZ_SPAN_LITERAL_FROM_STR("test_client_id");
+static const az_span test_sample_service_id = AZ_SPAN_LITERAL_FROM_STR("test_service_id");
+static const az_span test_sample_executor_id = AZ_SPAN_LITERAL_FROM_STR("test_executor_id");
+static const az_span test_sample_command_id = AZ_SPAN_LITERAL_FROM_STR("test_command_id");
+
+// Calculated hashes shown below. UPDATE whenever tokens change.
+static const uint32_t az_mqtt5_rpc_client_id_hash = 3426466449;
+static const uint32_t az_mqtt5_rpc_service_id_hash = 4175641829;
```
CIPop, line 31: These should be removed - we must define the hashes in a single place and verify that the prod code produces the same hashes.

CIPop (approved):
Thanks @RLeclair

sdk/tests/core/test_az_mqtt5_topic_parser.c
```diff
@@ -0,0 +1,138 @@
...
+static const az_span test_sample_client_id = AZ_SPAN_LITERAL_FROM_STR("test_client_id");
+static const az_span test_sample_service_id = AZ_SPAN_LITERAL_FROM_STR("test_service_id");
+static const az_span test_sample_executor_id = AZ_SPAN_LITERAL_FROM_STR("test_executor_id");
+static const az_span test_sample_command_id = AZ_SPAN_LITERAL_FROM_STR("test_command_id");
+
+static void test_az_mqtt5_topic_parser_calculate_hash_success(void** state)
+{
+  (void)state;
+
+  az_span invoker_client_id_key = AZ_SPAN_LITERAL_FROM_STR(_az_MQTT5_TOPIC_PARSER_CLIENT_ID_KEY);
```
CIPop, line 35: Nit: extract method to get a span over C-string minus the null.

CIPop (approved):

PR #2685, opened 2023-11-13
Expose command_name and service_id, do not fail if command respose has empty body

ewertons:


sdk/inc/azure/core/az_mqtt5_rpc_server.h
```diff
@@ -164,6 +164,14 @@ typedef struct
...
    */
   az_span content_type;
+  /**
+   * @brief The service id of the request.
+   */
+  az_span service_id;
+  /**
+   * @brief The command name of the request.
+   */
+  az_span command_name;
```
vaavva, line 174: @RLeclair was there a reason we didn't include the command name in this before?

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -244,6 +244,8 @@ AZ_INLINE az_result _handle_request(az_mqtt5_rpc_server* this_policy, az_mqtt5_r
             .request_data = data->payload,
             .request_topic = data->topic,
             .content_type = content_type_str,
+            .command_name = req_event_data->command_name,
```
vaavva, line 247: Does this need to be copied/will there be any issues with the span getting deallocated?

vaavva (approved):

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
ewertons: These data are only valid in the scope of the callback. The target of the callback must copy the information if they need it. After cb and _handle_request return this data is disregarded.

sdk/inc/azure/core/az_mqtt5_rpc_server.h
RLeclair: This should have been included here, the change is correct

RLeclair (approved):

ewertons:
Found the issue with test failures. Missing test was not updated to expect empty payloads in command responses. 

I modified and renamed test_az_mqtt5_rpc_client_recv_response_no_payload_failure to test_az_mqtt5_rpc_client_recv_response_no_payload_success

PR #2694, opened 2023-11-21
MQTT 5 Paho Improvements

RLeclair:
Added/fixed the following:
- If no payload is received on a publish, `AZ_SPAN_EMPTY` will be returned by paho MQTT layer.
- Paho MQTT layer's publish callback does not send a PUBACK message to the pipeline on qos0.
- Added extra unit tests to test the prior two.

Note that new UTs are disabled for Mosquitto. The publish callback for Mosquitto does not provide information on whether the message was sent with qos0 or qos1.

### DIFF
--- a/sdk/src/azure/platform/az_pahoasync5.c
+++ b/sdk/src/azure/platform/az_pahoasync5.c
@@ -136,12 +136,15 @@ static void _az_pahoasync5_on_publish_success(void* client, MQTTAsync_successDat
 {
   az_mqtt5* me = (az_mqtt5*)client;
 
-  az_result ret = az_mqtt5_inbound_puback(
-      me, &(az_mqtt5_puback_data){ .id = response->token, .reason_code = response->reasonCode });
-
-  if (az_result_failed(ret))
+  if (response->alt.pub.message.qos != AZ_MQTT5_QOS_AT_MOST_ONCE)
   {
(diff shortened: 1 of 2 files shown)


sdk/src/azure/platform/az_pahoasync5.c
```diff
@@ -136,12 +136,15 @@ static void _az_pahoasync5_on_publish_success(void* client, MQTTAsync_successDat
...
-  if (az_result_failed(ret))
+  if (response->alt.pub.message.qos != AZ_MQTT5_QOS_AT_MOST_ONCE)
   {
-    _az_pahoasync5_critical_error();
+    az_result ret = az_mqtt5_inbound_puback(
+        me, &(az_mqtt5_puback_data){ .id = response->token, .reason_code = response->reasonCode });
+
+    if (az_result_failed(ret))
+    {
+      _az_pahoasync5_critical_error();
```
CIPop, line 146: While this is correct, in most cases it should never happen. There is a bug in the pipeline inbound message - for any az_result failed on passing the event into the pipeline, it should transform it and pass it to the application callback as an `AZ_HFSM_ERROR` event. 
See the `_az_event_pipeline_timer_callback` implementation.

CIPop (approved):
The comment is about a bug in the post_inbound implementation (not necessarily related to this PR).

PR #2696, opened 2023-11-22
Make content type optional for RPC requests & responses

ewertons:
For reference:

This also follows this discussion: https://github.com/microsoft/mqtt-patterns/issues/27


sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -505,19 +497,16 @@ static az_result ready(az_event_policy* me, az_event event)
               event_data->correlation_id),
           &this_policy->_internal.property_bag);
 
-      if (!_az_span_is_valid(event_data->content_type, 1, false))
+      if (!az_span_is_content_equal(event_data->content_type, AZ_SPAN_EMPTY))
```
CIPop, line 500: Also, please add the fact that it can be AZ_SPAN_EMPTY within the `az_mqtt5_rpc_client_invoke_req_event_data` documentation in `az_mqtt5_rpc_client.h`.

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -425,14 +425,6 @@ send_resp_inbound_if_topic_matches(az_mqtt5_rpc_client* this_policy, az_event ev
       // read the error message if there is one
       resp_data.error_message = error_message_val_str;
     }
-    else if (
-        !az_span_is_content_equal(recv_data->payload, AZ_SPAN_EMPTY)
```
vaavva, line 429: In this scenario, content type will be AZ_SPAN_EMPTY, not random bytes, correct?

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
CIPop: Nit: `az_mqtt5_rpc_client_invoke_begin`, as a public API, should take all these arguments as parameters instead of the struct to be in-line with existing IoT functions and the codec. /cc @vaavva

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -236,8 +239,7 @@ AZ_INLINE az_result _handle_request(
   az_span correlation_data_bindata = az_mqtt5_property_get_binarydata(&correlation_data);
   az_span content_type_str = az_mqtt5_property_get_string(&content_type);
 
-  if (az_span_ptr(response_topic_str) != NULL && az_span_ptr(correlation_data_bindata) != NULL
-      && az_span_ptr(content_type_str) != NULL)
+  if (az_span_ptr(response_topic_str) != NULL && az_span_ptr(correlation_data_bindata) != NULL)
```
CIPop, line 242: @vaavva I think we should reconsider all these checks and allow the RPC HFSMs to make progress as much as they can based on the information available.
We could log warnings in the case where the protocol is broken but the information can still be relayed to the application as incomplete (`AZ_SPAN_EMPTY` for missing fields).

CIPop (approved):
Thanks @ewertons. Left a comment to add documentation. 
The other comments are for future enhancements.

sdk/tests/core/test_az_mqtt5_rpc_server_hfsm.c
```diff
@@ -296,14 +296,60 @@ static void test_az_mqtt5_rpc_server_recv_request_specific_endpoint_success(void
 #endif // TRANSPORT_PAHO
 }
 
+static void test_az_mqtt5_rpc_server_recv_request_no_cotent_type_success(void** state)
```
RLeclair, line 299: nit

```suggestion
static void test_az_mqtt5_rpc_server_recv_request_no_content_type_success(void** state)
```

RLeclair (approved):
LGTM

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
ewertons: We wouldn't know that. And we didn't check for that before either (e.g., testing if the CT is a known type).

sdk/tests/core/test_az_mqtt5_rpc_server_hfsm.c
ewertons: Taken.

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
ewertons: I'll add another PR for that.

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
ewertons: Done. Added:

  /**
   * @brief The content type of the request. It can be AZ_SPAN_EMPTY.
   */
  az_span content_type;

CIPop (approved):

PR #2697, opened 2023-11-27
Update vcpkg commit SHA to 2023-11-16 vcpkg-tool release

antkmsft:
Currently, the CI is failing, because msys, which is one of the dependencies for cmocka, can no longer be found at the URL that vcpkg port has.
The simplest way to try to get that fixed, is to update vcpkg, including the tool and all its libraries.
This PR updates the SHA to https://github.com/microsoft/vcpkg/tree/43cf47eccfbe27006cf9534a5db809798f8c37fe, which is the SHA that we are currently using in the C++ repo.

The one that has been used before, is more than a year old - it is from October 19, 2022.


antkmsft:
(check enforcer hang, so overriding it. The CI is all green)

CIPop (approved):
Thanks @antkmsft !

PR #2709, opened 2023-12-13
Telemetry HFSM

vaavva:
- Adds Telemetry producer and consumer HFSMs and Samples for Paho and Mosquitto


sdk/inc/azure/core/az_mqtt5_telemetry_consumer.h
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * @brief The MQTT5 Telemetry Consumer.
+ *
+ */
+typedef struct az_mqtt5_telemetry_consumer
+{
+  struct
+  {
+    /**
+     * @brief Telemetry Consumer HFSM for the MQTT5 Telemetry Consumer.
```
ericwol-msft, line 51:
```suggestion
     * @brief HFSM for the MQTT5 Telemetry Consumer.
```

sdk/inc/azure/core/az_mqtt5_telemetry_consumer.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    _az_event_client subclient;
+
+    /**
+     * @brief The MQTT5 connection linked to the MQTT5 Telemetry Consumer.
+     */
+    az_mqtt5_connection* connection;
+
+    /**
+     * @brief Timeout in seconds for subscribing acknowledgement.
```
ericwol-msft, line 67:
```suggestion
     * @brief Timeout, in seconds, for subscribing acknowledgement.
```

sdk/inc/azure/core/az_mqtt5_telemetry_consumer.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    int32_t subscribe_timeout_in_seconds;
+
+    /**
+     * @brief the message id of the pending subscribe for the telemetry topic.
```
ericwol-msft, line 72:
```suggestion
     * @brief The message id of the pending subscribe for the telemetry topic.
```

sdk/inc/azure/core/az_mqtt5_telemetry_consumer.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    int32_t pending_subscription_id;
+
+    /**
+     * @brief timer used for the subscribe of the telemetry topic.
```
ericwol-msft, line 77:
```suggestion
     * @brief Timer used for the subscribe of the telemetry topic.
```

sdk/inc/azure/core/az_mqtt5_telemetry_consumer.h
```diff
@@ -0,0 +1,177 @@
...
+ * @retval Other on other failures creating/sending the subscribe message.
+ */
+AZ_NODISCARD az_result
+az_mqtt5_telemetry_consumer_subscribe_begin(az_mqtt5_telemetry_consumer* client);
+
+/**
+ * @brief Triggers an #AZ_MQTT5_EVENT_UNSUB_REQ event from the application.
+ *
+ * @note This should be called from the application to unsubscribe from the telemetry topic. This
+ * may be used if the application doesn't want to receive telemetry anymore.
```
ericwol-msft, line 133:
```suggestion
 * may be used if the application doesn't want to receive telemetry messages anymore.
```

sdk/inc/azure/core/az_mqtt5_telemetry_producer.h
```diff
@@ -0,0 +1,220 @@
...
+     */
+    _az_event_client subclient;
+
+    /**
+     * @brief The MQTT5 connection linked to the MQTT5 Telemetry Producer.
+     */
+    az_mqtt5_connection* connection;
+
+    /**
+     * @brief Timeout in seconds for publishing (must be > 0).
```
ericwol-msft, line 88:
```suggestion
     * @brief Timeout, in seconds, for publishing (must be > 0).
```

sdk/inc/azure/core/az_mqtt5_telemetry_producer.h
```diff
@@ -0,0 +1,220 @@
...
+     */
+    int32_t publish_timeout_in_seconds;
+
+    /**
+     * @brief The message id of the pending publish.
```
ericwol-msft, line 93:
```suggestion
     * @brief The MQTT message id of the pending publish.
```

sdk/inc/azure/core/az_mqtt5_telemetry_producer.h
```diff
@@ -0,0 +1,220 @@
...
+   */
+  az_span telemetry_payload;
+
+  /**
+   * @brief The telemetry name.
+   */
+  az_span telemetry_name;
+
+  /**
+   * @brief The qos for this telemetry message.
```
ericwol-msft, line 143:
```suggestion
   * @brief The QOS for this telemetry message.
```

sdk/samples/core/mosquitto_telemetry_producer_sample.c
```diff
@@ -0,0 +1,218 @@
...
+  (void)client;
+  (void)callback_context;
+  az_app_log_callback(event.type, AZ_SPAN_FROM_STR("APP/callback"));
+  switch (event.type)
+  {
+    case AZ_MQTT5_EVENT_CONNECT_RSP:
+    {
+      az_mqtt5_connack_data* connack_data = (az_mqtt5_connack_data*)event.data;
+      printf(LOG_APP "CONNACK: reason=%d\n", connack_data->connack_reason);
+      break;
```
ericwol-msft, line 68: add sample_finished = true; here too?

sdk/samples/core/paho_telemetry_producer_sample.c
```diff
@@ -0,0 +1,206 @@
...
+          = (az_mqtt5_telemetry_producer_error_rsp_event_data*)event.data;
+      printf(
+          LOG_APP_ERROR "Failure sending telemetry : %s Reason Code: %d\n",
+          az_span_ptr(recv_data->error_message),
+          recv_data->reason_code);
+      break;
+    }
+
+    case AZ_HFSM_EVENT_ERROR:
+      printf(LOG_APP_ERROR "Error Event\n");
```
ericwol-msft, line 90: add comments what this means

sdk/samples/core/mosquitto_telemetry_producer_sample.c
```diff
@@ -0,0 +1,218 @@
...
+          = (az_mqtt5_telemetry_producer_error_rsp_event_data*)event.data;
+      printf(
+          LOG_APP_ERROR "Failure sending telemetry : %s Reason Code: %d\n",
+          az_span_ptr(recv_data->error_message),
+          recv_data->reason_code);
+      break;
+    }
+
+    case AZ_HFSM_EVENT_ERROR:
+      printf(LOG_APP_ERROR "Error Event\n");
```
ericwol-msft, line 90: add comments what this means

sdk/samples/core/paho_telemetry_producer_sample.c
```diff
@@ -0,0 +1,206 @@
...
+  {
+    printf(LOG_APP "Waiting...\r");
+    fflush(stdout);
+
+    // sends a telemetry message every 15 seconds. This cadence/how it is triggered should be
+    // customized for your solution.
+    if (i % 15 == 0)
+    {
+      // TODO: Payload should be generated and serialized
+      LOG_AND_EXIT_IF_FAILED(telemetry_send_begin(
```
ericwol-msft, line 181: this always returns AZ_OK

ericwol-msft (approved):

sdk/samples/core/paho_telemetry_producer_sample.c
vaavva: yep - log and exit if failed should only be triggered here on an unrecoverable error

sdk/samples/core/paho_telemetry_producer_sample.c
vaavva: This is a generic error event sent from the HFSM, so it could mean many different things. There's a work item to ensure this event type has details, but until that is complete I don't want to add a comment on how to parse this yet.

sdk/samples/core/paho_telemetry_producer_sample.c
```diff
@@ -0,0 +1,206 @@
...
+          telemetry_name,
+          AZ_SPAN_FROM_STR("{\"TelemetryTimestamp\":1691530585198,\"FuelLevel\":\"50\"}"),
```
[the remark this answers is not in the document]
vaavva: Would require the json serializer to be hooked up to this sample - will track this as a future work item.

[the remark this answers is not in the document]
vaavva: Added

sdk/src/azure/core/az_mqtt5_telemetry_consumer_hfsm.c
```diff
@@ -0,0 +1,398 @@
...
+      }
+      break;
+    }
+
+    case AZ_MQTT5_EVENT_PUBACK_RSP:
+    case AZ_EVENT_MQTT5_CONNECTION_OPEN_REQ:
+    case AZ_MQTT5_EVENT_CONNECT_RSP:
+      break;
+
+    default:
```
RLeclair, line 261: In your puml diagram you also handle the AZ_MQTT_SESSION_EXPIRY event, is there a reason why it is not shown here?

sdk/src/azure/core/az_mqtt5_telemetry_consumer_hfsm.c
```diff
@@ -0,0 +1,398 @@
...
+      break;
+  }
+
+  return ret;
+}
+
+/**
+ * @brief Failure state - locks up all execution of this hfsm
+ */
+static az_result faulted(az_event_policy* me, az_event event)
```
RLeclair, line 272: To follow the convention in the `az_mqtt5_connection_hfsm`s faulted state it would be good to have this state only return `AZ_ERROR_HFSM_INVALID_STATE` and have the api callback on the transition back in line 217

sdk/src/azure/core/az_mqtt5_telemetry_producer_hfsm.c
```diff
@@ -0,0 +1,436 @@
...
+
+  return ret;
+}
+
+/**
+ * @brief Failure state - locks up all execution of this policy
+ */
+static az_result faulted(az_event_policy* me, az_event event)
+{
+  az_result ret = AZ_OK;
```
RLeclair, line 341: Same comment as on hfsm for consumer

sdk/tests/core/test_az_mqtt5_telemetry_producer_hfsm.c
```diff
@@ -0,0 +1,430 @@
...
+          .telemetry_name = AZ_SPAN_FROM_STR(TEST_TELEMETRY_NAME) };
+
+  assert_int_equal(
+      az_mqtt5_telemetry_producer_send_begin(&test_telemetry_producer, &test_telemetry_data),
+      AZ_ERROR_ARG);
+
+  assert_int_equal(ref_pub_req, 0);
+}
+
+static void test_az_mqtt5_telemetry_producer_send_begin_qos_1_timeout(void** state)
```
RLeclair, line 371: Should we have a test to see if we can send a qos 0 message while waiting for a qos 1 message to return?

RLeclair (commented):
LGTM!

sdk/docs/core/resources/telemetry_consumer.puml
```diff
@@ -0,0 +1,22 @@
+@startuml TELEMETRY_CONSUMER
+
+state TelemetryConsumer {
+
+  state Ready
+  Ready : <b>telemetry_consumer_subscribe_begin/</b> start sub timer, ^MQTT_SUB_REQ(...)
```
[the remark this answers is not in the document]
vaavva: I don't think so. The telemetry_consumer_subscribe_begin is going to be an api call from the application, which triggers the other events. The MQTT_SUB_REQ sent out doesn't get seen by the hfsm, so it can't act on that event. We could create another custom event just so the timer start and sending of the sub request happens within the HFSM instead of the api call, but this feels a bit unnecessarily layered to me - I'll think on it though.

sdk/src/azure/core/az_mqtt5_telemetry_consumer_hfsm.c
vaavva: Good call - I'll take a look at this once we have the connection management story solidified a bit more and note that this isn't implemented yet

sdk/src/azure/core/az_mqtt5_telemetry_consumer_hfsm.c
vaavva: Left a comment about this on your PR as well - the entry state should succeed to indicate that faulted was entered correctly and then any other events after that should fail. Let's sync on this once you test in the connection_hfsm

sdk/docs/core/resources/telemetry_consumer.puml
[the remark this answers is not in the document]
RLeclair: That makes sense, I was basing my comment on what you did for the RPC client's `az_mqtt5_rpc_client_subscribe_begin`

Another benefit of taking this approach is that if the HFSM goes to a faulted state and the user tries to send out a subscribe_begin we won't start a timer since the case would be handled by the default of the `faulted` state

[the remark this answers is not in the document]
vaavva: Good call - it would be good to not send the subscribe if we're in faulted - I didn't add this extra layer in telemetry originally since there's only one valid state and subscribe would always behave the same way in it. But in faulted it shouldn't, so I'll separate that out.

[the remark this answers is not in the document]
vaavva: updated in the new PR

PR #2738, opened 2024-02-06
Unit tests connection management

RLeclair:
- Adding unit tests for connection management
- Includes retries
- Added config files for samples


vaavva (commented):
Will take a look at the rest of the unit tests tomorrow!

sdk/inc/azure/core/az_mqtt5_connection_config.h
```diff
@@ -38,9 +39,10 @@
 
 #ifndef AZ_MQTT5_CONNECTION_MAX_CONNECT_ATTEMPTS
 /**
- * @brief The maximum number of retry attempts to connect to the MQTT 5 broker.
+ * @brief The maximum number of retry attempts to connect to the MQTT 5 broker. No maximum number of
+ * retry attempts if set to -1.
  */
-#define AZ_MQTT5_CONNECTION_MAX_CONNECT_ATTEMPTS (-1)
+#define AZ_MQTT5_CONNECTION_MAX_CONNECT_ATTEMPTS (2)
```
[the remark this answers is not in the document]
RLeclair: Good catch, was doing some testing with this value and missed changing it back

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
```diff
@@ -248,7 +248,18 @@ static az_result started(az_event_policy* me, az_event event)
...
-      _az_RETURN_IF_FAILED(_az_hfsm_transition_peer((_az_hfsm*)me, started, faulted));
+      if (_az_hfsm_transition_peer((_az_hfsm*)me, started, faulted) != AZ_OK)
+      {
+        az_platform_critical_error();
+      }
+
+      ret = az_event_policy_send_inbound_event(
+          (az_event_policy*)this_policy,
+          (az_event){ .type = AZ_HFSM_EVENT_ERROR,
+                      .data = &(az_hfsm_event_data_error){ .error_type = event.type,
```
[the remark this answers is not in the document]
RLeclair: The type of the event is `AZ_HFSM_EVENT_ERROR` but the internal error has the actual error, similar to what is done for _az_FAULTED_IF_FAILED

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -134,6 +136,10 @@ static az_result test_subclient_policy_1_root(az_event_policy* me, az_event even
     case AZ_MQTT5_EVENT_SUB_REQ:
       ref_sub_req++;
       break;
+    case AZ_HFSM_EVENT_TIMEOUT:
```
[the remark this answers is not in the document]
RLeclair: I added them as refs, good point

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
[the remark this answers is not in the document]
vaavva: but that information would be in .sender_event.type, right?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -241,36 +252,161 @@ AZ_INLINE void test_az_mqtt5_connection_setup(
...
+      az_mqtt5_inbound_connack(
+          mock_mqtt5,
+          &(az_mqtt5_connack_data){ .connack_reason = AZ_MQTT5_CONNACK_SUCCESS,
+                                    .tls_authentication_error = false }),
+      AZ_OK);
+  // One pass through the policy collection, another through the callback.
+  assert_int_equal(ref_conn_rsp_ok_prior + 2, ref_conn_rsp_ok);
+}
+
+AZ_INLINE void test_az_mqtt5_connection_connected_connecting(az_mqtt5* mock_mqtt5)
```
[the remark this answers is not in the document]
RLeclair: There is a connected_disconnecting below this inline function, this one goes back to connecting because it receives a disconnect packet so it will try to re-connect

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -241,36 +252,161 @@ AZ_INLINE void test_az_mqtt5_connection_setup(
...
+                    } });
+  }
+
+  if (az_result_failed(ret))
+  {
+    assert_true(false);
+  }
+}
+
+static void test_az_mqtt5_connection_options_init(void** state)
```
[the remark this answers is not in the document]
RLeclair: Moved to setup!

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -213,6 +223,7 @@ AZ_INLINE void test_az_mqtt5_connection_setup(
       az_mqtt5_connection_init(
           test_conn, NULL, mock_mqtt5, test_mqtt_connection_callback, test_conn_options, NULL),
       AZ_OK);
+  test_conn->_internal.connection_timer._internal.pipeline = &test_conn->_internal.event_pipeline;
```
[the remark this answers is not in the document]
RLeclair: This is to setup the timer callback, the initialization is currently being mocked so it has to be done manually as shown here

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -279,29 +415,30 @@ static void test_az_mqtt5_connection_disabled_init_success(void** state)
...
-  test_connection_options.password_buffer = AZ_SPAN_FROM_STR(TEST_PASSWORD);
-  test_connection_options.port = TEST_PORT;
+  test_connection_no_certs_options.disable_sdk_connection_management = false;
+  test_connection_no_certs_options.hostname = AZ_SPAN_FROM_STR(TEST_HOSTNAME);
+  test_connection_no_certs_options.client_id_buffer = AZ_SPAN_FROM_STR(TEST_CLIENT_ID);
+  test_connection_no_certs_options.username_buffer = AZ_SPAN_FROM_STR(TEST_USERNAME);
+  test_connection_no_certs_options.password_buffer = AZ_SPAN_FROM_STR(TEST_PASSWORD);
+  test_connection_no_certs_options.port = TEST_PORT;
 
   test_az_mqtt5_connection_setup(
```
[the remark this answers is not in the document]
RLeclair: None is needed, all of the variables are constrained to that test

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+  assert_int_equal(az_mqtt5_connection_close(&test_connection), AZ_OK);
+  _az_mqtt5_connection_disconnecting_idle(&mock_mqtt);
+
+  assert_int_equal(ref_disconn_req, 1);
+  assert_int_equal(ref_disconn_rsp, 2);
+  assert_int_equal(ref_conn_error, 0);
+}
+
+// Testing that when connecting no publish requests are allowed.
+static void test_az_mqtt5_connection_connecting_pub_request_success(void** state)
```
vaavva, line 685:
```suggestion
static void test_az_mqtt5_connection_connecting_pub_request_failure(void** state)
```

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+  assert_int_equal(
+      az_mqtt5_inbound_recv(
+          &mock_mqtt,
+          &(az_mqtt5_recv_data){ .topic = AZ_SPAN_FROM_STR(TEST_TOPIC),
+                                 .qos = AZ_MQTT5_QOS_AT_MOST_ONCE,
+                                 .payload = AZ_SPAN_FROM_STR(TEST_PAYLOAD),
+                                 .id = 1 }),
+      AZ_OK);
+
+  // TODO_L: Once event filtering is implemented, add more details.
```
vaavva, line 753: can you do the same checks here as in `test_az_mqtt5_connection_connected_pub_rcv_success`?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+      &mock_client_2,
+      &mock_client_hfsm_1,
+      &mock_client_hfsm_2);
+
+  _az_mqtt5_connection_idle_connecting(&test_connection);
+  _az_mqtt5_connection_connecting_connected(&mock_mqtt);
+
+  will_return(__wrap_az_platform_clock_msec, 0);
+  will_return(__wrap_az_platform_clock_msec, 0);
+  will_return(__wrap_az_mqtt5_outbound_connect, AZ_OK);
```
vaavva, line 863: Why does the disconnect trigger an outbound_connect?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+      az_mqtt5_inbound_connack(
+          &mock_mqtt,
+          &(az_mqtt5_connack_data){ .connack_reason = AZ_MQTT5_CONNACK_SUCCESS,
+                                    .tls_authentication_error = false }),
+      AZ_OK);
+
+  assert_int_equal(ref_conn_rsp_ok, 2);
+  assert_int_equal(ref_conn_rsp_err, 0);
+  assert_int_equal(ref_conn_error, 0);
+  assert_int_equal(ref_disconn_req, 1);
```
vaavva, line 1072: Should this be 2 if an additional disconnect request is sent?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+  _az_mqtt5_connection_connecting_reconnecttimeout(&mock_mqtt, AZ_MQTT5_CONNACK_UNSPECIFIED_ERROR);
+
+  az_mqtt5_connection_open(&test_connection);
+  assert_int_equal(ref_conn_rsp_ok, 0);
+  assert_int_equal(ref_conn_rsp_err, 1);
+  assert_int_equal(ref_conn_req, 2);
+}
+
+// Testing that when reconnecting, an ok connect response from the broker leads to a faulted state.
+static void test_az_mqtt5_connection_reconnecttimeout_connect_rsp_ok_failure(void** state)
```
vaavva, line 1151: Shouldn't an ok connect response during a reconnect be what we're hoping for? (instead of going to a faulted state)

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+      &test_connection,
+      &test_connection_options,
+      &mock_mqtt,
+      &mock_mqtt_options,
+      &mock_client_1,
+      &mock_client_2,
+      &mock_client_hfsm_1,
+      &mock_client_hfsm_2);
+
+  _az_mqtt5_connection_idle_connecting(&test_connection);
```
vaavva, line 1397: If there aren't any additional ref checks in this, isn't this scenario tested in other tests?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+  _az_mqtt5_connection_connecting_reconnecttimeout(&mock_mqtt, AZ_MQTT5_CONNACK_NOT_AUTHORIZED);
+  _az_mqtt5_connection_reconnecttimeout_connecting(&test_connection);
+
+  assert_int_equal(test_connection._internal.client_certificate_index, 1);
+  assert_int_equal(ref_timeout, 1);
+}
+
+// Testing that when reconnecting, if the retry count is exhausted, the connection goes into an
+// idle state.
```
vaavva, line 1338: From the puml, doesn't retry count exhausted go to faulted?

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -434,28 +668,917 @@ static void test_az_mqtt5_connection_open_idle_success(void** state)
...
+  _az_mqtt5_connection_idle_connecting(&test_connection);
+  _az_mqtt5_connection_connecting_reconnecttimeout(&mock_mqtt, AZ_MQTT5_CONNACK_UNSPECIFIED_ERROR);
+  _az_mqtt5_connection_reconnecttimeout_connecting(&test_connection);
+  _az_mqtt5_connection_connecting_connected(&mock_mqtt);
+  _az_mqtt5_connection_connected_disconnecting(&test_connection);
+
+  assert_int_equal(ref_timeout, 1);
+}
+
+// Testing that a successful connection, exhausted re-connection, connection and disconnection
```
vaavva, line 1434: same question about exhausted to faulted here

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -241,36 +280,159 @@ AZ_INLINE void test_az_mqtt5_connection_setup(
       AZ_OK);
 }
 
+AZ_INLINE void _az_mqtt5_connection_idle_connecting(az_mqtt5_connection* test_conn)
```
vaavva, line 283: Can you add comments describing these functions as well?

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: There currently is no event filtering in the root state so I can't check that the pub rcv was sent to the application but I will add the other two checks

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: When in a connected state, if we receive a disconnect from the broker we will attempt to reconnect. Good catch though, the comment was not indicative of that

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: The `ref_disconn_req` variable only gets increased when the application calls connection close which isn't happening in this case

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: We arrived at the error because of a connect response error so it would be odd for us to receive a connect response ok when we haven't sent out another connect request yet

sdk/tests/core/test_az_mqtt5_connection.c
vaavva: Should it be renamed ref_conn_close_req then?

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: This was a happy case test scenario, but you're right we have tested the exact scenario in a previous test case

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: No, I realize that the puml has not updated, if you look at the .puml file it shows the updated retry count to idle transition but not the rendered image

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: See above!

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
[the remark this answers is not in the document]
RLeclair: Yes it would, the reasoning behind that is that the error sent inbound should be generic given that there are going to be state machines dependent on the connection so they should also know to go to faulted. They also may not know how to handle the error

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -292,6 +295,7 @@ AZ_INLINE void _az_mqtt5_connection_idle_connecting(az_mqtt5_connection* test_co
   assert_int_equal(ref_conn_req_prior + 1, ref_conn_req);
 }
 
+// Helper function to move the connection from connecting to connected.
```
vaavva, line 298:
```suggestion
// Helper function to move the connection from connecting to connected by sending a successful inbound connack.
```

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -282,6 +284,7 @@ AZ_INLINE void _az_mqtt5_connection_setup(
       AZ_OK);
 }
 
+// Helper function to move the connection from idle to connecting.
```
vaavva, line 287:
```suggestion
// Helper function to move the connection from idle to connecting by opening the connection.
```

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -358,6 +366,7 @@ AZ_INLINE void _az_mqtt5_connection_faulted_check(az_mqtt5_connection* test_conn
       AZ_ERROR_HFSM_INVALID_STATE);
 }
 
+// Helper function to move the connecting to a reconnecting state.
```
vaavva, line 371:
```suggestion
// Helper function to move the connection from connecting to a reconnecting state by sending an inbound connack with the passed in reason.
```

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -374,6 +383,7 @@ AZ_INLINE void _az_mqtt5_connection_connecting_reconnecttimeout(
   assert_int_equal(ref_conn_rsp_err_prior + 1, ref_conn_rsp_err);
 }
 
+// Helper function to move the from a reconnecting state to a connecting state.
```
vaavva, line 388:
```suggestion
// Helper function to move the connection from a reconnecting state to a connecting state by triggering a reconnect timeout.
```

vaavva (approved):

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
```diff
@@ -297,6 +308,72 @@ AZ_INLINE int32_t _get_elapsed_time(int64_t current_time_msec, int64_t start_tim
...
+  int32_t random_num = 0;
+  _az_event_pipeline* pipeline = (_az_event_pipeline*)&conn->_internal.event_pipeline;
+  _az_event_pipeline_timer* timer = (_az_event_pipeline_timer*)&conn->_internal.connection_timer;
+
+  if (az_result_failed(az_platform_get_random(&random_num)))
+  {
+    az_platform_critical_error();
+  }
+  double normalized_random_num = random_num / (double)RAND_MAX;
+  int32_t random_jitter_msec
```
ewertons, line 323: There is a potential int32_t overflow here, similar to what we are fixing in azure-iot-sdk-c.

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
```diff
@@ -297,6 +308,72 @@ AZ_INLINE int32_t _get_elapsed_time(int64_t current_time_msec, int64_t start_tim
...
+
+      return ret;
+    }
+    else
+    {
+      az_platform_critical_error();
+    }
+  }
+
+  conn->_internal.reconnect_counter++;
```
ewertons, line 349: Potential type overflow

ewertons (approved):
Approved with change requests.

sdk/samples/core/az_mqtt5_connection_user_config.h
```diff
@@ -0,0 +1,13 @@
+#ifndef _az_MQTT5_CONNECTION_USER_CONFIG_H
+#define _az_MQTT5_CONNECTION_USER_CONFIG_H
+
+// Uncomment the following macros to change the default MQTT connection configuration.
+
+// #define AZ_MQTT5_CONNECTION_MIN_RETRY_DELAY_MSEC (1000)
```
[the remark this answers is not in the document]
RLeclair: Can change to that, this was based on a suggestion from a previous comment. I do agree though that I should delete these comments and just leave a note specifying what needs to be done to change default values

sdk/samples/core/az_mqtt5_user_config.h
```diff
@@ -0,0 +1,9 @@
+#ifndef _az_MQTT5_USER_CONFIG_H
+#define _az_MQTT5_USER_CONFIG_H
+
+// Uncomment the following macros to change the default MQTT configuration.
+
+// #define AZ_MQTT5_DEFAULT_MQTT_CONNECT_KEEPALIVE_SECONDS (240)
```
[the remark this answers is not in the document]
RLeclair: See above

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
RLeclair: Fixed in new commit! Thanks for pointing that out

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
RLeclair: Fixed in my new commit

sdk/samples/core/az_mqtt5_connection_user_config.h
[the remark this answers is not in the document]
ewertons: Is there any bad reason not to have the defines uncommented with default values? That way we also guarantee that our builds will test them as well. If they are commented out, they are ignored throughout the code when it's built.

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
```diff
@@ -297,6 +308,93 @@ AZ_INLINE int32_t _get_elapsed_time(int64_t current_time_msec, int64_t start_tim
...
+  if (az_result_failed(az_platform_get_random(&random_num)))
+  {
+    az_platform_critical_error();
+  }
+  double normalized_random_num = random_num / (double)RAND_MAX;
+  int32_t random_jitter_msec = 0;
+  if ((int32_t)normalized_random_num != 0
+      && (int32_t)AZ_MQTT5_CONNECTION_MAX_RANDOM_JITTER_MSEC != 0)
+  {
+    if ((INT32_MAX / (int32_t)normalized_random_num)
```
ewertons, line 327: This resolves the type overflow issue. As a suggestion, it would be a lot cleaner (and safer? since they are more tested) if we ported (i.e., copied) the safe_math.h library from c-utility and used those safe_* functions.

PR #2742, opened 2024-02-12
Adding connection event relay policy

RLeclair:
- Added policy at the inbound end of the connection pipeline to relay events to the user instead of relying on a direct callback.
- Edited tests for connection, rpc, and telemetry to adjust for this change.
- Created new events to signal opening and closing the connection. 
- Updated HFSM connection diagram.


sdk/docs/core/resources/mqtt_connection.puml
```diff
@@ -40,11 +40,11 @@ state ConnectionManagement {
         Disconnecting : <b>AZ_EVENT_MQTT5_CONNECTION_CLOSE_REQ</b>/ ignore
 
         [*] -> Connecting
-        Connecting --> Connected : <b>MQTT_CONNECT_RSP_OK</b>/\n send_inbound(MQTT_CONNECT_RSP_OK)
+        Connecting --> Connected : <b>MQTT_CONNECT_RSP_OK</b>/\n send_inbound(MQTT_CONNECTION_OPEN_RSP)
```
vaavva, line 43: Do these changed need to be made in the samples too?

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -80,8 +80,8 @@ static az_result root(az_event_policy* me, az_event event)
 
     case AZ_HFSM_EVENT_ERROR:
     {
-      _az_RETURN_IF_FAILED(_az_mqtt5_connection_api_callback(
-          this_policy->_internal.connection,
+      _az_RETURN_IF_FAILED(az_event_policy_send_inbound_event(
+          (az_event_policy*)this_policy,
           (az_event){ .type = AZ_HFSM_EVENT_ERROR, .data = event.data }));
       // TODO: uncomment when this won't fail every time
       // if (az_result_failed(az_event_policy_send_inbound_event(me, event)))
```
vaavva, line 87:
```suggestion
      _az_RETURN_IF_FAILED(az_event_policy_send_inbound_event(
          me,
         event));
```
can remove the rest of that comment as well, github just wouldn't let me select it

vaavva: And for the rest of this file, can remove the commented out send_inbound_event calls and simplify the first parameter of the new send_inbound call

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -69,8 +69,8 @@ static az_result root(az_event_policy* me, az_event event)
       }
       else
```
vaavva, line 70: I think this whole else can be removed now that there is an inbound policy (it should do the same thing as the if statement above it). I can investigate this

vaavva: (same for other root state AZ_HFSM_EVENT_ERRORs)

sdk/tests/core/test_az_mqtt5_connection.c
```diff
@@ -109,44 +106,27 @@ static az_result test_subclient_policy_1_root(az_event_policy* me, az_event even
       ref_conn_close_req++;
       break;
     case AZ_HFSM_EVENT_ERROR:
-      ref_conn_error++;
-      break;
-    case AZ_MQTT5_EVENT_CONNECT_RSP:
-    {
-      az_mqtt5_connack_data* data = (az_mqtt5_connack_data*)event.data;
```
vaavva, line 116: Do these connack reasons still need to be checked?

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
```diff
@@ -557,16 +560,14 @@ static az_result reconnect_timeout(az_event_policy* me, az_event event)
...
+      if (az_result_failed(az_event_policy_send_inbound_event(
+              (az_event_policy*)this_policy,
+              (az_event){ .type = AZ_EVENT_MQTT5_CONNECTION_CLOSED_IND, .data = NULL })))
       {
         // Callback failed: fault the connection object.
         _az_RETURN_IF_FAILED(_az_hfsm_send_event(
             (_az_hfsm*)me, (az_event){ .type = AZ_HFSM_EVENT_ERROR, .data = NULL }));
       }
-      _az_RETURN_IF_FAILED(az_event_policy_send_inbound_event(
-          (az_event_policy*)this_policy,
```
vaavva, line 568: Is this event checked for in any other inbound HFSM policies that now need to look for the new one/stop looking for the old one?

sdk/inc/azure/core/az_mqtt5_connection.h
```diff
@@ -103,6 +113,41 @@ typedef az_result (*az_mqtt5_connection_callback)(
     az_event event,
     void* event_callback_context);
 
+/**
+ * @brief MQTT 5 connection event relay policy. This policy relays relevant events to the
+ * application.
+ */
+typedef struct
```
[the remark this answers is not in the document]
RLeclair: There is a dependency between the event relay policy and connection going both ways so it is better to keep the event relay policy within the scope of az_mqtt5_connection.h

sdk/src/azure/core/az_mqtt5_connection.c
```diff
@@ -40,6 +40,52 @@ AZ_NODISCARD az_mqtt5_connection_options az_mqtt5_connection_options_default()
   };
 }
 
+static az_result _az_mqtt5_connection_event_relay_process_outbound_event(
+    az_event_policy* policy,
+    az_event const event)
+{
+  _az_RETURN_IF_FAILED(az_event_policy_send_outbound_event(policy, event));
```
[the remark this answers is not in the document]
RLeclair: That is because we need a function type of `az_event_policy_handler` for the handler, `az_event_policy_send_outbound_event` does not fit the signature

sdk/src/azure/core/az_mqtt5_connection.c
```diff
@@ -116,12 +162,23 @@ AZ_NODISCARD az_result az_mqtt5_connection_init(
...
         &client->_internal.policy_collection,
         (az_event_policy*)&client->_internal.mqtt5_policy,
-        NULL));
+        (az_event_policy*)&client->_internal.event_relay_policy));
   }
 
+  client->_internal.event_relay_policy._internal.connection = client;
+  client->_internal.event_relay_policy._internal.relay_policy.outbound_policy
+      = (az_event_policy*)&client->_internal.policy_collection;
+  client->_internal.event_relay_policy._internal.relay_policy.inbound_policy = NULL;
```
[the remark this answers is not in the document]
RLeclair: We don't have a policy init function, all the policies we have (mqtt, collection, etc.) do this in their own initializers

sdk/docs/core/resources/mqtt_connection.puml
RLeclair: Good point! I will run those and make the edits

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
RLeclair: Changed! Good point, also looked into the other files

sdk/tests/core/test_az_mqtt5_connection.c
RLeclair: Not anymore, because of that the subclient also has `assert(false)` as a default

RLeclair: Not anymore, open and failed connections are sent through the new indicator events

sdk/src/azure/core/az_mqtt5_connection_hfsm.c
RLeclair: Good point! I checked again and didn't find any other HFSMs that use it

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -253,9 +253,10 @@ az_result mqtt_callback(az_mqtt5_connection* client, az_event event, void* callb
   az_app_log_callback(event.type, AZ_SPAN_FROM_STR("APP/callback"));
   switch (event.type)
   {
-    case AZ_MQTT5_EVENT_CONNECT_RSP:
+    case AZ_EVENT_MQTT5_CONNECTION_OPEN_IND:
     {
-      az_mqtt5_connack_data* connack_data = (az_mqtt5_connack_data*)event.data;
+      az_event* connack_event = (az_event*)event.data;
```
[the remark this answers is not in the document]
RLeclair: In case there was ever an event of a different type that signifies that the connection has been opened. I realize that for this indicator it will always be because of a connack event but for other indicators it might be more than one event. I agree with changing it to just the data, it'll simplify things

sdk/samples/core/mosquitto_multiple_rpc_sample.c
```diff
@@ -253,9 +253,10 @@ az_result mqtt_callback(az_mqtt5_connection* client, az_event event, void* callb
   az_app_log_callback(event.type, AZ_SPAN_FROM_STR("APP/callback"));
   switch (event.type)
   {
-    case AZ_MQTT5_EVENT_CONNECT_RSP:
+    case AZ_EVENT_MQTT5_CONNECTION_OPEN_IND:
```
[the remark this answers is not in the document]
RLeclair: Good catch, I think I missed this when rebasing, I had updated them before

sdk/samples/core/paho_multiple_rpc_sample.c
```diff
@@ -270,7 +270,7 @@ az_result mqtt_callback(az_mqtt5_connection* client, az_event event, void* callb
       break;
     }
 
-    case AZ_MQTT5_EVENT_DISCONNECT_RSP:
+    case AZ_EVENT_MQTT5_CONNECTION_CLOSED_IND:
```
vaavva, line 273: AZ_EVENT_MQTT5_CONNECTION_OPEN_IND in the paho samples too?

sdk/src/azure/core/az_mqtt5_rpc_client_hfsm.c
```diff
@@ -664,8 +666,11 @@ static az_result faulted(az_event_policy* me, az_event event)
   {
     case AZ_HFSM_EVENT_ENTRY:
     {
-      _az_RETURN_IF_FAILED(az_event_policy_send_inbound_event(
-          me, (az_event){ .type = AZ_HFSM_EVENT_ERROR, .data = NULL }));
+      if (az_result_failed(az_event_policy_send_inbound_event(
+              me, (az_event){ .type = AZ_HFSM_EVENT_ERROR, .data = NULL })))
+      {
+        az_platform_critical_error();
```
vaavva, line 672: Not sure if we need this on the faulted entry - the root one is more of a catch all if errors are bubbling up to that level (which they generally shouldn't because they should be handled in a sub-state), the application needs to be notified, potentially forcefully. I think a single hfsm going to faulted doesn't necessarily require the same force of notification

RLeclair: I misinterpreted the earlier comment, reverted the changes for faulted in this and the other HFSM files

sdk/inc/azure/core/az_mqtt5_connection.h
```diff
@@ -74,24 +74,34 @@ enum az_event_type_mqtt5_connection
...
    * @brief Event representing a request to close the MQTT 5 connection.
    *
    */
-  AZ_EVENT_MQTT5_CONNECTION_CLOSE_REQ = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT5, 20),
+  AZ_EVENT_MQTT5_CONNECTION_CLOSE_REQ = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT5, 23),
+
+  /**
+   * @brief Event representing the MQTT 5 connection being closed.
+   */
+  AZ_EVENT_MQTT5_CONNECTION_CLOSED_IND = _az_MAKE_EVENT(_az_FACILITY_CORE_MQTT5, 24),
```
vaavva, line 93: New events need to be added to the log listener in the samples folder

sdk/src/azure/core/az_mqtt5_telemetry_consumer_hfsm.c
```diff
@@ -60,21 +60,9 @@ static az_result root(az_event_policy* me, az_event event)
 
     case AZ_HFSM_EVENT_ERROR:
     {
-      az_mqtt5_telemetry_consumer* this_policy = (az_mqtt5_telemetry_consumer*)me;
-      if ((az_event_policy*)me->inbound_policy != NULL)
+      if (az_result_failed(az_event_policy_send_inbound_event(me, event)))
```
vaavva, line 63: add handling (to properly ignore) for AZ_EVENT_MQTT5_CONNECTION_CLOSED_IND and AZ_EVENT_MQTT5_CONNECTION_OPEN_IND for the two telemetry hfsms like you did for the rpc hfsms

vaavva (approved):

PR #2745, opened 2024-02-14
Update vcpkg commit SHA to match the C++ repo (#2744)

antkmsft:
Cherry-picking this commit from `main`, PR #2744:

>If you want, you can match the vcpkg SHA that we have in the C++ repo (https://github.com/Azure/azure-sdk-for-cpp/pull/5347).
>If we run into the same problems, we are on the same page when we are talking about possible solutions.
>Plus, I think, it is overall good, from time to time, to update dependencies to more recent versions.


vaavva:
This is addressed in https://github.com/Azure/azure-sdk-for-c/pull/2754

PR #2752, opened 2024-02-22
[HFSM] RPC server unsubscribe

vaavva:
Adds ability to unsubscribe to the RPC Server topic. Matches feature on the RPC Client

### DIFF
--- a/sdk/inc/azure/core/az_mqtt5_rpc_server.h
+++ b/sdk/inc/azure/core/az_mqtt5_rpc_server.h
@@ -184,7 +184,7 @@ typedef struct
 AZ_NODISCARD az_result az_mqtt5_rpc_server_register(az_mqtt5_rpc_server* client);
 
 /**
- * @brief Triggers an #AZ_MQTT5_EVENT_RPC_SERVER_UNSUB_REQ event from the application.
+ * @brief Triggers an #AZ_MQTT5_EVENT_UNSUB_REQ event from the application.
  *
  * @note This should be called from the application to unsubscribe to the request topic. This will
  * prevent the application from receiving commands unless it subscribes again. This may be used if
--- a/sdk/inc/azure/core/az_mqtt5_rpc_server.h
+++ b/sdk/inc/azure/core/az_mqtt5_rpc_server.h
@@ -200,7 +200,7 @@ AZ_NODISCARD az_result az_mqtt5_rpc_server_register(az_mqtt5_rpc_server* client)
  * @return An #az_result value indicating the result of the operation.
  * @retval #AZ_OK The event was triggered successfully.
  * @retval #AZ_ERROR_NOT_SUPPORTED if the server is not connected.
- * @retval Other on other failures creating/sending the unsubscribe message.
+ * @retval Other on failures creating/sending the unsubscribe message.
  */
 AZ_NODISCARD az_result az_mqtt5_rpc_server_unsubscribe_begin(az_mqtt5_rpc_server* client);
 
--- a/sdk/inc/azure/core/az_mqtt5_rpc_server.h
+++ b/sdk/inc/azure/core/az_mqtt5_rpc_server.h
@@ -40,12 +40,7 @@ enum az_event_type_mqtt5_rpc_server
    * @brief Event representing the RPC server requesting the execution of a command by the
    * application.
    */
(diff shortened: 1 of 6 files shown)


sdk/inc/azure/core/az_mqtt5_rpc_server.h
```diff
@@ -183,6 +188,22 @@ typedef struct
...
+ * @note This should be called from the application to unsubscribe to the request topic. This will
+ * prevent the application from receiving commands unless it subscribes again. This may be used if
+ * the application doesn't want to receive command requests anymore.
+ *
+ * @param[in] client The #az_mqtt5_rpc_server to use.
+ *
+ * @return An #az_result value indicating the result of the operation.
+ * @retval #AZ_OK The event was triggered successfully.
+ * @retval #AZ_ERROR_NOT_SUPPORTED if the server is not connected.
+ * @retval Other on other failures creating/sending the unsubscribe message.
```
ericwol-msft, line 203:
```suggestion
 * @retval Other on failures creating/sending the unsubscribe message.
```

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -424,20 +439,47 @@ static az_result waiting(az_event_policy* me, az_event event)
...
 
   if (_az_LOG_SHOULD_WRITE(event.type))
   {
     _az_LOG_WRITE(event.type, AZ_SPAN_FROM_STR("az_rpc_server/faulted"));
   }
+  switch (event.type)
+  {
+    case AZ_HFSM_EVENT_ENTRY:
+    {
+      // if ((az_event_policy*)this_policy->inbound_policy != NULL)
```
ericwol-msft, line 453: cod still needed?

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -424,20 +439,47 @@ static az_result waiting(az_event_policy* me, az_event event)
...
+  }
 
   return ret;
 }
 
+AZ_NODISCARD az_result az_mqtt5_rpc_server_unsubscribe_begin(az_mqtt5_rpc_server* client)
+{
+  if (client->_internal.connection == NULL)
+  {
+    // This API can be called only when the client is attached to a connection object.
```
ericwol-msft, line 475:
```suggestion
    // This API should be called only when the client is attached to a connection object.
```

ericwol-msft (approved):

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
vaavva: This will get uncommented in Raul's relay PR, but I can go ahead and remove it for now!

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
```diff
@@ -424,20 +439,42 @@ static az_result waiting(az_event_policy* me, az_event event)
...
 }
 
+AZ_NODISCARD az_result az_mqtt5_rpc_server_unsubscribe_begin(az_mqtt5_rpc_server* client)
+{
+  if (client->_internal.connection == NULL)
+  {
+    // This API should be called only when the client is attached to a connection object.
+    return AZ_ERROR_NOT_SUPPORTED;
+  }
+  return _az_hfsm_send_event(
```
RLeclair, line 473: I think we should be using `az_event_policy_send_outbound` here rather than `az_hfsm_send_event`

RLeclair (approved):
LGTM, left a small comment

sdk/src/azure/core/az_mqtt5_rpc_server_hfsm.c
vaavva: This is how we're doing it for the other api calls right now. Send_outbound only sends it to the outbound, it doesn't send it to this policy

RLeclair: I was thinking of something similar to what is done for `az_mqtt5_rpc_server_register` so moving the preparation of the package to the function for unsubscribing

vaavva: oh, I see! And I think it makes sense to be able to unsubscribe from faulted, which would be the only other limitation this would cause

PR #2763, opened 2024-03-01
[HFSM] [RPC] RPC Client to faulted on pub timeout

vaavva:
- This PR ensures that the RPC Client will be faulted if one of it's publishes times out and any requests related to that RPC Client are faulted as well and can be free'd


sdk/src/azure/core/az_mqtt5_request_hfsm.c
```diff
@@ -108,9 +108,22 @@ static az_result root(az_event_policy* me, az_event event)
...
-        az_platform_critical_error();
+        // if the RPC Client goes to faulted, fault all requests so they can be cleaned up.
+        az_mqtt5_rpc_client_rsp_event_data resp_data
+            = { .response_payload = AZ_SPAN_EMPTY,
+                .status = AZ_MQTT5_RPC_STATUS_UNKNOWN,
+                .error_message = AZ_SPAN_FROM_STR("RPC Client Faulted."),
+                .content_type = AZ_SPAN_EMPTY,
+                .correlation_id = this_policy->_internal.correlation_id };
+
+        _az_CRITICAL_IF_FAILED(_az_hfsm_transition_substate((_az_hfsm*)me, root, faulted));
```
RLeclair, line 122: I don't think this transition is reflected on the Request puml

RLeclair (approved):
LGTM!

sdk/src/azure/core/az_mqtt5_request_hfsm.c
vaavva: Wasn't 100% sure how to signify this was from root, but it's captured in the puml!
image

PR #2781, opened 2024-03-15
Turning the EnforceMaxLifeOfIssues Rule Off

ronniegeraghty:
This rule addresses a problem that isn't present in the C Repo.

### DIFF
--- a/.github/event-processor.config
+++ b/.github/event-processor.config
@@ -22,5 +22,5 @@
   "IdentifyStalePullRequests": "On",
   "CloseAddressedIssues": "On",
   "LockClosedIssues": "On",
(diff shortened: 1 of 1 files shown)


ahsonkhan (approved):

.github/event-processor.config
```diff
@@ -22,5 +22,5 @@
   "IdentifyStalePullRequests": "On",
   "CloseAddressedIssues": "On",
   "LockClosedIssues": "On",
```
ahsonkhan, line 24: Does this mean all closed issues (even manual ones) get locked?

ronniegeraghty: After a certain amount of time, yes

PR #3070, opened 2025-03-17
Fix strict-prototypes warning

RickWinter:
The current MAC builds use a clang compiler which enforces strict-prototypes. To avoid needing to make every `()` signature into `(void)` I turned on -std=c2x since the 23 standard handles the empty param signature correctly.

This also meant I had to fix a large chunk of IoT tests that had invalid signatures.


danieljurek (approved):

ewertons (approved):
For the iot-related code changes, it's approved.

antkmsft (approved):

PR #3531, opened 2026-06-26
Add az_span_dtoa_with_fractional for fixed-precision double formatting (#2157)

### LINKED ISSUE #2157: az_span_dtoa does not honor number of fractional digits
**Describe the bug**
az_span_dtoa seems to optimize the conversion and ignores leading zeros.

According to @ahsonkhan , the function was designed like that to be optimized for json (and produce a shorter content), but that breaks the expectation of the developer when using this function, it is not expected that it have any relationship with json.

**To Reproduce**
run:
```
double value = 1.00;
uint8_t buffer[16];
az_span buffer_span = AZ_SPAN_FROM_BUFFER(buffer);

az_span_dtoa(buffer_span, value, 2, &buffer_span);
```

Here `buffer` ends up with "1" instead of "1.00".

***Code Snippet***
In this branch I have a proposed fix as well as a test app for verifying the behavior.
https://github.com/ewertons/azure-sdk-for-c/tree/ewertons/strictdtoa 

**Expected behavior**

When az_span_dtoa is called for each of the cases below (considering 2 fractional digits), it should produce:

value=1.00, result="1.00"
value=1.01, result="1.01"
value=1.001, result="1.00"

**Screenshots**

When az_span_dtoa is called for each of the cases below (considering 2 fractional digits), it actually produces:

value=1.00, result="1"
value=1.01, result="1.01"
value=1.001, result="1"

- OS: Any
- IDE: Any
- Version of the Library used: https://github.com/Azure/azure-sdk-for-c/commit/df6f73eb38103f7c6e3a56b6c5ccf6559bc8ee20

**Additional context**

**Information Checklist**
Please make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report

- [x] Bug Description Added
- [x] Repro Steps Added
- [x] Setup information Added

ewertons:
Fixes #2157.

## Problem

`az_span_dtoa` intentionally strips non-significant trailing fractional zeros, so `1.0` with `fractional_digits = 2` renders as `"1"` instead of `"1.00"`. This is correct/compact for interoperable JSON numbers, but callers that need fixed-precision text output (e.g. a fixed number of decimal places) cannot get it.

## Approach (opt-in, non-breaking)

Changing the default behavior of `az_span_dtoa` would alter every existing payload, break ~15 existing `test_az_span.c` golden expectations (`1.0 -> "1"`, `0 -> "0"`, `1.5 -> "1.5"`, ...), and change the JSON writer's number output contract. As noted in the issue thread (and by `@ahsonkhan`), the default should not change.

Instead, this PR adds an **opt-in** public function that keeps trailing zeros:

```c
AZ_NODISCARD az_result az_span_dtoa_with_fractional(
    az_span destination,
    double source,
    int32_t fractional_digits,
    az_span* out_span);
```

`az_span_dtoa_with_fractional` always writes exactly `fractional_digits` digits after the decimal point (padding with trailing zeros), e.g. `1.0 -> "1.00"`, `1.001 -> "1.00"`, `1.01 -> "1.01"`. When `fractional_digits == 0`, no decimal point is written. `az_span_dtoa` is unchanged.

`fractional_digits` is a required parameter (not derived from `source`) because a `double` is an IEEE-754 binary value and carries no decimal precision — `1.0`, `1.00`, and `1.000` are byte-identical, so the desired number of fractional digits must come from the caller.

The shared conversion logic is factored into a static helper `_az_span_dtoa(..., bool keep_trailing_zeros, ...)`; both public functions are thin wrappers, so behavior and preconditions for `az_span_dtoa` are byte-for-byte identical.

## Tests

Adds `az_span_dtoa_with_fractional_succeeds`, covering the reported cases (`1.00`, `1.01`, `1.001` at 2 digits), integer/zero padding, preserved trailing zeros (`1.5 -> "1.50"`), leading zeros (`1.05 -> "1.05"`), negatives, sub-fractional values, other widths, and the `fractional_digits == 0` (no decimal point) case. All core tests pass.

## Note

This adds a new public API symbol, so it will need to go through API review (APIView). It is deliberately **not** a change to the existing `az_span_dtoa` default. A CHANGELOG entry is included under `1.6.0-beta.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features Added
 
+- [[#2157](https://github.com/Azure/azure-sdk-for-c/issues/2157)] Added `az_span_dtoa_fixed()`, which converts a `double` to text while preserving trailing zeros so the output always has exactly the requested number of fractional digits (e.g. `1.0` with 2 fractional digits produces `"1.00"`). `az_span_dtoa()` continues to strip non-significant trailing zeros.
+
 ### Breaking Changes
 
 ### Bugs Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features Added
 
-- [[#2157](https://github.com/Azure/azure-sdk-for-c/issues/2157)] Added `az_span_dtoa_fixed()`, which converts a `double` to text while preserving trailing zeros so the output always has exactly the requested number of fractional digits (e.g. `1.0` with 2 fractional digits produces `"1.00"`). `az_span_dtoa()` continues to strip non-significant trailing zeros.
+- [[#2157](https://github.com/Azure/azure-sdk-for-c/issues/2157)] Added `az_span_dtoa_with_fractional()`, which converts a `double` to text while preserving trailing zeros so the output always has exactly the requested number of fractional digits (e.g. `1.0` with 2 fractional digits produces `"1.00"`). `az_span_dtoa()` continues to strip non-significant trailing zeros.
 
 ### Breaking Changes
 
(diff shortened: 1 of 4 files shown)


xirzec (approved):
approving for CI purposes but check with @LarryOsterman before hitting merge please

LarryOsterman (approved):
I'm not a fan of the mandatory final pointer parameter pattern, but that's separate from the PR.

PR #3538, opened 2026-06-27
Fix memory-safety bugs in core parsing and POSIX clock

ewertons:
## Summary

Code review of the library ahead of repo lock surfaced **five latent memory-safety / correctness defects**. None duplicate existing GitHub issues. Each fix is minimal, confined to the malformed-input / boundary paths, and validated against the core and IoT unit-test suites (all passing).

## Fixes

| File | Function | Defect | Class |
|------|----------|--------|-------|
| `az_json_token.c` | `az_json_string_unescape` | Bounds guard `position > destination_size` allowed a 1-byte write at `index == destination_size` (and returned an over-long span). Changed to `>=`. | OOB write (CWE-787) |
| `az_http_response.c` | `az_http_response_get_next_header` | Header-value scan read `reader[offset]` in `while (true)` with no bound (the sibling field-name/OWS loops bound-check). A response whose last header value lacks a terminating CR read past the buffer. Added `offset >= reader_size` guard. | OOB read (CWE-125) |
| `az_http_response.c` | `_az_get_http_status_line` | When LF immediately follows the status code, `offset == 0` made the reason-phrase slice request end index `-1` (negative-size span). Guarded with `offset > 0 ? offset - 1 : 0`. | Invalid span |
| `az_posix.c` | `az_platform_clock_msec` | `ts.tv_sec * 1000` evaluated in `time_t` width (constants are `int` enums) then cast to `int64_t` too late. On 32-bit `time_t` targets, monotonic uptime > ~24.8 days overflows, corrupting retry backoff, SAS expiry and context timeouts. Cast to `int64_t` before the arithmetic. | Integer overflow (CWE-190) |
| `az_http_request.c` | `az_http_request_init` | `url_length` was never validated against the `url` span size, so the query scan (and later `az_span_slice` calls) could read OOB on caller misuse. Added `_az_PRECONDITION_RANGE(0, url_length, az_span_size(url))`, matching sibling init functions. | OOB read / hardening |

## Validation

- Builds clean (MSVC, incl. Code Analysis): `az_core`, `az_iot_hub`, `az_iot_provisioning`, `az_iot_adu`.
- `ctest` — **5/5 suites pass** (core, adu, common, hub, provisioning), 0 failures.

> Note: `az_posix.c` is not compiled on Windows; the change is a trivial, inspected cast. The other four files are exercised by the passing suites.

### DIFF
--- a/sdk/src/azure/core/az_http_response.c
+++ b/sdk/src/azure/core/az_http_response.c
@@ -108,7 +108,7 @@ _az_get_http_status_line(az_span* ref_span, az_http_response_status_line* out_st
   }
 
   // save reason-phrase in status line now that we got the offset. Remove 1 last chars(\r)
-  out_status_line->reason_phrase = az_span_slice(*ref_span, 0, offset - 1);
+  out_status_line->reason_phrase = az_span_slice(*ref_span, 0, offset > 0 ? offset - 1 : 0);
   // move position of reader after reason-phrase (parsed done)
   *ref_span = az_span_slice_to_end(*ref_span, offset + 1);
   // CR LF
@@ -249,8 +249,15 @@ AZ_NODISCARD az_result az_http_response_get_next_header(
   {
     int32_t offset = 0;
     int32_t offset_value_end = offset;
+    int32_t const reader_size = az_span_size(*reader);
     while (true)
     {
+      // Guard against reading past the end of the buffer when the header value is not
+      // terminated by a carriage return (malformed or truncated response).
+      if (offset >= reader_size)
+      {
+        return AZ_ERROR_HTTP_CORRUPT_RESPONSE_HEADER;
+      }
       uint8_t c = az_span_ptr(*reader)[offset];
       offset += 1;
       if (c == '\r')
--- a/sdk/src/azure/core/az_http_response.c
+++ b/sdk/src/azure/core/az_http_response.c
@@ -250,18 +250,18 @@ AZ_NODISCARD az_result az_http_response_get_next_header(
     int32_t offset = 0;
     int32_t offset_value_end = offset;
     int32_t const reader_size = az_span_size(*reader);
-    while (true)
+    // Scan the header value, bounding the scan by the size of the reader so a malformed or
+    // truncated response (one whose last header value lacks a terminating carriage return)
+    // cannot read past the end of the buffer. A non-positive reader size leaves the loop body
+    // unexecuted and is reported as a corrupt header below.
+    bool found_carriage_return = false;
+    for (; offset < reader_size; offset += 1)
     {
-      // Guard against reading past the end of the buffer when the header value is not
-      // terminated by a carriage return (malformed or truncated response).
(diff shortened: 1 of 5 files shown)


sdk/src/azure/core/az_http_response.c
```diff
@@ -249,8 +249,15 @@ AZ_NODISCARD az_result az_http_response_get_next_header(
   {
     int32_t offset = 0;
     int32_t offset_value_end = offset;
+    int32_t const reader_size = az_span_size(*reader);
     while (true)
     {
+      // Guard against reading past the end of the buffer when the header value is not
+      // terminated by a carriage return (malformed or truncated response).
+      if (offset >= reader_size)
```
LarryOsterman: Can the span size of `reader` be negative? If so, this check fails.

LarryOsterman (approved):

sdk/src/azure/core/az_http_response.c
ewertons: Hey Larry, so no, the reader is parser.remaining, a size that is guaranteed to be >= 0 by the az_span_create precondition, it can't be negative. I reworked the scan into a for (; offset < reader_size; ...) bound (matching the field-name/OWS loops above), which is structurally safe even for a non-positive size: the body simply doesn't execute and it returns a corrupt-header error. Added unit tests for the truncated / missing-CR value cases that exercise this bound.

sdk/src/azure/core/az_http_response.c
```diff
@@ -249,12 +249,19 @@ AZ_NODISCARD az_result az_http_response_get_next_header(
...
+    // Scan the header value, bounding the scan by the size of the reader so a malformed or
+    // truncated response (one whose last header value lacks a terminating carriage return)
+    // cannot read past the end of the buffer. A non-positive reader size leaves the loop body
+    // unexecuted and is reported as a corrupt header below.
+    bool found_carriage_return = false;
+    for (; offset < reader_size; offset += 1)
     {
       uint8_t c = az_span_ptr(*reader)[offset];
-      offset += 1;
       if (c == '\r')
```
LarryOsterman, line 261: Is it safe to assume that you'll always see `\r\n`? Or should you check for `\n` as well as `\r`?

LarryOsterman (approved):

sdk/src/azure/core/az_http_response.c
ewertons: Per HTTP/1.1 (RFC 9112 §2.2) well-formed headers do end in CRLF, so on conformant input this path behaves exactly as before. But this is a response parser consuming bytes off the wire, so it can't assume the peer is conformant — a truncated read or a non-conformant/malicious server can send a value with no terminating CR. The change doesn't try to accept malformed input; it just makes the scan fail safely (AZ_ERROR_HTTP_CORRUPT_RESPONSE_HEADER) instead of reading past the buffer (CWE-125). So it's defense-in-depth / memory safety on untrusted input, not an attempt to support malformed HTTP.

